PR #4, opened 2022-12-02
Improve the performance of changing the composer mode.

jmartinesp:
Note: this will only be noticeable once a new version of the RTE library is used and the crash when calling `renderComposerMode` is fixed.

Also, improved the UI for full screen mode.

Its Element Android counterpart is https://github.com/vector-im/element-android/pull/7691.


bmarty (approved):
Thanks!

PR #17, opened 2022-12-09
TestID semantics tag for login flow

michaelkaye:
I was working through a PoC of allowing the element-x-android app to be automated via a trafficlight adapter to see how hard it was and if we add testId tags then it's actually fairly trivial with things off the shelf.

Using uiautomator2 and appium seems to be fairly simple, provided we set appropriate resource-ids for it to click through. We should be able to use locally attached emulators/devices or third party services to be automated.

Here is an example of the consumer of these tags in this code base: https://github.com/vector-im/trafficlight-adapter-appium/blob/main/trafficlight_adapter_appium/test.py#L33 . It's not turned into a proper trafficlight adapter yet as it doesn't poll the server for actions to perform (atm all it can do is login :) 

Are these testIds something we should put in the code? I think they're a nice way to split out the testing points from the UI/UX we create around it, and definitely avoid trying to match based on button names or position in a layout. I think having them makes writing jetpack compose tests easier as well; it can consume the testId in a matcher.

These tags might also be a good place to start a real-world timing test for inital login and viewing of a large encrypted room, for instance, which might help work out timings. The same tags can be used to make sure that Jetpack Compose tests can find the same things to click on.

I think there should be a way to put the `androidx.compose.ui.semantics.testTagsAsResourceId` at a top level of the app layout, but i'm not sure how to make that work properly.

Bleh, anyway: would love it if we could get these tags in, i don't think they've got a particular downside. Happy to chat on matrix about it ^^.


bmarty:
Replaced by #52

PR #28, opened 2023-01-04
UI tests

bmarty:
TBC


ganfra (approved):
LGTM

bmarty:
We may iterate later, because `@Preview` are not nice rendered in the IDE. (ElementXTheme is not applied anymore).
We may use `@ShowkaseComposable(skip = true)` to have preview only for the ide and not for the tests.

PR #43, opened 2023-01-18
Setup Flipper.

### LINKED ISSUE #40: Flipper
Set up [Flipper](https://fbflipper.com/) on the project.

bmarty:
Closes #40

This is quite limited, since network on DB are handle by the Rust SDK, so this cannot really be plugged to Flipper. Also there is no Plugin to investigate Jetpack Compose structure (and this will not be supported by the core team, see https://github.com/facebook/flipper/issues/2366).

Can still be useful to take screen shot or video, watch logs, inspect shared preferences and investigate crashes.

This PR set up the tool, so later we may add more plugins.

### DIFF
--- a/app/src/debug/java/io/element/android/x/flipper/VectorFlipperProxy.kt
+++ b/app/src/debug/java/io/element/android/x/flipper/VectorFlipperProxy.kt
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2023 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


app/src/debug/java/io/element/android/x/flipper/VectorFlipperProxy.kt
```diff
@@ -0,0 +1,54 @@
...
+import com.facebook.flipper.android.utils.FlipperUtils
+import com.facebook.flipper.plugins.crashreporter.CrashReporterPlugin
+import com.facebook.flipper.plugins.inspector.DescriptorMapping
+import com.facebook.flipper.plugins.inspector.InspectorFlipperPlugin
+import com.facebook.flipper.plugins.sharedpreferences.SharedPreferencesFlipperPlugin
+import com.facebook.soloader.SoLoader
+
+class VectorFlipperProxy(
+    private val context: Context,
+) {
```
bmarty, line 31: @ganfra I am not sure that adding this class with alter ego no-op versions is the best way to handle no-op version. WDYT?

ganfra: I'd say we have 2 options: 
- do not call the initializer if not debug
- isEnabled could also check if it's debug or not ?

PR #47, opened 2023-01-18
Feature/fga/presenter tests

ganfra:
This is first setup to test a presenter.
This includes some refactoring to add Interfaces in the matrix module.


gradle/libs.versions.toml
```diff
@@ -110,6 +111,9 @@ test_mockk = { module = "io.mockk:mockk", version.ref = "test_mockk" }
 test_barista = { module = "com.adevinta.android:barista", version.ref = "test_barista" }
 test_hamcrest = { module = "org.hamcrest:hamcrest", version.ref = "test_hamcrest" }
 test_orchestrator = { module = "androidx.test:orchestrator", version.ref = "test_orchestrator" }
+test_turbine = { module = "app.cash.turbine:turbine", version.ref = "test_turbine"}
+test_truth = "com.google.truth:truth:1.1.3"
```
bmarty, line 115: Not the same format than others?

libraries/architecture/build.gradle.kts
```diff
@@ -18,16 +18,16 @@
 @Suppress("DSL_SCOPE_VIOLATION")
 plugins {
     id("io.element.android-compose-library")
-    alias(libs.plugins.molecule)
 }
 
 android {
-    namespace = "io.element.android.x.architecture"
+    namespace = "io.element.android.x.libraries.architecture"
```
bmarty, line 24: we should change all the other `libraries` namespaces then.

libraries/architecture/src/main/kotlin/io/element/android/x/architecture/Bindings.kt
```diff
@@ -39,7 +38,7 @@ fun <T : Any> Context.bindings(klass: Class<T>): T {
 fun <T : Any> Node.bindings(klass: Class<T>): T {
     // search dagger components in node hierarchy
     return generateSequence(this, Node::parent)
-        .filterIsInstance<DaggerComponentOwner>()
+        .filterIsInstance<io.element.android.x.di.DaggerComponentOwner>()
```
bmarty, line 41: Maybe revert the change on this file? (just update the import)

libraries/matrix/src/main/kotlin/io/element/android/x/matrix/FakeMatrixClient.kt
```diff
@@ -0,0 +1,69 @@
...
+import io.element.android.x.matrix.core.UserId
+import io.element.android.x.matrix.media.FakeMediaResolver
+import io.element.android.x.matrix.media.MediaResolver
+import io.element.android.x.matrix.room.FakeMatrixRoom
+import io.element.android.x.matrix.room.InMemoryRoomSummaryDataSource
+import io.element.android.x.matrix.room.MatrixRoom
+import io.element.android.x.matrix.room.RoomSummaryDataSource
+import org.matrix.rustcomponents.sdk.MediaSource
+
+class FakeMatrixClient(override val sessionId: SessionId) : MatrixClient {
```
bmarty, line 30: This class must be in the `test` folder.

libraries/matrix/src/main/kotlin/io/element/android/x/matrix/room/FakeMatrixRoom.kt
```diff
@@ -0,0 +1,66 @@
...
+package io.element.android.x.matrix.room
+
+import io.element.android.x.matrix.core.EventId
+import io.element.android.x.matrix.core.RoomId
+import io.element.android.x.matrix.timeline.FakeMatrixTimeline
+import io.element.android.x.matrix.timeline.MatrixTimeline
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.emptyFlow
+
+class FakeMatrixRoom(
```
bmarty, line 26: Move to `test` folder.

libraries/matrix/src/main/kotlin/io/element/android/x/matrix/room/InMemoryRoomSummaryDataSource.kt
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.x.matrix.room
+
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.emptyFlow
+
+class InMemoryRoomSummaryDataSource : RoomSummaryDataSource {
```
bmarty, line 22: `test` folder.

libraries/matrix/src/main/kotlin/io/element/android/x/matrix/timeline/FakeMatrixTimeline.kt
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+package io.element.android.x.matrix.timeline
+
+import io.element.android.x.matrix.core.EventId
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.emptyFlow
+import org.matrix.rustcomponents.sdk.TimelineListener
+
+class FakeMatrixTimeline : MatrixTimeline {
```
bmarty, line 24: `test`

bmarty (changes requested):
Thanks. Some classes should be moved to the `test` sourceset. Else LGTM!

gradle/libs.versions.toml
ganfra: Yes, I think we should use this format when there is no shared version between libs. What do you think?

libraries/architecture/build.gradle.kts
ganfra: Oh I thought it was already the case. What do you prefer?

libraries/matrix/src/main/kotlin/io/element/android/x/matrix/FakeMatrixClient.kt
ganfra: We can't put in test folder as it'll be available only internally. These classes are used on other modules, so we have to access them. But what we can do is create a matrix-test module

gradle/libs.versions.toml
bmarty: This is definitely more readable 👍 
Also if possible we should prioritize usage of `bom`.

libraries/architecture/build.gradle.kts
bmarty: not a strong pref, the risk of name collision is low even if we do not include `libraries`. But maybe for coherency keep the same path than the module path.

libraries/matrix/src/main/kotlin/io/element/android/x/matrix/FakeMatrixClient.kt
bmarty: OK. Maybe use this pattern to fix that?
```
dependencies {
  testCompile project(':libraries:matrix').sourceSets.test.output
}
```

ganfra: In an ideal world we'd use textFixtures gradle feature, but it's not working over AGP
Adding a module is still cleaner than relying on sourceSets.test, otherwise we might end up with UnitTests from matrix module

libraries/architecture/build.gradle.kts
ganfra: will do a clean-up in a separate PR as there are some mix between package name, namespace and module hierarchy.

bmarty (approved):
Excellent work! Thanks!

PR #51, opened 2023-01-19
Initial feature template module

ganfra:
Initial state of the feature template module.
This includes: 
- gradle file with most used dependencies for a feature module
- main classes : Presenter, View, Events, State and Node
- a unit test class for the Presenter


features/template/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+// TODO: Remove once https://youtrack.jetbrains.com/issue/KTIJ-19369 is fixed
+@Suppress("DSL_SCOPE_VIOLATION")
+plugins {
+    id("io.element.android-compose-library")
+    alias(libs.plugins.ksp)
+    alias(libs.plugins.anvil)
+}
+
+android {
+    namespace = "io.element.android.x.features.template"
```
bmarty, line 26: Maybe add some TODO everywhere that the developer will have to change something?

features/template/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+    implementation(project(":anvilannotations"))
+
+    implementation(project(":libraries:di"))
+    implementation(project(":libraries:core"))
+    implementation(project(":libraries:architecture"))
+    implementation(project(":libraries:matrix"))
+    implementation(project(":libraries:matrixui"))
+    implementation(project(":libraries:designsystem"))
+    implementation(project(":libraries:elementresources"))
+    implementation(project(":libraries:ui-strings"))
```
bmarty, line 44: Maybe create  a `fun DependencyHandlerScope.allLibrariesForFeatures()`? (or reuse `allLibraries()`).

I am not sure if we want to be certain that there is no useless dependencies...

features/template/src/main/kotlin/io/element/android/x/features/template/TemplateNode.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.bumble.appyx.core.modality.BuildContext
+import com.bumble.appyx.core.node.Node
+import com.bumble.appyx.core.plugin.Plugin
+import dagger.assisted.Assisted
+import dagger.assisted.AssistedInject
+import io.element.android.x.anvilannotations.ContributesNode
+import io.element.android.x.architecture.presenterConnector
+import io.element.android.x.di.AppScope
+
+// Change to use the right Scope for your feature.
```
bmarty, line 32:
```suggestion
// TODO Change to use the right Scope for your feature.
```

bmarty (approved):
Thanks. Some minor remarks, else LGTM!

features/template/build.gradle.kts
ganfra: Would be better to avoid using useless dependencies yes. I prefer to let the user pick the right ones

PR #52, opened 2023-01-20
Add test tags

bmarty:
Replacement for #17 

Please read the rationale in #17.

@michaelkaye can you confirm that this is still fulfilling your needs? Thanks!


libraries/testtags/src/main/kotlin/io/element/android/x/testtags/TestTags.kt
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package io.element.android.x.testtags
+
+@JvmInline
+value class TestTag internal constructor(val value: String)
+
+object TestTags {
+    /**
+     * OnBoarding screen
```
ElementBot, line 24: - :warning: The first sentence of this KDoc does not end with the correct punctuation.

libraries/testtags/src/main/kotlin/io/element/android/x/testtags/TestTags.kt
```diff
@@ -0,0 +1,43 @@
...
+     */
+    val onBoardingSignIn = TestTag("onboarding-sign_in")
+
+    /**
+     * Login screen
```
ElementBot, line 29: - :warning: The first sentence of this KDoc does not end with the correct punctuation.

libraries/testtags/src/main/kotlin/io/element/android/x/testtags/TestTags.kt
```diff
@@ -0,0 +1,43 @@
...
+     */
+    val loginChangeServer = TestTag("login-change_server")
+    val loginEmailUsername = TestTag("login-email_username")
+    val loginPassword = TestTag("login-password")
+    val loginContinue = TestTag("login-continue")
+
+    /**
+     * Change server screen
```
ElementBot, line 37: - :warning: The first sentence of this KDoc does not end with the correct punctuation.

ganfra (approved):

michaelkaye:
Sorry for not responding so far; just wanted to say i've seen this and am going to spend time now to check it does what we need.

michaelkaye:
So this looks fine by me; have made a clean build and appium and maestro are able to see the IDs of the various tags on android using an unsigned release build.

You should look forward to some more PRs from me to add extra test tags so we can automate more parts of the app.

On the topic of not being able to use these testTags in maestro, i did find the ones on android working just fine in `maestro studio`, for example:

```
- tapOn:
    id: "login-change_server"
```

Though I'm not sure whether that's what you mean with iOS EditText; 

It does look like the iOS code is already using [accessibilityIdentifier](https://developer.apple.com/documentation/swiftui/view/accessibilityidentifier(_:)) as a way to label things to click on in their UI tests; that is  [already doing what I need](https://github.com/vector-im/element-x-ios/blob/eeb69d19712eaf3ff20094743edad20e2c2f51d4/ElementX/Sources/Screens/Authentication/LoginScreen/View/LoginScreen.swift#L87), just with different strings than android.

bmarty:
OK, thanks for feedback.
On `maestro studio` I can see the same, but IIRC, when running the test it was not working. Maybe a silly mistake on my side, I will try again (later).

> just with different strings than android.

Ideally the `id`s will be identical between iOS and Android, but I guess this is what you want too :). Feel free to change the `id`s here if you want to.

PR #56, opened 2023-01-23
Rageshake fixes

bmarty:
- Use dedicated GH project (it's working, see https://github.com/vector-im/element-x-android-rageshakes/issues/1)
- Fix crash when sending bugreport. Java classes must be in java folder. It compiles, but it crashes at runtime 🥳 
- Add some log about the versions


app/src/main/kotlin/io/element/android/x/info/Logs.kt
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package io.element.android.x.info
+
+import io.element.android.x.BuildConfig
+import timber.log.Timber
+import java.text.SimpleDateFormat
+import java.util.*
+
+class Logs {
```
ganfra, line 24: I wouldn't create a class for that just the method

ganfra (approved):

app/src/main/kotlin/io/element/android/x/info/Logs.kt
bmarty: OK, I will change. In the future, we may want to inject classes in the constructor, to get the SDK version or instance.

PR #57, opened 2023-01-24
Feature/fga/gradle perf

ganfra:
This PR takes care of: 
- Moving some module around 
- Make core kotlin lib instead of android module
- Remove some useless dependencies
- change some gradle settings to improve build perf


libraries/textcomposer/src/main/kotlin/io/element/android/libraries/textcomposer/RichTextComposerLayout.kt
```diff
@@ -118,7 +117,8 @@ class RichTextComposerLayout @JvmOverloads constructor(
         }
     }
 
-    private val dimensionConverter = DimensionConverter(resources)
+    private val dimensionConverter =
+        io.element.android.libraries.androidutils.ui.DimensionConverter(resources)
```
bmarty, line 121: Maybe do some cleanup here (re-add the import).

bmarty (approved):
LGTM. We should maybe add a tool to detect useless dependency?

bmarty (approved):

PR #58, opened 2023-01-24
Add animation between screens

bmarty:


app/src/main/kotlin/io/element/android/x/node/NotLoggedInFlowNode.kt
```diff
@@ -62,7 +63,7 @@ class NotLoggedInFlowNode(
         return when (navTarget) {
             NavTarget.OnBoarding -> node(buildContext) {
                 OnBoardingScreen(
-                    onSignIn = { backstack.replace(NavTarget.LoginFlow) }
+                    onSignIn = { backstack.push(NavTarget.LoginFlow) }
```
bmarty, line 66: I think it's more friendly to be able to go back to the OnBoarding screen.

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/animation/ScreenTransition.kt
```diff
@@ -0,0 +1,31 @@
...
+import androidx.compose.animation.core.spring
+import androidx.compose.runtime.Composable
+import com.bumble.appyx.core.navigation.transition.ModifierTransitionHandler
+import com.bumble.appyx.navmodel.backstack.BackStack
+import com.bumble.appyx.navmodel.backstack.transitionhandler.rememberBackstackSlider
+
+@Composable
+fun <NavTarget> getDefaultTransitionHandler(): ModifierTransitionHandler<NavTarget, BackStack.State> {
+    return rememberBackstackSlider(
+        transitionSpec = { spring(stiffness = Spring.StiffnessMediumLow) },
```
bmarty, line 29: Using `StiffnessMediumLow` for a faster animation than the default one.

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/animation/ScreenTransition.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import androidx.compose.animation.core.Spring
+import androidx.compose.animation.core.spring
+import androidx.compose.runtime.Composable
+import com.bumble.appyx.core.navigation.transition.ModifierTransitionHandler
+import com.bumble.appyx.navmodel.backstack.BackStack
+import com.bumble.appyx.navmodel.backstack.transitionhandler.rememberBackstackSlider
+
+@Composable
+fun <NavTarget> getDefaultTransitionHandler(): ModifierTransitionHandler<NavTarget, BackStack.State> {
```
ganfra, line 27: I think it should be called `rememberDefaultTransitionHandler` to make it clear

bmarty: Naming convention is a bit mixing up here, since this method also returns something. So it does something (`remember`) and it return something (`get`).
But I think you are right, on Compose, it's better to use `remember` here.

ganfra: Yes all the `remember` methods return something :-)

PR #67, opened 2023-02-01
Feature/fga/fix restoration

ganfra:
This PR takes care of : 
- Fixing crashes about missing session on config change/process death
- Handling uiMode changes through compose (switching dark/light theme)
- Removing PresenterConnector and usage of Molecule in feature modules (only necessary for UT presenters)
- Improving AvatarData management


app/src/main/AndroidManifest.xml
```diff
@@ -31,7 +31,8 @@
         tools:targetApi="33">
         <activity
             android:name=".MainActivity"
-            android:configChanges="orientation|screenSize|screenLayout|keyboardHidden"
+            android:theme="@style/Theme.ElementX.Splash"
+            android:configChanges="orientation|screenSize|screenLayout|keyboardHidden|uiMode"
```
bmarty, line 35: `uiMode` is added. OK. I am surprised to see that change. What about language change? Will it crash the app?

bmarty (approved):
LGTM, thanks!

app/src/main/AndroidManifest.xml
ganfra: The crash is fixed for any cases (state and process restoration).
It's just that compose already takes care of uiMode changes, so no need for a configChange.

PR #98, opened 2023-02-20
Store session data in a secure way

### LINKED ISSUE #84: Store user credentials in the secure storage.
The current session info is currently stored in shared preferences in plain text (see this [TODO](https://github.com/vector-im/element-x-android/blob/develop/libraries/matrix/src/main/kotlin/io/element/android/libraries/matrix/session/PreferencesSessionStore.kt#L42)).

We should securely store that data instead. After some discussion, we agreed to use a SQLite DB for that with SQLCipher to ensure its contents are encrypted.

Tasks:

- [x] Integrate SQLDelight.
- [x] Make it work with SQLCipher.
- [x] Migrate the current `PreferenceSessionStore` component to use this DB instead.

jmartinesp:
## What

Some important changes were added with this PR:

- `PreferenceSessionStore` is now backed by a SQLite DB with a SQLCipher driver instead of a `DataStore`. This same DB can be used in the future to store other platform-specific data.
- The SQLCipher driver configuration and passphrase handling are implemented in a new `:libraries:encrypteddb` module. Passphrase is generated using a `SecureRandom` and stored using AndroidX's [EncryptedFile](https://developer.android.com/reference/androidx/security/crypto/EncryptedFile).
- To use `androidx.security.crypto` we need to bump minSdkVersion to 23. There haven't been any complaints coming from PS about this, so I believe this is fine: there are very few users in APIs < 23 according to our available data.
- These APIs were supposed to user `SessionId`, but this object isn't part of the SDK and wasn't properly defined. As we believed it's no longer useful, APIs using `SessionId` have been modified to use `UserId` instead.

## Why
Closes #84 .

## Tests

1. If you had Element X previously installed, you probably need to uninstall it or at least delete its data.
2. Sign in as usual.

If the app works, that should be all.


libraries/encrypted-db/src/main/kotlin/io/element/encrypteddb/EncryptedDatabase.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.squareup.sqldelight.db.SqlDriver
+import io.element.encrypteddb.passphrase.PassphraseProvider
+import net.sqlcipher.database.SupportFactory
+
+/**
+ * Creates an encrypted version of the [SqlDriver] using SQLCipher's [SupportFactory].
+ * @param passphraseProvider Provides the passphrase needed to use the SQLite database with SQLCipher.
+ */
+class SqlCipherDriverFactory(
+    private val passphraseProvider: PassphraseProvider,
```
jmartinesp, line 30: My first idea was trying to pass this using DI, but as there might be several instances of this in the future and its configuration is not trivial, maybe it would be better to just pass a `passphrase: String` parameter retrieved in the `provides...` method that creates this driver instead.

libraries/matrix/src/main/kotlin/io/element/android/libraries/matrix/di/MatrixModule.kt
```diff
@@ -33,4 +38,14 @@ object MatrixModule {
     fun providesRustAuthenticationService(baseDirectory: File): AuthenticationService {
         return AuthenticationService(baseDirectory.absolutePath)
     }
+
+    @Provides
+    @SingleIn(AppScope::class)
+    fun provideMatrixDatabase(@ApplicationContext context: Context): Database {
+        val name = "matrix_database"
```
jmartinesp, line 45: Maybe `credentials_database` instead? What are we planning to store there in the future?

jmartinesp:
Maybe we should add some way to retrieve or print the passphrase in debug builds so we can inspect the DB if needed?

libraries/encrypted-db/src/main/kotlin/io/element/encrypteddb/passphrase/RandomSecretPassphraseProvider.kt
```diff
@@ -0,0 +1,59 @@
...
+import java.security.SecureRandom
+
+/**
+ * Provides a secure passphrase for SQLCipher by generating a random secret and storing it into an [EncryptedFile].
+ * @param secretSize Length of the generated secret.
+ * @param file Destination file where the key will be stored.
+ * @param context Android [Context], used by [EncryptedFile] for cryptographic operations.
+ * @param alias Alias of the key used to encrypt & decrypt the [EncryptedFile]'s contents.
+ */
+class RandomSecretPassphraseProvider(
```
ElementBot, line 31: - :warning: Documentation of RandomSecretPassphraseProvider is outdated

jmartinesp:
The `codecov/patch` job is failing, I believe this is because I changed lots of lines using `SessionId` and it's recognising them as  new lines with no coverage 😞 .

libraries/matrix/src/main/kotlin/io/element/android/libraries/matrix/di/MatrixModule.kt
ganfra: Yeah I guess we could have an other name. Also the `Database` class should have a better name?

libraries/matrix/src/main/kotlin/io/element/android/libraries/matrix/session/PreferencesSessionStore.kt
```diff
@@ -16,91 +16,62 @@
...
 import javax.inject.Inject
 
-private val Context.dataStore: DataStore<Preferences> by preferencesDataStore(name = "elementx_sessions")
-
-// TODO It contains the access token, so it has to be stored in a more secured storage.
-private val sessionKey = stringPreferencesKey("session")
-
 @SingleIn(AppScope::class)
 @ContributesBinding(AppScope::class)
 class PreferencesSessionStore @Inject constructor(
```
ganfra, line 33: We should probably have this implementation in a separate module? 
It's not really the responsability of the matrix-module to know that. 
But we can do that later when we'll expose a matrix-api module

In the meantime, it should probably be named `SqliteSessionStore`

libraries/matrix/src/main/sqldelight/io/element/android/libraries/matrix/session/SessionData.sq
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE SessionData (
```
ganfra, line 1: Same remarks than for the SessionStore impl, we should move that to a module as it doesn't belong to the matrix module to know that.

ganfra (commented):
Good work, have some remarks/questions

libraries/matrix/src/main/kotlin/io/element/android/libraries/matrix/di/MatrixModule.kt
jmartinesp: Yes, I will change it to something better.

app/src/main/kotlin/io/element/android/x/di/MatrixClientsHolder.kt
```diff
@@ -21,7 +21,7 @@ import io.element.android.libraries.di.AppScope
 import io.element.android.libraries.di.SingleIn
 import io.element.android.libraries.matrix.MatrixClient
 import io.element.android.libraries.matrix.auth.MatrixAuthenticationService
-import io.element.android.libraries.matrix.core.SessionId
+import io.element.android.libraries.matrix.core.UserId
```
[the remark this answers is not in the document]
jmartinesp: From our previous discussion I didn't think there was a distinction: `SessionId` was built from user id + device id in EA, but I thought we didn't need to do that now? And if we don't, `SessionId` and `UserId` are equivalent, so when I asked if we could just remove `SessionId` in that case we agreed on doing that. 

I mean, we can bring back `SessionId` as some hash that identifies the session but keeps the actual `UserId` hidden, but that's not the way either the Rust SDK or the `Session` and `SessionData` classes and tables currently work. The Rust SDK won't accept anything but a `UserId` as input AFAIK and every time we want to use it we'd have to translate or map from `SessionId` to `UserId` again.

gradle/libs.versions.toml
```diff
@@ -117,6 +119,11 @@ jsoup = { module = "org.jsoup:jsoup", version.ref = "jsoup" }
 appyx_core = { module = "com.bumble.appyx:core", version.ref = "appyx" }
 molecule-runtime = { module = "app.cash.molecule:molecule-runtime", version.ref = "molecule" }
 timber = "com.jakewharton.timber:timber:5.0.1"
+sqldelight-driver-android = { module = "com.squareup.sqldelight:android-driver", version.ref = "sqldelight" }
+sqldelight-driver-jvm = { module = "com.squareup.sqldelight:sqlite-driver", version.ref = "sqldelight" }
+sqldelight-coroutines = { module = "com.squareup.sqldelight:coroutines-extensions", version.ref = "sqldelight" }
+sqlcipher = { module = "net.zetetic:android-database-sqlcipher", version.strictly = "4.5.3" }
+sqlite = { module = "androidx.sqlite:sqlite", version.strictly = "2.3.0" }
```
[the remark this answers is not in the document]
jmartinesp: I thought it was a good way to have the version separated from the maven identifier, but it's probably easier to just use the joint notation 👍 .

libraries/matrix/src/main/sqldelight/io/element/android/libraries/matrix/session/SessionData.sq
jmartinesp: Should I just create a new 'elementdb' library module containing this? Or maybe a `storage` feature module?

app/src/main/kotlin/io/element/android/x/di/MatrixClientsHolder.kt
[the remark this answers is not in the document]
ganfra: It's up to you, it's just that `UserId` is way more generic and will be used for lots of different things. It's a matter of API design, but the result is the same. I'm ok with both choices.

bmarty (commented):
Thanks for this PR!
More comments, still open to discussion!

libraries/session-storage/src/main/kotlin/io/element/android/libraries/sessionstorage/DatabaseSessionStore.kt
```diff
@@ -0,0 +1,56 @@
...
+    override suspend fun storeData(sessionData: SessionData) {
+        database.sessionDataQueries.insertSessionData(sessionData)
+    }
+
+    override suspend fun getLatestSession(): SessionData? {
+        return database.sessionDataQueries.selectFirst()
+            .executeAsOneOrNull()
+    }
+
+    override suspend fun getSession(sessionId: SessionId): SessionData? {
```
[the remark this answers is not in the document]
jmartinesp: That was added after splitting the module from the `matrix` one, so I guessed we could just add a mapping to SessionId here. However, since we'll rollback the SessionId -> UserId change I'll what else can I do here.

libraries/session-storage/src/main/sqldelight/io/element/android/libraries/matrix/session/SessionData.sq
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE SessionData (
+    userId TEXT NOT NULL PRIMARY KEY,
```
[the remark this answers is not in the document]
jmartinesp: I wanted to avoid this because I saw no point on doing an extra hash operation to create an id when you already have a valid id for free. I could understand using those for values that the user or some attacker might see such as the data folder, but I'm not really convinced about why it's necessary to have that same overhead for each query.

libraries/matrix/src/main/kotlin/io/element/android/libraries/matrix/auth/MatrixAuthenticationService.kt
```diff
@@ -17,15 +17,15 @@
...
-import io.element.android.libraries.matrix.core.SessionId
+import io.element.android.libraries.matrix.core.UserId
 import kotlinx.coroutines.flow.Flow
 
 interface MatrixAuthenticationService {
     fun isLoggedIn(): Flow<Boolean>
-    suspend fun getLatestSessionId(): SessionId?
-    suspend fun restoreSession(sessionId: SessionId): MatrixClient?
+    suspend fun getLatestUserId(): UserId?
+    suspend fun restoreSession(userId: UserId): MatrixClient?
```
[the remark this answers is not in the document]
ganfra: Yeah by the way, I think we should remove `getLatestSessionId` from the MatrixAuthenticationService and move it somewhere else, it's not really the responsability of the matrix sdk to know that IMO. But we can change later.

libraries/session-storage/src/main/sqldelight/io/element/android/libraries/matrix/session/SessionData.sq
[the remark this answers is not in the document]
bmarty: OK, let's keep it like this then, since this table will not be a big table, so using potentially big strings for primary key is not a problem.

jmartinesp:
After merging `develop` I had to change `sample/minimal` so instead of using `PreferencesSessionStore` that no longer exists in this PR uses a custom `InMemorySessionStore`.

bmarty (approved):
I added a last commit for code clarity, I think the PR can be merged now. Thanks!

PR #99, opened 2023-02-20
Remove roadmap section from README

Johennes:
We don't currently have a public roadmap and the current wording might set the wrong expectations. The EX iOS repository doesn't have this section either. So I think it'll be good to remove it for now.

CC @ara4n


jmartinesp (approved):
LGTM.

Johennes:
Thanks @opusforlife2. I fixed it with https://github.com/vector-im/element-x-android/commit/6592a1a2ebc3ab68bf932d23c7dbfb3e87a61ddc.

PR #117, opened 2023-02-27
CI: make sure Paparazzi test failures stop the test flow

jmartinesp:
Changes to CI:
- Run Paparazzi tests just once when `koverMergedReport` is used.
- Make sure a Paparazzi failure breaks the test flow and blocks merging.
- Change `./gradlew check` call in `quality.yml` to custom `runQualityChecks` task.

This should also improve overall build times in the test and quality flows and avoid duplicate test runs.


.github/workflows/tests.yml
```diff
@@ -21,6 +22,27 @@ jobs:
       cancel-in-progress: true
     steps:
       - uses: actions/checkout@v3
+        with:
+          lfs: 'true'
```
bmarty, line 26: `lfs: 'true'` was maybe the only thing which was missing to make the screenshot test not failing, WDYT?

tests/uitests/src/test/kotlin/io/element/android/tests/uitests/ScreenshotTest.kt
```diff
@@ -67,7 +67,7 @@ class ScreenshotTest {
 
     @get:Rule
     val paparazzi = Paparazzi(
-        maxPercentDifference = 0.0,
+        maxPercentDifference = 0.1,
```
bmarty, line 70: https://github.com/vector-im/element-android/commit/28f4e79b43fa10bcd228c1da2bf19bcd290b4e46

.github/workflows/tests.yml
bmarty: OK, so this is not enough, we have to explicitly invoke `verifyPaparazziDebug` task for the test to fail :/

tests/uitests/build.gradle.kts
```diff
@@ -42,3 +42,7 @@ dependencies {
     allLibraries()
     allFeatures()
 }
+
+tasks.withType<Test> {
+    enabled = project.hasProperty("runPaparazzi")
+}
```
bmarty, line 48: It seems that [the coverage is not computed](https://github.com/vector-im/element-x-android/actions/runs/4290584239/jobs/7474748317) if we exclude this module from the global `test` task. So I would remove this `runPaparazzi` trick.

bmarty (commented):
Thanks, please see my comment about `runPaparazzi`.

.github/workflows/tests.yml
jmartinesp: Yep. Also, in my case the tolerance had to be > `0.0` to work, as there are very tiny difference in maybe 6-8 pixels between my generated screenshots and the ones from the CI. It might be due to some color callibration of the OS? I'm not sure, to be honest, but they didn't match 100%.

Maybe something like `0.01` could work?

tests/uitests/build.gradle.kts
jmartinesp: Then we'll have to run it twice to make it fail, or maybe use some other trick to detect a failure in verifyPaparazzi like verifying if the `failures` folder exists?

bmarty: Yes, it's a shame... let's run them twice for now...

bmarty: Paparazzi not verifying by default seems to be by design: https://github.com/cashapp/paparazzi/issues/701#issuecomment-1404295807 . The workaround to change this behavior could not work IIUC.

.github/workflows/tests.yml
```diff
@@ -21,8 +22,27 @@ jobs:
...
+        with:
+          name: reports
+          path: tests/uitests/build/reports/tests/testDebugUnitTest/
+          retention-days: 1
+
+      - name: Upload Screenshot Failure Differences
+        uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: reports
```
bmarty, line 43: maybe use a distinct name?

.github/workflows/tests.yml
```diff
@@ -21,8 +22,27 @@ jobs:
...
+          path: tests/uitests/out/failures/
+          retention-days: 1
```
bmarty, line 45: Maybe keep the artifact a bit longer?

.github/workflows/tests.yml
```diff
@@ -20,19 +21,50 @@ jobs:
...
+      - name: ⏬ Checkout with LFS
+        uses: actions/checkout@v3
+        with:
+          lfs: 'true'
+
+      - name: ⚙️ Run unit & screenshot tests, generate kover report
+        run: ./gradlew koverMergedReport $CI_GRADLE_ARG_PROPERTIES -Pci-build=true
+
+      - name: 📈 Verify coverage
+        run: ./gradlew koverMergedVerify $CI_GRADLE_ARG_PROPERTIES -Pci-build=true
```
bmarty, line 33: `koverMergedReport` and `koverMergedVerify` will both run the tests IIRC. It's a shame there is not a single task to do both without running the test twice...

jmartinesp: But we just launched `verifyPaparazzi` from the previous `koverMergedReport`, so if I'm not mistaken it should be up to date and not run again. At least, I don't see any delay when running these, I'll double check just in case.

bmarty (approved):
Thanks!
I see that you have recorded the screenshots?

jmartinesp:
> Thanks! I see that you have recorded the screenshots?

Yes, I recorded them again to make sure the `maxPercentDifference = 0.01` was enough to make them work even if the OS used to record the images doesn't match the one that verifies them.

jmartinesp:
I just force pushed to try to force codecov to run. For some reason, it's not running in any of my PRs since a couple of days ago...

PR #92, opened 2023-02-17
Configure Diawi and split APK

bmarty:
This PR:
- configure Diawi so that arm64 APK are uploaded there and a comment is added to the PR with a handy QR code to download the app
- split the output into 4 APK, one per arch. Mainly because APK must be < 75MB to be uploaded to Diawi using a free plan, but we would need to do that at some point anyway.

APK versionCodes are as expected, see below:

```shell
> $ANDROID_HOME/build-tools/30.0.2/aapt dump badging ./app/build/outputs/apk/debug/app-arm64-v8a-debug.apk |grep package 
package: name='io.element.android.x.debug' versionCode='12' versionName='1.0' compileSdkVersion='33' compileSdkVersionCodename='13'
> $ANDROID_HOME/build-tools/30.0.2/aapt dump badging ./app/build/outputs/apk/debug/app-armeabi-v7a-debug.apk |grep package 
package: name='io.element.android.x.debug' versionCode='11' versionName='1.0' compileSdkVersion='33' compileSdkVersionCodename='13'
> $ANDROID_HOME/build-tools/30.0.2/aapt dump badging ./app/build/outputs/apk/debug/app-x86-debug.apk |grep package  
package: name='io.element.android.x.debug' versionCode='13' versionName='1.0' compileSdkVersion='33' compileSdkVersionCodename='13'
> $ANDROID_HOME/build-tools/30.0.2/aapt dump badging ./app/build/outputs/apk/debug/app-x86_64-debug.apk |grep package 
package: name='io.element.android.x.debug' versionCode='14' versionName='1.0' compileSdkVersion='33' compileSdkVersionCodename='13'
> $ANDROID_HOME/build-tools/30.0.2/aapt dump badging ./app/build/outputs/apk/debug/app-universal-debug.apk |grep package  
package: name='io.element.android.x.debug' versionCode='10' versionName='1.0' compileSdkVersion='33' compileSdkVersionCodename='13'
```

PR #136, opened 2023-03-06
Split some library modules

bmarty:
In the context of #92:

- Split module `dateformatter` into `api`, `impl` and `test`.
   - The existing `FakeDaySeparatorFormatter` has been moved to the `test`module.
- Split module `session-storage` into `api` and `impl`.
   - This allow us to create a second `impl` module, to host `InMemorySessionStore` in it's own module implementation

Other existing library modules do not really need to be splitted, they are "real" libraries (and not "services").


libraries/session-storage/impl/src/main/kotlin/io/element/android/libraries/sessionstorage/impl/SessionDataMapper.kt
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.libraries.sessionstorage.impl
+
+import io.element.android.libraries.sessionstorage.api.SessionData
+
+internal fun SessionData.toDbModel(): io.element.android.libraries.matrix.session.SessionData {
```
jmartinesp, line 21: If we're going to have 2 instances of classes called `SessionData` maybe it would be better to rename one of them? The `api` one could be `SessionModel` or `SessionDetails` maybe?

jmartinesp (approved):
LGTM, I just have a comment about the duplicate `SessionData` classes.

libraries/session-storage/impl/src/main/kotlin/io/element/android/libraries/sessionstorage/impl/SessionDataMapper.kt
bmarty: Yes, it's not ideal. I wanted to avoid renaming. But since this is hidden in the impl module, I guess it's fine.

Issue #91, opened 2023-02-17
Real multi module project.


bmarty:
# NOTE: This is a draft issue, still in discussion.

As a developer and as a company who provide application for others, we want to have a real and clean multi modules project.

- We want to be able to tweak the application by adding or removing feature(s) on-demand, at compilation time.
- In #82 some Fake classes have been duplicated because classes in `test` sourceset are not exposed to other modules. To fix that we need dedicated `test` module, where Fake classes will be in the `main` sourceset. This will allow other modules to add `testImplementation` dependency to such modules.

To be able to do that we need to split library and all the feature modules into 3 sub modules:

- 1 `api` module where interface are defined;
- 1 `test` module to put Fake classes and fixtures;
-  1 (or several) `impl` module(s) where (default) implementation are placed.

The main `app` module will have an `implementation` dependency to `api` and one of the `impl` for each part (= parent module) of the project

To avoid having too many lines in the project view of the IDE, we can put sub-module as subfolders of the current module folder.

## Pros

- Changing that will help to have faster compilation time, since if we change things in the `impl` modules, the other modules which only depends on `api` do not need to be built again;
- Easier to switch between different implementations of each module if necessary. Can be very useful for forks for instance to fully disable at compilation time a feature by replacing the default impl with a no op one. `api` should expose a method `fun isEnable()` or similar to be able to do that though;
- avoid duplicated Fake classes.

## Cons

- moar modules, can become a bit tricky to have a clear vision of the whole module hierarchy.

## Plan

- [x] do it for the `matrix` module since it's the most use dep across the app (#125)
- [x] do it for modules which have Fake classes which must be exposed to others (and so remove duplicated Fake classes)

## DOD

- [x] All `TODO Remove this duplicated class when we will rework modules.` have been handled

PR #173, opened 2023-03-07
Feature/fga/improve node architecture

ganfra:
This PR handles https://github.com/vector-im/element-x-android/issues/131

Related: #91


bmarty (approved):
LGTM, thanks for the huge work.
I have added 2 small commit. Also we have a mix between module referenced using `:` and module referenced using `project`, which may hard to find all the dependencies. We may harmonize later.

PR #179, opened 2023-03-07
Attempt to check org/group membership before danger complains about signoff.

michaelkaye:
Rather than maintaining an allow-list of users who have the ability to skip the signoff for other reasons (eg, contract with the company), we can check for membership of the internal group that matches.

This should allow us to reduce the size of the allow-list to just those unexpected cases, eg bots.

### DIFF
--- a/tools/danger/dangerfile.js
+++ b/tools/danger/dangerfile.js
@@ -88,7 +88,25 @@ const allowList = [
     "yostyle",
 ]
 
-const requiresSignOff = !allowList.includes(user)
+// We ignore signoff requirements for employees of Element
+// as copyright assignment is covered by our employee contracts.
+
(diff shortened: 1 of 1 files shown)


michaelkaye:
So: 04477ca checks for membership of the entire organisation and works - I've checked with users on the allow list, just in the org, and not in the org.

The most recent commit tightens it to just those in the vector-core group - however the ElementRobot user isn't a member of that so is unable to determine if other users are a member of it. 

Should we be using https://github.com/RiotRobot instead? We'd need to swap out the access token for that. I don't know if it has other permissions we should retain.

michaelkaye:
I believe this fundamentally works; but there's a question over the access permissions for it.

PR #182, opened 2023-03-07
Fix the case when a valid homeserver url can't be extracted from the MXID

jmartinesp:
## Changes

We now provide the actual homeserver url to the Rust SDK's  `ClientBuilder`.

## Steps to test:

1. Set the home server to `https://large-account.lab.element.dev`.
2. Log in with the test account credentials (ping me if you don't know them).

If you get past the log in screen (although the room list would take ages to load) it's working.

### DIFF
--- a/changelog.d/182.bugfix
+++ b/changelog.d/182.bugfix
@@ -1,0 +1,1 @@
+Fix the case when a valid homeserver url can't be extracted from the MXID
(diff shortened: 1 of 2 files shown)


ganfra:
Could you fix the knit issue in the same time? related to the status section added to the doc

Also, can you take the opportunity to release the rust objects from memory : 
- the `Client` in the `login` method 
- the `ClientBuilder` in the `restore` method

bmarty:
> Could you fix the knit issue in the same time? related to the status section added to the doc

Quality issue has been fixed on develop by d59ab524972c93fd295df8f828978bb70e68c3d7

Florian14 (approved):
Tested ok, code lgtm

PR #194, opened 2023-03-09
Fix nightly build

bmarty:
Upload tested OK locally, also installed the generate nightly the regular way and the app is not crashing anymore 🎉 

CC @michaelkaye for the change about `browserstack`.


jmartinesp (approved):
If the CI passes, LGTM!

Is towncrier necessary here though? Will it create the changelog used for the nightly version?

bmarty:
> Is towncrier necessary here though? Will it create the changelog used for the nightly version?

Yes, and it will be visible in the App Tester app and at other location

image

But ideally it would contains diff with the previous nightly...

PR #196, opened 2023-03-09
Address design comments on the change server screen

jmartinesp:
These changes should fix the design issues mentioned in https://github.com/vector-im/element-x-android/issues/101#issuecomment-1460232247.


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/Color.kt
```diff
@@ -68,4 +68,4 @@ val ElementGreen = Color(0xFF0DBD8B)
 val ElementOrange = Color(0xFFD9B072)
 val Vermilion = Color(0xFFFF5B55)
 
-val LinkColor = Color(0xFF054F6E)
+val LinkColor = Color(0xFF0086E6)
```
jmartinesp, line 71: This seems to be the new color in Figma.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/ClickableLinkText.kt
```diff
@@ -40,14 +41,15 @@ import io.element.android.libraries.designsystem.theme.components.Text
...
 fun ClickableLinkText(
     text: AnnotatedString,
-    linkAnnotationTag: String,
-    onClick: () -> Unit,
-    onLongClick: () -> Unit,
     interactionSource: MutableInteractionSource,
     modifier: Modifier = Modifier,
+    linkAnnotationTag: String = "",
+    onClick: () -> Unit = {},
+    onLongClick: () -> Unit = onClick,
```
jmartinesp, line 52: Here:
- I made `linkAnnotationTag` as the string annotation is `""` by default for URL annotations.
- `onClick` can be null if we're only using URL annotations, as they will automatically open their URLs as the default action.
- `onLongClick` behaves the same, but I thought we might want it to default to the same action as `onClick`. I can change it otherwise.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/dialogs/ConfirmationDialog.kt
```diff
@@ -39,18 +37,20 @@ import io.element.android.libraries.ui.strings.R as StringR
 fun ConfirmationDialog(
     title: String,
     content: String,
+    onSubmitClicked: () -> Unit,
+    onDismiss: () -> Unit,
```
jmartinesp, line 41: These 2 lambdas should never be null, as we will *always* want the visible buttons in some dialog to trigger an action.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/dialogs/ConfirmationDialog.kt
```diff
@@ -39,18 +37,20 @@ import io.element.android.libraries.ui.strings.R as StringR
...
     modifier: Modifier = Modifier,
     submitText: String = stringResource(id = StringR.string.ok),
     cancelText: String = stringResource(id = StringR.string.action_cancel),
     thirdButtonText: String? = null,
-    onSubmitClicked: () -> Unit = {},
-    onCancelClicked: () -> Unit = {},
+    emphasizeSubmitButton: Boolean = false,
```
jmartinesp, line 46: Added this to make the positive action 'bolder' than the rest, as per the Figma designs.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/dialogs/ConfirmationDialog.kt
```diff
@@ -39,18 +37,20 @@ import io.element.android.libraries.ui.strings.R as StringR
...
+    onCancelClicked: () -> Unit = onDismiss,
```
jmartinesp, line 47: I made it default to `onDismiss`, as I believed the same action should be triggered by default.

jmartinesp (commented):
A few comments to explain my changes.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/ClickableLinkText.kt
bmarty: > onLongClick behaves the same, but I thought we might want it to default to the same action as onClick. I can change it otherwise.

I would prefer `onLongClick` to have no effect by default. I think it is better UX.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/dialogs/ConfirmationDialog.kt
```diff
@@ -106,6 +113,9 @@ private fun ContentToPreview() {
     ConfirmationDialog(
         title = "Title",
         content = "Content",
-        thirdButtonText = "Disable"
+        thirdButtonText = "Disable",
+        onSubmitClicked = {},
+        onDismiss = {},
+        emphasizeSubmitButton = true,
```
bmarty, line 119: Maybe add another preview with `false`. Not sure this is mandatory though. We have a `BooleanProvider` if necessary.

bmarty (approved):
The close button color change is strange on https://github.com/vector-im/element-x-android/pull/196/files#diff-c31b73c53b1167244e6bb75ea742cbc471809e73427a8cba329d0c6487bbfb34

image

Same for the reveal pwd action

image

and the checkbox:

image

Note: this are just remarks, nothing blocking at this stage of the project (and I have not checked the Figma).

jmartinesp:
> The close button color change is strange on https://github.com/vector-im/element-x-android/pull/196/files#diff-c31b73c53b1167244e6bb75ea742cbc471809e73427a8cba329d0c6487bbfb34

I might have messed up that change. The active indicator line in TextFields should be `onSurfaceVariant` according to the MD guidelines, but on Figma it seems like it is `tertiary` instead (which should apparently be `Gray_300`, not White).

Maybe there's a clash of these colors in the components used, as we're still using a mix of Material Theme and the old compound colors?

PR #201, opened 2023-03-10
Improve ci

### LINKED ISSUE #121: Limit our usage of maestro cloud
To avoid running out of 100 runs per month in the free plan, we should avoid running maestro cloud each time a commit is pushed to an open pull request. In February we reached this limit before the end of month, and this will getting worse with more developers working on the project.

Not sure how to fix this properly. Either we:

- add a label to the pull request so that we trigger maestro run only if the label is present, but in this case every new commit will trigger a run (if we do not think about removing the label)

- manually trigger a maestro cloud run when the PR is ready to be merged, as a final step (how?).

What can be also done:
- [ ] ensure maestro is not run on Draft PR;
- [ ] ensure maestro is not run on PR touching only non-code files such as `md` files for instance.

### LINKED ISSUE #195: Make sure the CI compile the samples/minimal app.
Or it can get broken by mistake.

bmarty:
Closes #195 
Closes #121

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,3 +62,5 @@ jobs:
         run: ./gradlew compileReleaseSources $CI_GRADLE_ARG_PROPERTIES
       - name: Compile nightly sources
         run: ./gradlew compileNightlySources $CI_GRADLE_ARG_PROPERTIES
+      - name: Compile samples minimal
+        run: ./gradlew :samples:minimal:assemble $CI_GRADLE_ARG_PROPERTIES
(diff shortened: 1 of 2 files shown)


bmarty:
So once not approved, this is what we get:

image

Ideally we would see a pending check about maestro.

Let's see what's happening once the PR is approved.

mnaturel (approved):

bmarty:
Once approved:

image

Next month we will make the Maestro check `Required` again, and see what is happening during the time it is not run.

PR #209, opened 2023-03-13
Deactivate dependabot

bmarty:
We are now relying on Renovate. But Dependabot is tied to the project, and I do not want to fully switch the alerts off:

image

So just tweak the configuration file and hopefully it will be enough (we will now once the PR is merged)

Also try to skip the CI on this PR:

[ci skip]

^ does not seem to work (even if I have seen that in other PR on other project and would be very handy), it has to be in the commit message. https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,8 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    ignore:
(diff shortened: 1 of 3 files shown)


mnaturel (approved):
LGTM!

Issue #96, opened 2023-02-17
Show or create direct message room


alfogrillo:

Florian14:
blocked by https://github.com/matrix-org/matrix-rust-sdk/issues/1768

PR #211, opened 2023-03-13
[Start chat] Show a single result when searching for a matrixId

Florian14:
Content:

- The user can type a matrix identifier in the search field
- Exactly one user with the typed identifier will be shown under the search field
- The implementation of the search and get profile queries is not part of this ticket

Context:

#96 

Demo:

Note: Toasts are only here for the purpose of the demo 

[search_by_matrix_id.webm](https://user-images.githubusercontent.com/11990514/224756278-72e303f2-7562-4f2d-b9cd-90e9551feec9.webm)


features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootView.kt
```diff
@@ -123,12 +127,14 @@ fun CreateRoomRootViewTopBar(
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun CreateRoomSearchBar(
-    text: String,
+    query: String,
     placeHolderTitle: String,
+    results: List<MatrixUser>,
```
ElementBot, line 132: - :warning: The Compose Compiler cannot infer the stability of a parameter if a List<MatrixUser> is used in it, even if the item type is stable.
You should use Kotlinx Immutable Collections instead: `results: ImmutableList<MatrixUser>` or create an `@Immutable` wrapper for this class: `@Immutable data class ResultsList(val items: List<MatrixUser>)`

See https://mrmans0n.github.io/compose-rules/rules/#avoid-using-unstable-collections for more information.

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootPresenter.kt
```diff
@@ -17,23 +17,71 @@
...
             when (event) {
+                is CreateRoomRootEvents.UpdateSearchQuery -> searchQuery = event.query
+                is CreateRoomRootEvents.StartDM -> handleStartDM(event.matrixUser)
                 CreateRoomRootEvents.CreateRoom -> Unit // Todo Handle create room action
                 CreateRoomRootEvents.InvitePeople -> Unit // Todo Handle invite people action
             }
         }
 
+        LaunchedEffect(searchQuery) {
+            searchResults.value = if (MatrixPatterns.isUserId(searchQuery)) {
```
mnaturel, line 56: Should we remove the first assignement expression and only keep the following code?
```
searchResults.value = if (searchQuery.isNotEmpty()) {
    performSearch(searchQuery)
} else {
    persistentListOf()
}
```

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootPresenter.kt
```diff
@@ -17,23 +17,71 @@
...
     override fun present(): CreateRoomRootState {
+        var searchQuery by rememberSaveable { mutableStateOf("") }
+        val searchResults: MutableState<ImmutableList<MatrixUser>> = remember {
+            mutableStateOf(persistentListOf())
+        }
 
         fun handleEvents(event: CreateRoomRootEvents) {
             when (event) {
+                is CreateRoomRootEvents.UpdateSearchQuery -> searchQuery = event.query
+                is CreateRoomRootEvents.StartDM -> handleStartDM(event.matrixUser)
```
mnaturel, line 49: Just a thought, I think for the case of `StartDM`, we will not use a `ViewEvent`. I guess we would rather use a callback on the node entry point to navigate to the corresponding screen. I had to do this for the `CreateRoom` case for instance.

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootPresenter.kt
Florian14: Nope, the value is updated twice for this reason:

we want to update the search results as soon as the query has changed, if this is not a matrixId, we clear the list, otherwise, we display a "fake" result with this matrixId so the user would be able to start a chat with this matrixId.

Then, we perform the search query (this is an asynchronous operation that can take a few seconds) and we update the query results.

we'll also need some debounce for the search query but not on the first value update

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootPresenter.kt
Florian14: Yes, I had this in mind, this is a temporary code. Also, I don't know if we should keep a viewEvent and just trigger the node callback when handling this event?

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootStateProvider.kt
```diff
@@ -27,5 +30,7 @@ open class CreateRoomRootStateProvider : PreviewParameterProvider<CreateRoomRoot
 }
 
 fun aCreateRoomRootState() = CreateRoomRootState(
```
mnaturel, line 32: Maybe we we should also have state cases with:
- a search query field and and empty results
- an empty search query and empty results

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootView.kt
```diff
@@ -175,7 +182,14 @@ fun CreateRoomSearchBar(
         },
         shape = if (!active) SearchBarDefaults.dockedShape else SearchBarDefaults.fullScreenShape,
         colors = if (!active) SearchBarDefaults.colors() else SearchBarDefaults.colors(containerColor = Color.Transparent),
-        content = {},
+        content = {
+            results.forEach {
```
mnaturel, line 186: Do we have a known limit on the search? I wonder if we should use `LazyColumn` instead?

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootView.kt
```diff
@@ -199,6 +213,20 @@ fun CreateRoomActionButtonsList(
...
 }
 
+@Composable
+fun CreateRoomSearchResultItem(
+    matrixUser: MatrixUser,
+    modifier: Modifier = Modifier,
+    onClick: () -> Unit = {},
+) {
+    MatrixUserRow(
+        modifier = modifier.heightIn(min = 56.dp),
```
mnaturel, line 223: Should we use a fixed height instead? `modifier.height(56.dp)`

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/avatar/AvatarSize.kt
```diff
@@ -18,11 +18,15 @@ package io.element.android.libraries.designsystem.components.avatar
 
 import androidx.compose.ui.unit.dp
 
-enum class AvatarSize(val value: Int) {
-    SMALL(32),
-    MEDIUM(40),
-    BIG(48),
-    HUGE(96);
+sealed class AvatarSize(open val value: Int) {
```
mnaturel, line 21: What do you think of updating the type of the value to `Dp`? I think it would make things clearer when expecially using the `Custom` class.

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootStateProvider.kt
mnaturel: And maybe we should also move the `isSearchActive` inside the `CreateRoomRootState` to cover the different UI in the screenshots tests?

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootPresenter.kt
mnaturel: Okay I see, maybe we could add some comment to explain that, what do you think?

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootPresenter.kt
mnaturel: In my case I entirely remove the `CreateRoom` viewEvent and I have just used a callback.

ganfra: It depends if you have logic first before triggering the navigation

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/avatar/AvatarSize.kt
ganfra: Yeah the change makes sense

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootEvents.kt
```diff
@@ -16,7 +16,11 @@
 
 package io.element.android.features.createroom.impl.root
 
+import io.element.android.libraries.matrix.ui.model.MatrixUser
```
ganfra, line 19: A side note (not for you then) but we should probably put the MatrixUser inside the `matrix` module instead of `matrix.ui`, I find it weird to be seperated.

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootPresenter.kt
```diff
@@ -17,23 +17,71 @@
...
+
         return CreateRoomRootState(
-            eventSink = ::handleEvents
+            eventSink = ::handleEvents,
+            searchQuery = searchQuery,
+            searchResults = searchResults.value,
         )
     }
+
+    private fun performSearch(query: String): ImmutableList<MatrixUser> {
```
ganfra, line 73: All the logic around selecting/searching for a user could probably be in a different module as we'll certainly need this in other features

mnaturel: Yes we had that in mind, I can try to handle this point since I will implement multi selection of user reusing what we have in this PR.

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootStateProvider.kt
bmarty: Yes, aCreateRoomRootState() should return the "default" state, and then we can copy it (in the provider) to create more states.

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootView.kt
Florian14: By doing this, the text was cropped. I updated the component paddings for a better rendering and do not set the height anymore. Also, I would prefer to not set a fixed height to prevent the same kind of issue, especially if the text size is higher on the user's device.

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootStateProvider.kt
Florian14: done https://github.com/vector-im/element-x-android/pull/211/commits/9c0164e9bd7317495b8253b9be8fcc56ba90256d

bmarty (approved):
LGTM

PR #216, opened 2023-03-15
Fix behaviour of Enter and Tab hardware keys in TextFields

jmartinesp:
## Changes

Make sure to consume both key up and key down events for Enter and Tab hardware keys in TextFields using `Modifier.onTabOrEnterKeyFocusNext`.

## Why

If we only consumed the key down event as we were doing, the Tab and Enter characters were still being added to the TextFields even if the focus was moved to the next item in its parent. This is especially annoying when using the app in an emulator. 

However, If we consume both events no character is added.

## Tests

Using an emulator, complete the login form using only a keyboard and using the Enter key to go to the next field/button.

### DIFF
--- a/changelog.d/216.bugfix
+++ b/changelog.d/216.bugfix
@@ -1,0 +1,1 @@
+Make sure `Modifier.onTabOrEnterKeyFocusNext` doesn't add line breaks or tabs to TextFields.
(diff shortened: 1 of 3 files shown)


bmarty:
If this fixes the `Hello\` weird bug, that's great! Thanks!

mnaturel (approved):
LGTM!

PR #218, opened 2023-03-15
Introduce AppNavigationStateService.

bmarty:
Part of handling notification, we need a way to share the current state of the application, especially to know if a room is currently displayed or not.

Also introduce a `services` folder, to host libraries which have `api` and `impl` modules.

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/RoomFlowNode.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/RoomFlowNode.kt
@@ -35,6 +35,7 @@ import io.element.android.libraries.architecture.NodeInputs
 import io.element.android.libraries.architecture.inputs
 import io.element.android.libraries.di.SessionScope
 import io.element.android.libraries.matrix.api.room.MatrixRoom
+import io.element.android.services.appnavstate.api.AppNavigationStateService
 import kotlinx.parcelize.Parcelize
 import timber.log.Timber
 
@@ -43,6 +44,7 @@ class RoomFlowNode @AssistedInject constructor(
     @Assisted buildContext: BuildContext,
     @Assisted plugins: List<Plugin>,
     private val messagesEntryPoint: MessagesEntryPoint,
+    private val appNavigationStateService: AppNavigationStateService,
 ) : BackstackNode<RoomFlowNode.NavTarget>(
     backstack = BackStack(
         initialElement = NavTarget.Messages,
@@ -68,11 +70,13 @@ class RoomFlowNode @AssistedInject constructor(
             onCreate = {
                 Timber.v("OnCreate")
                 plugins<LifecycleCallback>().forEach { it.onFlowCreated(inputs.room) }
+                appNavigationStateService.onNavigateToRoom(inputs.room.roomId)
             },
             onDestroy = {
                 Timber.v("OnDestroy")
                 inputs.room.close()
                 plugins<LifecycleCallback>().forEach { it.onFlowReleased(inputs.room) }
+                appNavigationStateService.onLeavingRoom()
             }
         )
     }
--- a/libraries/core/src/main/kotlin/io/element/android/libraries/core/log/logger/LoggerTag.kt
+++ b/libraries/core/src/main/kotlin/io/element/android/libraries/core/log/logger/LoggerTag.kt
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2023 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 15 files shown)


services/appuistate/api/src/main/kotlin/io/element/android/services/appuistate/api/AppUiStateService.kt
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.services.appuistate.api
+
+import kotlinx.coroutines.flow.StateFlow
+
+interface AppUiStateService {
```
ganfra, line 21: Rename to AppNavigationState? found it more accurate. Same for the sealed interface.

services/appuistate/api/src/main/kotlin/io/element/android/services/appuistate/api/AppUiStateService.kt
```diff
@@ -0,0 +1,35 @@
...
+    val appUiStateFlow: StateFlow<AppUiState>
+
+    fun onNavigateToSession(sessionId: String)
```
ganfra, line 24: Use `SessionId ` ? and create `SpaceId` and `ThreadId` so we are all aligned

appnav/src/main/kotlin/io/element/android/appnav/RoomFlowNode.kt
```diff
@@ -68,11 +70,13 @@ class RoomFlowNode @AssistedInject constructor(
             onCreate = {
                 Timber.v("OnCreate")
                 plugins<LifecycleCallback>().forEach { it.onFlowCreated(inputs.room) }
+                appUiStateService.onNavigateToRoom(inputs.room.roomId.value)
```
ganfra, line 73: If you use the RoomFlowNode, it'll include the future room details, and not just the messages screen.

ganfra (commented):
Some remarks

appnav/src/main/kotlin/io/element/android/appnav/RoomFlowNode.kt
bmarty: I will change. Also I can see that when the app goes to BG when this screen is displayed, the onDestroy is not called, so this is not what we expect.

services/appuistate/api/src/main/kotlin/io/element/android/services/appuistate/api/AppUiStateService.kt
bmarty: It would require to add a dep on `: libraries:matrix:api`, do we really want to do that?

ganfra: Yes, why we wouldn't?

appnav/src/main/kotlin/io/element/android/appnav/RoomFlowNode.kt
ganfra: You can use `resume/pause` then

services/appuistate/api/src/main/kotlin/io/element/android/services/appuistate/api/AppUiStateService.kt
bmarty: Because `appnavstate` module does not really need a lot from this library. But OK...

appnav/src/main/kotlin/io/element/android/appnav/RoomFlowNode.kt
bmarty: But it will break when user will open a Thread...

bmarty: Actually, I will revert and think about another approach.

libraries/core/src/main/kotlin/io/element/android/libraries/core/log/logger/LoggerTag.kt
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Parent class for custom logger tags. Can be used with Timber :
+ *
+ * val loggerTag = LoggerTag("MyTag", LoggerTag.VOIP)
+ * Timber.tag(loggerTag.value).v("My log message")
+ */
+open class LoggerTag(name: String, parentTag: LoggerTag? = null) {
+
+    object SYNC : LoggerTag("SYNC")
```
ganfra, line 27: Should we remove all these? it's on the rust sdk side now

ganfra (approved):

PR #227, opened 2023-03-17
Add localazy.json

pixlwave:
Thought I'd open this up as an initial starting point for adding strings from Localazy based on what I've figured out for iOS integration.

With this setup in the project, you can install the [CLI](https://localazy.com/docs/cli/installation) and then run `localazy download` to fetch the latest strings.


bmarty:
Thanks, I will merge this PR and will create a new one to iterate.

PR #230, opened 2023-03-20
[Verify Session] Design fixes

jmartinesp:
## Changes:

- Use default colors for snackbar.
- Present the 'verification successful' snackbar for a longer time.
- Change text size for buttons to `16.sp`.
- Add more top vertical spacing en the verify session screen.

## Why

Fixes https://github.com/vector-im/element-x-android/issues/89#issuecomment-1473620720 and https://github.com/vector-im/element-x-android/issues/89#issuecomment-1473851299.


jmartinesp:
> * The distance between the icon at the top of each page and the top of the screen needs to be increased to match the design - can you set it to 132px from the top of the screen?

I can't really set `132px` as a hardcoded value to the top, as I don't have control of the size of the status bar in each device, but I just realised I messed up by a few pixels, I had 76px to the status bar and it should have been 80px (16px to the top app bar + 64px of top app bar height, although there's none in this screen).

> * I have issues starting the emoji verification - sometimes after accepting the request on my other device nothing happens, and I have to reset the flow and go through it again for the emoji verification to appear

I sometimes see that too, I'm not sure if it's just some undelivered message between the devices, to be honest.

> * Sometimes after confirming the emojis match, the flow is stuck on the loading screen

That never happened to me, but it could also be caused by either some request being dropped or some fault in the logic for transitioning between states in this screen. Could you rageshake next time this happens? I'll try to reproduce it too.

> * The snackbar shows for too long now - could you reduce it by 2 seconds?

Not really, the only options I have are `Duration.Short`,  which was the previous duration, `Duration.Long`, the current one and `Duration.Indefinite` for non-dismissable snackbars.

ganfra:
> Not really, the only options I have are `Duration.Short`, which was the previous duration, `Duration.Long`, the current one and `Duration.Indefinite` for non-dismissable snackbars.

Actually we could manage this by using Duration.Indefinite in the ui and managing the duration in the Presenter

jmartinesp:
I have uploaded some fixes (top padding, font size in the emoji items, font size in the cancel and don't match buttons), and also some code to be able to read the state of the verification process more easily, so next time you rageshake we have some better logs.

callumu:
@jmartinesp 

80 px to to the top of the status bar is good - can you set it as if the status bar exists?

> I sometimes see that too, I'm not sure if it's just some undelivered message between the devices, to be honest.

Okay sure it could just be this - I occasionally get this issue with the legacy app also

In that case could you change it back to `Duration.Short`? I didn't realise that the alternative was very long

jmartinesp:
> @jmartinesp
> 
> 80 px to to the top of the status bar is good - can you set it as if the status bar exists?
> 
> > I sometimes see that too, I'm not sure if it's just some undelivered message between the devices, to be honest.
> 
> Okay sure it could just be this - I occasionally get this issue with the legacy app also
> 
> In that case could you change it back to `Duration.Short`? I didn't realise that the alternative was very long

All this should be fixed in the latest apk.

tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_3,NEXUS_5,1.0,en].png
Comment (Florian14) [inline tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_3,NEXUS_5,1.0,en].png:1]: With the new padding, we don't see anymore the labels on the second row, I suppose we can scroll but this is not ideal imho, wdyt @callumu?

libraries/core/build.gradle.kts
```diff
@@ -30,6 +30,7 @@ java {
 dependencies {
     implementation(libs.coroutines.core)
     implementation(platform(libs.network.okhttp.bom))
+    implementation(libs.kotlin.reflect)
```
Florian14, line 33: Why do you need this lib? I'm not sure you're using it and I'm not sure we want to have it in the project (it will increase the size of the APK)

Florian14:
@jmartinesp was this merge intentional? It includes changes that do not seem related to this PR:

image

tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_3,NEXUS_5,1.0,en].png
Comment (callumu) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_3,NEXUS_5,1.0,en].png:1]: Could you add a screenshot so I can see what you mean?

libraries/core/build.gradle.kts
jmartinesp: > @jmartinesp was this merge intentional? It includes changes that do not seem related to this PR:
> 
> image

I added it so @callumu  could share the logs containing info about the `StateMachine` for the issues he mentioned,  but I should have reverted that yesterday. I'll do that now.

tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_3,NEXUS_5,1.0,en].png
Comment (jmartinesp) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_3,NEXUS_5,1.0,en].png:1]: We could either:
- Make the bottom buttons scroll with the rest of the contents too.
- Reduce the paddings for smaller display sizes. The device used for the screenshots is a Nexus 5, with 4.95" display and 1080x1794 resolution `@3x` (xxhdpi) if I'm not mistaken.

Comment (callumu) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_3,NEXUS_5,1.0,en].png:1]: Yes this isn't ideal. Is this for all screen sizes, or only for the small ones?

Comment (jmartinesp) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_3,NEXUS_5,1.0,en].png:1]: I can't see the issue on a Pixel 2 emulator, but it is there in the Nexus 5, which is relatively small for modern devices, but not *that* small:

<details><summary>Screenshot</summary>

</details>

Comment (callumu) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_3,NEXUS_5,1.0,en].png:1]: This looks much better, but does it affect the bigger screens at all?

Comment (jmartinesp) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_3,NEXUS_5,1.0,en].png:1]: It does 😢 . For some reason, it's always taking the minimum size.

Comment (callumu) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_3,NEXUS_5,1.0,en].png:1]: Noooooo. Can we scale the size of the emojis instead?

Comment (jmartinesp) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_3,NEXUS_5,1.0,en].png:1]: Actually, I think I got it. It won't be as easy as with the `heightIn` I mentioned before, but I can reduce the paddings if I detect that the screen height is smaller than 720dp. Alternatively, we can also make the emojis smaller, but since they're the core part of the verification process, I think it's better to leave them as they are?

Comment (callumu) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_3,NEXUS_5,1.0,en].png:1]: I think the solution you suggested is good, thanks!

features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/VerifySelfSessionView.kt
```diff
@@ -293,3 +298,15 @@ private fun ContentToPreview(state: VerifySelfSessionState) {
         goBack = {},
     )
 }
+
+@Composable
+private fun Modifier.shrinkableHeight(
```
ElementBot, line 303: - :warning: Modifier factory functions should not be marked as @Composable, and should use composed instead
- :warning: Modifier factory functions should not be marked as @Composable, and should use composed instead
- :warning: Modifier factory functions should not be marked as @Composable, and should use composed instead
- :warning: Using @Composable builder functions for modifiers is not recommended, as they cause unnecessary recompositions.
You should use Modifier.composed { ... } instead, as it limits recomposition to just the modifier instance, rather than the whole function tree.

See https://mrmans0n.github.io/compose-rules/rules/#avoid-modifier-extension-factory-functions for more information.

tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_0,NEXUS_5,1.0,en].png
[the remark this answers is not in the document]
Comment (Florian14) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_0,NEXUS_5,1.0,en].png:1]: ... and also on the following screens 😬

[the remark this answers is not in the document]
Comment (jmartinesp) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_0,NEXUS_5,1.0,en].png:1]: I don't think it's broken, I made an special case for devices with small displays so everything would fit (see https://github.com/vector-im/element-x-android/pull/230#discussion_r1143776903) and the Nexus 5 used for screenshots matches this.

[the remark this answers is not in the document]
Comment (jmartinesp) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_0,NEXUS_5,1.0,en].png:1]: The improvement can be seen in [some other screenshots](https://github.com/vector-im/element-x-android/pull/230/files/a4315acd92986d4c9c4c8764978bb0fb1a47554b#diff-f614aeb622883c22e1aa08f3b33206d43a779f47628ae6e2936fec686aeec1b1).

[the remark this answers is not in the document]
Comment (jmartinesp) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_0,NEXUS_5,1.0,en].png:1]: It's not ideal, but I believe it's the only way to make everything fit without scrolling...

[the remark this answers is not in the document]
Comment (Florian14) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_0,NEXUS_5,1.0,en].png:1]: Just did some tests with Nexus 5 preview, I restored the fixed paddings for the top and the bottom (80dp / 40dp) and kept the shrinkableHeight for the content padding, the result seems good to me:
image

wdyt?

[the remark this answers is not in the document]
Comment (Florian14) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_0,NEXUS_5,1.0,en].png:1]: I also tested to set the top & bottom Spacers of the content with `heightIn(max = 56.dp)` (without a min), it seems also ok for bigger screens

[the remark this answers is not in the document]
Comment (Florian14) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.verifysession.impl_null_DefaultGroup_VerifySelfSessionViewDarkPreview_0_null_0,NEXUS_5,1.0,en].png:1]: > Those look way better! There might be some issues with devices with smaller width as the content might become scrollable, but I'm not even sure we should really care too much about those. Could you upload your changes?

Yes, I think it's better to keep the title & the buttons at a fixed position for consistency with the other screens

Florian14 (approved):
LGTM, feel free to merge if my changes are ok for you :)

PR #231, opened 2023-03-20
Localazy

bmarty:
Following #227 , update the README files and remove automation to sync strings from EA.


pixlwave (approved):
LGTM 👍

jmartinesp (approved):
LGTM too. Just a question, since we deleted the GH workflow: are we going to manually import these strings from now on?

bmarty:
> LGTM too. Just a question, since we deleted the GH workflow: are we going to manually import these strings from now on?

Strings will have to be imported, only the EAx ones. But thy will need to follow the naming rules (I am checking this), so some change will be required on the codebase I guess.

PR #235, opened 2023-03-20
Improve `ButtonWithProgress` component, replace login and change server buttons

jmartinesp:
## What
- Modifies `ButtonWithProgress` (maybe we could rename it to `ProgressButton`?) so it takes a `text` and `progressText` parameters to set the text to display and a `showProgress` one to decide whether to show the text and the progress indicator or not.
- Also modifies login and server selection screens to use this component instead of the floating progress indicator.

## Test

1. Open either the login or server selection screens.
2. Submit their forms.


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/button/ButtonWithProgress.kt
```diff
@@ -43,6 +44,7 @@ import io.element.android.libraries.designsystem.theme.components.Text
 @Composable
 fun ButtonWithProgress(
```
ElementBot, line 45: - :warning: Parameters in a composable function should be ordered following this pattern: params without defaults, modifiers, params with defaults and optionally, a trailing function that might not have a default param.
Current params are: [onClick: () -> Unit, showProgress: Boolean = false, modifier: Modifier = Modifier, enabled: Boolean = true, shape: Shape = ElementButtonDefaults.shape, colors: ButtonColors = ElementButtonDefaults.buttonColors(), elevation: ButtonElevation? = ElementButtonDefaults.buttonElevation(), border: BorderStroke? = null, contentPadding: PaddingValues = ElementButtonDefaults.ContentPadding, interactionSource: MutableInteractionSource = remember { MutableInteractionSource() }, text: String?, progressText: String? = text] but should be [onClick: () -> Unit, text: String?, modifier: Modifier = Modifier, showProgress: Boolean = false, enabled: Boolean = true, shape: Shape = ElementButtonDefaults.shape, colors: ButtonColors = ElementButtonDefaults.buttonColors(), elevation: ButtonElevation? = ElementButtonDefaults.buttonElevation(), border: BorderStroke? = null, contentPadding: PaddingValues = ElementButtonDefaults.ContentPadding, interactionSource: MutableInteractionSource = remember { MutableInteractionSource() }, progressText: String? = text].

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/button/ButtonWithProgress.kt
```diff
@@ -86,7 +97,5 @@ internal fun ButtonWithProgressDarkPreview() = ElementPreviewDark { ContentToPre
 
 @Composable
 private fun ContentToPreview() {
-    ButtonWithProgress(onClick = {}) {
-        Text(text = "Button with progress")
-    }
+    ButtonWithProgress(text = "Button with progress", onClick = {})
```
bmarty, line 100: Maybe set `showProgress = true` for a better the preview

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/button/ButtonWithProgress.kt
```diff
@@ -44,17 +45,21 @@ import io.element.android.libraries.designsystem.theme.components.Text
 fun ButtonWithProgress(
```
ElementBot, line 45: - :warning: Parameters in a composable function should be ordered following this pattern: params without defaults, modifiers, params with defaults and optionally, a trailing function that might not have a default param.
Current params are: [onClick: () -> Unit, modifier: Modifier = Modifier, showProgress: Boolean = false, enabled: Boolean = true, shape: Shape = ElementButtonDefaults.shape, colors: ButtonColors = ElementButtonDefaults.buttonColors(), elevation: ButtonElevation? = ElementButtonDefaults.buttonElevation(), border: BorderStroke? = null, contentPadding: PaddingValues = ElementButtonDefaults.ContentPadding, interactionSource: MutableInteractionSource = remember { MutableInteractionSource() }, text: String?, progressText: String? = text] but should be [onClick: () -> Unit, text: String?, modifier: Modifier = Modifier, showProgress: Boolean = false, enabled: Boolean = true, shape: Shape = ElementButtonDefaults.shape, colors: ButtonColors = ElementButtonDefaults.buttonColors(), elevation: ButtonElevation? = ElementButtonDefaults.buttonElevation(), border: BorderStroke? = null, contentPadding: PaddingValues = ElementButtonDefaults.ContentPadding, interactionSource: MutableInteractionSource = remember { MutableInteractionSource() }, progressText: String? = text].

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/changeserver/ChangeServerView.kt
```diff
@@ -114,7 +112,7 @@ fun ChangeServerView(
         topBar = {
             TopAppBar(
                 title = {},
-                navigationIcon = { BackButton(onClick = onBackPressed, enabled = interactionEnabled) }
+                navigationIcon = { BackButton(onClick = onBackPressed, enabled = !isLoading) }
```
bmarty, line 115: I do not really like to block the user from going back, even if there is a loading in progress... I think the user should always be able to go back and in this case we should cancel the action.

But this is not new with this PR, so not blocking here.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/root/LoginRootView.kt
```diff
@@ -318,15 +311,15 @@ internal fun LoginForm(
         Spacer(Modifier.height(28.dp))
 
         // Submit
-        Button(
+        ButtonWithProgress(
+            text = stringResource(R.string.screen_login_submit),
+            showProgress = isLoading,
             onClick = ::submit,
-            enabled = interactionEnabled && state.submitEnabled,
+            enabled = state.submitEnabled || isLoading,
```
bmarty, line 318: Same remark about enabled button state.

features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/VerifySelfSessionView.kt
```diff
@@ -257,21 +256,12 @@ internal fun BottomMenu(screenState: VerifySelfSessionState, goBack: () -> Unit)
...
-            ) {
-                positiveButtonTitle?.let { Text(stringResource(it), style = ElementTextStyles.Button) }
-            }
-        }
+        ButtonWithProgress(
+            text = positiveButtonTitle?.let { stringResource(it) },
+            showProgress = isVerifying,
+            modifier = Modifier.fillMaxWidth(),
+            onClick = { positiveButtonEvent?.let { eventSink(it) } }
+        )
```
bmarty, line 264: Definitely better!

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/button/ButtonWithProgress.kt
```diff
@@ -42,19 +43,23 @@ import io.element.android.libraries.designsystem.theme.components.Text
 
 @Composable
 fun ButtonWithProgress(
+    text: String?,
     onClick: () -> Unit,
     modifier: Modifier = Modifier,
+    showProgress: Boolean = false,
```
bmarty, line 49: Maybe document that when `showProgress` is set to true, the lambda `onClick` will not be invoked?

bmarty (commented):
Thanks! The API for `ButtonWithProgress` is better, I have some remarks though.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/changeserver/ChangeServerView.kt
jmartinesp: You're right, I'll remove it.

jmartinesp:
Those remarks should be fixed now in https://github.com/vector-im/element-x-android/pull/235/commits/4f26118988b9f2a2c8f283d972d183723e6df87a and https://github.com/vector-im/element-x-android/pull/235/commits/eb8f515e5831ca6b6b9a57f3c7f0e9d72f498c29.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/changeserver/ChangeServerState.kt
```diff
@@ -23,5 +23,5 @@ data class ChangeServerState(
     val changeServerAction: Async<Unit>,
     val eventSink: (ChangeServerEvents) -> Unit,
 ) {
-    val submitEnabled = homeserver.isNotEmpty() && changeServerAction is Async.Uninitialized
+    val submitEnabled: Boolean get() = homeserver.isNotEmpty() && (changeServerAction is Async.Uninitialized || changeServerAction is Async.Loading)
```
bmarty, line 26: Should be enabled when `changeServerAction is Async.Failure` too. So in fact I think the test can be removed, and it should be simply:

```suggestion
    val submitEnabled: Boolean get() = homeserver.isNotEmpty()
```

WDYT?

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/root/LoginRootState.kt
```diff
@@ -27,8 +27,8 @@ data class LoginRootState(
     val formState: LoginFormState,
     val eventSink: (LoginRootEvents) -> Unit
 ) {
-    val submitEnabled =
-        formState.login.isNotEmpty() && formState.password.isNotEmpty() && loggedInState != LoggedInState.LoggingIn
+    val submitEnabled: Boolean get() =
+        formState.login.isNotEmpty() && formState.password.isNotEmpty() && loggedInState !is LoggedInState.ErrorLoggingIn
```
bmarty, line 31: In case of error, the button has to be enabled, no? To let the user retry? So maybe simplify with:

```suggestion
        formState.login.isNotEmpty() && formState.password.isNotEmpty()
```

bmarty (commented):
More remarks (sorry!)

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/root/LoginRootState.kt
jmartinesp: Not really, when the error is received an alert dialog is displayed, and that error is cleared as soon as the dialog disappears. Technically, it's not needed to disabled the button since the alert will be blocking the UI, but it seemed to be clearer that way that you couldn't advance until the error was accepted.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/changeserver/ChangeServerState.kt
jmartinesp: Similar to the comment below, in this case the retry is blocked and the error is displayed until the user changes the info in the text field, either adding or removing some text. As soon as that happens, the error is cleared and the user can retry.

bmarty (approved):
Thanks!

PR #236, opened 2023-03-21
Make `ChangeServerPresenter.submit` fail as expected

jmartinesp:
`AuthenticationService.setHomeserver` now returns a `Result<Unit>`, so it wasn't throwing an error as we previously expected in `ChangeServerPresenter.submit`. This caused any change in homeserver to be reported to the presenter as successful, even when it wasn't.


Florian14 (approved):
SGTM

PR #241, opened 2023-03-22
[Verify Session] Add missing 'start verification' screen.

jmartinesp:
Also fix retries sometimes not working because of `SessionVerificationController.setDelegate` being called several times.


bmarty (approved):
LGTM. You can merge, I will handle the new added string (move it to localazy).

PR #271, opened 2023-03-29
[Session Verification] Set a fixed number of emojis per row

jmartinesp:
## What

Modifies the layout of the session verification flow so instead of having as much emojis as can fit in each row, we always have N rows of up to 4 emojis each. This should make the layout always match those of other platforms.

## Why

See https://github.com/vector-im/element-x-android/issues/89#issuecomment-1486784809.

## Test

Start a session verification, advance until the emojis are displayed.


ganfra (approved):
LGTM thanks

PR #273, opened 2023-03-29
Sliding Sync - Use the AllRooms window with growing_full_sync mode

Florian14:
### Context:

Unlike [Element X iOS](https://github.com/vector-im/element-x-ios/blob/e15b36b916100ae01e7a8eaea862679c094c398b/ElementX/Sources/Services/Client/ClientProxy.swift#L292-L307), we didn't implement the SS strategy with multiple views, there is an [issue](https://github.com/vector-im/element-x-android/issues/143) to track this missing feature. This logic should be moved to the [rust-sdk](https://github.com/matrix-org/matrix-rust-sdk/issues/1644), we're waiting for this implementation to integrate it into Element X Android.

It seems that the account data is not synced with the current strategy ("CurrentlyVisibleRooms" window with `selective` sliding sync mode).
Because of that, we cannot fetch existing DMs in the account data, even worse, any change in the account data will erase the existing data.

### Content:

This PR switches the "CurrentlyVisibleRooms" window with the "AllRooms" window which is an additional window on the iOS side. This change is afecting the performance of the app, we will no longer have the advantages of the previous mode but this change ensures that we have the correct data and prevent breaking them.

Fortunately, this is a temporary workaround until the multiple windows are supported in the sdk.

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
@@ -83,29 +83,46 @@ class RustMatrixClient constructor(
         )
     }
 
-    private val visibleRoomsSlidingSyncList = SlidingSyncListBuilder()
-        .timelineLimit(limit = 1u)
+    // The account data is not synchronized with the VisibleRooms window, we have to use the AllRooms window in the meantime
+    // We're waiting for the multiple windows support in the rust sdk to remove this workaround
+//    private val visibleRoomsSlidingSyncList = SlidingSyncListBuilder()
+//        .timelineLimit(limit = 1u)
+//        .requiredState(
+//            requiredState = listOf(
+//                RequiredState(key = "m.room.avatar", value = ""),
+//                RequiredState(key = "m.room.encryption", value = ""),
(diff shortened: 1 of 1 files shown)


jmartinesp (approved):
LGTM. It's true that the performance is a bit worse, and I also saw some of the 'dupe rooms' issues that iOS was mentioning recently. Hopefully, by the time it's integrated in the Rust SDK it might be fixed.

Florian14:
Force pushed to retrigger the CI

Florian14:
After some tests at the top of #272, this is still not ideal. It takes time to receive the DM room after the creation. If I leave the room, it takes time to receive the updated account data. So If I start a new DM a few seconds after leaving, I open the previous room... The perf is really bad, I tested on an account with no joined rooms and a few left ones... I expected to get at least good performances on a (very) small account :/

ganfra:
Can we disable DM creation for now, waiting to have a proper SS? As long as the feature is ready, I'd say it's ok, but maybe not my call.

Florian14:
> Can we disable DM creation for now, waiting to have a proper SS? As long as the feature is ready, I'd say it's ok, but maybe not my call.

Yeah, will do... Finally, I don't think it's a SS issue but more a rust-sdk issue, let's close this PR and disable the start DM flow while there are issues with the account data

PR #279, opened 2023-03-30
Makes MatrixClient Closeable outside of logout function.

ganfra:


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
```diff
@@ -179,18 +179,18 @@ class RustMatrixClient constructor(
...
-        close()
         try {
             client.logout()
         } catch (failure: Throwable) {
             Timber.e(failure, "Fail to call logout on HS. Still delete local files.")
         }
         baseDirectory.deleteSessionDirectory(userID = client.userId())
         sessionStore.removeSession(client.userId())
-        client.destroy()
+        close()
```
bmarty, line 193: Not blocking (and not new), but is it the responsibility to the class to close itself?

bmarty (approved):
LGTM, thanks, just a small question.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
ganfra: I'd say when it's a logout, yes, but both are valid options

ganfra:
I don't understand the code quality check failure

bmarty:
> I don't understand the code quality check failure

You can ignore for now.

ganfra:
So I can merge?

PR #294, opened 2023-04-04
Disable nightly workflow for forks

jmartinesp:
Last night the nightly workflow tried to run in my fork and failed since it's missing some secrets. Also, it's something we don't want to happen at all in forks.


bmarty (approved):
Thanks!

PR #297, opened 2023-04-04
Feat: show invites list

csmith:
Show a list of invites with accept/deny buttons

https://user-images.githubusercontent.com/189133/232783682-52369648-10af-44ca-b9e8-2819d16787aa.mp4


features/invitelist/api/src/main/kotlin/io/element/android/features/invitelist/api/InviteListEntryPoint.kt
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package io.element.android.features.invitelist.api
+
+import com.bumble.appyx.core.modality.BuildContext
+import com.bumble.appyx.core.node.Node
+import com.bumble.appyx.core.plugin.Plugin
+import io.element.android.libraries.architecture.FeatureEntryPoint
+
+interface InviteListEntryPoint : FeatureEntryPoint {
```
jmartinesp, line 24: You don't need to change it, but this could just extend `SimpleFeatureEntryPoint` and have a simpler implementation.

features/invitelist/impl/src/main/kotlin/io/element/android/features/invitelist/impl/components/InviteSummaryRow.kt
```diff
@@ -0,0 +1,202 @@
...
+        text = buildAnnotatedString {
+            val placeholder = "$"
+            val text = stringResource(R.string.screen_invites_invited_you, placeholder)
+            val nameIndex = text.indexOf(placeholder)
+
+            // Text before the placeholder
+            append(text.take(nameIndex))
+
+            // Avatar and display name
+            appendInlineContent("avatar")
```
jmartinesp, line 167: I would have just used a Row here, as it seems easier. Is there any reason why you chose this instead?

jmartinesp (approved):
The code LGTM, just a couple of comments. It seems like you need to solve some issues with `knit` in the README and broken tests, too.

Also, I noticed the buttons are quite bigger than in the designs. You'll probably want to either make them smaller or ask the designers to change the designs if this is not possible.

features/invitelist/api/src/main/kotlin/io/element/android/features/invitelist/api/InviteListEntryPoint.kt
csmith: Oh, nice. Will swap to that when I fix the tests tomorrow :+1:

features/invitelist/impl/src/main/kotlin/io/element/android/features/invitelist/impl/components/InviteSummaryRow.kt
csmith: I'm not sure we can rely on the displayname always being at the start of the translated string, which is why there's the mess with the placeholder and splitting the string up (the English is currently "%1$s invited you" but I don't see why some other language with odd grammar couldn't end up as "xxx %1$s yyy" or "xxx yyy %1$s").

At that point it could be a row of something like [Text?] [Avatar] [Display name] [Text?] but the optional texts made it feel a bit fiddly, and I'm not sure how well that'd work for screen readers and the like if one short sentence is spread across many UI elements :thinking:

jmartinesp: Oh, I thought the avatar had to be next to the message, I didn't realise it had to be next to the user's display name as in a mention. I understand why we needed the extra complexity here then.

features/invitelist/api/src/main/kotlin/io/element/android/features/invitelist/api/InviteListEntryPoint.kt
csmith: Ah, seems like that wouldn't work because of the callback? :thinking:

csmith:
> It seems like you need to solve some issues with knit in the README and broken tests, too.

I made the fatal mistake of not rerunning the tests after resolving a merge conflict. Fixed. :joy:

> Also, I noticed the buttons are quite bigger than in the designs.

Good spot! Fixed.

Issue #300, opened 2023-04-05
[Room Details] Add room member details screen


jmartinesp:
Implement a room member screen to be accessed when you select a user in the room member list screens.

PR #302, opened 2023-04-05
[Room Details] Implement member details screen

jmartinesp:
## What

Implement room member details view, displaying basic user info and a share user action.

## Why

Implements #300 .

## Test

Open a room, open its settings by tapping in the room name, select 'People' option, then select any user.


features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/details/RoomMemberDetailsView.kt
```diff
@@ -0,0 +1,177 @@
...
+import io.element.android.libraries.designsystem.preview.ElementPreviewLight
+import io.element.android.libraries.designsystem.theme.LocalColors
+import io.element.android.libraries.designsystem.theme.components.Scaffold
+import io.element.android.libraries.designsystem.theme.components.Text
+import io.element.android.libraries.designsystem.theme.components.TopAppBar
+import io.element.android.libraries.ui.strings.R as StringR
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun RoomMemberDetailsView(
```
ElementBot, line 60: - :warning: Parameters in a composable function should be ordered following this pattern: params without defaults, modifiers, params with defaults and optionally, a trailing function that might not have a default param.
Current params are: [state: RoomMemberDetailsState, modifier: Modifier = Modifier, onShareUser: () -> Unit, goBack: () -> Unit] but should be [state: RoomMemberDetailsState, onShareUser: () -> Unit, modifier: Modifier = Modifier, goBack: () -> Unit].

Damaj301damaj-lol:
Screenshot_20230406-091102_ElementX dbg
Its great and all, however small suggestion: it was never called "block user" on element, it is always called "ignore user"

Damaj301damaj-lol:
Screenshot_20230406-091253_Element
On element android

Aquathing:
> Screenshot_20230406-091102_ElementX dbg Its great and all, however small suggestion: it was never called "block user" on element, it is always called "ignore user"

I think "Block user" is better as it is way more familiar to users not coming from Element

Damaj301damaj-lol:
> 
> > Its great and all, however small suggestion: it was never called "block user" on element, it is always called "ignore user"
> 
> I think "Block user" is better as it is way more familiar to users not coming from Element

Well how about users that already use element? I don't really know the demographic of element-x so i can't answer myself

Aquathing:
> > 
> > > Its great and all, however small suggestion: it was never called "block user" on element, it is always called "ignore user"
> > 
> > I think "Block user" is better as it is way more familiar to users not coming from Element
> 
> Well how about users that already use element? I don't really know the demographic of element-x so i can't answer myself 

Users already using Element will have no trouble understanding what "Block user" means

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListNode.kt
```diff
@@ -21,22 +21,33 @@ import androidx.compose.ui.Modifier
...
 
 @ContributesNode(RoomScope::class)
 class RoomMemberListNode @AssistedInject constructor(
     @Assisted buildContext: BuildContext,
     @Assisted plugins: List<Plugin>,
+    private val room: MatrixRoom,
     private val presenter: RoomMemberListPresenter,
 ) : Node(buildContext, plugins = plugins) {
 
+    private val callback = plugins<RoomDetailsFlowNode.Callback>().first()
```
ganfra, line 42: The callback should be defined at this node level, not in the FlowNode

ganfra (approved):
Just one small remark, otherwise LGTM!

Issue #312, opened 2023-04-11
[Room details] Add DM sections to room details


jmartinesp:
Display custom sections in the room details screen when the room is considered a DM with another user.

PR #313, opened 2023-04-11
[Room details] DM designs

jmartinesp:
## What 

Implement DM designs for rooms that we consider DMs (rooms with 2 users, direct and encrypted).

This PR is based on #302, so it must be merged later. The important commit is  46a8697a3ac5e84ed41ef549778598ee6d5d9e6f.

Changes:
- Use different header section component.
- Hide 'People' section.
- Replace 'leave' action with 'block user'.

*Note: block user does nothing at the moment*.

## Why

Implements #312 .

## Designs:

image


features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/RoomDetailsView.kt
```diff
@@ -80,33 +85,50 @@ fun RoomDetailsView(
...
+                    roomAlias = state.roomAlias
+                )
+                RoomShareSection(onShareUser = onShareRoom)
+            } else {
+                RoomMemberHeaderSection(
+                    avatarUrl = state.roomAvatarUrl ?: state.dmMember.avatarUrl,
+                    userId = state.dmMember.userId,
+                    userName = state.roomName
+                )
+                RoomMemberShareSection(onShareUser = { onShareMember(state.dmMember) })
```
ganfra, line 102: Avoid lambda to avoid recomposition

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/RoomDetailsView.kt
```diff
@@ -80,33 +85,50 @@ fun RoomDetailsView(
...
         ) {
-            HeaderSection(
-                avatarUrl = state.roomAvatarUrl,
-                roomId = state.roomId,
-                roomName = state.roomName,
-                roomAlias = state.roomAlias
-            )
-
-            ShareSection(onShareRoom = onShareRoom)
+            if (state.dmMember == null) {
```
ganfra, line 88: I'd add a computed property in the state instead of checking nullability in the view, like `val showAsDM: Boolean`

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/RoomDetailsView.kt
```diff
@@ -80,33 +85,50 @@ fun RoomDetailsView(
...
                 TopicSection(roomTopic = state.roomTopic)
             }
 
-            val memberCount = (state.memberCount as? Async.Success<Int>)?.state
-            MembersSection(
-                memberCount = memberCount,
-                isLoading = state.memberCount.isLoading(),
-                openRoomMemberList = openRoomMemberList
-            )
+            if (state.dmMember == null) {
```
ganfra, line 109: Introducing a sealed interface to avoid checking on nullability would probably be cleaner

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListNode.kt
```diff
@@ -21,22 +21,33 @@ import androidx.compose.ui.Modifier
...
 
 @ContributesNode(RoomScope::class)
 class RoomMemberListNode @AssistedInject constructor(
     @Assisted buildContext: BuildContext,
     @Assisted plugins: List<Plugin>,
+    private val room: MatrixRoom,
     private val presenter: RoomMemberListPresenter,
 ) : Node(buildContext, plugins = plugins) {
 
+    private val callback = plugins<RoomDetailsFlowNode.Callback>().first()
```
ganfra, line 42: Same than other PR, the Callback should be defined in this node, not the parent one

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/details/RoomMemberDetailsNode.kt
```diff
@@ -0,0 +1,80 @@
...
+
+    @Composable
+    override fun View(modifier: Modifier) {
+        val context = LocalContext.current
+        val state = presenter.present()
+        RoomMemberDetailsView(
+            state = state,
+            modifier = modifier,
+            goBack = { navigateUp() },
+            onShareUser = { onShareUser(context) }
```
ganfra, line 77: use method references instead of lambdas?

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/MatrixRoom.kt
```diff
@@ -32,12 +33,17 @@ interface MatrixRoom: Closeable {
...
+    val isDirect: Boolean
     val isPublic: Boolean
 
     suspend fun members() : List<RoomMember>
 
     suspend fun memberCount(): Int
 
+    fun getMember(userId: UserId): RoomMember?
+
+    fun getDmMember(currentUserId: UserId): RoomMember?
```
ganfra, line 45: the current user is already known by the sdk, so you shouldn't need to pass it

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/RoomMember.kt
```diff
@@ -16,16 +16,22 @@
 
 package io.element.android.libraries.matrix.api.room
 
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
```
ganfra, line 22: Why do we need it to be Parcelable?

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/RoomMember.kt
```diff
@@ -16,16 +16,22 @@
...
     val powerLevel: Long,
-    val normalizedPowerLevel: Long
-)
+    val normalizedPowerLevel: Long,
+    val isIgnored: Boolean,
+) : Parcelable
 
-enum class RoomMembershipState {
+@Parcelize
+enum class RoomMembershipState : Parcelable {
```
ganfra, line 35: An enum is serializable by default so we don't need to add parcelable

ganfra (approved):
Added some small remarks

bmarty (commented):
Nice work, my 2 cents.

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/RoomDetailsView.kt
jmartinesp: This `dmMember` comes from the presenter, so I need to pass it through the callback lambda somehow. Is there a better way to do this?

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/RoomMember.kt
jmartinesp: It's used in a NavTarget, and I think those need to be Parcelable.

PR #318, opened 2023-04-12
Update the Rust SDK to v0.1.7

jmartinesp:


libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/timeline/item/event/EventContent.kt
```diff
@@ -26,7 +27,7 @@ sealed interface EventContent
 
 data class MessageContent(
     val body: String,
-    val inReplyTo: UserId?,
+    val inReplyTo: EventId?,
```
jmartinesp, line 30: Looking at the original data from the Rust SDK and [ReplyToContent](https://github.com/vector-im/element-android/blob/e0b93c2d2cabfbe1f8166f20719df7612670e0cc/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/relation/ReplyToContent.kt) in EA, I *believe* this should be an `EventId`, but I'm not 100% sure.

ganfra (approved):
Thanks!

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/item/event/EventMessageMapper.kt
```diff
@@ -69,7 +69,7 @@ class EventMessageMapper {
         }
         MessageContent(
             body = message.body(),
-            inReplyTo = message.inReplyTo()?.let { UserId(it) },
+            inReplyTo = message.inReplyTo()?.eventId?.let { EventId(it) },
```
bmarty, line 72: We have `fun String.asEventId()` that does more check. Maybe change to

```suggestion
            inReplyTo = message.inReplyTo()?.eventId?.let {it.asEventId() },
```

?

jmartinesp: I can change it, but this is what I currently see for `String.asEventId()`:

```kotlin
fun String.asEventId() = EventId(this)
```

Should it be doing anything else?

ganfra: The check is added on  an other PR https://github.com/vector-im/element-x-android/pull/315

PR #319, opened 2023-04-13
Feature/fga/clean up

ganfra:
Just some small clean up around : 
- RoomDetails features
- AppNavigationState


services/appnavstate/api/src/main/kotlin/io/element/android/services/appnavstate/api/AppNavigationState.kt
```diff
@@ -21,26 +21,34 @@ import io.element.android.libraries.matrix.api.core.SessionId
 import io.element.android.libraries.matrix.api.core.SpaceId
 import io.element.android.libraries.matrix.api.core.ThreadId
 
-sealed interface AppNavigationState {
-    object Root : AppNavigationState
+/**
+ * Can represent the current global app navigation state.
+ * @param owner mostly a Node identifier associated with the state.
+ */
+sealed class AppNavigationState(open val owner: String) {
```
ganfra, line 28: Adding a `owner` parameter to check when calling `onLeaving` methods is still using the same owner than his companion `onNavigate`.
Why this is needed : for now we rely on lifecycle methods of the node, which are async.
If you navigate quickly between nodes, `onCreate` of the new node is called before `onDestroy` of the previous node. So we assume if we don't get the same owner, we can skip the `onLeaving` action as we already replaced it. That's a hack for now, I'd like to have something better where we just listen to nodes directly.

bmarty: Thanks for explaining, but maybe add this useful comment in the code, it will be more visible in the future. Comments in Pull request should be reserved for ephemeral remarks

bmarty (approved):
LGTM, thanks!

PR #321, opened 2023-04-13
Handle navigation on notification click

bmarty:
This PR add deep navigation to the application when clicking on notification.

It has same result if the app was launch or not. If the app was launched, the app is resume and is navigating to the desired target. Previous navigation is lost.

- navigate to the room list when the group notification is clicked (i.e. there are 2 notifications for 2 different rooms and there are collapsed)
- navigate to a room  (i.e. handle click on a room notification), with backstack to the room list.
- navigation to thread is not supported yet (there is a TODO in the code)

Not handled yet: for now the last session is restored, i.e. if a notification for another session is clicked, I am not sure about the effect. Some change will have to be done. We will handle that when we will think about multi-session.

Tested on API 33 emulator OK.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -40,6 +40,14 @@
 
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
+            <!-- Handle deep-link for notification, uncomment to be able to test deeplink with ./tools/adb/deeplink.sh -->
+            <!--intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data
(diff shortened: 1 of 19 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -40,6 +40,14 @@
 
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
+            <!-- Handle deep-link for notification, uncomment to be able to test deeplink with ./tools/adb/deeplink.sh -->
+            <!--intent-filter>
```
ganfra, line 44: Maybe we should keep it, I don't think this would cause any issue

ganfra (approved):
Thanks, I like it!

Issue #111, opened 2023-02-24
Create a room and show it


alfogrillo:

Florian14:
I close this issue, there is some remaining work related to the avatar. These changes are tracked in dedicated issues #122 and #123

PR #324, opened 2023-04-14
Handle create room action

Florian14:
#111 

### Content

- Create a room on the server and open it
- Show an error dialog with a retry button if the creation fails
- Wait for the room to be received from the sync before returning the result from the create room method

Note: Avatar upload will be handled separately

### DIFF
--- a/features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/CreateRoomFlowNode.kt
+++ b/features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/CreateRoomFlowNode.kt
@@ -68,16 +68,16 @@ class CreateRoomFlowNode @AssistedInject constructor(
                         backstack.push(NavTarget.NewRoom)
                     }
 
-                    override fun onOpenRoom(roomId: RoomId) {
-                        plugins<CreateRoomEntryPoint.Callback>().forEach { it.onOpenRoom(roomId) }
+                    override fun onStartChatSuccess(roomId: RoomId) {
+                        plugins<CreateRoomEntryPoint.Callback>().forEach { it.onSuccess(roomId) }
                     }
                 }
                 createNode<CreateRoomRootNode>(context = buildContext, plugins = listOf(callback))
             }
             NavTarget.NewRoom -> {
                 val callback = object : ConfigureRoomNode.Callback {
-                    override fun onRoomCreated(roomId: RoomId) {
-                        plugins<CreateRoomEntryPoint.Callback>().forEach { it.onOpenRoom(roomId) }
+                    override fun onCreateRoomSuccess(roomId: RoomId) {
+                        plugins<CreateRoomEntryPoint.Callback>().forEach { it.onSuccess(roomId) }
                     }
                 }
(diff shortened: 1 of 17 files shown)


features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/CreateRoomFlowNode.kt
```diff
@@ -74,7 +75,12 @@ class CreateRoomFlowNode @AssistedInject constructor(
                 createNode<CreateRoomRootNode>(context = buildContext, plugins = listOf(callback))
             }
             NavTarget.NewRoom -> {
-                createNode<ConfigureRoomFlowNode>(context = buildContext, plugins = emptyList())
+                val callback = object : ConfigureRoomNode.Callback {
+                    override fun onRoomCreated(roomId: RoomId) {
+                        plugins<CreateRoomEntryPoint.Callback>().forEach { it.onOpenRoom(roomId) }
```
bmarty, line 80: Maybe rename to `onRoomCreated` (it has more sense to me regarding the module contract)

```suggestion
                        plugins<CreateRoomEntryPoint.Callback>().forEach { it.onRoomCreated(roomId) }
```

Or maybe the module can return an existing room sometimes (ex: dm already exist)?

In this case, maybe rename to `onSuccess` or `onDone`.

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/configureroom/ConfigureRoomPresenter.kt
```diff
@@ -39,21 +52,47 @@ class ConfigureRoomPresenter @Inject constructor(
             }
         }
 
+        val localCoroutineScope = rememberCoroutineScope()
+        val createRoomAction: MutableState<Async<RoomId>> = remember { mutableStateOf(Async.Uninitialized) }
+
+        fun createRoom(config: CreateRoomConfig) {
+            createRoomAction.value = Async.Uninitialized
```
bmarty, line 59: Nit: setting to `Uninitialized` could be done in the fun `createRoom()`

bmarty (approved):
LGTM, thanks!

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/configureroom/ConfigureRoomPresenter.kt
Florian14: https://github.com/vector-im/element-x-android/pull/324/commits/f566a543e4913b84e6ea596e685ee1dafe2eeeea

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/CreateRoomFlowNode.kt
Florian14: https://github.com/vector-im/element-x-android/pull/324/commits/2403775857c8b791a64a4e107a25c5bbd9cd6abc

PR #331, opened 2023-04-17
Let renovate upgrade the compose compiler and group related upgrades about Kotlin.

bmarty:


jmartinesp (commented):
So if I'm not mistaken, this will group Kotlin, KSP and compose compiler version upgrades into a single 'renovate/kotlin' branch?

jmartinesp (approved):
Supposing it groups updates for these libraries together, LGTM.

bmarty:
> So if I'm not mistaken, this will group Kotlin, KSP and compose compiler version upgrades into a single 'renovate/kotlin' branch?

Yes, this is what I understand too. Sorry, I could have been more verbose in the description.

Thanks for the review, let's merge and see!

PR #334, opened 2023-04-18
Feature/fga/feature flag

ganfra:
This PR introduces a FeatureFlagService so we : 
- Check if a Feature is enabled/disabled
- Enable/disable a Feature at runtime when in debug/nightly mode
- Have a foundation for using Remote FeatureFlag service (like firebase)


bmarty (approved):
LGTM, thanks!

PR #337, opened 2023-04-18
Fix more R8 issues

jmartinesp:
These values are actually from OkHttp: https://raw.githubusercontent.com/square/okhttp/master/okhttp/src/jvmMain/resources/META-INF/proguard/okhttp3.pro

The weird thing is, it [mentions here](https://square.github.io/okhttp/features/r8_proguard/) those shouldn't be needed when using R8, which we use since the upgrade to AGP 8.0 that actually broke minification.


bmarty (approved):
Thanks

Issue #339, opened 2023-04-19
[Room details] Block/ignore user


jmartinesp:
Add the ability to block and unblock users from both the room details screen when in a DM room and from the member details screen.

The action should present a confirmation dialog first, and blocking a user should hide their messages in a DM room.

PR #340, opened 2023-04-19
[Room Details] Block & unblock user

jmartinesp:
## What

Adds shared UI and logic code to block and unblock users from both room details screen and member details screen.

## Why

Implements #339 .

## Tests

You'll need to test this both in a DM room and a regular room.

1. Select a user you want to ignore.
2. Open the DM or regular room. Make sure it contains messages from that user (other events, including undecrypted, ones don't count).
3. Go to the room details. If you're on a regular room navigate to the member details screen for the user you chose before.
4. Block the user, come back to the timeline.
5. Their messages should have disappeared from the timeline. In the DM screen the composer shouldn't be present either.
6. Go again to the user's details, unblock them.
7. The messages should be back in the timeline, the composer should be visible again.

## Video

[block_unblock.webm](https://user-images.githubusercontent.com/480955/233005413-d19f70db-5237-4f2f-81ed-3b465d83c418.webm)

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/textcomposer/MessageComposerPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/textcomposer/MessageComposerPresenter.kt
@@ -19,13 +19,10 @@ package io.element.android.features.messages.impl.textcomposer
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.MutableState
-import androidx.compose.runtime.collectAsState
-import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.saveable.rememberSaveable
 import io.element.android.libraries.architecture.Presenter
-import io.element.android.libraries.core.bool.orTrue
 import io.element.android.libraries.core.data.StableCharSequence
 import io.element.android.libraries.core.data.toStableCharSequence
 import io.element.android.libraries.matrix.api.room.MatrixRoom
@@ -36,7 +33,7 @@ import javax.inject.Inject
 
 class MessageComposerPresenter @Inject constructor(
     private val appCoroutineScope: CoroutineScope,
-    private val room: MatrixRoom,
+    private val room: MatrixRoom
 ) : Presenter<MessageComposerState> {
 
     @Composable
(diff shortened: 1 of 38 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/textcomposer/MessageComposerPresenter.kt
```diff
@@ -48,6 +51,11 @@ class MessageComposerPresenter @Inject constructor(
             mutableStateOf(MessageComposerMode.Normal(""))
         }
 
+        val dmMember by room.getDmMember().collectAsState(initial = null)
```
ganfra, line 54: There will be some changes here cause I'm modifying a bit the api, but I'll handle them.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/textcomposer/MessageComposerPresenter.kt
```diff
@@ -48,6 +51,11 @@ class MessageComposerPresenter @Inject constructor(
...
+
+        // Hide the composer if it's a DM and the other user is ignored
+        val isVisible = (dmMember?.isIgnored?.not()).orTrue()
```
ganfra, line 57: Why? Is it a requirement from product?
Otherwise I'd keep it for now, keep it simple.

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/details/RoomMemberDetailsPresenter.kt
```diff
@@ -36,11 +44,35 @@ class RoomMemberDetailsPresenter @AssistedInject constructor(
...
 
-//        fun handleEvents(event: RoomMemberDetailsEvents) {
-//            when (event) {
-//            }
-//        }
+        fun handleEvents(event: RoomMemberDetailsEvents) {
+            when (event) {
+                is RoomMemberDetailsEvents.BlockUser -> {
+                    if (event.needsConfirmation) {
+                        confirmationDialog = ConfirmationDialog.Block
```
ganfra, line 55: Please extract some methods here using extension on `CoroutineScope`

ganfra (commented):
Some small remarks

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/textcomposer/MessageComposerPresenter.kt
jmartinesp: It's mentioned in the issue, but after talking to the iOS team, it seems like we're postponing implementing this until we also take power levels into account. I'll revert it.

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/details/RoomMemberDetailsPresenter.kt
jmartinesp: I've tried doing that in https://github.com/vector-im/element-x-android/pull/340/commits/96bbd77fcbedcd8047c298e2a4e63c7a5ebd000b, but it's not as easy to do it as in `ChangeServerFlowPresenter`.

ganfra (approved):

PR #341, opened 2023-04-19
Fix: improve offline indicator designs

jmartinesp:
Changes the font weight for the offline indicator text from regular to medium, as asked by @callumu [here](https://github.com/vector-im/element-x-android/issues/141#issuecomment-1514349134).

Also does some cleanup of unneeded suppressions of warnings in the `network-indicator` module and other build files.


bmarty (approved):
LGTM, thanks for the cleanup. I guess the screenshot tests need to be recorded again.

PR #343, opened 2023-04-19
Avoid crash on placeholder RoomSummary

ganfra:
Make RoomId nullable on RoomSummary UiModel to avoid crash on placeholder.


jmartinesp (approved):
LGTM!

PR #348, opened 2023-04-21
Accepting and declining invites

csmith:
Hook up accept and decline buttons in the invites UI. Accept will attempt to accept and then navigate to the room; decline shows a confirmation dialog.

[inviteaccepting.webm](https://user-images.githubusercontent.com/189133/233606036-487b71da-7cb5-47f5-9d51-9a2929b9c1f1.webm)

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
@@ -227,6 +227,10 @@ class LoggedInFlowNode @AssistedInject constructor(
                     override fun onBackClicked() {
                         backstack.pop()
                     }
+
+                    override fun onInviteAccepted(roomId: RoomId) {
+                        backstack.push(NavTarget.Room(roomId))
+                    }
                 }
 
                 inviteListEntryPoint.nodeBuilder(this, buildContext)
--- a/features/invitelist/impl/src/main/kotlin/io/element/android/features/invitelist/impl/InviteListState.kt
+++ b/features/invitelist/impl/src/main/kotlin/io/element/android/features/invitelist/impl/InviteListState.kt
@@ -18,9 +18,20 @@ package io.element.android.features.invitelist.impl
 
 import androidx.compose.runtime.Immutable
 import io.element.android.features.invitelist.impl.model.InviteListInviteSummary
+import io.element.android.libraries.architecture.Async
+import io.element.android.libraries.matrix.api.core.RoomId
 import kotlinx.collections.immutable.ImmutableList
 
 @Immutable
 data class InviteListState(
-    val inviteList: ImmutableList<InviteListInviteSummary>
+    val inviteList: ImmutableList<InviteListInviteSummary>,
+    val declineConfirmationDialog: InviteDeclineConfirmationDialog = InviteDeclineConfirmationDialog.Hidden,
+    val acceptedAction: Async<RoomId> = Async.Uninitialized,
+    val declinedAction: Async<Unit> = Async.Uninitialized,
+    val eventSink: (InviteListEvents) -> Unit = {}
 )
+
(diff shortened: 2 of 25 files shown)


appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -227,6 +227,10 @@ class LoggedInFlowNode @AssistedInject constructor(
                     override fun onBackClicked() {
                         backstack.pop()
                     }
+
+                    override fun onInviteAccepted(roomId: RoomId) {
+                        backstack.push(NavTarget.Room(roomId))
```
bmarty, line 232: Maybe also pop the backstack if the invitation list is empty?

features/invitelist/impl/src/main/kotlin/io/element/android/features/invitelist/impl/InviteListState.kt
```diff
@@ -18,9 +18,20 @@ package io.element.android.features.invitelist.impl
...
 data class InviteListState(
-    val inviteList: ImmutableList<InviteListInviteSummary>
+    val inviteList: ImmutableList<InviteListInviteSummary>,
+    val confirmDialog: InviteConfirmationDialog = InviteConfirmationDialog.Hidden,
+    val acceptedAction: Async<RoomId> = Async.Uninitialized,
+    val declinedAction: Async<Unit> = Async.Uninitialized,
+    val eventSink: (InviteListEvents) -> Unit = {}
 )
+
+sealed interface InviteConfirmationDialog {
```
bmarty, line 34: For clarity maybe rename to 
```suggestion
sealed interface InviteDeclineConfirmationDialog {
```

`InviteListState.confirmDialog` could be renamed too.

features/invitelist/impl/src/test/kotlin/io/element/android/features/invitelist/impl/InviteListPresenterTests.kt
```diff
@@ -159,26 +407,35 @@ class InviteListPresenterTests {
...
+        return this
+    }
+
+    private suspend fun FakeRoomSummaryDataSource.withDirectChatInvitation(): FakeRoomSummaryDataSource {
+        postRoomSummary(
+            listOf(
+                RoomSummary.Filled(
+                    RoomSummaryDetails(
+                        roomId = A_ROOM_ID,
+                        name = A_USER_NAME,
```
bmarty, line 419:
```suggestion
                        name = A_ROOM_NAME,
```

libraries/matrix/test/src/main/kotlin/io/element/android/libraries/matrix/test/FakeMatrixClient.kt
```diff
@@ -51,10 +51,11 @@ class FakeMatrixClient(
     private var createDmResult: Result<RoomId> = Result.success(A_ROOM_ID)
     private var createDmFailure: Throwable? = null
     private var findDmResult: MatrixRoom? = FakeMatrixRoom()
+    private var getRoomResults = mutableMapOf<RoomId, MatrixRoom>()
```
bmarty, line 54: Neat: This can be a `val` instead of a `var`

bmarty (approved):
LGTM, just a few minor remarks. Thanks!

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
csmith: This is currently how it's done on iOS. I guess it might be a bit weird if an invite is removed for some remote reason (e.g. you're uninvited) and that causes a navigation while you're looking at the page?

I'll leave it for now but bear it in mind when the design is reviewed to see if we need to change it on both platforms

bmarty (approved):
Thanks for the update!

PR #353, opened 2023-04-25
Feature/fga/some room related fixes

ganfra:
Make some fixes/improvements around room including : 
- Fix action list bottomsheet background color
- Loading room members for the timeline
- Change/reduce API on room about members
- Avoid using @ on InitialAvatar
- Avoid calling rust when timeline is not ready to use
- Introduce ContentType for each timeline item for better perf


bmarty (approved):
LGTM, nice improvements, please check the test coverage issue!

Issue #360, opened 2023-04-26
[Media Upload] Implement pickers


jmartinesp:
We need to be able to allow the user to pick media from their gallery, files or directly from the camera.

This needs to be done through different system pickers and grouped together in a nice, reusable, interface.

du64:
It would be nice if we can send multiple images and preview them in the composer box before sending

jmartinesp:
> It would be nice if we can send multiple images and preview them in the composer box before sending

I believe it's something that's planned, but since we need to more or less re-implement this feature from scratch we'll start with the easy paths first.

PR #361, opened 2023-04-26
Add `PickerLauncher` wrapper for media/file pickers.

jmartinesp:
## Changes

* Adds a `:library:mediapickers` module containing a `PickerLauncher` wrapper around Compose's default `ActivityResultLaunchers`.
* Create different `PickerTypes` to open different system pickers.
* Add an example usage to the composer.

## Why

Implements #360.


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/textcomposer/MessageComposerView.kt
```diff
@@ -53,13 +70,42 @@ fun MessageComposerView(
...
+    } else {
+        val context = LocalContext.current
+        val tmpFile = remember {
+            val filename = UUID.randomUUID().toString()
+            File(context.cacheDir, filename)
+        }
+        val tmpFileUri = remember(tmpFile) {
+            val authority = "${context.packageName}.fileprovider"
+            FileProvider.getUriForFile(context, authority, tmpFile)
+        }
```
jmartinesp, line 97: These lines could probably be extracted to a helper function so they can be used for both camera photo and video pickers, Also, I'm not sure if we might want to use a filename based on a constant instead of using a hash here.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/textcomposer/MessageComposerView.kt
```diff
@@ -53,13 +70,42 @@ fun MessageComposerView(
...
+        },
         composerCanSendMessage = state.isSendButtonVisible,
         composerText = state.text?.charSequence?.toString(),
         isInDarkMode = !ElementTheme.colors.isLight,
         modifier = modifier
     )
 }
 
+@Composable
+internal fun CameraPhotoPicker(onResult: (Uri?) -> Unit): PickerLauncher<Uri, Boolean> {
```
ElementBot, line 84: - :warning: Composable functions that return a value should start with a lowercase letter.
While useful and accepted outside of @Composable functions, this factory function convention has drawbacks that set inappropriate expectations for callers when used with @Composable functions.

See https://mrmans0n.github.io/compose-rules/rules/#naming-composable-functions-properly for more information.
- :warning: Composable functions with a return type should start with a lowercase letter
- :warning: Composable functions with a return type should start with a lowercase letter
- :warning: Composable functions with a return type should start with a lowercase letter

ElementBot, line 84: - :warning: Composable functions that return a value should start with a lowercase letter.
While useful and accepted outside of @Composable functions, this factory function convention has drawbacks that set inappropriate expectations for callers when used with @Composable functions.

See https://mrmans0n.github.io/compose-rules/rules/#naming-composable-functions-properly for more information.

ganfra (approved):
Thanks!

PR #374, opened 2023-05-02
Allow custom push gateway to use non-default port

jonnyandrew:
When a custom UnifiedPush push gateway is detected, it's gateway URL is constructed from URL components contained within the push key. However, if the gateway URL contains a custom port, this is ignored and the custom push gateway doesn't work.

After this change, the correct port is used.

See vector-im/element-android#8376


bmarty (approved):
Thanks!

jonnyandrew:
Thanks for the review @bmarty. I can't actually merge because I don't have write access - are you able to grant that?

PR #376, opened 2023-05-02
Search for users to start a new DM.

### LINKED ISSUE #95: Search a user to start a chat

csmith:
Hooks up the create room UI to the matrix client to get search results. Searches are debounced for 500ms and only executed when 3 or more characters are entered.

Wrap the result state so we can distinguish between "no results because we haven't searched yet" and
"no results because the API returned nothing", and add a "No results found" message in the UI for the latter case.

Closes #95

[usersearch.webm](https://user-images.githubusercontent.com/189133/235708313-f3bf2f93-53df-4484-880e-02f78b81050c.webm)

### DIFF
--- a/features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/AllMatrixUsersDataSource.kt
+++ b/features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/AllMatrixUsersDataSource.kt
@@ -24,16 +24,16 @@ import io.element.android.libraries.matrix.api.usersearch.MatrixUserProfile
 import io.element.android.libraries.matrix.ui.model.MatrixUser
 import javax.inject.Inject
 
-// TODO this is empty as we currently don't have an endpoint to perform user search
 class AllMatrixUsersDataSource @Inject constructor(
     private val client: MatrixClient
 ) : UserListDataSource {
     override suspend fun search(query: String): List<MatrixUser> {
-        val res = client.searchUsers(query, SEARCH_RESULTS)
+        val res = client.searchUsers(query, MAX_SEARCH_RESULTS)
         return res.getOrNull()?.results?.map(::toMatrixUser).orEmpty()
     }
 
     override suspend fun getProfile(userId: UserId): MatrixUser? {
+        // TODO: hook up to matrix client
         return null
     }
 
@@ -48,6 +48,6 @@ class AllMatrixUsersDataSource @Inject constructor(
     )
 
     companion object {
-        private const val SEARCH_RESULTS = 5L
+        private const val MAX_SEARCH_RESULTS = 5L
     }
 }
--- a/features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/AllMatrixUsersDataSource.kt
+++ b/features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/AllMatrixUsersDataSource.kt
@@ -33,7 +33,7 @@ class AllMatrixUsersDataSource @Inject constructor(
     }
 
     override suspend fun getProfile(userId: UserId): MatrixUser? {
-        // TODO: hook up to matrix client
+        // TODO hook up to matrix client
         return null
     }
 
--- a/features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/AllMatrixUsersDataSource.kt
+++ b/features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/AllMatrixUsersDataSource.kt
@@ -17,17 +17,37 @@
 package io.element.android.features.createroom.impl
 
 import io.element.android.features.userlist.api.UserListDataSource
+import io.element.android.libraries.designsystem.components.avatar.AvatarData
+import io.element.android.libraries.matrix.api.MatrixClient
 import io.element.android.libraries.matrix.api.core.UserId
+import io.element.android.libraries.matrix.api.usersearch.MatrixUserProfile
 import io.element.android.libraries.matrix.ui.model.MatrixUser
 import javax.inject.Inject
 
 // TODO this is empty as we currently don't have an endpoint to perform user search
-class AllMatrixUsersDataSource @Inject constructor() : UserListDataSource {
+class AllMatrixUsersDataSource @Inject constructor(
+    private val client: MatrixClient
+) : UserListDataSource {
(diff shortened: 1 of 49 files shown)


features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/AllMatrixUsersDataSource.kt
```diff
@@ -17,17 +17,37 @@
...
+        username = matrixUserProfile.displayName,
+        avatarData = AvatarData(
+            id = matrixUserProfile.userId.value,
+            name = matrixUserProfile.displayName,
+            url = matrixUserProfile.avatarUrl,
+        )
+    )
+
+    companion object {
+        private const val SEARCH_RESULTS = 5L
```
jmartinesp, line 51: Nit: `MAX_SEARCH_RESULTS` or maybe `SEARCH_RESULTS_COUNT`?

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/AllMatrixUsersDataSource.kt
```diff
@@ -17,17 +17,37 @@
...
 
 import io.element.android.features.userlist.api.UserListDataSource
+import io.element.android.libraries.designsystem.components.avatar.AvatarData
+import io.element.android.libraries.matrix.api.MatrixClient
 import io.element.android.libraries.matrix.api.core.UserId
+import io.element.android.libraries.matrix.api.usersearch.MatrixUserProfile
 import io.element.android.libraries.matrix.ui.model.MatrixUser
 import javax.inject.Inject
 
 // TODO this is empty as we currently don't have an endpoint to perform user search
```
jmartinesp, line 27: It seems like we do now 😅 ?

features/userlist/api/src/main/kotlin/io/element/android/features/userlist/api/UserListState.kt
```diff
@@ -21,11 +21,17 @@ import kotlinx.collections.immutable.ImmutableList
...
     val selectionMode: SelectionMode,
     val eventSink: (UserListEvents) -> Unit,
 ) {
     val isMultiSelectionEnabled = selectionMode == SelectionMode.Multiple
 }
+
+sealed interface UserSearchResultState {
+    object NotSearching : UserSearchResultState
+    object NoResults : UserSearchResultState
+    data class Results(val results: ImmutableList<MatrixUser>) : UserSearchResultState
```
jmartinesp, line 36: Maybe we could add some docs for these?

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/usersearch/MatrixUserProfile.kt
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.libraries.matrix.api.usersearch
+
+import io.element.android.libraries.matrix.api.core.UserId
+
+data class MatrixUserProfile(
```
jmartinesp, line 21: Isn't this almost the same as `MatrixUser`? Do we need to have the 2 of them?

jmartinesp (commented):
Thanks, and sorry for forgetting to add screenshot tests to some components of userlist 😅 .

I have a couple of small comments and also some doubts about `MatrixUserProfile`, since we have an almost identical `MatrixUser` class already. Maybe we should keep only one of them?

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/usersearch/MatrixUserProfile.kt
csmith: Yes, the only difference is that `MatrixUser` has an `AvatarData` member instead of just the URL. But that difference means that `MatrixUser` sits in the `matrixui` library and depends on the design system, so can't really be pulled in to the `matrix` module.

For comparison when we're dealing with rooms the `matrix` module has `RoomMember` which contains these details and the extra room-specific info, and then later is converted into a `MatrixUser` when presented in the UI.

Having the slight duplication seemed like the best option for now, unless there's some nicer way to handle this that I haven't seen?

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/AllMatrixUsersDataSource.kt
csmith: D'oh! Fixed.

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/AllMatrixUsersDataSource.kt
```diff
@@ -17,17 +17,37 @@
...
+    private val client: MatrixClient
+) : UserListDataSource {
     override suspend fun search(query: String): List<MatrixUser> {
-        return emptyList()
+        val res = client.searchUsers(query, MAX_SEARCH_RESULTS)
+        return res.getOrNull()?.results?.map(::toMatrixUser).orEmpty()
     }
 
     override suspend fun getProfile(userId: UserId): MatrixUser? {
+        // TODO: hook up to matrix client
```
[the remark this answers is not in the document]
jmartinesp: If you want to keep the TODO here, just remove the `:` character.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/usersearch/MatrixUserProfile.kt
csmith: Yeah, looks like MatrixUser being Parcelable isn't required. Dropping that in favour of a data class in the `matrix` module with extension funcs in the `matrixui` makes sense to me.

Are you happy to land this change as-is then I'll do the MatrixUser->MatrixUserProfile refactor in a separate PR later today?

jmartinesp: Sure, let's do that.

jmartinesp (approved):
LGTM.

PR #380, opened 2023-05-03
[Media upload] Add media pickers to the Room screen and the composer

### LINKED ISSUE #360: [Media Upload] Implement pickers

jmartinesp:
## Changes

- Modifies `ModalBottomSheetLayout` to match designs.
- Attaches the different media pickers to the room and composer components in `MessagesView` and `MessageComposerPresenter`.
- Works around a bug where the keyboard wouldn't be hidden when calling `focusManager.clearFocus()` because the focused component is inside an `AndroidView`.
- Remove translations moved to a different module.
- Modify localazy script to be able to exclude regexes too (`screen_room_*` clashes with `screen_room_details_*` and `screen_room_member_details_*`).

## Why

Closes #360 .

## Tests

Get inside a room, play with the different pickers. Right now they don't do anything with the retrieved media, that should be handled in future PRs.


tools/localazy/generateLocalazyConfig.py
```diff
@@ -36,10 +36,13 @@ def convertModuleToPath(name):
...
 for entry in config["modules"]:
     # Create action for the default language
+    excludeRegex = regexToAlwaysExclude
+    if "excludeRegex" in entry:
+        excludeRegex += entry["excludeRegex"]
     action = baseAction | {
         "output": convertModuleToPath(entry["name"]) + "/src/main/res/values/localazy.xml",
         "includeKeys": list(map(lambda i: "REGEX:" + i, entry["includeRegex"])),
-        "excludeKeys": list(map(lambda i: "REGEX:" + i, regexToAlwaysExclude)),
+        "excludeKeys": list(map(lambda i: "REGEX:" + i, excludeRegex)),
```
bmarty, line 45: Thanks for the enhancement. You also have to update the action for the translations, [this line on develop](https://github.com/vector-im/element-x-android/blob/develop/tools/localazy/generateLocalazyConfig.py#L54):
From
```
            "excludeKeys": list(map(lambda i: "REGEX:" + i, regexToAlwaysExclude)),
```
to
```
            "excludeKeys": list(map(lambda i: "REGEX:" + i, excludeRegex)),
```

bmarty: Why do you compute again `excludeRegex`?

jmartinesp: 🤦  I don't know why I thought they were 2 different functions. It should be fixed now.

features/messages/impl/src/main/res/values/localazy.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
+  <string name="screen_room_attachment_source_camera">"Camera"</string>
+  <string name="screen_room_attachment_source_camera_photo">"Take photo"</string>
+  <string name="screen_room_attachment_source_camera_video">"Record a video"</string>
+  <string name="screen_room_attachment_source_files">"Attachment"</string>
+  <string name="screen_room_attachment_source_gallery">"Photo &amp; Video Library"</string>
+  <string name="screen_room_error_failed_processing_media">"Failed processing media to upload, please try again."</string>
+</resources>
```
bmarty, line 9: I would not use `screen_room_` prefix for those string, maybe something more generic like `screen_attachment_picker_`. And move to a new module, see my other remark.

bmarty (approved):
Nice work! A few non-blocking remarks.

features/logout/api/src/main/res/values/localazy.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
   <string name="screen_signout_confirmation_dialog_content">"Are you sure you want to sign out?"</string>
-  <string name="screen_signout_confirmation_dialog_submit">"Sign out"</string>
   <string name="screen_signout_confirmation_dialog_title">"Sign out"</string>
   <string name="screen_signout_in_progress_dialog_content">"Signing out…"</string>
+  <string name="screen_signout_confirmation_dialog_submit">"Sign out"</string>
```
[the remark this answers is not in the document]
jmartinesp: It was actually automatic. I also found it quite weird.

features/messages/impl/src/main/res/values/localazy.xml
jmartinesp: I wanted to use some `common_*` strings for these, but we agreed with the iOS team to use `screen_room_` and then duplicate & link the strings if needed in Localazy.

PR #381, opened 2023-05-03
Move and refactor MatrixUser

csmith:
Instead of living in matrixui and having an AvatarData, this can reside in the matrix module and just have the URL. An extension method in matrixui can then provide the AvatarData when required.

This removes some small duplication, and pushes the UI-specific information (i.e., what size of avatar is going to be rendered) further down the stack. It also aligns the field names with those used by the rust SDK (e.g. "displayName" instead of "userName").


libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/user/MatrixUser.kt
```diff
@@ -16,10 +16,13 @@
 
 package io.element.android.libraries.matrix.api.user
 
+import android.os.Parcelable
 import io.element.android.libraries.matrix.api.core.UserId
+import kotlinx.parcelize.Parcelize
 
+@Parcelize
```
jmartinesp, line 23: It feels a bit weird to make this `Parcelable`, but I guess it's fine since it's the Android SDK after all, not a pure Kotlin one? I think we discussed this in the past (and for this same class if I'm not mistaken) with @ganfra .

csmith: Yeah, agree it's a bit weird.

We only need it now for rememberSaveable calls in compose, so we could instead implement a [Saver](https://developer.android.com/reference/kotlin/androidx/compose/runtime/saveable/Saver) in the `matrixui` module and use that whenever we call rememberSaveable with a MatrixUser. It's a bit prone to error, though, as it just results in a runtime crash if you pass a "non-savable" object to rememberSaveable. What do you think?

jmartinesp: Let's keep it `Parcelable`, as discussed in the room.

features/roomlist/impl/src/main/kotlin/io/element/android/features/roomlist/impl/components/RoomListTopBar.kt
```diff
@@ -107,7 +107,7 @@ private fun DefaultRoomListTopBar(
                     modifier = Modifier.testTag(TestTags.homeScreenSettings),
                     onClick = onOpenSettings
                 ) {
-                    Avatar(matrixUser.avatarData, contentDescription = stringResource(StringR.string.common_settings))
+                    Avatar(matrixUser.getAvatarData(), contentDescription = stringResource(StringR.string.common_settings))
```
jmartinesp, line 110: I wonder if we should be using `derivedStateOf` to avoid generating an `AvatarData` object in each recompilation. Maybe it's not worth the trouble.

jmartinesp (approved):
LGTM, just a minor suggestion, but maybe it's not even needed.

features/roomlist/impl/src/main/kotlin/io/element/android/features/roomlist/impl/components/RoomListTopBar.kt
csmith: Yeah, fair point. Done.

PR #382, opened 2023-05-03
Fix screenshot tests for userlists

csmith:
These were recorded before Paparazzi's rendering mode was changed to SHRINK.


jmartinesp (approved):
LGTM, although I wonder how the screenshot tests passed in that PR.

PR #391, opened 2023-05-04
Make danger check view changes

csmith:
Add a check that if a file with @Preview or @LargeHeightPreview in it is changed, then the corresponding build file includes the showkase processor.

Also change the check that prompts about screenshots to use the same @Preview logic instead of checking for "/layout" in the path which doesn't work with compose.


jmartinesp:
The Maestro tests actually succeeded, but they never reported back to the action, I think.

PR #399, opened 2023-05-05
Fix navigation broken

### LINKED ISSUE #397: [Create Room] Bug in navigation, cannot skip a second time the "Add people" screen
- Click on the add room FAB
- Click on "Create room"
- Skip
- Press back
- Skip -> No-op

see a video of the issue:

CannotSkipTwice

As a side note, I am wondering if entering the room detail (name, etc.) before adding people would have more sense (not of scope of this issue though!).

Florian14:
Fix #397


bmarty (approved):
Thanks!
I will test the fix on the next nightly.

Florian14:
> Thanks! I will test the fix on the next nightly.

It was introducing another issue on the room creation by not propagating anymore the `plugins` (which was the root cause of this issue). I improved the fix with the help of ganfra. We should not propagate the plugins from the FlowNode to the Node but filter the callbacks that we want to propagate from the plugins instead.

PR #404, opened 2023-05-09
Custom Compose Preview annotation

julioromano:
Adds `@ElementPreviews` which is a shorthand for:
```
@Preview(name = "Day mode")
@Preview(name = "Night mode", uiMode = Configuration.UI_MODE_NIGHT_YES)
```
Should be used in connection with the now public `fun ElementPreview()` composable.


csmith:
Could you add `io.element.android.libraries.designsystem.preview.ElementPreviews` to [the list in dangerfile.js](https://github.com/vector-im/element-x-android/blob/develop/tools/danger/dangerfile.js#L133-L136) so we catch any usages of it in modules without showkase configured? :)

jmartinesp:
Sorry, but I don't think this is working as intended. It does work for previews in the IDE, but it seems like Paparazzi can't provide a custom value for `isSystemInDarkTheme` by default. 

We might be able to work around something from the Showkase metadata, using its `componentKey` to search for some strings and then setting a custom `LocalConfiguration.current.uiMode` with `Configuration.UI_MODE_NIGHT_YES`, but I don't know if it's worth the effort and extra complexity.

bmarty:
@julioromano I think we can close this, now?

PR #405, opened 2023-05-09
Fix a few FFI leaks

csmith:
These are instances where we obtain an FFIObject and don't call Close on it to release the underlying reference on the Rust side.

The worst instance here was leaking an object per room member every time we refreshed the member list

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomMemberMapper.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomMemberMapper.kt
@@ -24,7 +24,7 @@ import org.matrix.rustcomponents.sdk.RoomMember as RustRoomMember
 
 object RoomMemberMapper {
 
-    fun map(roomMember: RustRoomMember): RoomMember =
+    fun map(roomMember: RustRoomMember): RoomMember = roomMember.use {
         RoomMember(
             UserId(roomMember.userId()),
             roomMember.displayName(),
@@ -35,6 +35,7 @@ object RoomMemberMapper {
             roomMember.normalizedPowerLevel(),
             roomMember.isIgnored(),
         )
+    }
 
     fun mapMembership(membershipState: RustMembershipState): RoomMembershipState =
         when (membershipState) {
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomMemberMapper.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomMemberMapper.kt
@@ -26,14 +26,14 @@ object RoomMemberMapper {
 
     fun map(roomMember: RustRoomMember): RoomMember = roomMember.use {
         RoomMember(
-            UserId(roomMember.userId()),
-            roomMember.displayName(),
-            roomMember.avatarUrl(),
-            mapMembership(roomMember.membership()),
-            roomMember.isNameAmbiguous(),
-            roomMember.powerLevel(),
-            roomMember.normalizedPowerLevel(),
-            roomMember.isIgnored(),
+            UserId(it.userId()),
+            it.displayName(),
+            it.avatarUrl(),
+            mapMembership(it.membership()),
+            it.isNameAmbiguous(),
+            it.powerLevel(),
+            it.normalizedPowerLevel(),
+            it.isIgnored(),
         )
     }
 
--- a/samples/minimal/src/main/kotlin/io/element/android/samples/minimal/RoomListScreen.kt
+++ b/samples/minimal/src/main/kotlin/io/element/android/samples/minimal/RoomListScreen.kt
@@ -33,17 +33,16 @@ import io.element.android.libraries.dateformatter.impl.LocalDateTimeProvider
 import io.element.android.libraries.designsystem.utils.SnackbarDispatcher
 import io.element.android.libraries.matrix.api.MatrixClient
 import io.element.android.libraries.matrix.api.core.RoomId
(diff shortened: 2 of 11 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomMemberMapper.kt
```diff
@@ -24,7 +24,7 @@ import org.matrix.rustcomponents.sdk.RoomMember as RustRoomMember
 
 object RoomMemberMapper {
 
-    fun map(roomMember: RustRoomMember): RoomMember =
+    fun map(roomMember: RustRoomMember): RoomMember = roomMember.use {
```
bmarty, line 27: You are not using `it` in the `use` block, which is fine, but could be seen as a bug.
What do other devs think?

samples/minimal/src/main/kotlin/io/element/android/samples/minimal/RoomListScreen.kt
```diff
@@ -80,6 +80,7 @@ class RoomListScreen(
                     initialize()
                     paginateBackwards(20, 50)
                     dispose()
+                    room.close()
```
bmarty, line 83: A bit weird to call this in the apply block on a timeline, using `use` was not possible?
But on the `minimal` sample I guess this is fine.

bmarty (approved):
LGTM, thanks! Do you know a quick way to detect missing usage of `use`? I mean without using the profiler. Maybe with Leak Canary? I am also wondering if the CI could help to detect such errors.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomMemberMapper.kt
jmartinesp: It's true that we should probably use `it` to make it more explicit, but I don't see much harm in using `roomMember` directly.

bmarty:
@jmartinesp do you know why Maestro action [has been skipped](https://github.com/vector-im/element-x-android/actions/runs/4928295558?pr=405)? The PR has been approved, this is strange.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomMemberMapper.kt
julioromano: Even if in this case it appears to be safe, it may lead to bad practice if generalized.
- Not using the lambda's param could break smart casting at times (e.g. when `roomMember` is a nullable function or property accessor or when it is defined in a different module). 
- From the reviewer's PoV it will increase the cognitive load: the bigger the lambda becomes, the more care will be required to keep track of the variables used outside of it.
- On the contrary using `it` enforces locality and makes reasoning with the code simpler (i.e. the reader can be logically sure that `it` will never change throughout the lambda)

Florian14: Not sure it's a problem, sometimes we use this syntax to have a stream instead of some if conditions on the nullability or some predicates. I agree the scope is not exactly the same, so we have to be vigilant, but most of the time there will be no consequence from my point of view.

Sometimes, it's better to rename the `it` variable for clarity, `roomMember.use { roomMember -> ... }` will work in this case but I prefer using the original field instead. `rm` or `member` will introduce new names and more complexity.

For simple cases, using `it` is fine.

To answer the original question, I don't think we need a coding rule, I am for letting the developer make the decision

julioromano: I do agree a rule on this would be a bit too much though.

Florian14:
> @jmartinesp do you know why Maestro action [has been skipped](https://github.com/vector-im/element-x-android/actions/runs/4928295558?pr=405)? The PR has been approved, this is strange.

@bmarty It seems the job is triggered on each comment on the PR:
image

But you can see that the job has already been executed previously:
image

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomMemberMapper.kt
csmith: I don't have a super strong opinion on this, but will point out all the other mappers I looked at do it in the same way currently so we'd want to change those if we decide we should use the block parameter everywhere.

I think I slightly prefer this way for this particular case, as you can glance at the function, see the parameter name, then glance down and see how it's used without having to think about the `use` at all (it's basically just a side effect you can ignore for the most part).

csmith:
> Do you know a quick way to detect missing usage of `use`? I mean without using the profiler. Maybe with Leak Canary? I am also wondering if the CI could help to detect such errors.

No, I spent a while looking yesterday and couldn't find anything good. There doesn't seem to be a static analysis tool that does closeable-checking and also supports kotlin properly. And I don't think LeakCanary could work even because the kotlin-side objects would get GC'd normally :(

I think we'd probably have to write something to generate code to explicitly track them, which is probably more work than it's worth at the minute

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomMemberMapper.kt
csmith: Consensus seems to be "use `it`, but we don't need it to be a hard rule". I've updated the various mappers that used this pattern to use `it`.

bmarty (approved):
Thanks for the update!

PR #411, opened 2023-05-10
[Media upload] Upload image, video and files

jmartinesp:
## Changes

- Add `sendImage`,  `sendVideo`, `sendFile` and `sendAudio` methods to `MatrixRoom`.
- Add uploading functionality to `MessageComposerPresenter`.
- Add Blurhash library (needed for uploading).
- Fix several issues with media pre processing.
- Show snackbar message when upload or pre-processing fails.

## Why
Closes https://github.com/vector-im/element-x-android/issues/414.

## Test

Go into a room, try to send images & videos from gallery, from your camera and then some other files.

**Note**: while you can *upload* these files, there is not media preview in the timeline yet, so we should verify these are correctly received and displayed from some other app.

### DIFF
--- a/features/messages/impl/src/test/kotlin/io/element/android/features/messages/textcomposer/MessageComposerPresenterTest.kt
+++ b/features/messages/impl/src/test/kotlin/io/element/android/features/messages/textcomposer/MessageComposerPresenterTest.kt
@@ -57,6 +57,7 @@ import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.android.awaitFrame
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.test.runCurrent
 import kotlinx.coroutines.test.runTest
 import kotlinx.coroutines.test.setMain
 import org.junit.Test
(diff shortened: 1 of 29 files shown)


features/messages/impl/src/test/kotlin/io/element/android/features/messages/textcomposer/MessageComposerPresenterTest.kt
```diff
@@ -285,16 +297,83 @@ class MessageComposerPresenterTest {
...
+                )
+            )
+        ))
         moleculeFlow(RecompositionClock.Immediate) {
             presenter.present()
         }.test {
             val initialState = awaitItem()
             initialState.eventSink(MessageComposerEvents.PickAttachmentSource.FromGallery)
+            // Wait for the launched upload coroutine to run
+            delay(1)
```
Florian14, line 334: What about using `TestScope.runCurrent()` instead of delaying with a "random" value? It's seems more accurate to me but not a big deal.

Florian14 (approved):
LGTM, not tested yet

features/messages/impl/src/test/kotlin/io/element/android/features/messages/textcomposer/MessageComposerPresenterTest.kt
jmartinesp: That's way better! I knew there was something like that but I couldn't remember its name. Thanks!

Florian14:
Tested ok, there is no indication that we are sending a media until we receive it from the sync. I suppose this will be improved later?

jmartinesp:
> Tested ok, there is no indication that we are sending a media until we receive it from the sync. I suppose this will be improved later?

Yes, we mentioned this earlier, there are no designs for the media upload or a way to get an upload progress at the moment.

PR #426, opened 2023-05-15
Do not let Renovate assign a reviewer anymore

bmarty:
Now that we have a CODEOWNERS file, there is no need for Renovate to assign a reviewer to assign a reviewer. It will be done automatically by GitHub.

(I think - see below) It leads to double assignment to the team, which is not handled fine by GitHub, for instance in [this PR](https://github.com/vector-im/element-x-android/pull/421):

image

This is also maybe due to this weird timeline:

image

Anyway we can merge this PR and see what's happening on new PRs from Renovate in the coming days.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -4,7 +4,6 @@
     "config:base"
   ],
   "labels": ["dependencies"],
-  "reviewers": ["team:element-x-android-reviewers"],
   "ignoreDeps": ["string:app_name"],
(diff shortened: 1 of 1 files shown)


bmarty:
Note: for this PR, we can see this automatic step (v), and we should see the same step for PR from Renovate. If it is not the case, we will need to revert this change.

image

Florian14 (approved):

PR #430, opened 2023-05-16
New copy for members search results

julioromano:
Context: https://matrix.to/#/!dypRwVXIkJaTAACHPd:element.io/$Obt7MbgGtHBqJHpZqcdd6OQ5XgxxNXSJofnFG1NyWWM?via=matrix.org&via=element.io&via=t2l.io

Related to: https://github.com/vector-im/element-x-android/issues/246

### DIFF
--- a/features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListView.kt
+++ b/features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListView.kt
@@ -71,6 +71,7 @@ import io.element.android.libraries.matrix.ui.components.MatrixUserRow
 import kotlinx.collections.immutable.ImmutableList
 import io.element.android.libraries.ui.strings.R as StringR
 
+@OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun RoomMemberListView(
     state: RoomMemberListState,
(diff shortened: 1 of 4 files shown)


features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListView.kt
```diff
@@ -71,6 +71,7 @@ import io.element.android.libraries.matrix.ui.components.MatrixUserRow
 import kotlinx.collections.immutable.ImmutableList
 import io.element.android.libraries.ui.strings.R as StringR
 
+@OptIn(ExperimentalMaterial3Api::class)
```
Florian14, line 74: According to the linter, there is no experimental API

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListView.kt
```diff
@@ -149,6 +150,29 @@ private fun RoomMemberList(
     }
 }
 
+@Composable
+private fun RoomMemberSearchResult(
```
Florian14, line 154: I think you wanted to rename the method but the previous one is still present

Florian14: Would we have a single method with a param for the header title?

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListView.kt
julioromano: Indeed, my AS was out of sync.

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListView.kt
julioromano: Refactored

julioromano:
All suggestion addressed!

Florian14 (approved):
SGTM

PR #443, opened 2023-05-22
Timeline state event formatting

bmarty:
And group adjacent state events with collapse / expand actions.

This PR introduce new `:libraries:eventformatter` modules to share code between room list last message formatter and formatter of state event for timeline.

Animation

Ideally the collapsed item would expand to the bottom direction, but this is due to the `reversed` state of the LazyColumn. I tried to modify this behavior, but with no acceptable result so far. We can revisit this later.

Also once expanded, a Column is added to the LazyList to display all the event, so there is no recycling. Tested on MatrixHQ with more than 60 collapsed Event without noticing any performance impact. The app is just rendering text, so I think the impact is low.

Also we could have an alternative solution to manage the collapsed/expanded state to the View, using `rememberSaveable`, it's also working, but if we want to improve the code later it's probably better to keep this boolean in the state.

Note: there is no actions yet when long pressing on state event, but there will be some in the future (`Share` for instance).

Some state event are displayed without formatting, we need more info from the SDK to be able to display valuable information.

Later we can consider improving the timeline start, for now there is no specific treatment on it:

|Collapsed|Expanded|
|-|-|
|image|image|


libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/StateContentFormatter.kt
```diff
@@ -0,0 +1,219 @@
...
+                    "Custom event ${content.eventType}"
+                }
+            }
+            OtherState.PolicyRuleRoom -> when (renderingMode) {
+                RenderingMode.RoomList -> {
+                    Timber.v("Filtering timeline item for room state change: $content")
+                    null
+                }
+                RenderingMode.Timeline -> {
+                    "PolicyRuleRoom"
```
jmartinesp, line 97: Are we sure we want to add these? They look quite weird in the timeline. 
image

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/group/GroupHeaderView.kt
```diff
@@ -0,0 +1,134 @@
...
+@Composable
+fun GroupHeaderView(
+    text: String,
+    isExpanded: Boolean,
+    isHighlighted: Boolean,
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier
+) {
+    val backgroundColor = if (isHighlighted) {
+        ElementTheme.colors.messageHighlightedBackground
```
jmartinesp, line 57: Maybe we should change the text color if the view is highlighted too? It looks quite weird in the screenshots.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/MessageEventBubble.kt
```diff
@@ -84,6 +84,7 @@ fun MessageEventBubble(
 
     fun Modifier.offsetForItem(): Modifier {
         return if (state.isMine) {
+            // FIXME setting y offset to -12.dp can overlap a state event displayed above.
```
jmartinesp, line 87: This FIXME is intended for future work, right?

jmartinesp (commented):
Thanks for splitting the formatter into several classes! I have a couple of questions before approving the PR. Also, the tests are failing because of the weird timeout issue we saw, so we should probably rebase/merge the latest changes to fix that.

libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/StateContentFormatter.kt
bmarty: Yes, for now the application is displaying what the SDK is providing, i.e. not a lot.
I prefer to revisit this later.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/group/GroupHeaderView.kt
bmarty: I have ask Amsha for a design input.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/MessageEventBubble.kt
bmarty: Yes, I prefer to merge this PR faster and revisit.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/group/GroupHeaderView.kt
bmarty: (Answer: there is no design yet, so this will be handled later)

jmartinesp (approved):
LGTM.

bmarty:
@jmartinesp I have force-pushed the branch and answered your questions. Will merge if CI is happy.

bmarty:
(checking the TimelinePresenter mising test)

bmarty:
(I have found a bug when writing unit test, working on it)

bmarty:
@jmartinesp can I ask you to check my 2 last commits?

jmartinesp (approved):
The new changes look fine. Nice catch!

PR #447, opened 2023-05-23
Wysiwyg 2.2.0

bmarty:
Replacement for #435.

Composer tested Ok, even if we do not have the latest features (see https://github.com/vector-im/element-android/pull/8440)


jmartinesp (approved):
LGTM, it seems to work fine.

PR #449, opened 2023-05-23
Reduce time to run test from 15 minutes to about 8

bmarty:
Motivation: test run by the CI should be as fast as possible.

Reduce time to run test from 15 minutes to about 8 by disabling creating artifact with Kover merged report (this artifact is not very useful, when we want to check the coverage, we run the task locally) and uploading to sonar (analysis on develop is enough for now).

Instead those 2 steps are now run once a day in a separate GitHub action.

See for instance the log a the Test task:

image

The 2 removed steps should save about 8 minutes per run.

### DIFF
--- a/.github/workflows/nightlyReports.yml
+++ b/.github/workflows/nightlyReports.yml
@@ -1,0 +1,43 @@
+name: Nightly reports
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Every nights at 5
+    - cron: "0 5 * * *"
+
+# Enrich gradle.properties for CI/CD
+env:
+  GRADLE_OPTS: -Dorg.gradle.jvmargs="-Xmx3072m -Dfile.encoding=UTF-8 -XX:+HeapDumpOnOutOfMemoryError" -Dkotlin.daemon.jvm.options="-Xmx2560m" -Dkotlin.incremental=false
+  CI_GRADLE_ARG_PROPERTIES: --stacktrace -PpreDexEnable=false --max-workers 4
+
+jobs:
+  nightlyReports:
+    name: Create kover report artifact and upload sonar result.
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


bmarty:
Before:

image

After:

image

Better than expected!

.github/workflows/nightlyReports.yml
jmartinesp, line 1: Maybe we can add this to the nightly workflow so we can benefit from the build cache? If we're worried about these jobs failing and cancelling the workflow maybe we can add some flags to prevent it.

jmartinesp (approved):
I don't want to block this so I'm approving it, but I have a couple of comments that could improve the PR.

.github/workflows/nightlyReports.yml
bmarty: Yes, I was not sure and preferred to split into 2 different workflows. Also they can be manually triggered independently easily (I mean without the pain to add trigger parameters) if necessary.

.github/workflows/tests.yml
```diff
@@ -85,12 +81,8 @@ jobs:
...
             **/build/reports/tests/*UnitTest/
 
-      - name: 🔊 Publish results to Sonar
-        env:
-          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
-          ORG_GRADLE_PROJECT_SONAR_LOGIN: ${{ secrets.SONAR_TOKEN }}
-        if: ${{ always() && env.SONAR_TOKEN != '' && env.ORG_GRADLE_PROJECT_SONAR_LOGIN != '' }}
-        run: ./gradlew sonar $CI_GRADLE_ARG_PROPERTIES
+      - name: 🔊 Publish results to Sonar (disabled)
+        run: echo "This is now done only once a day, see nightlyReports.yml"
```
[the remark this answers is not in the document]
bmarty: Not really, just to give info that the task has not been run. It does not cost a lot.

PR #458, opened 2023-05-24
Notification content

bmarty:
Map more data from the Rust SDK: notification content and avatars.
Still not matching was is displayed by Element Android, also sometimes the room name is not resolved by the SDK: https://github.com/matrix-org/matrix-rust-sdk/issues/1980 .
Also display the account name in the notification title, for better support of multi account.
There is a `annotateForDebug` which help to understand what is displayed in the notification, since it can depends on OS version and devices. This is disabled by default.

image


jmartinesp (approved):
LGTM, thanks!

PR #461, opened 2023-05-25
Display a notice if Matrix ID isn't resolved

### LINKED ISSUE #424: [Search] Display a notice if the matrix ID is not resolved
When creating a DM, a room, or inviting members, if a matrix identifier is entered in the search field, then, if any account is resolved behind this matrix ID, a notice should be added in the search result to inform the user that this identifier may not exist.

**ACs**
- an exclamation mark button shows on the right of the display name ([iOS design](https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=1761-98358&t=04ew9UXBL2zirzPY-4))
- Tap on the above button will show an informational bubble (system platform popovers should be preferred to show this info)

csmith:
If we can't get the profile of a user after an mxid was searched for, show a warning under their ID to say the invite probably won't be delivered.

Closes #424

## Demo

[device-2023-05-25-130609.webm](https://github.com/vector-im/element-x-android/assets/189133/9794f1f6-e008-4e40-8d83-e06a8b4a0d93)


libraries/usersearch/api/src/main/kotlin/io/element/android/libraries/usersearch/api/UserRepository.kt
```diff
@@ -16,10 +16,9 @@
 
 package io.element.android.libraries.usersearch.api
 
-import io.element.android.libraries.matrix.api.user.MatrixUser
 import kotlinx.coroutines.flow.Flow
 
 interface UserRepository {
```
ganfra, line 21: It's not new, but is `UserRepository` a find name? 
And do we really need to have a module just for that? Could we not embedded that inside the Matrix module?

ganfra (approved):
One remark/question, otherwise LGTM!

libraries/usersearch/api/src/main/kotlin/io/element/android/libraries/usersearch/api/UserRepository.kt
csmith: Florian suggested the name. It does some business logic to combine data from several sources, so seems reasonable?

It could go in the Matrix module, but it felt more like a "specific thing we do on top of Matrix in a small number of places" rather than a general purpose thing. If we pushed everything that had anything to do with Matrix into the Matrix module, 90% of the app could be in there :sweat_smile:

csmith: (I'll go ahead and merge this to get it out the way. If we want to move it, it'll be easier in a separate PR with no changes to the logic)

PR #462, opened 2023-05-25
Enable one last ignored test for LeaveRoomPresenterImpl

julioromano:
Most of the tests in `LeaveRoomPresenterImplTest.kt` where using `UnconfinedTestDispatcher` which was conflating some of the state returned by the presenter. This prevented to test one specific case which had been left with an `@Ignore` annotation.
This PR switches to `StandardTestDispatcher` so that the ignored test case can work and also fixes some other test cases whose behavior is now more correct under the `StandardTestDispatcher`.

Also updates our test factory method for `CoroutineDispatchers` to more easily obtain a `CoroutineDispatchers` which uses `StandardTestDispatcher`.


bmarty (approved):
LGTM, thanks!

PR #465, opened 2023-05-26
Display timestamps for text messages

jmartinesp:
## What

- Adds a timestamp to all text messages.
- Displays sending errors on the timestamp (although I believe local echo is not yet implemented).
- Displays a prefix in the timestamp if a messages has been edited.

## Why

See #464 .

## Screenshots

| Light mode | Dark mode |
|-|-|
|Screenshot_1685086803|Screenshot_1685086809|
|image|image|


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -232,6 +256,36 @@ fun TimelineItemEventRow(
     }
 }
 
+@Composable
+private fun TimestampView(
+    sentTime: String,
+    isEdited: Boolean,
+    sendingFailed: Boolean,
+    onResendClicked: () -> Unit,
```
bmarty, line 264: Maybe some wording discussion, but at this point, `TimestampView` cannot know the action that will be done once it is clicked. I would rename to `onTimestampClicked`, or just `onClicked`.
In the same spirit, `sendingFailed` could be rename to `isError` (or maybe something better)

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -232,6 +256,36 @@ fun TimelineItemEventRow(
...
+    modifier: Modifier = Modifier
+) {
+    val tint = if (sendingFailed) ElementTheme.colors.textActionCritical else null
+    Row(modifier = modifier.clickable(enabled = sendingFailed, onClick = onResendClicked)){
+        if (isEdited) {
+            Text(
+                stringResource(StringR.string.common_edited_suffix),
+                style = ElementTextStyles.Regular.caption2,
+                color = tint ?: MaterialTheme.colorScheme.secondary,
+            )
```
bmarty, line 274: (Quick note: this will be done in the future, but clicking on this text should open the edition history.)

bmarty (approved):
LGTM, there are some conflicts to fix, nothing hard at first look.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
jmartinesp: Changed them in https://github.com/vector-im/element-x-android/pull/465/commits/a37a5523ceab1494f2bfefef4d64e191c2acc59f, it should be better now.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
jmartinesp: There might be a corner case where we have an edited message that couldn't be send, maybe? I've asked what we should do in this case too.

PR #475, opened 2023-05-29
Revert `RenderingMode` to `NORMAL` in `ScreenshotTest`

jmartinesp:
The screenshot test customization was useful to get better screenshots of small components or large screens, but it seems like using any other `RenderingMode` than `RenderingMode.NORMAL` causes memory leaks, and we have reached a point 
in some branches where it's impossible to make `verifyPaparazziDebug` pass given these OOM issues.

For this reason and since we don't have a better alternative at the moment, I want to revert those changes so the screenshot tests don't become a blocker.


julioromano (approved):

jmartinesp:
Unfortunately, we have confirmed with @julioromano this only postpones the issue: after ~850 screenshot tests, the OOMs are back. Still, we'd be blocked without this in the near future, so this gives us some extra time.

PR #479, opened 2023-05-30
Fix flaky media tests.

csmith:
Creating a mock during the test run seems to sometimes take a large amount of time when running the tests in parallel (possibly some kind of resource contention?)

Instead, perform the mocking in the test class so it's part of the setup, not the  actual test runs.


csmith:
> should we add some checks in lint to avoid mockk android stuff in test exec?

That would be good, I'm just not sure it's doable with static analysis :grimacing:

Issue #480, opened 2023-05-30
[Room details] Open room member details when tapping on its avatar or username in the timeline


jmartinesp:
As the title says, we should open the profile screen of a member if we tap on either its username or avatar in the timeline.

jmartinesp:
I forgot to add a link to the [original enhancement request](https://github.com/vector-im/element-meta/issues/1701).

It was decided not to do that as the components have small touch targets and it would add extra complexity. But yes, it could be a bit confusing for previous users of Element Android.

PR #482, opened 2023-05-31
[Room details] Open room member details when clicking on user data in timeline

jmartinesp:
## What

* Modifies `RoomDetailsEntryPoint` so it takes an `Inputs` object with an initial element.
* When tapping on users' UI in the timeline (name or avatar) the room member details screen will be opened.

## Why

Solves #480 .

##  Test

Go to a room with some other user, tap on either their name or avatar. The room member details screen should be opened.

### DIFF
--- a/features/roomdetails/api/src/main/kotlin/io/element/android/features/roomdetails/api/RoomDetailsEntryPoint.kt
+++ b/features/roomdetails/api/src/main/kotlin/io/element/android/features/roomdetails/api/RoomDetailsEntryPoint.kt
@@ -16,11 +16,26 @@
 
 package io.element.android.features.roomdetails.api
 
+import android.os.Parcelable
 import com.bumble.appyx.core.modality.BuildContext
 import com.bumble.appyx.core.node.Node
 import com.bumble.appyx.core.plugin.Plugin
 import io.element.android.libraries.architecture.FeatureEntryPoint
+import io.element.android.libraries.architecture.NodeInputs
+import io.element.android.libraries.matrix.api.core.UserId
+import kotlinx.parcelize.Parcelize
 
 interface RoomDetailsEntryPoint : FeatureEntryPoint {
-    fun createNode(parentNode: Node, buildContext: BuildContext, plugins: List<Plugin>): Node
+
+    sealed interface InitialTarget : Parcelable {
+        @Parcelize
+        object RoomDetails : InitialTarget
+
+        @Parcelize
+        data class RoomMemberDetails(val roomMemberId: UserId) : InitialTarget
+    }
+
+    data class Inputs(val initialElement: InitialTarget) : NodeInputs
+
+    fun createNode(parentNode: Node, buildContext: BuildContext, inputs: Inputs, plugins: List<Plugin>): Node
 }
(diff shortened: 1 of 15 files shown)


features/roomdetails/api/src/main/kotlin/io/element/android/features/roomdetails/api/RoomDetailsEntryPoint.kt
```diff
@@ -16,11 +16,32 @@
...
+
+        @Parcelize
+        object RoomMemberList : NavTarget
+
+        @Parcelize
+        object InviteMembers : NavTarget
+
+        @Parcelize
+        data class RoomMemberDetails(val roomMemberId: UserId) : NavTarget
+    }
```
bmarty, line 42: I understand that we do not want to duplicate `NavTarget`, but this is a bit strange to expose all the targets to the API module. Some targets may stay internal.

I would create a new type here:

```kotlin
    sealed interface InitialTarget {
        // Default to RoomDetails
        object Default : InitialTarget

        data class RoomMemberDetails(val roomMemberId: UserId) : InitialTarget
    }
```

and in the impl module, a conversion fun:

```kotlin
private fun RoomDetailsEntryPoint.InitialTarget.toInternalTarget(): RoomDetailsFlowNode.NavTarget {
    return when (this) {
        RoomDetailsEntryPoint.InitialTarget.Default -> RoomDetailsFlowNode.NavTarget.RoomDetails
        is RoomDetailsEntryPoint.InitialTarget.RoomMemberDetails -> RoomDetailsFlowNode.NavTarget.RoomMemberDetails(roomMemberId)
    }
}
```

Not sure if it worth to change it now though.

bmarty (commented):
A quick remark about `api` and `impl` module content.

features/roomdetails/api/src/main/kotlin/io/element/android/features/roomdetails/api/RoomDetailsEntryPoint.kt
ganfra: Yes I'd have done it like that too

jmartinesp: Yes, I moved stuff around just to check if it worked, but I forgot to do a proper mapping here. I'll move it back to impl and add the `InitialTarget` here.

jmartinesp: Done in https://github.com/vector-im/element-x-android/pull/482/commits/42dd2726899a9835915f9302e3ece2ce94477589

features/roomdetails/api/src/main/kotlin/io/element/android/features/roomdetails/api/RoomDetailsEntryPoint.kt
```diff
@@ -16,11 +16,26 @@
...
+        @Parcelize
+        object RoomDetails : InitialTarget
+
+        @Parcelize
+        data class RoomMemberDetails(val roomMemberId: UserId) : InitialTarget
+    }
+
+    data class Inputs(val initialElement: InitialTarget) : NodeInputs
+
+    fun createNode(parentNode: Node, buildContext: BuildContext, inputs: Inputs, plugins: List<Plugin>): Node
```
ganfra, line 40: I'd have use the Builder pattern instead so you can hide the `NodeInputs` implementation behind.

jmartinesp: I thought about doing that, but since the `Inputs` are mandatory it seemed better to enforce them here at compile time than crashing on runtime if they're not present.

jmartinesp:
Squashed and force-pushed since Danger got stuck...

bmarty (approved):
Thanks for the update!

PR #493, opened 2023-05-31
Let the CI record the screenshots test if necessary (#492)

### LINKED ISSUE #492: [DevX] Iterate on the screenshot test
## Problematic

When a developer records the reference images for the screenshot test, they use their own machine with different env, and it ends up with small change on the generated images (which are binary files), even if the image itself is identical.

It adds unwanted changes in the PR content.

## Solution

To fix this, we should generate the image using the same environment, and the CI would fit this need.

We can get inspired by what have be done here: https://proandroiddev.com/no-emulator-needed-for-screenshot-tests-of-compose-previews-on-ci-d292fa0eb26e#5b38:

>  -  It compares the screenshots with the existing ones. If they are the same, the check is green.
 >  - If they are different, it pushes the different screenshots to a different branch based on the PR branch.
  > - It posts a comment to the PR with a link to compare the two branches. GitHub does a pretty good job of showing differences between the images.
   > - If the change is intentional, you can simply merge the branch back to the PR branch. If not, you should fix the regression bug.

The link above contains workflow and link to script to achieve that.

Also the [documentation](https://github.com/vector-im/element-x-android/blob/develop/docs/screenshot_testing.md) will have to be updated.

bmarty:
Draft, let see what's happening now.

Closes #492


bmarty:
We found a (better?) solution in #520

PR #494, opened 2023-05-31
Check code

bmarty:
Import tool from Element Android which help to detect forbidden pattern in our codebase.

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -12,6 +12,14 @@ env:
   CI_GRADLE_ARG_PROPERTIES: --stacktrace -PpreDexEnable=false --max-workers 2 --no-daemon --warn
 
 jobs:
+  checkScript:
+    name: Search for forbidden patterns
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run code quality check suite
+        run: ./tools/check/check_code_quality.sh
+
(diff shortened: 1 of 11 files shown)


julioromano:
I see some code changes that appear unrelated to the addition of the check_code_quality.sh script. Can you offer some more context?

julioromano (commented):
I see some code changes that appear unrelated to the addition of the check_code_quality.sh script. What are those for?

bmarty:
> I see some code changes that appear unrelated to the addition of the check_code_quality.sh script. What are those for?

- We should not use `System.currentTimeMillis()` directly, it's better to inject a Clock.
- Using `"$o.s"` is forbidden because it can lead to unexpected log.
- For the rest, this is formatting.

julioromano:
> > I see some code changes that appear unrelated to the addition of the check_code_quality.sh script. What are those for?
> 
> * We should not use `System.currentTimeMillis()` directly, it's better to inject a Clock.
> * Using `"$o.s"` is forbidden because it can lead to unexpected log.
> * For the rest, this is formatting.

I see, but is it the `check_code_quality.sh` that fixed those errors or was it you manually?

julioromano (approved):
I'll LGTM this in the meantime to not become a blocker. Though please try to answer my comment above.

bmarty:
I fixed the issue manually.

PR #502, opened 2023-06-01
Onboarding iteration

bmarty:
Figma: https://www.figma.com/file/o9p34zmiuEpZRyvZXJZAYL/FTUE?type=design&node-id=611-61333&t=ANPtrLshqx4O1ePA-0

Also introduce page `OnBoardingPage` from the atomic design system, to be able to reuse it for other screens.

(We should have less warning on the compilation since we are not using deprecated accompanist library anymore).


features/onboarding/impl/src/main/kotlin/io/element/android/features/onboarding/impl/OnBoardingScreen.kt
```diff
@@ -17,151 +17,153 @@
...
 import io.element.android.libraries.designsystem.theme.components.Text
 import io.element.android.libraries.testtags.TestTags
 import io.element.android.libraries.testtags.testTag
-import kotlinx.coroutines.delay
-import kotlinx.coroutines.launch
 
+// Ref: https://www.figma.com/file/o9p34zmiuEpZRyvZXJZAYL/FTUE?type=design&node-id=133-5427&t=5SHVppfYzjvkEywR-0
 @Composable
 fun OnBoardingScreen(
+    canLoginWithQrCode: Boolean,
```
ganfra, line 55: Maybe we should have a proper state and presenter?

ganfra (approved):
Small remarks otherwise LGTM

PR #503, opened 2023-06-01
Update android.gradle.plugin to v8.0.2

bmarty:
Replaces #496

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -3,7 +3,7 @@
 
 [versions]
 # Project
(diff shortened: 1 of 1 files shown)


jmartinesp:
I believe we need to migrate to the new kover DSL format, I'm getting lots of errors locally.

bmarty:
@jmartinesp yes, sorry for the mess up, let's upgrade dep one by one. I have force push with only the first commit.

jmartinesp (approved):
LGTM

PR #511, opened 2023-06-01
In-app PDF renderer

ganfra:
First version of the PDF renderer, with basic scroll and zoom supports.
We could probably add more features later like search and scroll page position.


ganfra:
The failing test is the same than was failing on develop, cause I didn't merge the fix in my branch.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/media/local/pdf/PdfRendererManager.kt
```diff
@@ -0,0 +1,68 @@
...
+import kotlinx.coroutines.withContext
+
+class PdfRendererManager(
+    private val parcelFileDescriptor: ParcelFileDescriptor,
+    private val width: Int,
+    private val coroutineScope: CoroutineScope,
+) {
+
+    private val mutex = Mutex()
+    private var pdfRenderer: PdfRenderer? = null
```
jmartinesp, line 37: 👀  TIL

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/media/local/pdf/PdfPage.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    private val renderWidth = maxWidth
+    private val renderHeight: Int
+    private var loadJob: Job? = null
+
+    init {
+        pdfRenderer.openPage(pageIndex).use { page ->
+            renderHeight = (page.height * (renderWidth.toFloat() / page.width)).toInt()
+        }
+    }
```
jmartinesp, line 56: This seems a bit confusing, it looks as if the PDF renderer would move to that page instead of loading it in memory. I know it's because Android's PdfRenderer API has a weird naming, but maybe a comment could help?

jmartinesp (approved):
LGTM, thanks!

I had some issues in my real device, but as far as I could tell, it was the issue with the Rust SDK downloading to a folder with no permissions.

Issue #513, opened 2023-06-02
[Tools] Upgrade kover to 0.7.0


bmarty:
Upgrading kover to [0.7.0](https://github.com/Kotlin/kotlinx-kover/releases/tag/v0.7.0) and up requires some work since there are some API change.

Migration guide: https://github.com/Kotlin/kotlinx-kover/blob/v0.7.0/docs/gradle-plugin/migrations/migration-to-0.7.0.md

This is not urgent, and for now we can ignore this dependency so that Renovate does not try to upgrade it. I will create a quick PR about that.

bmarty:
Done in #1782

PR #514, opened 2023-06-02
Ignore kover dep

bmarty:
Let Renovate ignore Kover dep.
See #513


bmarty:
(not waiting for CI here)

PR #516, opened 2023-06-02
Fix SearchBar colors. Also fix tertiary color in dark mode.

jmartinesp:
@callumu mentioned that some colors used in `SearchBar` are wrong, so I fixed them.

I also discovered we were using the wrong tertiary color in dark mode (it was pure white before, for some reason).


csmith (approved):

callumu:
These look great, thanks for making the changes

PR #519, opened 2023-06-02
Design tweaks for create/join rooms

csmith:
This is a complete mix of small tweaks. The individual commits describe what they're doing, and the screenshot tests show the result.

The only notable things are:

- Using `PlatformTextStyle(includeFontPadding=false)` in a few places where we care about vertical alignment of text. The font padding [isn't needed with Compose and is going to default to false in the future](https://medium.com/androiddevelopers/fixing-font-padding-in-compose-text-768cd232425b) but for now we have to specify it explicitly or have a few random pixels of padding in places we don't want them :angry: 

- SelectedUsersList now has some funky logic to spread out the users a little such that the last entry in the list will always be half visible, to indicate that there are more entries and that it can be scrolled

- I noticed a bunch of real users in search results had blank avatars because their avatarUrl was empty (instead of unset), so I've changed to using initials in those cases as well.


features/invitelist/impl/src/main/res/values/localazy.xml
```diff
@@ -5,5 +5,5 @@
   <string name="screen_invites_decline_direct_chat_message">"Are you sure you want to decline to chat with %1$s?"</string>
   <string name="screen_invites_decline_direct_chat_title">"Decline chat"</string>
   <string name="screen_invites_empty_list">"No Invites"</string>
-  <string name="screen_invites_invited_you">"%1$s invited you"</string>
-</resources>
+  <string name="screen_invites_invited_you">"%1$s (%2$s) invited you"</string>
```
ElementBot, line 8: - :no_entry_sign: Inconsistent number of arguments in formatting string `screen_invites_invited_you`; found both 1 here and 2 in values/localazy.xml

jmartinesp:
I don't like the new lists design much, but matters of taste aside, I think it's a bit weird to have different designs in [this list](https://github.com/vector-im/element-x-android/pull/519/commits/7437a2348e31d22478234d0e0300982ee53f16f6#diff-2c86530f634e3533f81458171a12d70d5758027c90dd4b303dbb137d84d7dced) and [this other one](https://github.com/vector-im/element-x-android/pull/519/commits/7437a2348e31d22478234d0e0300982ee53f16f6#diff-56e7b6f629d31ce42b714bbedee76fd9bade59848d3babdcf12421adeda5e337). That said, I'll get to review the code.

csmith:
> I think it's a bit weird to have different designs in [this list](https://github.com/vector-im/element-x-android/pull/519/commits/7437a2348e31d22478234d0e0300982ee53f16f6#diff-2c86530f634e3533f81458171a12d70d5758027c90dd4b303dbb137d84d7dced) and [this other one](https://github.com/vector-im/element-x-android/pull/519/commits/7437a2348e31d22478234d0e0300982ee53f16f6#diff-56e7b6f629d31ce42b714bbedee76fd9bade59848d3babdcf12421adeda5e337).

Oh, you're right. The invite screen is [meant to have dividers too](https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=4058-166572&t=N81Virim3RTJGwTx-0). I'll fix.

features/invitelist/impl/src/main/kotlin/io/element/android/features/invitelist/impl/components/InviteSummaryRow.kt
```diff
@@ -145,17 +145,17 @@ internal fun DefaultInviteSummaryRow(
                 OutlinedButton(
                     content = { Text(stringResource(StringR.string.action_decline), style = ElementTextStyles.Button) },
                     onClick = onDeclineClicked,
-                    modifier = Modifier.weight(1f),
-                    contentPadding = PaddingValues(horizontal = 24.dp, vertical = 7.dp),
+                    modifier = Modifier.weight(1f).heightIn(max = 36.dp),
```
jmartinesp, line 148: I believe this min `40.dp` is because Material Design components are supposed to have a minimum touch target size of 40px, but if the design team gave it the OK it's not that much of a difference.

jmartinesp (approved):
LGTM, just some minor suggestions and a comment.

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootView.kt
```diff
@@ -185,11 +191,15 @@ fun CreateRoomActionButton(
         verticalAlignment = Alignment.CenterVertically,
     ) {
         Icon(
-            modifier = Modifier.alpha(0.5f), // FIXME align on Design system theme (removing alpha should be fine)
```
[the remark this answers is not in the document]
csmith: Ah, nice. Done.

features/invitelist/impl/src/main/kotlin/io/element/android/features/invitelist/impl/components/InviteSummaryRow.kt
csmith: Yeah. Fortunately compose expands the touch target automagically if the component is smaller than `48dp`.

This lets you tap just above or below the button to trigger it:

image

so it basically acts like a 48dp high button  while still conforming to the designs

features/invitelist/impl/src/main/kotlin/io/element/android/features/invitelist/impl/components/InviteSummaryRow.kt
```diff
@@ -173,47 +173,36 @@ internal fun DefaultInviteSummaryRow(
...
-                        avatarData = sender.avatarData.copy(size = AvatarSize.Custom(16.dp)),
-                        modifier = Modifier.align(Alignment.Center)
-                    )
+    ) {
+        Avatar(
+            avatarData = sender.avatarData.copy(size = AvatarSize.Custom(16.dp)),
+        )
+        Text(
+            text = buildAnnotatedString {
+                val placeholder = "$"
```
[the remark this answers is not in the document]
csmith: Ah, good call. I already simplified the avatar into a Row, but I left the overly complicated annotated string builder alone :sweat_smile: 

It still needs to be annotated to make the display name bold/primary coloured, but I've made it a lot more straight forward now :+1:

PR #527, opened 2023-06-05
Preparatory work to integrate new FTUE screen

bmarty:
Create some Pages and Molecules.

### DIFF
--- a/features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/VerifySelfSessionView.kt
+++ b/features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/VerifySelfSessionView.kt
@@ -21,10 +21,10 @@ import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.systemBarsPadding
 import androidx.compose.foundation.layout.widthIn
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.TextButton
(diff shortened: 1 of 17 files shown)


features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/VerifySelfSessionView.kt
```diff
@@ -121,58 +110,22 @@ internal fun HeaderContent(verificationFlowStep: FlowStep, modifier: Modifier =
...
-        Box(
-            modifier = Modifier
-                .size(width = 70.dp, height = 70.dp)
-                .align(Alignment.CenterHorizontally)
-                .background(
-                    color = LocalColors.current.quinary,
-                    shape = RoundedCornerShape(14.dp)
-                )
-        ) {
-            Spacer(modifier = Modifier.height(68.dp))
```
bmarty, line 135: This line has been removed, since I found it useless. Did I miss something?

features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/VerifySelfSessionView.kt
```diff
@@ -302,15 +250,3 @@ private fun ContentToPreview(state: VerifySelfSessionState) {
...
-    min: Dp,
-    max: Dp,
-    minScreenHeight: Int = 720
-): Modifier = composed {
-    if (LocalConfiguration.current.screenHeightDp >= minScreenHeight) {
-        then(Modifier.height(max))
-    } else {
-        then(Modifier.height(min))
-    }
-}
```
bmarty, line 316: I also found that this was useless, because we are using weight and we are centering content, but maybe it was not. @jmartinesp any thoughts? I could move it to the design module if necessary. Let me know.
Screenshot test is still happy, but it maybe has some effect on other screen sizes?

bmarty:
(Manually assigning to @jmartinesp since I have got questions for him :) )

features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/VerifySelfSessionView.kt
jmartinesp: I think this was done to fix some weird issues in small displays, but I think those were already noticeable in the screenshots tests, so if we don't see them maybe it's fixed now. There were some changes to use a fixed number or emojis per row to make sure the content size was constant, so maybe that helped with this issue too.

features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/VerifySelfSessionView.kt
jmartinesp: I can't remember, to be honest. If you test it in a small display (Nexus 5?) and it works fine, then it's probably not needed anymore.

jmartinesp (changes requested):
I got this in a Nexus 5 emulator with these changes, I believe this is the issue the `shrinkableHeightModifier` was supposed to fix.

Screenshot_1686038871

bmarty:
This is the preview on develop with `heightDp = 640` for the preview and there is no scroll (I think):

image

So I am not sure what's worse.

We may iterate later on this screen to add better support for small screens?

Also I have heard from @BillCarsonFr  that verification with emojis will disappear in favor of verification with digits. I do not know if this is still a plan though.

jmartinesp:
You're right, it's also broken on `develop`. This was working fine on a Nexus 5 at the time as far as I recall, so we probably changed some component later that affected this screen too.

Anyway, we can probably re-visit it later since the new code is simpler and doesn't make the issue worse.

PR #530, opened 2023-06-05
Use member count instead of counting members

csmith:
For the room details screen, use the member count as supplied by
matrix instead of waiting for the entire member list to be
retrieved and then manually adding up all the relevant users.

This removes the loading state of the member count, relying on
a spinner on the member list itself if the user actually wants
to see the members. (The performance of that will be improved
separately on the rust side in the future)

Closes https://github.com/vector-im/element-x-android/issues/505

## Before

[device-2023-06-05-161736.webm](https://github.com/vector-im/element-x-android/assets/189133/3b32040b-3961-469c-9701-d9410acc3a20)

## After

[device-2023-06-05-162221.webm](https://github.com/vector-im/element-x-android/assets/189133/496a13d8-888b-40a7-be00-9c571fc4de78)


csmith:
I deleted one of the screenshot tests (which tested the loading state for the member count) and the remaining ones got reordered, so the diffs are a bit weird

jmartinesp (approved):
LGTM, thanks for the simplification. When we built this, Mauro was trying to get this value from the Rust side, but it seemed to be a bit flaky at the time. I guess it has improved since then, since they seem to be using the same input on their side now.

Also, are we sure we want to display the sum of both the joined and invited members and not just the joined member count?

csmith:
> I guess it has improved since then, since they seem to be using the same input on their side now.

Yeah, it seems good now.

> Also, are we sure we want to display the sum of both the joined and invited members

No, very unsure! :joy:  Fixed, thanks.

PR #526, opened 2023-06-05
[Message actions] New UI for message composer and editing

jmartinesp:
## Changes

- Replaces the Rich Text Editor implementation of the message composer with a simplified Compose TextField, as the current designs don't include the RTE features at the moment. The plan is to integrate them again with some proper Compose components in the near future.
- Adds a custom 'editing' composer mode UI.

## Why

Closes #484 .

## Screenshots

| Light mode | Dark mode |
| - | - |
| ||

*Note:* for some reason, the 'multline' corner radius sometimes doesn't render properly in the screenshots, as seen above.

PR #545, opened 2023-06-06
[Message actions] New UI for replies

### LINKED ISSUE #485: [Message actions] Update composer & messages UI for replies
## Changes

Follow the [designs in Figma](https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=4367%3A168017&t=ZYzKB8ZPEHPfZTXN-1).

- Display a 'preview' of the message to reply to in the composer.
- Display a similar preview in messages that reply to another message.

jmartinesp:
## Changes

- Add 'reply to' message summaries to both the composer and the timeline items.
- Set min size for images in the timeline.
- Create separate layouts for items with only the media item (media items replying to another message use the regular layout).
- Add `EqualWidthColumn` for chat bubble contents.
- Move `TimelineItem___Row` views to separate files since `TimelineView` was growing too much.
- Split previews for `TextComposer`.

## Why:

Closes #485 .

**Note:** based on #526 , so that PR needs to be merged first.

## Tests

Open a chat room and send some messages and media items (images, videos, files), then reply to those with both messages and media items.

**Note**: for some reason, it's not possible to reply with media to a message in the Rust SDK, so we can't do this in our apps yet. You can test that by using Element Web.


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -101,12 +72,25 @@ fun TimelineView(
...
     val lazyListState = rememberLazyListState()
+
+    fun inReplyToClicked(eventId: EventId) {
+        val index = state.timelineItems.indexOfFirst { it.identifier() == eventId.value }
+        if (index >= 0) {
+            coroutineScope.launch {
+                lazyListState.animateScrollToItem(index)
+            }
+        } else {
+            // TODO should we paginate first?
```
jmartinesp, line 90: I'm not sure about what we should do here, to be honest. Should we block the UI until we've paginated back to that item? What if we somehow reference a non-existing one?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemImageView.kt
```diff
@@ -28,14 +28,17 @@ import io.element.android.features.messages.impl.timeline.model.event.TimelineIt
...
 import io.element.android.libraries.matrix.ui.media.MediaRequestData
+import kotlin.math.max
 
 @Composable
 fun TimelineItemImageView(
     content: TimelineItemImageContent,
     modifier: Modifier = Modifier,
 ) {
+    // TODO place this value somewhere else?
+    val minHeight = max(100, content.height ?: 0)
```
jmartinesp, line 39: Do we have somewhere where we add these kind of constants?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/messagesummary/MessageSummaryFormatter.kt
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.features.messages.impl.utils.messagesummary
+
+import io.element.android.features.messages.impl.timeline.model.TimelineItem
+
+interface MessageSummaryFormatter {
+    fun format(event: TimelineItem.Event): String
```
jmartinesp, line 22: This references an `impl` only class, so I had to add it to the `impl` module.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/ColorsDark.kt
```diff
@@ -46,6 +46,7 @@ fun elementColorsDark() = ElementColors(
     gray400 = Compound_Gray_400_Dark,
     gray1400 = Compound_Gray_1400_Dark,
     textActionCritical = TextColorCriticalDark,
+    accentColor = Color(0xFF0DBD8B),
```
jmartinesp, line 49: It's the same color in both themes.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/MatrixTimelineItemMapper.kt
```diff
@@ -30,6 +37,12 @@ class MatrixTimelineItemMapper(
         val asEvent = it.asEvent()
         if (asEvent != null) {
             val eventTimelineItem = eventTimelineItemMapper.map(asEvent)
+
+
+            if (eventTimelineItem.hasNotLoadedInReplyTo() && eventTimelineItem.eventId != null) {
```
jmartinesp, line 42: Maybe it's not the best point to do this? It's where iOS fetches this data, but maybe we can do this in `MatrixClient` instead.

jmartinesp:
Since I made some changes to add extra mappings to timeline events and message contents, maybe it would be nice to have an experienced second pair of eyes taking a look at it: can you do a quick review @ganfra ?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
ganfra: We should be able to call the context API somehow with the rust-sdk, so it'll fail if the event doesn't exist (which should not happen I guess?)

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemImageView.kt
ganfra: Not for now, but we could have dimension param in the ElementTheme?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
jmartinesp: We're having a kick off meeting with iOS and design today so I'll ask about how this should work both UX and logic wise.

jmartinesp: Apparently we're removing this logic altogether for now since we can't just back-paginate indefinitely.

ganfra: But like I said, the `context` api should return the event and some of the prev and next events. It's what is used for permalink too. But probably not available yet in the rust sdk?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
```diff
@@ -0,0 +1,361 @@
...
+        ContentAndTimestampView(overlayTimestamp = true, modifier = modifier)
+    }
+
+    /** Used for every other type of message, groups the different components in a Column with some space between them. */
+    @Composable
+    fun CommonLayout(
+        inReplyToDetails: InReplyTo.Ready?,
+        modifier: Modifier = Modifier
+    ) {
+        EqualWidthColumn(modifier = modifier, spacing = 8.dp) {
```
ganfra, line 250: Should we not avoid using `EqualWidthColumn` if the inReply is null?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
jmartinesp: Yes, I think SlidingSync can only handle back pagination right now and this would be like jumping in time. We'd probably have to add some kind of "load next/load previous" marks between the latest items and those retrieved with the API.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
jmartinesp: I just changed `EqualWidthColumn` so it doesn't re-measure everything if all the items already have the same width. If we only have 1 child (this case) that should already be the case and it will just work like a simple Column.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/BlurHashAsyncImage.kt
```diff
@@ -0,0 +1,100 @@
...
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.asImageBitmap
+import androidx.compose.ui.layout.ContentScale
+import coil.compose.AsyncImage
+import com.vanniktech.blurhash.BlurHash
+
+@Composable
+fun BlurHashAsyncImage(
```
julioromano, line 40: This can be simplified down to something like:

```
@Composable
fun BlurHashAsyncImage(
    model: Any?,
    blurHash: String?,
    contentDescription: String?,
    contentScale: ContentScale = ContentScale.Fit,
) {
    AsyncImage(
        model = ImageRequest.Builder(LocalContext.current)
            .data(model)
            .crossfade(true)
            .build(),
        contentDescription = contentDescription,
        modifier = modifier,
        placeholder = blurHash?.let { hash ->
            BlurHash.decode(hash, 10, 10)?.asImageBitmap()?.let(::BitmapPainter)
        },
        contentScale = contentScale,
    )
}
```

Previews also work with this impl so we can add one.
What d'you think?

Or perhaps do you think this is too trivial and we can just create an utility function to build the Painter from a blurhash string?
Something like:
```
fun blurHashPainter(blurHash: String?): Painter? = blurHash?.let { hash ->
    BlurHash.decode(hash, 10, 10)?.asImageBitmap()?.let(::BitmapPainter)
}
```
And then use it with Coil's AsyncImage?

julioromano (commented):
Just some suggestions wherever I could be hopefully useful

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/BlurHashAsyncImage.kt
jmartinesp: Actually, I just moved this class to the design system module from a feature one since it's needed in several places. I don't mind simplifying the implementation if it has a similar UX.

julioromano: > Actually, I just moved this class to the design system module from a feature one since it's needed in several places. I don't mind simplifying the implementation if it has a similar UX.

The preview looks the same, I didn't check the crossfade animation though.

julioromano: There's one difference though, the original `BlurHashAsyncImage` applies `fillMaxSize()` to its inner AsyncImage.
We could do the same here but I just thought it'd be too opinionated for being a library composable.

julioromano: Anyhow I thought this was a new class, not a move, if you prefer I can do a separate rework PR once this is merged

jmartinesp: Yeah, that seems for the best. Also, the cross fade sometimes looks kind of weird, starting from a black square.

jmartinesp:
@julioromano  I uploaded some changes to address your comments, mostly removing unnecessary/duplicate code and resources.

julioromano (approved):

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
ganfra: Ok nice, thanks!

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/BlurHashAsyncImage.kt
julioromano: > Is the Drawable being released from memory by the placeholder api?
Nope

> Also I think the painter should be remembered?
Yepp

Good thing Jorge didn't listen to me ;)

julioromano: @ganfra Why the need of calling `Bitmap.recycle()` though? Its doc string says it shouldn't normally be called.

PR #546, opened 2023-06-06
GH Workflows: move deps check, fix nightly

csmith:
Move the dependency check job from the quality workflow that runs on every PR to the nightly reports workflow. This sometimes flakes as it does a _lot_ of HTTP requests. It's less intrusive if that happens outside of the PR workflow.

Also change the nightly job that runs tests to use LFS, so it actually has access to screenshots :)

The sonar task is still broken, but it seems like an upstream bug: https://community.sonarsource.com/t/sonar-plugin-4-1-0-3113-4-2-0-3129-errors/91568


jmartinesp (approved):
LGTM!

PR #551, opened 2023-06-06
Feature/fga/media viewer actions

ganfra:
This PR handle the following points :
- Add actions on MediaViewer (OpenWith, SaveOnDisk and ShareWith)
- Display a placeholder for non-renderable files in MediaViewer
- Makes some minor adjustments over media

Figma : https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=5117-177717&t=XcF9ru9U2jGOLaJS-0


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/media/local/AndroidLocalMediaActions.kt
```diff
@@ -0,0 +1,161 @@
...
+class AndroidLocalMediaActions @Inject constructor(
+    @ApplicationContext private val context: Context,
+    private val coroutineDispatchers: CoroutineDispatchers,
+    private val buildMeta: BuildMeta,
+) : LocalMediaActions {
+
+    private var activityContext: Context? = null
+
+    @Composable
+    override fun Configure() {
```
ganfra, line 55: Not sure this is the best solution, but it was quite fast to handle like this... the Intent.Chooser are not working properly without an Activity Context.
We could inject the activity in the DI but it'll need a lot of changes.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/media/local/LocalMediaView.kt
```diff
@@ -186,3 +212,50 @@ fun MediaPDFView(
...
+            Box(
+                modifier = Modifier
+                    .size(72.dp)
+                    .clip(CircleShape)
+                    .background(MaterialTheme.colorScheme.onBackground),
+                contentAlignment = Alignment.Center,
+            ) {
+                Icon(
+                    imageVector = Icons.Outlined.Attachment,
+                    contentDescription = "OpenFile",
```
csmith, line 239: Should be a string resource so it gets translated

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/media/local/LocalMediaView.kt
```diff
@@ -186,3 +212,50 @@ fun MediaPDFView(
...
+fun MediaFileView(
+    uri: Uri?,
+    info: MediaInfo?,
+    onReady: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    LaunchedEffect(Unit) {
+        if(uri != null) {
+            onReady()
+        }
```
csmith, line 226: Probably out of scope for this PR, but this kind of callback feels a bit weird to me.

I think the more typical way would be to pass in a mutable state object, like how a `LazyColumn` takes a `LazyListState` that is populated with its state?

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/utils/Snackbar.kt
```diff
@@ -56,14 +59,33 @@ fun handleSnackbarMessage(
...
+            launch {
                 snackbarDispatcher.clear()
             }
         }
     }
     return snackbarMessage
 }
 
+@Composable
+fun rememberSnackbarHostState(snackbarMessage: SnackbarMessage?): SnackbarHostState {
```
csmith, line 71: Nice! :heart_eyes:

csmith (approved):
Nice, thanks. LGTM with a few small comments.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/media/local/AndroidLocalMediaActions.kt
csmith: Yeah, agree DI'ing the activity seems like the best approach long term but doesn't make sense in this PR. Maybe add a todo or open an issue so we can track it?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/media/helper/fileExtensionAndSize.kt
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.features.messages.impl.media.helper
+
+fun formatFileExtensionAndSize(name: String, size: String?): String {
+    val fileExtension = name.substringAfterLast('.', "").uppercase()
```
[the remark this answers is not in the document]
ganfra: Thats a good point, I guess we could check something yeah. Let me think about that.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/media/local/LocalMediaView.kt
ganfra: I agree, I can change that.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/media/local/LocalMediaView.kt
ganfra: I think there is no need for contentDescription here as it doesn't convey anything important, I'll remove it

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/media/helper/fileExtensionAndSize.kt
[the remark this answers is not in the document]
ganfra: Extension file in unix system can be anything, so I don't know what we should do exactly... any thoughts @amshakal ?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/media/viewer/MediaViewerPresenter.kt
```diff
@@ -68,29 +79,93 @@ class MediaViewerPresenter @AssistedInject constructor(
...
             }.onSuccess {
                 localMedia.value = Async.Success(it)
             }.onFailure {
                 localMedia.value = Async.Failure(it)
             }
     }
+
+    private fun CoroutineScope.saveOnDisk(localMedia: Async<LocalMedia>) = launch {
+        when (localMedia) {
+            is Async.Success -> {
```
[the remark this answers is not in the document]
ganfra: I don't have strong preference over this, but maybe early return is ok in this scenario?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/media/helper/fileExtensionAndSize.kt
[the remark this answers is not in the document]
ganfra: Looks like other messenger apps are checking some validity and defaulting to BIN if not.

Could do something like that :

```
val fileExtension = name.substringAfterLast('.', "")
    val safeExtension = if (MimeTypeMap.getSingleton().hasExtension(fileExtension)) {
        fileExtension.uppercase()
    } else {
        "BIN"
    }
```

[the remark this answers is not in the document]
amshakal: This is a tough one. How do we handle this today?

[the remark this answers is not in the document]
ganfra: What do you mean by today? In Element Android?

PR #553, opened 2023-06-07
Fix changing room avatar from details screen

### LINKED ISSUE #550: Editing name / topic from details failed

csmith:
The presenter was expecting the MediaProcessor to return a MediaUploadInfo.Image, but it actually returns
MediaUploadInfo.AnyFile because we're not compressing avatars (so it doesn't process the file and return more detailed info).

This check/cast was entirely pointless, so change to just working on whatever we're given. The pickers constrain which types of file the user select, so we should be reasonably happy the files are images.

Also actually log error details when updating the details, so we know what's going wrong.

Closes #550


bmarty (approved):
LGTM!

PR #568, opened 2023-06-08
[Message Actions] Add emoji reactions option

### LINKED ISSUE #563: [Message Actions] Add emoji reactions option to the context menu in the timeline
Changes needed:

- Send a reaction when one of the emojis is clicked in the context menu of the timeline.
- Allow selecting an arbitrary emoji from a picker in that same menu.
- Mark already sent reactions as highlighted, so clicking on them again will remove the reaction.

Out of scope:

- Displaying emojis based on past usage.

jmartinesp:
## Changes

- Added logic to send reactions for events.
- Added [Vanniktech's Emoji library](https://github.com/vanniktech/Emoji) for the emoji picker.
- Created component `EmojiPicker` based on that.
- Added a wrapping `CustomReactionBottomSheet` for the picker.
- Fixed `ModalBottomSheet` implementation in `MessagesView`, which broke animations.

## Why

Closes #563 .

## Screenshots

| Light mode | Dark mode |
| - | - |
|Screenshot_1686296492|Screenshot_1686296521|

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
@@ -131,6 +131,12 @@ fun MessagesView(
         }
     }
 
+    fun onEmojiReactionClicked(emoji: String, event: TimelineItem.Event) {
+        if (event.eventId == null) return
+        isMessageActionsBottomSheetVisible = false
+        state.eventSink(MessagesEvents.SendReaction(emoji, event.eventId))
+    }
+
     fun onDismissActionListBottomSheet() {
         isMessageActionsBottomSheetVisible = false
     }
@@ -172,7 +178,8 @@ fun MessagesView(
         state = state.actionListState,
         isVisible = isMessageActionsBottomSheetVisible,
         onDismiss = ::onDismissActionListBottomSheet,
-        onActionSelected = ::onActionSelected
+        onActionSelected = ::onActionSelected,
+        onEmojiReactionClicked = ::onEmojiReactionClicked,
     )
 }
 
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
@@ -34,8 +34,11 @@ import androidx.compose.foundation.layout.wrapContentHeight
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.ArrowBack
 import androidx.compose.material3.ExperimentalMaterial3Api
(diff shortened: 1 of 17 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/actionlist/ActionListView.kt
```diff
@@ -118,7 +133,9 @@ fun ActionListView(
 private fun SheetContent(
```
ElementBot, line 133: - :warning: Parameters in a composable function should be ordered following this pattern: params without defaults, modifiers, params with defaults and optionally, a trailing function that might not have a default param.
Current params are: [state: ActionListState, modifier: Modifier = Modifier, onActionClicked: (TimelineItemAction) -> Unit, onEmojiReactionClicked: (String) -> Unit, onCustomReactionClicked: () -> Unit] but should be [state: ActionListState, onActionClicked: (TimelineItemAction) -> Unit, onEmojiReactionClicked: (String) -> Unit, modifier: Modifier = Modifier, onCustomReactionClicked: () -> Unit].

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
```diff
@@ -168,11 +186,44 @@ fun MessagesView(
...
     ActionListView(
         state = state.actionListState,
+        sheetState = actionListViewBottomSheetState,
         isVisible = isMessageActionsBottomSheetVisible,
-        onDismiss = ::onDismissActionListBottomSheet,
-        onActionSelected = ::onActionSelected
+        onDismiss = { coroutineScope.launch { onDismissActionListBottomSheet() } },
+        onActionSelected = ::onActionSelected,
+        onCustomReactionClicked = { event ->
+            reactingToEventId = event.eventId
```
csmith, line 197: The data flow here feels a bit weird. Should this be pulled into the state maybe? :thinking:

csmith (approved):
Nice, LGTM

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
jmartinesp: Are you talking about the `reactingToEventId` being declared here? I thought about moving it to some *State class, but that required creating a dedicated event and it didn't seem to be worth it.

csmith: Yeah, I found it a bit hard to follow where it was being set/unset/used as they're all in callbacks on different components. I think having it in the state with event(s) as needed would be a little clearer, but I don't feel terribly strongly about it -- if you don't think it's worth it then happy for it to go as is :)

PR #573, opened 2023-06-12
Fix timeline item padding and corners

jmartinesp:
Changes:

- Add content paddings to `TimelineView` of 16dp horizontally, following the designs.
- Make the offset in `MessageEventBubble` for incoming messages of just 16dp too (was 20dp before).
- Change corner radius in `MessageEventBubble` to 12dp to match the designs (previously it was 16dp).

The Figma designs are [here](https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=1819%3A99356&t=lhwd9NNTRGMs5kB0-1).


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/MessageEventBubble.kt
```diff
@@ -88,7 +89,7 @@ fun MessageEventBubble(
         return if (state.isMine) {
             this
         } else {
-            offset(x = 20.dp)
+            offset(x = BUBBLE_OWN_OFFSET)
         }
```
csmith, line 93: Is this constant named right? It seems like it's being used for items that are _not_ our own? :thinking:

csmith (approved):

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/MessageEventBubble.kt
jmartinesp: Ooops, nice catch.

jmartinesp: Fixed in https://github.com/vector-im/element-x-android/pull/573/commits/769fe337ebab6772fed88c48b385cd084848e53c.

PR #594, opened 2023-06-14
Extract dependencies from group `com.google.auto.service` into libs.versions.toml`

bmarty:
To avoid multiple PR to upgrade the 2 deps such has #581 and #583.


jmartinesp (approved):
LGTM

PR #596, opened 2023-06-14
[Message Actions] Retry sending failed messages

### LINKED ISSUE #487: [Message actions] Allow re-sending failed messages
## Changes

Implement the [designs from Figma](https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=4367%3A168017&t=ZYzKB8ZPEHPfZTXN-1).

If a message failed to send, we should allow launching a manual retry when selecting the message.

jmartinesp:
## Changes

- Added a retry sending menu that appears when tapping on the timestamp view of a failed to send message: it allows you to retry sending the message or delete them from your local echoes.
- Fixed an issue that retrieved the original state of the event instead of the current one when using this feature, so the `sendState` sometimes wasn't updated.
- Added extra touch area to the timestamp.

## Why

Fixes #487.

## Test

- Open a room.
- Enable airplane mode (disable any network connection).
- Send a couple of messages.
- Disable airplane mode, wait until you have internet connection.
- Retry sending one of the messages and removing the other one.

It could be interesting to also try re-sending the message with no network connection, although it worked fine in my tests.

**Note:** there is an issue currently that causes the sliding sync loop to be stopped when network connection is lost and it never recovers. This means that sending messages from the previously offline device works, but you won't see any incoming messages or the room list updated.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesStateProvider.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesStateProvider.kt
@@ -21,6 +21,7 @@ import io.element.android.features.messages.impl.actionlist.anActionListState
 import io.element.android.features.messages.impl.messagecomposer.aMessageComposerState
 import io.element.android.features.messages.impl.timeline.aTimelineItemList
 import io.element.android.features.messages.impl.timeline.aTimelineState
+import io.element.android.features.messages.impl.timeline.components.retrysendmenu.RetrySendMenuState
 import io.element.android.features.messages.impl.timeline.model.event.aTimelineItemTextContent
 import io.element.android.libraries.core.data.StableCharSequence
 import io.element.android.libraries.designsystem.components.avatar.AvatarData
@@ -47,6 +48,10 @@ fun aMessagesState() = MessagesState(
     ),
     timelineState = aTimelineState().copy(
         timelineItems = aTimelineItemList(aTimelineItemTextContent()),
+    ),
+    retrySendMenuState = RetrySendMenuState(
+        selectedEvent = null,
+        eventSink = {},
     ),
     actionListState = anActionListState(),
     hasNetworkConnection = true,
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
@@ -34,14 +34,10 @@ import androidx.compose.foundation.layout.wrapContentHeight
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.ArrowBack
 import androidx.compose.material3.ExperimentalMaterial3Api
-import androidx.compose.material3.SheetState
-import androidx.compose.material3.SheetValue
 import androidx.compose.material3.SnackbarHost
-import androidx.compose.material3.SnackbarHostState
 import androidx.compose.material3.rememberModalBottomSheetState
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
-import androidx.compose.runtime.SideEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
@@ -59,7 +55,6 @@ import androidx.compose.ui.tooling.preview.PreviewParameter
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import io.element.android.features.messages.impl.actionlist.ActionListEvents
-import io.element.android.features.messages.impl.actionlist.ActionListState
 import io.element.android.features.messages.impl.actionlist.ActionListView
 import io.element.android.features.messages.impl.actionlist.model.TimelineItemAction
 import io.element.android.features.messages.impl.attachments.Attachment
@@ -67,6 +62,8 @@ import io.element.android.features.messages.impl.messagecomposer.AttachmentsStat
 import io.element.android.features.messages.impl.messagecomposer.MessageComposerView
 import io.element.android.features.messages.impl.timeline.TimelineView
 import io.element.android.features.messages.impl.timeline.components.CustomReactionBottomSheet
+import io.element.android.features.messages.impl.timeline.components.retrysendmenu.RetrySendMenuEvents
+import io.element.android.features.messages.impl.timeline.components.retrysendmenu.RetrySendMessageMenu
 import io.element.android.features.messages.impl.timeline.model.TimelineItem
 import io.element.android.features.networkmonitor.api.ui.ConnectivityIndicatorView
 import io.element.android.libraries.androidutils.ui.hideKeyboard
@@ -85,6 +82,7 @@ import io.element.android.libraries.designsystem.utils.rememberSnackbarHostState
 import io.element.android.libraries.matrix.api.core.EventId
 import io.element.android.libraries.matrix.api.core.UserId
 import io.element.android.libraries.matrix.api.timeline.item.TimelineItemDebugInfo
+import io.element.android.libraries.matrix.api.timeline.item.event.EventSendState
 import kotlinx.collections.immutable.ImmutableList
 import kotlinx.coroutines.launch
 import timber.log.Timber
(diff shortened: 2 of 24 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/retrysendmenu/RetrySendMessageMenu.kt
```diff
@@ -0,0 +1,134 @@
...
+            sheetState = sheetState,
+            onDismissRequest = {
+                coroutineScope.launch {
+                    sheetState.hide()
+                    onDismiss()
+                }
+            }
+        ) {
+            ListItem(headlineContent = {
+                Text("Your message failed to send", fontWeight = FontWeight.Medium)
```
bmarty, line 92: i18n ?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/retrysendmenu/RetrySendMessageMenu.kt
```diff
@@ -0,0 +1,134 @@
...
+                modifier = Modifier.clickable {
+                    coroutineScope.launch {
+                        sheetState.hide()
+                        onRetry()
+                    }
+                }
+            )
+            ListItem(
+                headlineContent = {
+                    Text("Remove")
```
bmarty, line 107: i18n ?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/TimelineItem.kt
```diff
@@ -52,6 +52,7 @@ sealed interface TimelineItem {
     data class Event(
         val id: String,
         val eventId: EventId? = null,
+        val transactionId: String? = null,
```
bmarty, line 55: Does it worth creating a `TransactionId` value class, as per `EventId`?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
```diff
@@ -264,6 +265,20 @@ class RustMatrixRoom(
         }
     }
 
+    override suspend fun retrySendMessage(transactionId: String): Result<Unit> =
+        withContext(Dispatchers.IO) {
```
bmarty, line 269:
```suggestion
        withContext(coroutineDispatchers.io) {
```

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesStateProvider.kt
```diff
@@ -48,6 +49,10 @@ fun aMessagesState() = MessagesState(
     timelineState = aTimelineState().copy(
         timelineItems = aTimelineItemList(aTimelineItemTextContent()),
     ),
+    retrySendMenuState = RetrySendMenuState(
+        selectedEvent = null,
+        eventSink = {},
+    ),
```
bmarty, line 55: Maybe create a `fun aRetrySendMenuState(): RetrySendMenuState`

bmarty (commented):
Nice work, thanks!
A few remarks after a static review.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
```diff
@@ -245,9 +252,10 @@ private fun AttachmentStateView(
 fun MessagesViewContent(
```
ElementBot, line 252: - :warning: Parameters in a composable function should be ordered following this pattern: params without defaults, modifiers, params with defaults and optionally, a trailing function that might not have a default param.
Current params are: [state: MessagesState, modifier: Modifier = Modifier, onMessageClicked: (TimelineItem.Event) -> Unit, onUserDataClicked: (UserId) -> Unit, onMessageLongClicked: (TimelineItem.Event) -> Unit, onTimestampClicked: (TimelineItem.Event) -> Unit] but should be [state: MessagesState, onMessageClicked: (TimelineItem.Event) -> Unit, onUserDataClicked: (UserId) -> Unit, onMessageLongClicked: (TimelineItem.Event) -> Unit, modifier: Modifier = Modifier, onTimestampClicked: (TimelineItem.Event) -> Unit].

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -69,9 +69,10 @@ import kotlinx.coroutines.launch
 fun TimelineView(
```
ElementBot, line 69: - :warning: Parameters in a composable function should be ordered following this pattern: params without defaults, modifiers, params with defaults and optionally, a trailing function that might not have a default param.
Current params are: [state: TimelineState, modifier: Modifier = Modifier, onUserDataClicked: (UserId) -> Unit, onMessageClicked: (TimelineItem.Event) -> Unit, onMessageLongClicked: (TimelineItem.Event) -> Unit, onTimestampClicked: (TimelineItem.Event) -> Unit] but should be [state: TimelineState, onUserDataClicked: (UserId) -> Unit, onMessageClicked: (TimelineItem.Event) -> Unit, onMessageLongClicked: (TimelineItem.Event) -> Unit, modifier: Modifier = Modifier, onTimestampClicked: (TimelineItem.Event) -> Unit].

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/retrysendmenu/RetrySendMessageMenu.kt
jmartinesp: Oops. I got so focused on trying to fix the UX issues that I completely forgot to change these placeholders. Thanks!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/TimelineItem.kt
jmartinesp: I'm not sure if there is some format we should enforce here. If there isn't, I don't think it's worth it.

tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.messages.impl.timeline.components.retrysendmenu_null_DefaultGroup_RetrySendMessageMenuPreview_0_null,NEXUS_5,1.0,en].png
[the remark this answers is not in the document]
Comment (jmartinesp) [reply tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.messages.impl.timeline.components.retrysendmenu_null_DefaultGroup_RetrySendMessageMenuPreview_0_null,NEXUS_5,1.0,en].png:1]: I can create the states, but `ModalBottomSheet` can't be previewed at the moment, it seems to be a bug in the material library.

jmartinesp:
I think I addressed all the issues, could you take another look @bmarty ?

libraries/matrix/test/src/main/kotlin/io/element/android/libraries/matrix/test/TestData.kt
```diff
@@ -37,6 +37,7 @@ val A_ROOM_ID_2 = RoomId("!aRoomId2:domain")
 val A_THREAD_ID = ThreadId("\$aThreadId")
 val AN_EVENT_ID = EventId("\$anEventId")
 val AN_EVENT_ID_2 = EventId("\$anEventId2")
+val A_TRANSACTION_ID = "aTransactionId"
```
ElementBot, line 40: - :warning: A_TRANSACTION_ID can be a `const val`.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/TimelineItem.kt
bmarty: It just to have more typed code, and so enforce usage at compilation time. Using `String` every where can lead to errors.

bmarty (approved):
Thanks for the update. Just a remark about creating `TransactionId` type. Not a blocker though!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/TimelineItem.kt
ganfra: I don't think this is an important value like `EventId` or `RoomId`, so I'd keep it as `String` IMO.

PR #597, opened 2023-06-15
Feature/fga/fix reply timeline UI

ganfra:
2 small fixes around reply in timeline : 
- Replied sender name was using wrong one 
- Ripple on reply view was missing the rounded clip


jmartinesp (approved):
LGTM, thanks for the fixes!

PR #598, opened 2023-06-15
Inline Async extension functions

julioromano:
Just a tiny perf improvement.

### DIFF
--- a/libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
+++ b/libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
@@ -36,17 +36,20 @@ sealed interface Async<out T> {
     }
 }
 
-suspend fun <T> (suspend () -> T).execute(state: MutableState<Async<T>>, errorMapping: ((Throwable) -> Throwable)? = null) {
+suspend inline fun <T> (suspend () -> T).execute(
+    state: MutableState<Async<T>>,
+    errorMapping: ((Throwable) -> Throwable) = { it },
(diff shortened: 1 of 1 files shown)


libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
```diff
@@ -36,17 +36,20 @@ sealed interface Async<out T> {
     }
 }
 
-suspend fun <T> (suspend () -> T).execute(state: MutableState<Async<T>>, errorMapping: ((Throwable) -> Throwable)? = null) {
```
julioromano, line 39: I was also thinking if it'd be useful to offer these APIs as extensions over `MutableState<Async<T>>` too.
This could make code like:
```
suspend {
    mediaSender.sendMedia(mediaAttachment.localMedia.uri, mediaAttachment.localMedia.info.mimeType, mediaAttachment.compressIfPossible)
}.executeResult(sendActionState)
```

also writeable as:
```
sendActionState.executeResult {
    mediaSender.sendMedia(mediaAttachment.localMedia.uri, mediaAttachment.localMedia.info.mimeType, mediaAttachment.compressIfPossible)
}
```

Not actually sure it it looks better or more ergonomical, maybe it's just an impression of mine. Looking for feedback in this respect.

bmarty: The form `sendActionState.executeResult` looks nice to me.

ganfra (approved):

PR #599, opened 2023-06-15
Feature/fga/timeline media improvements

ganfra:
This PR offers 2 improvements on timeline media handling: 
- Use thumbnailSource  for images when available instead of mediaSource 
- Use `loadMediaFile` instead of `loadMediaContent` with Coil. This will reduce the memory pressure and avoid lots of GC in the timeline.


jmartinesp (approved):
LGTM! 

File thumbnails are now displayed a bit differently, making them look wrong in the timeline, but they never looked that great to being with, so it's something we can fix later.

An example:

image

PR #600, opened 2023-06-15
Move logic of different BottomSheets in `MessagesView` to presenters

jmartinesp:
## Changes

As @csmith mentioned in a previous PR, having the logic for displaying these bottom sheet menus in `MessagesView` was kind of awkward so as I found a small issue to find with `ActionListView` when displaying an event's debug info, it looked like  the perfect time to solve this.

- Split `CustomReactionBottomSheet` from `EmojiPicker` and add a presenter to it.
- Improve `ActionListPresenter` and use it as a source of truth to display the bottom sheet.
- Add `SheetState.hide(coroutineScope: CoroutineScope, then: suspend () -> Unit)` to try to unify how we dismiss modal bottom sheets.
- Remove lots of state code from `MessagesView`.


ganfra (approved):
LGTM, thanks for splitting things!

PR #617, opened 2023-06-16
Feature/fga/fix media pre processing

ganfra:


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemAspectRatioBox.kt
```diff
@@ -24,22 +24,23 @@ import androidx.compose.runtime.Composable
...
     height: Int?,
-    aspectRatio: Float,
+    aspectRatio: Float?,
     modifier: Modifier = Modifier,
     contentAlignment: Alignment = Alignment.TopStart,
     content: @Composable BoxScope.() -> Unit,
 ) {
-    // TODO should probably be moved to an ElementTheme.dimensions
-    val maxHeight = min(300, height ?: 0)
+    val maxHeight = minOf(300, maxOf(100, height ?: Int.MAX_VALUE))
```
bmarty, line 36: coercein?

Also could be nice to have some const for 100 and 300 rather than using magic number. Thanks!

bmarty (approved):
LGTM, thanks!

PR #627, opened 2023-06-17
Add analytics events for room creation

Johennes:


Johennes:
CC @giomfo

Johennes:
I wanted to add the screen tracking in this PR, too, but I think we'll probably want a convenience function to easily do that in all screens. Sadly, I couldn't quite figure out the magic to inject the analytics service into a standalone function (or if that would be the right thing to do to begin with). @csmith maybe you could advise on this?

csmith:
> I wanted to add the screen tracking in this PR, too, but I think we'll probably want a convenience function to easily do that in all screens. Sadly, I couldn't quite figure out the magic to inject the analytics service into a standalone function (or if that would be the right thing to do to begin with). @csmith maybe you could advise on this?

I wonder if we can just hook in to the navigation library to get notified of every screen view.  The docs mention a [child aware interactor](https://bumble-tech.github.io/appyx/apps/childaware/) which sounds like it might be useful, but the examples don't make much sense to me :thinking: 

If we can't do it in a single place, then I think it should just be done in each Node, and they can just have the AnalyticsService injected as they're all relatively small and focused.

(There's no way to inject something into a single method, the only way to keep it isolated would be to have some global state somewhere. Sometimes that's the least-bad option, but I don't think we need to resort to it here)

csmith (approved):
I snuck in some tests, everything else LGTM :)

Johennes:
> I snuck in some tests

Nice!! Thank you! 🤩

Johennes:
> I wonder if we can just hook in to the navigation library to get notified of every screen view. The docs mention a [child aware interactor](https://bumble-tech.github.io/appyx/apps/childaware/) which sounds like it might be useful, but the examples don't make much sense to me 🤔

I gave this a short look but I think this is beyond my Android "skills" currently. I've went for the more primitive injection via each node. Does that look ok?

csmith:
Not quite what I was thinking :grin:. I've pushed a small change. WDYT?

Johennes:
> Not quite what I was thinking 😁. I've pushed a small change. WDYT?

Oh, yes that looks better than routing the service down into the view. 🤯

Johennes:
Is the quality gate known to be broken? 0% coverage can't be correct. 🤔

csmith:
> Is the quality gate known to be broken? 0% coverage can't be correct. thinking

Sonar is new, and it doesn't seem to have passed for any PR that actually adds code :upside_down_face:. I guess that's why it's not a required status yet! [Codecov has the real coverage](https://app.codecov.io/gh/vector-im/element-x-android/pull/627) and is happy with it.

PR #631, opened 2023-06-19
Take into acount send permission (power level)

### LINKED ISSUE #602: Hide message editor when user's power level is not sufficient to send message.
This message is displayed on Element Web:

image

Currently it's possible to type and send message, the local echo is displayed, but the message is not sent (rejected by the HS), and no error is displayed.

bmarty:
Fixes #602

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
@@ -19,6 +19,7 @@ package io.element.android.features.messages.impl
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.State
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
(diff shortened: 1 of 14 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
```diff
@@ -218,4 +222,13 @@ class MessagesPresenter @Inject constructor(
             MessageComposerEvents.SetMode(composerMode)
         )
     }
+
+    @Composable
+    private fun getCanSendEvent(type: MessageEventType): State<Boolean> {
+        val canSendEvent = remember(type) { mutableStateOf(false) }
```
ganfra, line 228: I guess we could add this method to matrix-ui like [this](https://github.com/vector-im/element-x-android/blob/6497a3f86ceaa36c9e0c0571bea80e77fcacf221/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/room/MatrixRoomMembers.kt).

Also you should probably use `produceState` method (which does the same internally)

```
return produceState(initialValue = false, key1 = type){
        room.canSendEvent(type).getOrElse { false }
 }
```

ganfra (approved):
One remark otherwise LGTM

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
bmarty: Good point, thanks! a8fbcb0b1676d7a447037a68b284c2ea8c9a18e9

ganfra:
Thanks, lets merge as soon as CI is happy!

PR #633, opened 2023-06-19
Async API improvements

julioromano:
Proposing this API as discussed in: https://github.com/vector-im/element-x-android/pull/598/files#r1230664392

### DIFF
--- a/libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
+++ b/libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
@@ -36,24 +36,46 @@ sealed interface Async<out T> {
     }
 }
 
+suspend inline fun <T> MutableState<Async<T>>.execute(
+    errorMapping: (Throwable) -> Throwable = { it },
+    block: () -> T,
+): Unit = execute(state = this, errorMapping = errorMapping, block = block)
+
 suspend inline fun <T> (suspend () -> T).execute(
     state: MutableState<Async<T>>,
-    errorMapping: ((Throwable) -> Throwable) = { it },
-) {
-    try {
-        state.value = Async.Loading()
-        val result = this()
-        state.value = Async.Success(result)
-    } catch (error: Throwable) {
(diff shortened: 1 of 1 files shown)


bmarty (commented):
Playing a bit with the new API, I found something weird, see the inline comment below:

image

Text version:

```kotlin
    private suspend fun sendMedia(
        mediaAttachment: Attachment.Media,
        sendActionState: MutableState<Async<Unit>>,
    ) {
        // Original code -> OK
        suspend {
            mediaSender.sendMedia(mediaAttachment.localMedia.uri, mediaAttachment.localMedia.info.mimeType, mediaAttachment.compressIfPossible)
        }.executeResult(sendActionState)
        // Change and use `execute` instead of `executeResult` -> Not compiling -> OK
        suspend {
            mediaSender.sendMedia(mediaAttachment.localMedia.uri, mediaAttachment.localMedia.info.mimeType, mediaAttachment.compressIfPossible)
        }.execute(sendActionState)
        // Use new extension `executeResult`, compiling -> OK
        sendActionState.executeResult {
            mediaSender.sendMedia(mediaAttachment.localMedia.uri, mediaAttachment.localMedia.info.mimeType, mediaAttachment.compressIfPossible)
        }
        // Use new extension `execute`, compiling -> OK, but should be error
        sendActionState.execute {
            mediaSender.sendMedia(mediaAttachment.localMedia.uri, mediaAttachment.localMedia.info.mimeType, mediaAttachment.compressIfPossible)
        }
    }
```

WDYT?

julioromano:
A little bit unfortunate but it is working as expected.

In the last case `execute()` is invoked over a `MutableState<Async<Unit>>` so the return type of the lambda block has to be `Unit`.
In Kotlin lambda implicitly return, so in case of Unit returning lambdas a it's as if the compiler adds a last line with `return Unit`

Consider this snippet:
```
val lambdaReturnsUnit: () -> Unit = {
    5 + 5
}
```
It is perfectly valid, for the same reason. I do agree it looks a bit odd.

julioromano:
Superseded by https://github.com/vector-im/element-x-android/pull/672

PR #634, opened 2023-06-19
Create PR to sync Localazy strings in the name of `ElementBot`, to th…

bmarty:
…at created PR can trigger other GitHub actions.

Ref: https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs

### DIFF
--- a/.github/workflows/sync-localazy.yml
+++ b/.github/workflows/sync-localazy.yml
@@ -31,3 +31,5 @@ jobs:
             - Update Strings from Localazy
           branch: sync-localazy
           base: develop
+        env:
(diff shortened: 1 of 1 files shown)


jmartinesp:
We can't know if this works fine until we merge it and run the sync workflow, right?

bmarty:
> We can't know if this works fine until we merge it and run the sync workflow, right?

Yes, I am afraid so...

jmartinesp (approved):
LGTM then, let's see how it goes

PR #635, opened 2023-06-19
[Message Actions] Forward messages

### LINKED ISSUE #486: [Message actions] Forward messages
## Changes

Implement the [designs from Figma](https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=4367%3A168017&t=ZYzKB8ZPEHPfZTXN-1).

- Create a new flow & screens to forward a message to one or several rooms.
- If the message was forwarded to a single room, we should navigate to it afterwards. Otherwise we should go back to the original chat room.

jmartinesp:
## Changes

- Added `ForwardMessages` feature inside `:features:messages`.
- Created `RoomContentForwarder` to forward an event to another room.

**Note**: while forwarding supports selecting multiple target rooms, for now we're limiting to be able to select just one room for feature parity with iOS.

## Why 

Closes #486 .

## Testing

Select an event from a room with long press, forward it to another room.


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/forward/ForwardMessagesPresenter.kt
```diff
@@ -0,0 +1,135 @@
...
+        var selectedRooms by remember { mutableStateOf(persistentListOf<RoomSummaryDetails>()) }
+        var query by remember { mutableStateOf<String>("") }
+        var isSearchActive by remember { mutableStateOf(false) }
+        var results: SearchBarResultState<ImmutableList<RoomSummaryDetails>> by remember { mutableStateOf(SearchBarResultState.NotSearching()) }
+        val forwardingActionState: MutableState<Async<ImmutableList<RoomId>>> = remember { mutableStateOf(Async.Uninitialized) }
+
+        val summaries by client.roomSummaryDataSource.roomSummaries().collectAsState()
+
+        LaunchedEffect(query, summaries) {
+            val filteredSummaries = summaries.filterIsInstance<RoomSummary.Filled>()
```
jmartinesp, line 73: We should also filter out:

1. Duplicates, as duplicate IDs will crash the LazyColumn, and we're not guaranteed there won't be any.
2. Rooms we cannot write to (take powerlevel into account, if possible).

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomContentForwarder.kt
```diff
@@ -0,0 +1,86 @@
...
+        eventId: EventId,
+        toRoomIds: List<RoomId>,
+        timeoutMs: Long = 5000L
+    ) {
+        val content = fromRoom.getTimelineEventContentByEventId(eventId.value)
+        val targetSlidingSyncRooms = toRoomIds.mapNotNull { slidingSync.getRoom(it.value) }
+        val targetRooms = targetSlidingSyncRooms.mapNotNull { room -> room.use { it.fullRoom() } }
+        val failedForwardingTo = mutableSetOf<RoomId>()
+        val results = targetRooms.map { room ->
+            coroutineScope.launch {
```
ganfra, line 59: Not sure if we should use `async/await` instead.
We have a `parallelMap` extension function.

FTR, not the same subject but I think we need to move away from `runCatching` in all our `suspend` methods as it'll avoid to propagate `CancellationException`.

ganfra (approved):
Thanks! 
Some small remarks, otherwise LGTM.
Will try and comment if I see issue.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/forward/ForwardMessagesPresenter.kt
jmartinesp: We shouldn't, but sadly we do. It seems like it happens on iOS too. When I asked @manuroe he mentioned they still have this issue with RoomList.

jmartinesp: Also, it seems like:

> Rooms we cannot write to (take powerlevel into account, if possible).

It's quite difficult to do right now with no performance hits. As far as I could see there is no option in slidingsync, sor for each item received in `RoomSummaryDataSource` we'd have to get its full room, the current member, and then its powerlevel/`canSend(eventType)`.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
```diff
@@ -143,11 +143,11 @@ class MessagesPresenter @Inject constructor(
     ) = launch {
         when (action) {
             TimelineItemAction.Copy -> notImplementedYet()
-            TimelineItemAction.Forward -> notImplementedYet()
             TimelineItemAction.Redact -> handleActionRedact(targetEvent)
             TimelineItemAction.Edit -> handleActionEdit(targetEvent, composerState)
             TimelineItemAction.Reply -> handleActionReply(targetEvent, composerState)
-            TimelineItemAction.Developer -> Unit // Handled at UI level
+            TimelineItemAction.Developer,
+            TimelineItemAction.Forward -> Unit // Handled at UI level
```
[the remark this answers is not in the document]
jmartinesp: Sounds good! I'm wondering if this should be a separate class with a reference to the Node or maybe just an interface the Node can implement.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/forward/ForwardMessagesPresenter.kt
```diff
@@ -0,0 +1,135 @@
...
+            forwardingSucceeded = forwardingSucceeded,
+            eventSink = { handleEvents(it) }
+        )
+    }
+
+    private fun CoroutineScope.forwardEvent(
+        eventId: EventId,
+        roomIds: ImmutableList<RoomId>,
+        isForwardMessagesState: MutableState<Async<ImmutableList<RoomId>>>,
+    ) = launch {
```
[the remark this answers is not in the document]
jmartinesp: Not with the current implementation (if everything goes well, `Result.success(Unit)` is returned, otherwise a `Result.failure(ForwardEventException(failedRoomIds))` is returned instead, but it might not be the best implementation.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomContentForwarder.kt
jmartinesp: > Not sure if we should use async/await instead.

Done, it looks way simpler now, thanks!

> FTR, not the same subject but I think we need to move away from runCatching in all our suspend methods as it'll avoid to propagate CancellationException.

We probably should, maybe create our own version that just re-throws the `CancellationException` when found. For now, I've just done that manually in a new commit.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/MatrixRoom.kt
```diff
@@ -83,6 +84,8 @@ interface MatrixRoom : Closeable {
 
     suspend fun sendReaction(emoji: String, eventId: EventId): Result<Unit>
 
+    suspend fun forwardEvent(eventId: EventId, rooms: List<RoomId>): Result<Unit>
```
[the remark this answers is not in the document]
jmartinesp: See https://github.com/vector-im/element-x-android/pull/635#discussion_r1238063180. I was grouping the failures in a single exception type so we could later check it in the presenter and display a proper error for multi-selection. Since we canceled multi-selection for feature parity, right now it's not used.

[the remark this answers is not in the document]
jmartinesp: In any case, if this implementation seems difficult to use we can improve it, I'm just unsure what's the best DevX here.

PR #638, opened 2023-06-19
Change the token used by the localazy sync so workflows are triggered in the resulting PR

jmartinesp:


bmarty (approved):
Yes, hopefully this will work better. Thanks!

PR #639, opened 2023-06-19
Sync Strings

ElementBot:


bmarty (approved):
OK, CI triggered properly after merging #638 . I am still the author of the commit, maybe because I triggered the action manually? 🤔

PR #641, opened 2023-06-20
Create screenshots with ElementBot

jmartinesp:
This change triggers the usual CI workflows automatically when the record screenshots workflow is done.

### DIFF
--- a/.github/workflows/recordScreenshots.yml
+++ b/.github/workflows/recordScreenshots.yml
@@ -16,6 +16,7 @@ jobs:
       - name: ⏬ Checkout with LFS
         uses: actions/checkout@v3
         with:
+          persist-credentials: false
           lfs: 'true'
       - name: ☕️ Use JDK 17
         uses: actions/setup-java@v3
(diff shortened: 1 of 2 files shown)


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/Color.kt
```diff
@@ -25,7 +25,7 @@ val LightGrey = Color(0x993C3C43)
 @ShowkaseColor(name = "DarkGrey", group = "Material Design")
 val DarkGrey = Color(0x99EBEBF5)
 
-val AvatarGradientStart = Color(0xFF4CA1AF)
+val AvatarGradientStart = Color.Red
```
bmarty, line 28: Do not forget to revert this change :)

bmarty (commented):
OK, I think this is working as expected, thanks for the changes!
Will approve the PR after the clean up :)

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/Color.kt
jmartinesp: I'll do 1 more change to try to simplify the script a bit more and then revert this change if it works 👍

jmartinesp: Done!

bmarty (approved):
LGTM, thanks for having fixed this pain in the neck!

PR #645, opened 2023-06-20
Set max lines to 2 in replies for composer

jmartinesp:
## Changes

- Set `maxLines` to 2 for the message previews in replies in the composer, unless we're replying to a media event.


ganfra (approved):
LGTM, by the way, I'm reworking a bit how we build data for Reply

PR #647, opened 2023-06-21
Use github action that caches LFS files when checking out the repo

julioromano:
Uses https://github.com/nschloe/action-cached-lfs-checkout where we're using git-lfs.

This is a wrapper around `actions/checkout@v3` which uses `actions/cache@v3` to cache the files on LFS to avoid downloading them every time and spare LFS bandwidth.


bmarty (approved):
That's looking great. Unfortunately the CI is complaining, maybe merge this PR and wait a few day for our quota to be reset.

PR #648, opened 2023-06-21
Developer option to clear cache

### LINKED ISSUE #643: Add a developer option to clear cache.
Only available on debug build.

It already exists on EIx.

This is the added files by the RustSDK into the `sessions` folder, provided by the client app:

image

bmarty:
Add option for developer to clear cache.

This PR also contains some cleanup, see the details in the commit messages.

Tested OK, but there is an issue on sliding sync, so I cannot properly load the room list at the moment.

Also when it was working, the room Avatar were not loaded again properly after a clear cache, but OK after restarting the app. I have added a clear of the Coil cache, it may help, but I was not able to test it.

Closes #643


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
```diff
@@ -71,6 +73,14 @@ class RustMatrixAuthenticationService @Inject constructor(
         return sessionStore.isLoggedIn()
     }
 
+    override fun incrementCacheIdx() {
```
julioromano, line 76: These APIs feel opaque.
It is unclear what's the intent behind them.
The name "cache index" is misleading because it leads the reader into thinking this API has something to do with the cache.
Instead it seems it's only used to call `matrixClientsHolder.removeAll()` every time the cache is cleared.
If we could call `removeAll()` from the `ClearCacheUseCase` directly we wouldn't need these.

appnav/src/main/kotlin/io/element/android/appnav/RootFlowNode.kt
```diff
@@ -88,11 +90,17 @@ class RootFlowNode @AssistedInject constructor(
...
+                    matrixClientsHolder.removeAll()
+                }
+            ) { isLoggedIn, cacheIdx -> isLoggedIn to cacheIdx }
             .onEach { isLoggedIn ->
                 Timber.v("isLoggedIn=$isLoggedIn")
-                if (isLoggedIn) {
+                if (isLoggedIn.first) {
                     tryToRestoreLatestSession(
-                        onSuccess = { switchToLoggedInFlow(it) },
+                        onSuccess = { switchToLoggedInFlow(it, isLoggedIn.second) },
```
julioromano, line 103: It's unclear why you're passing the cache index here, it is not used in the `resolve()` method.

appnav/src/main/kotlin/io/element/android/appnav/RootFlowNode.kt
```diff
@@ -88,11 +90,17 @@ class RootFlowNode @AssistedInject constructor(
     private fun observeLoggedInState() {
         authenticationService.isLoggedIn()
             .distinctUntilChanged()
+            .combine(
+                authenticationService.cacheIdx().onEach {
+                    Timber.v("cacheIdx=$it")
+                    matrixClientsHolder.removeAll()
+                }
+            ) { isLoggedIn, cacheIdx -> isLoggedIn to cacheIdx }
             .onEach { isLoggedIn ->
```
julioromano, line 99: This lambda param is now misnamed, it's a Pair, maybe you might want to use the destructuring syntax here.

julioromano:
Just a note for the future:

This PR is actually 2 PRs:
1. A "clear cache" PR
2. A "lots of unrelated small optimizations" PR

I'd have preferred to have it split because of both the size it has reached and the complexity during review.
I see that you've divided the changes into several descriptive commits but this doesn't help the reviewer much because when browsing files in a commit I can't mark them as "reviewed" nor comment on the PR (I can comment on the commit but it's not the same thing).

What do you think? Should PRs be more focused in general?
- I prefer lots of small focused PRs than a big PR with lots of mixed changes in it.
- Lots of small PRs should also go quicker down the reviewers' pipeline because a dev is much more likely to context switch to LGTM a small PR, but if its a big one he/she will have to set aside some dedicated time for review.

.idea/dictionaries/bmarty.xml
```diff
@@ -2,6 +2,7 @@
   <dictionary name="bmarty">
     <words>
       <w>homeserver</w>
+      <w>showkase</w>
```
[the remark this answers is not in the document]
bmarty: Android Studio "quick fix" is adding words here, I do not know why, and if it's configurable. Anyway all the XML dicts are taken into account by AS. I can merge them later.

libraries/androidtools/impl/src/main/kotlin/io/element/android/libraries/androidtools/impl/AndroidFileSizeFormatter.kt
```diff
@@ -0,0 +1,53 @@
...
+import io.element.android.libraries.di.AppScope
+import io.element.android.libraries.di.ApplicationContext
+import javax.inject.Inject
+
+@ContributesBinding(AppScope::class)
+class AndroidFileSizeFormatter @Inject constructor(
+    @ApplicationContext private val context: Context,
+    ) : FileSizeFormatter {
+    override fun format(fileSize: Long, useShortFormat: Boolean): String {
+        // Since Android O, the system considers that 1ko = 1000 bytes instead of 1024 bytes.
```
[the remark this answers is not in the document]
bmarty: Yes, will fixed (code was copied/pasted from EA)

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
bmarty: The main role is to trigger a reload of the app, where the Flow is used. But I admit I can rework this part.

appnav/src/main/kotlin/io/element/android/appnav/RootFlowNode.kt
bmarty: If I do not pass the cacheIndex, Appyx will not reload the Node, because the state will be considered as unchanged. So it's like a cache buster.

appnav/src/main/kotlin/io/element/android/appnav/RootFlowNode.kt
bmarty: Yes, good catch I will rename it.

appnav/src/main/kotlin/io/element/android/appnav/RootFlowNode.kt
julioromano: Then it is even more opaque than I thought :)
Maybe adding a comment that explains this side effect could help the future us?

libraries/androidtools/impl/src/main/kotlin/io/element/android/libraries/androidtools/impl/AndroidFileSizeFormatter.kt
```diff
@@ -0,0 +1,53 @@
...
+import android.os.Build
+import android.text.format.Formatter
+import com.squareup.anvil.annotations.ContributesBinding
+import io.element.android.libraries.androidtools.api.FileSizeFormatter
+import io.element.android.libraries.di.AppScope
+import io.element.android.libraries.di.ApplicationContext
+import javax.inject.Inject
+
+@ContributesBinding(AppScope::class)
+class AndroidFileSizeFormatter @Inject constructor(
```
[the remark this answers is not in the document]
julioromano: Then why bother at all? :) Could we simply do without this piece of code?

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/developer/DeveloperSettingsState.kt
```diff
@@ -16,10 +16,12 @@
 
 package io.element.android.features.preferences.impl.developer
 
+import io.element.android.libraries.architecture.Async
 import io.element.android.libraries.featureflag.ui.model.FeatureUiModel
 import kotlinx.collections.immutable.ImmutableList
 
-data class DeveloperSettingsState(
+data class DeveloperSettingsState constructor(
```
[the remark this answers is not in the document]
julioromano: TIL: You can still call "Find Usages" on the default primary constructor just by command clicking or right clicking on the opening parenthesis (i.e. click on `(` instead of `constructor`).

bmarty:
> A "lots of unrelated small optimizations" PR

I see only 2f08fe039f6d19957360d90c2dba7c14a88e069c and 7fa3d91b18010a51f8dd67f0da4d4a2b3e989b30, for the rest, it was needed for this PR.

> What do you think? Should PRs be more focused in general?

Yes, I agree we should split PR into dedicated tasks, but I will not create a PR to change a `var` to a `val`. I prefer to add small commits to fix little issue when I see it. It should not disturb too much the reviewer :)

julioromano:
> It should not disturb too much the reviewer :)

It does not "disturb" but it raises the noise floor, making it more difficult to understand why some changes have been made.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/tasks/ClearCacheUseCase.kt
```diff
@@ -0,0 +1,62 @@
...
+import io.element.android.libraries.di.SessionScope
+import io.element.android.libraries.matrix.api.MatrixClient
+import io.element.android.libraries.matrix.api.auth.MatrixAuthenticationService
+import kotlinx.coroutines.withContext
+import okhttp3.OkHttpClient
+import javax.inject.Inject
+import javax.inject.Provider
+
+interface ClearCacheUseCase {
+    suspend fun execute()
```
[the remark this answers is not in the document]
bmarty: OK

libraries/androidtools/impl/src/main/kotlin/io/element/android/libraries/androidtools/impl/AndroidFileSizeFormatter.kt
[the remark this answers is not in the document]
bmarty: We can check what other Element clients are displaying, but I believe this change was done to avoid EleWeb / iOS and Android display different values for the file sizes. Also was done to avoid having a limit for file upload on homeserver and file size displayed to the user not consistent. See https://github.com/vector-im/element-android/pull/589 .

bmarty:
Thanks for the review @julioromano I have handled all your remark (I think)

julioromano (approved):

PR #662, opened 2023-06-23
Expose new `windowInsets` param from `ModalBottomSheet`

julioromano:
Part of new public API in compose.material3:1.1.1


jmartinesp (approved):
LGTM! I wanted to play with these insets and some paddings a bit in case we can improve the default implementation which doesn't cover the systembars with the scrim or the bottom sheet itself, so thanks for this first step!

PR #665, opened 2023-06-23
[Message Actions] Copy events to clipboard

### LINKED ISSUE #663: [Message Actions] Copy message to clipboard
When selecting 'Copy' in a timeline item's context menu, the markdown/plain text version of the body of that event should be copied into the clipboard and the user should see a confirmation (maybe not needed in the latest versions of Android, since a popup bubble will already appear?).

jmartinesp:
## Changes

- Adds `ClipboardHelper` in `libraries:androidutils` as a wrapper of `ClipboardManager` to be able to use it in presenters and unit tests.
- Adds logic for 'Copy' action in the timeline item context menu.

**Note:** copying media events is out of scope for this PR.

## Why

Closes #663 .

## Test

Go inside a room, long press a text based event or a state event, then select 'Copy'. The markdown/plain text version of the message should be copied to the clipboard. In Android versions < 33 a snackbar will be displayed saying 'message copied', while on version >= 33 there's no need to, as there is a clipboard popup provided by the OS.


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/event/TimelineItemEventContent.kt
```diff
@@ -22,3 +22,14 @@ import androidx.compose.runtime.Immutable
 sealed interface TimelineItemEventContent {
     val type: String
 }
+
+/**
+ * Only text based content and states can be copied
```
ElementBot, line 27: - :warning: The first sentence of this KDoc does not end with the correct punctuation.

app/src/main/kotlin/io/element/android/x/di/AppModule.kt
```diff
@@ -107,4 +109,10 @@ object AppModule {
     fun provideSnackbarDispatcher(): SnackbarDispatcher {
         return SnackbarDispatcher()
     }
+
+    @Provides
+    @SingleIn(AppScope::class)
```
ganfra, line 114: Use `ContributesBindings` in directly in the `AndroidClipboardHelper` constructor?

ganfra (approved):
One remark otherwise LGTM

app/src/main/kotlin/io/element/android/x/di/AppModule.kt
jmartinesp: `libraries:androidutils` doesn't have any DI, that's why I created it there instead. I can add Anvil support for the library instead if you think that's better.

ganfra: I think whenever we can it's easier to follow how it's used yes

bmarty: I have added `libraries:androidtools` module with DI to be able to do this.

jmartinesp: Well, I also added it 😅 . I guess we'll deal with the possible conflicts later.

bmarty: (but I will move my stuff to `libraries:androidutils` now).

jmartinesp: Ah, I didn't realise you meant `tools` and not `utils`,  my bad.

PR #666, opened 2023-06-23
Enable analytics

yostyle:


services/analyticsproviders/posthog/src/main/kotlin/io/element/android/services/analyticsproviders/posthog/PosthogAnalyticsProvider.kt
```diff
@@ -66,10 +66,10 @@ class PosthogAnalyticsProvider @Inject constructor(
...
 
     override fun updateUserProperties(userProperties: UserProperties) {
-        posthog?.identify(
-            REUSE_EXISTING_ID, userProperties.getProperties()?.toPostHogUserProperties(),
-            IGNORED_OPTIONS
-        )
+//        posthog?.identify(
+//            REUSE_EXISTING_ID, userProperties.getProperties()?.toPostHogUserProperties(),
+//            IGNORED_OPTIONS
+//        )
```
jmartinesp, line 72: This was intended, right?

jmartinesp (approved):
LGTM. I find saving userContent and then calling `initOrStop` kind of weird, but I guess it's safer this way.

services/analyticsproviders/posthog/src/main/kotlin/io/element/android/services/analyticsproviders/posthog/PosthogAnalyticsProvider.kt
yostyle: Yes for the moment analytics id is anonymous.

Issue #624, opened 2023-06-16
Add missing posthog events


giomfo:
### Your use case

The following posthog events are needed for the EX with at least the listed sub-fields:

- [x] [Composer](https://github.com/matrix-org/matrix-analytics-events/blob/main/schemas/Composer.json)
   - isEditing
   - isReply

- [x] [ViewRoom](https://github.com/matrix-org/matrix-analytics-events/blob/main/schemas/ViewRoom.json)
   - isDM

- [x] [CreatedRoom](https://github.com/matrix-org/matrix-analytics-events/blob/main/schemas/CreatedRoom.json)
   - isDM

- [x] [JoinedRoom](https://github.com/matrix-org/matrix-analytics-events/blob/main/schemas/JoinedRoom.json)
  - isDM

- [x] screen (I would consider [MobileScreen](https://github.com/matrix-org/matrix-analytics-events/blob/main/schemas/MobileScreen.json))
   - screenName
   - durationMs

### Are you willing to provide a PR?

Yes

giomfo:
@yostyle FYI a first PR has been created by @Johennes about CreatedRoom: https://github.com/vector-im/element-x-android/pull/627

PR #669, opened 2023-06-23
Add ViewRoom posthog event

yostyle:
#624

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
@@ -38,6 +38,7 @@ import com.bumble.appyx.navmodel.backstack.operation.replace
 import com.bumble.appyx.navmodel.backstack.operation.singleTop
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedInject
+import im.vector.app.features.analytics.plan.ViewRoom
 import io.element.android.anvilannotations.ContributesNode
 import io.element.android.appnav.loggedin.LoggedInNode
 import io.element.android.features.analytics.api.AnalyticsEntryPoint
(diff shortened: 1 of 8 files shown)


Johennes (approved):
Looks good as far as I can tell but keep in mind that I'm very much out of my depth on Android. I only added the room creation events and screens to help move things along in our team. 😅

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -238,6 +239,7 @@ class LoggedInFlowNode @AssistedInject constructor(
                         }
                     }
                 } else {
+                    analyticsService.capture(ViewRoom(isDM = room.isDirect))
```
julioromano, line 242: Perhaps we should put tracking ode in the node that it directly relates with (i.e. not in the parent node).
What if this was in `RoomFlowNode.init{}` ?

julioromano:
@yostyle can you please put this one on hold a bit until we try to understand if we should send it on the messages view or the whole RoomFlow intead?

julioromano:
@yostyle 
I've checked with the iOS folks, they send this event only when actually seeing the timeline.
So, in order to follow along, the most correct place for this would be `MessagesNode` right now.
Sorry for having you change this so many times :(

yostyle:
> @yostyle I've checked with the iOS folks, they send this event only when actually seeing the timeline. So, in order to follow along, the most correct place for this would be `MessagesNode` right now. Sorry for having you change this so many times :(

The good way could be to capture this event in `RoomFlowNode` and add the trigger argument in node inputs.

julioromano:
> > @yostyle I've checked with the iOS folks, they send this event only when actually seeing the timeline. So, in order to follow along, the most correct place for this would be `MessagesNode` right now. Sorry for having you change this so many times :(
> 
> The good way could be to capture this event in `RoomFlowNode` and add the trigger argument in node inputs.

I'd vouch to do without triggers for the moment as complexity will skyrocket otherwise.

PR #672, opened 2023-06-26
Async API improvements "v2"

julioromano:
**NB: This PR actually changes only 3 files in `libraries/architecture/`. All the other changes are automated refactors to fix the calling code.**

This is a proposal for improvements to our `Async` type as discussed in: https://github.com/vector-im/element-x-android/pull/598/files#r1230664392 and in other chats.

Please bear in mind it is just a proposal, I'd love to hear your feedback about it, especially when it comes to naming: I've tried to make parameter and function names use a terminology similar to what we find in the Kotlin stdlib and its `Result` type.

I'm inclined to like more the non-extension flavours of the new `run*` APIs, though I'd also like your feedback about what API shape you prefer. 

#### Functional
- Adds `exceptionOrNull()` API to complement the existing `dataOrNull()` API.
- Adds `isFailure()`, `isLoading()`, `isSuccess()` and `isUninitialized()` courtesy APIs.
- Renames `executeResult()` to `runUpdatingState()`:
	- Becomes the base API to which all the other similarly named APIs call into.
	- Makes it inline.
	- Adds contract.
	- Passes over any `prevData` to newre Async states.
	- Passes through the `block`s return value.
	- Adds unit tests.
- Renames `execute` to `runCatchingUpdatingState()` and makes it just call into `runUpdatingState()`
- Adds extension function overloads to the `run*` functions to accept `MutableState` as receiver

#### Cosmetics
- Reorders classes and methods in alphabetic order.
- Reorder parameter names to mimic conventions in Kotlin stdlib.
- Adds docstrings where useful.

### DIFF
--- a/libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
+++ b/libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
@@ -18,51 +18,119 @@ package io.element.android.libraries.architecture
 
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.Stable
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.InvocationKind
+import kotlin.contracts.contract
 
+/**
+ * Sealed type that allows to model an asynchronous operation.
+ */
 @Stable
 sealed interface Async<out T> {
+    data class Failure<out T>(
+        val exception: Throwable,
+        val prevData: T? = null,
+    ) : Async<T>
+
+    data class Loading<out T>(
+        val prevData: T? = null,
+    ) : Async<T>
+
+    data class Success<out T>(
+        val data: T,
+    ) : Async<T>
+
     object Uninitialized : Async<Nothing>
-    data class Loading<out T>(val prevState: T? = null) : Async<T>
-    data class Failure<out T>(val error: Throwable, val prevState: T? = null) : Async<T>
-    data class Success<out T>(val state: T) : Async<T>
 
-    fun dataOrNull(): T? {
-        return when (this) {
-            is Failure -> prevState
-            is Loading -> prevState
-            is Success -> state
-            Uninitialized -> null
-        }
+    fun dataOrNull(): T? = when (this) {
+        is Failure -> prevData
+        is Loading -> prevData
+        is Success -> data
+        Uninitialized -> null
     }
+
+    fun exceptionOrNull(): Throwable? = when (this) {
+        is Failure -> exception
+        else -> null
+    }
+
+    fun isFailure(): Boolean = this is Failure<T>
+
+    fun isLoading(): Boolean = this is Loading<T>
+
+    fun isSuccess(): Boolean = this is Success<T>
+
+    fun isUninitialized(): Boolean = this == Uninitialized
 }
 
-suspend inline fun <T> (suspend () -> T).execute(
+suspend inline fun <T> MutableState<Async<T>>.runCatchingUpdatingState(
+    exceptionTransform: (Throwable) -> Throwable = { it },
+    block: () -> T,
+): Result<T> = runUpdatingState(
+    state = this,
+    exceptionTransform = exceptionTransform,
+    resultBlock = {
+        runCatching {
+            block()
+        }
+    },
+)
+
+suspend inline fun <T> (suspend () -> T).runCatchingUpdatingState(
     state: MutableState<Async<T>>,
-    errorMapping: ((Throwable) -> Throwable) = { it },
-) {
-    try {
-        state.value = Async.Loading()
-        val result = this()
-        state.value = Async.Success(result)
-    } catch (error: Throwable) {
-        state.value = Async.Failure(errorMapping.invoke(error))
-    }
-}
-
-suspend inline fun <T> (suspend () -> Result<T>).executeResult(state: MutableState<Async<T>>) {
-    if (state.value !is Async.Success) {
-        state.value = Async.Loading()
-    }
-    this().fold(
-        onSuccess = {
-            state.value = Async.Success(it)
-        },
-        onFailure = {
-            state.value = Async.Failure(it)
+    exceptionTransform: (Throwable) -> Throwable = { it },
+): Result<T> = runUpdatingState(
+    state = state,
+    exceptionTransform = exceptionTransform,
+    resultBlock = {
+        runCatching {
+            this()
         }
-    )
-}
+    },
+)
 
-fun <T> Async<T>.isLoading(): Boolean {
-    return this is Async.Loading<T>
+suspend inline fun <T> MutableState<Async<T>>.runUpdatingState(
+    exceptionTransform: (Throwable) -> Throwable = { it },
+    resultBlock: () -> Result<T>,
+): Result<T> = runUpdatingState(
+    state = this,
+    exceptionTransform = exceptionTransform,
+    resultBlock = resultBlock,
+)
+
+/**
+ * Calls the specified [Result]-returning function [resultBlock]
+ * encapsulating its progress and return value into an [Async] while
+ * posting its updates to the MutableState [state].
+ *
+ * @state the [MutableState] to post updates to.
+ * @exceptionTransform a function to transform the exception before posting it.
+ * @resultBlock a suspending function that returns a [Result].
+ * @return the [Result] returned by [resultBlock].
+ */
+@OptIn(ExperimentalContracts::class)
+@Suppress("REDUNDANT_INLINE_SUSPEND_FUNCTION_TYPE")
+suspend inline fun <T> runUpdatingState(
+    state: MutableState<Async<T>>,
+    exceptionTransform: (Throwable) -> Throwable = { it },
+    resultBlock: suspend () -> Result<T>,
+): Result<T> {
+    contract {
+        callsInPlace(resultBlock, InvocationKind.EXACTLY_ONCE)
+    }
+    val prevData = state.value.dataOrNull()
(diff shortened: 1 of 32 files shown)


libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
```diff
@@ -18,51 +18,119 @@ package io.element.android.libraries.architecture
...
+import kotlin.contracts.contract
 
+/**
+ * Sealed type that allows to model an asynchronous operation.
+ */
 @Stable
 sealed interface Async<out T> {
+    data class Failure<out T>(
+        val exception: Throwable,
+        val prevData: T? = null,
```
jmartinesp, line 32: I'd add a comment here about what `prevData` is used for.

libraries/architecture/src/test/kotlin/io/element/android/libraries/architecture/AsyncKtTest.kt
```diff
@@ -0,0 +1,113 @@
...
+    override var value: T
+        get() = _deque.last()
+        set(value) {
+            _deque.addLast(value)
+        }
+
+    /**
+     * Returns the states that were set in the order they were set.
+     */
+    fun pop(): T = _deque.removeFirst()
```
jmartinesp, line 99: Minor detail, but should this be `popFirst()`? I was a bit confused when I saw it worked beckwards to what I expected.

jmartinesp (commented):
I added a few comments to discuss.

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
```diff
@@ -18,51 +18,119 @@ package io.element.android.libraries.architecture
...
-    if (state.value !is Async.Success) {
-        state.value = Async.Loading()
-    }
-    this().fold(
-        onSuccess = {
-            state.value = Async.Success(it)
-        },
-        onFailure = {
-            state.value = Async.Failure(it)
+suspend inline fun <T> MutableState<Async<T>>.runCatchingUpdatingState(
```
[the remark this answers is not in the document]
julioromano: The idea was to tell that this was the same as `runUpdatingState` but that it used `runCatching` to wrap the lambda.

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
```diff
@@ -18,51 +18,119 @@ package io.element.android.libraries.architecture
...
+    resultBlock = {
+        runCatching {
+            this()
+        }
+    },
+)
+
+suspend inline fun <T> MutableState<Async<T>>.runUpdatingState(
+    exceptionTransform: (Throwable) -> Throwable = { it },
+    resultBlock: () -> Result<T>,
```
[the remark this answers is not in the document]
julioromano: it is implicitly, due to the function being `inline` .

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
```diff
@@ -18,51 +18,119 @@ package io.element.android.libraries.architecture
...
+    exceptionTransform: (Throwable) -> Throwable = { it },
+    resultBlock: suspend () -> Result<T>,
+): Result<T> {
+    contract {
+        callsInPlace(resultBlock, InvocationKind.EXACTLY_ONCE)
+    }
+    val prevData = state.value.dataOrNull()
+    state.value = Async.Loading(prevData = prevData)
+    val result = resultBlock()
+    return if (result.isSuccess) {
```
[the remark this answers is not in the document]
jmartinesp: You could still use a `Result.success(data)` and `Result.failure(data)` as the result if I'm not mistaken, and you'd avoid a couple of forced unwraps.

libraries/architecture/src/test/kotlin/io/element/android/libraries/architecture/AsyncKtTest.kt
jmartinesp: I agree we want to add them as `last`, but if I call `pop()` on some collection, I expect it to return the last item added, not the first one, that's why I suggested using `popFirst()` instead, since I found it confusing: I had to actually check what this `pop` function was doing under the hood for the asserts to make sense.

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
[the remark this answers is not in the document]
julioromano: Oh indeed! Thanks for the tip, updated accordingly.

libraries/architecture/src/test/kotlin/io/element/android/libraries/architecture/AsyncKtTest.kt
julioromano: Oh, I now understand what you mean. Renamed to `popFirst()`.

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
julioromano: Added docstrings

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
```diff
@@ -18,51 +18,148 @@ package io.element.android.libraries.architecture
...
+    data class Loading<out T>(
+        val prevData: T? = null,
+    ) : Async<T>
+
+    /**
+     * Represents a successful operation.
+     *
+     * @param data the data returned by the operation.
+     */
+    data class Success<out T>(
```
ElementBot, line 56: - :warning: Documentation of Success is outdated

julioromano:
@ganfra @bmarty How do you like this? Any further suggestions?

ganfra (approved):
Thanks for the improvements!

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
```diff
@@ -18,51 +18,148 @@ package io.element.android.libraries.architecture
...
 sealed interface Async<out T> {
+
+    /**
+     * Represents a failed operation.
+     *
+     * @param exception the exception that caused the operation to fail.
+     * @param prevData the data returned by a previous successful run of the operation if any.
+     */
+    data class Failure<out T>(
+        val exception: Throwable,
```
bmarty, line 38: Since this is not an `Exception` (but a `Throwable`), I would not name it like this.

Maybe use `failure`, keep `error`, or just simple `throwable`. From `mvrx`, it's `error`: [code](https://github.com/airbnb/mavericks/blob/main/mvrx-common/src/main/java/com/airbnb/mvrx/Async.kt#L63).

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
```diff
@@ -18,51 +18,148 @@ package io.element.android.libraries.architecture
...
-    try {
-        state.value = Async.Loading()
-        val result = this()
-        state.value = Async.Success(result)
-    } catch (error: Throwable) {
-        state.value = Async.Failure(errorMapping.invoke(error))
+    /**
+     * Returns the exception that caused the operation to fail, or null otherwise.
+     */
+    fun exceptionOrNull(): Throwable? = when (this) {
```
bmarty, line 80: I would rename to `errorOrNull`

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
```diff
@@ -18,51 +18,148 @@ package io.element.android.libraries.architecture
...
+        else -> null
     }
+
+    fun isFailure(): Boolean = this is Failure<T>
+
+    fun isLoading(): Boolean = this is Loading<T>
+
+    fun isSuccess(): Boolean = this is Success<T>
+
+    fun isUninitialized(): Boolean = this == Uninitialized
```
bmarty, line 91: This could be useful too: https://github.com/airbnb/mavericks/blob/main/mvrx-common/src/main/java/com/airbnb/mvrx/Async.kt#L86C21-L86C21

bmarty (approved):
I mainly agree with all the changes, just some remarks about the naming.

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
julioromano: I agree, I took the naming from `Result.kt` as convention. I thought making it consistent with `Result` terminology would be best.
Though I have no strong opinions on this. Happy to improve it.

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
julioromano: Same as above.

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/Async.kt
julioromano: Yeah indeed. Let's leave it out for this first iteration. It's a non breaking change that can be added at any time.

julioromano:
Changed `exception` to `error` throughout all the PR.

julioromano:
> I mainly agree with all the changes, just some remarks about the naming.

Changed `exception` to `error` throughout all the PR.

PR #684, opened 2023-06-27
Quick PR to introduce `CommonStrings` type alias

bmarty:
In first commit, next commit are for the big replacement, and last commit is to update the check tool.


jmartinesp (approved):
LGTM! It seems quite better than the aliased import everywhere.

PR #687, opened 2023-06-27
Add JoinedRoom posthog event

yostyle:
#624

### DIFF
--- a/services/analytics/api/src/main/kotlin/io/element/android/services/analytics/api/extensions/JoinedRoomExt.kt
+++ b/services/analytics/api/src/main/kotlin/io/element/android/services/analytics/api/extensions/JoinedRoomExt.kt
@@ -32,11 +32,11 @@ fun Long?.toAnalyticsRoomSize(): JoinedRoom.RoomSize {
     }
 }
 
-fun MatrixRoom?.toAnalyticsJoinedRoom(trigger: JoinedRoom.Trigger?): JoinedRoom {
+fun MatrixRoom.toAnalyticsJoinedRoom(trigger: JoinedRoom.Trigger?): JoinedRoom {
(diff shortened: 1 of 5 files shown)


services/analytics/api/src/main/kotlin/io/element/android/services/analytics/api/extensions/JoinedRoomExt.kt
```diff
@@ -0,0 +1,42 @@
...
+}
+
+fun MatrixRoom?.toAnalyticsJoinedRoom(trigger: JoinedRoom.Trigger?): JoinedRoom {
+    return JoinedRoom(
+            isDM = this?.isDirect.orFalse(),
+            isSpace = MatrixPatterns.isSpaceId(this?.roomId?.value),
+            roomSize = this?.joinedMemberCount?.toAnalyticsRoomSize() ?: JoinedRoom.RoomSize.Two,
+            trigger = trigger
+    )
+}
```
csmith, line 42: The nullability and defaults here seem a bit weird to me. If the `MatrixRoom` is `null` we'll log that it's not a DM and has two people in it... is that not just completely bad data?

This could just have a non-null receiver (i.e. `MatrixRoom` not `MatrixRoom?`) which would probably fix that?

csmith (approved):
LGTM, just one question

csmith:
And the tests need updating so they compile :)

services/analytics/api/src/main/kotlin/io/element/android/services/analytics/api/extensions/JoinedRoomExt.kt
yostyle: You right. I updated that.

PR #691, opened 2023-06-27
Danger: remove "add screenshots" warning

csmith:
We have screenshot tests that basically negate the need for this, and when we're doing more complicated things we tend to use videos anyway which aren't caught by this check.


jmartinesp (approved):
LGTM

julioromano (approved):

PR #692, opened 2023-06-27
Feature/fga/room list api

ganfra:
This PR makes usage of rust-sdk `RoomListService`.
It also includes : 
- Small rework of `NetworkMonitor` 
- A `SyncService` 
- Auto-restart of `Sync` when background/foreground and Network.Online
- A single `RoomSummaryDataSource`


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
```diff
@@ -94,13 +82,15 @@ class RustMatrixClient constructor(
 
     override val sessionId: UserId = UserId(client.userId())
 
+    private val roomListService = client.roomListService()
```
bmarty, line 85: What it the diff between `roomListService` and `roomListServiceWithEncryption`?

libraries/core/src/main/kotlin/io/element/android/libraries/core/coroutine/ChildScopeOf.kt
```diff
@@ -0,0 +1,33 @@
...
+import kotlinx.coroutines.plus
+
+fun childScopeOf(
+    parentScope: CoroutineScope,
+    dispatcher: CoroutineDispatcher,
+    name: String,
+): CoroutineScope = run {
+    val supervisorJob = SupervisorJob(parent = parentScope.coroutineContext.job)
+    parentScope + dispatcher + supervisorJob + CoroutineName(name)
+}
```
bmarty, line 33: Maybe rewrite as a Coroutine extension like:

```
/**
 * Create a child scope of the current scope.
 * The child scope will be cancelled if the parent scope is cancelled.
 * The child scope will be cancelled if an exception is thrown in the parent scope.
 *
 * @param dispatcher the dispatcher to use for this scope.
 * @param name the name of the coroutine.
 */
fun CoroutineScope.childScope(
    dispatcher: CoroutineDispatcher,
    name: String,
): CoroutineScope = run {
    val supervisorJob = SupervisorJob(parent = coroutineContext.job)
    this + dispatcher + supervisorJob + CoroutineName(name)
}
```

bmarty (approved):
LGTM, tested OK an a small account, and on a big account 👍 

With an empty account, I still have this displayed, but we may handle this later:

image

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
ganfra: Actually, we should use `roomListServiceWithEncryption` as long as we don't have a dedicated sync for encryption

PR #695, opened 2023-06-27
Do not fail the whole build if Diawi upload fails

bmarty:
Uploading to Diawi is failing at the moment since the APK is greater than 75MB (after we integrate the MapLibre dependency).

This PR let the GitHub action finish even if Diawi upload step is failing. We should not prevent the PR from being merged because of that failure.

Successful if the CI run on this PR is all green despite the Diawi upload failure.


jmartinesp (approved):
LGTM, let's see how the CI behaves

julioromano (approved):

PR #696, opened 2023-06-27
Show a prompt to reinvite other party in a DM

csmith:
If you are looking at a DM where the other party has left then
when you focus the composer for the first time we'll show a
dialog asking if you want to reinvite the other party.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
@@ -21,11 +21,13 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.saveable.rememberSaveable
+import androidx.compose.runtime.setValue
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedFactory
 import dagger.assisted.AssistedInject
@@ -52,7 +54,9 @@ import io.element.android.features.messages.impl.utils.messagesummary.MessageSum
 import io.element.android.features.networkmonitor.api.NetworkMonitor
 import io.element.android.features.networkmonitor.api.NetworkStatus
 import io.element.android.libraries.androidutils.clipboard.ClipboardHelper
+import io.element.android.libraries.architecture.Async
 import io.element.android.libraries.architecture.Presenter
+import io.element.android.libraries.architecture.runCatchingUpdatingState
 import io.element.android.libraries.core.coroutine.CoroutineDispatchers
 import io.element.android.libraries.designsystem.components.avatar.AvatarData
 import io.element.android.libraries.designsystem.components.avatar.AvatarSize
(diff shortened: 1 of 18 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
```diff
@@ -108,6 +113,22 @@ class MessagesPresenter @AssistedInject constructor(
...
+
+        val showReinvitePrompt by remember(
+            hasDismissedInviteDialog,
+            composerState.hasFocus,
+            syncUpdateFlow,
+        ) {
+            derivedStateOf {
+                !hasDismissedInviteDialog && composerState.hasFocus && room.isDirect && room.activeMemberCount == 1L
+            }
+        }
```
csmith, line 130: I kinda want this to be in its own presenter, but that would depend on the composerState from the compose presenter and inter-dependent sub-presenters feels a bit gross. Any ideas? :thinking:

jmartinesp: I wouldn't mind it too much to have this on a separate presenter, but if we want to avoid that maybe we could move it to a separate `@Composable` function?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
```diff
@@ -176,8 +208,21 @@ class MessagesPresenter @AssistedInject constructor(
...
+        suspend {
+            room.updateMembers()
+
+            val memberList = when (val memberState = room.membersStateFlow.value) {
+                is MatrixRoomMembersState.Ready -> memberState.roomMembers
+                is MatrixRoomMembersState.Error -> memberState.prevRoomMembers.orEmpty()
+                else -> emptyList()
+            }
+
+            val member = memberList.first { it.userId != room.sessionId }
```
jmartinesp, line 221: I was wondering if this could happen in a room that was just created, but I think `room.isDirect` would be false in that case and this point would never be reached.

jmartinesp (commented):
I can't make it work locally. I created a DM with another user, made the other user leave the room, but `room.activeMemberCount` keeps returning `2` and the prompt is never displayed. Could this be caused by the new RoomList api?

csmith:
> I can't make it work locally. I created a DM with another user, made the other user leave the room, but `room.activeMemberCount` keeps returning `2` and the prompt is never displayed. Could this be caused by the new RoomList api?

Ah, I meant to mention that in the PR description but forgot :sweat:. It's a caching issue in the SDK which also affects iOS: https://github.com/matrix-org/matrix-rust-sdk/issues/2078. For now you need to clear the cache after the user leaves...

jmartinesp:
> Ah, I meant to mention that in the PR description but forgot 😓. It's a caching issue in the SDK which also affects iOS: [matrix-org/matrix-rust-sdk#2078](https://github.com/matrix-org/matrix-rust-sdk/issues/2078). For now you need to clear the cache after the user leaves...

It doesn't seem to work even after clearing cache, sorry. The `activeMemberCount` is still 2. Element Web does display the right member count in the list:

image

jmartinesp (approved):
Ok, after creating a whole new room it seems to be working fine now (after clearing cache). LGTM then, I'd try to get that logic in a separate method as I mentioned in the a comment, but it's no big deal if it's not moved.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
csmith: I'll leave as-is for now. It's only really `showReinvitePrompt` that could move into its own method, and that doesn't make much difference. I might refactor it into a presenter when I have some more time, but that can be in a separate PR.

PR #698, opened 2023-06-27
Add analytics tracking for some screens

yostyle:
#624

### DIFF
--- a/features/roomdetails/impl/build.gradle.kts
+++ b/features/roomdetails/impl/build.gradle.kts
@@ -47,6 +47,7 @@ dependencies {
     api(projects.services.apperror.api)
     implementation(libs.coil.compose)
     implementation(projects.features.leaveroom.api)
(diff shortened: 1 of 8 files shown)


julioromano (approved):
Just to understand if this was intentional:
Are we okay to send screen events in `onResume`?
This means that if a user navigated to a screen, the even will be sent.
And if he switches to another app and back to EX the same event will be sent again.
Another case is if the user navigates to a screen, then puts the phone to sleep and wakes it. On wake the screen event will be sent again.

yostyle:
> Just to understand if this was intentional: Are we okay to send screen events in `onResume`? This means that if a user navigated to a screen, the even will be sent. And if he switches to another app and back to EX the same event will be sent again. Another case is if the user navigates to a screen, then puts the phone to sleep and wakes it. On wake the screen event will be sent again.

Yes, it's the wanted behavior like Element Android.

PR #699, opened 2023-06-28
Fix crash when opening a URL with associated text

### LINKED ISSUE #517: [Crash] Click on a link with text crashes the app
### Steps to reproduce

1. Click on a link in a message with a text, for instance like this: [link](https://example)

#### What did you expect?

The link is opened in the native browser

#### What happened instead?

the app crashes

Thread: main, Exception: android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW dat=here }

Looks like the app is trying to use the text for the URL.

Also in prod the app should not crash and the exception should be caught.

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

Yes

Yes

jmartinesp:
## Changes

- Created a `SafeUriHandler` and provided it in `MainActivity`.
- Added a `openUrlInExternalApp` function to `SystemUtils.kt` that tries to open the Uri with any app in the system and shows a toast when it can't.
- Fixed the value passed to the link annotation in `AnnotatedString.Builder.appendLink`.

## Tests

- Open a chat with a link in the format `[link here](https://element.io)` or send a message that contains it.
- You can also add a message with a link with some unsupported schema.
- Try to open the links. If they either work or display a toast, the bug is fixed.

Closes #517

### DIFF
--- a/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/system/SystemUtils.kt
+++ b/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/system/SystemUtils.kt
@@ -242,6 +242,19 @@ fun startImportTextFromFileIntent(
     }
 }
 
+@Suppress("SwallowedException")
+fun Context.openUrlInExternalApp(
+    url: String,
+    errorMessage: String = getString(R.string.error_no_compatible_app_found),
+) {
+    val intent = Intent(Intent.ACTION_VIEW, Uri.parse(url))
+    try {
+        startActivity(intent)
+    } catch (activityNotFoundException: ActivityNotFoundException) {
+        toast(errorMessage)
+    }
+}
+
 // Not in KTX anymore
 fun Context.toast(resId: Int) {
     Toast.makeText(this, resId, Toast.LENGTH_SHORT).show()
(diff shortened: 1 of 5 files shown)


ganfra (approved):
LGTM, thanks!

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/system/SystemUtils.kt
```diff
@@ -242,6 +242,19 @@ fun startImportTextFromFileIntent(
...
 }
 
+@Suppress("SwallowedException")
+fun Context.openUrlInExternalApp(
+    url: String,
+    errorMessage: String = getString(R.string.error_no_compatible_app_found),
+) {
+    val intent = Intent(Intent.ACTION_VIEW, Uri.parse(url))
+    try {
+        startActivity(intent)
```
bmarty, line 252: I think the Context must be an Activity here, so maybe enforce that?

jmartinesp: Good point, thanks!

PR #700, opened 2023-06-28
Link Compound's `TypographyTokens` to `ElementTheme`.

jmartinesp:
Also add some docs about when we should use each set of tokens.

### DIFF
--- a/libraries/theme/src/main/kotlin/io/element/android/libraries/theme/ElementTheme.kt
+++ b/libraries/theme/src/main/kotlin/io/element/android/libraries/theme/ElementTheme.kt
@@ -48,17 +48,17 @@ object ElementTheme {
      * The current [ElementColors] provided by [ElementTheme]. Usage of these colors is discouraged.
      * In Figma, they usually have the `Zzz` prefix or have no name at all.
      */
-    val colors: ElementColors
+    val legacyColors: ElementColors
         @Composable
         @ReadOnlyComposable
-        get() = LocalColors.current
+        get() = LocalLegacyColors.current
 
     /**
      * The current [SemanticColors] provided by [ElementTheme].
      * These come from Compound and are the recommended colors to use for custom components.
      * In Figma, these colors usually have the `Light/` or `Dark/` prefix.
      */
-    val compoundColors: SemanticColors
+    val colors: SemanticColors
         @Composable
         @ReadOnlyComposable
         get() = LocalCompoundColors.current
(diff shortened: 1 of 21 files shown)


csmith (approved):

libraries/theme/src/main/kotlin/io/element/android/libraries/theme/ElementTheme.kt
```diff
@@ -62,19 +65,25 @@ object ElementTheme {
...
-    val typography: Typography
+    val materialTypography: Typography
         @Composable
         @ReadOnlyComposable
         get() = MaterialTheme.typography
+
+    /**
+     * Compound [Typography] tokens. In Figma, these have the `Android/font/` prefix.
+     */
+    val typography: TypographyTokens = TypographyTokens
```
julioromano, line 86: Can we adopt a more consistent naming scheme?

e.g.

`materialColors` + `materialTypography` (for material stuff)
`compoundColors` + `compoundTypography ` (for compound stuff)
`colors` (for legacy stuff)

OR

`materialColors` + `materialTypography` (for material stuff)
`colors` + `typography ` (for compound stuff)
`elementColors` (for legacy stuff)

The second example does away with prefixes for the compound stuff so that is naturally the "default" choice.

jmartinesp: Yeah, the 2nd option sounds nice. I thought about using `legacyColors` for those but I wasn't sure it was accurate as some of them are not legacy as in coming from the old app, but they are legacy as in we shouldn't be using it anymore.

bmarty: Can we just remove what should not be used anymore?

jmartinesp: We shouldn't use them anymore but they might still be in the designs and we'll need to use them if there's no replacement, that's why I'm keeping them here. Also, by having all of them referenced here it's easier to just search for usage of this field if we wanted to do a batch replace in the future.

jmartinesp: Some new changes in https://github.com/vector-im/element-x-android/pull/700/commits/076c282d751e52ee02b7406389a16256061155e3 to rename the helper properties and enforce their usage.

julioromano (approved):

PR #713, opened 2023-06-28
Send read receipts

### LINKED ISSUE #671: Send read receipts for read messages
At the moment we don't have read receipts so even if we've read all messages in a room it will still display as having unread messages. 

We can use `Room.sendReadReceipt(EventId)` for that.

jmartinesp:
## Changes

- Add a `sendReadReceipt` method to `MatrixTimeline`.
- Send a read receipt for the latest read message in the timeline. The implemented logic is:
  - The `LazyColumn` stopped scrolling.
  - The index of the first visible item in the timeline (in reverse order) is <= the index of the last event we sent a read receipt for so if we were at the latest message (0) and a new one comes, we can send another read receipt.
  - The event id of the event at that index doesn't match the last one we sent.

## Why

Closes #671 .

## Test

- Open a room with unread messages, the room will be marked as read soon after.
- When at the last message on a room, receive a message. The room should be marked as read automatically.
- When the last message of the room is no longer visible, receive a message. The room should **not** be marked as read in this case.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
@@ -39,6 +39,7 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.saveable.rememberSaveable
+import androidx.compose.runtime.setValue
 import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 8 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -85,6 +86,21 @@ fun TimelineView(
         // TODO implement this logic once we have support to 'jump to event X' in sliding sync
     }
 
+    var lastReadMarkerIndex by rememberSaveable { mutableStateOf(Int.MAX_VALUE) }
+    var lastReadMarkerId by rememberSaveable { mutableStateOf<EventId?>(null) }
+
+    // Send a read receipt when the user has finished scrolling to an item in an index <= the last one
+    // and the event ids are different.
+    LaunchedEffect(lazyListState.firstVisibleItemIndex, lazyListState.isScrollInProgress) {
```
ElementBot, line 94: - :warning: Frequently changing state should not be directly read in composable function

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -85,6 +86,24 @@ fun TimelineView(
...
     }
 
+    var lastReadMarkerIndex by rememberSaveable { mutableStateOf(Int.MAX_VALUE) }
+    var lastReadMarkerId by rememberSaveable { mutableStateOf<EventId?>(null) }
+
+    // Send a read receipt when the user has finished scrolling to an item in an index <= the last one
+    // and the event ids are different.
+    val firstVisibleIndex by remember { derivedStateOf { lazyListState.firstVisibleItemIndex } }
+    val isScrollFinished by remember { derivedStateOf { !lazyListState.isScrollInProgress } }
+    LaunchedEffect(firstVisibleIndex, isScrollFinished) {
```
julioromano, line 96: Would it make sense to put some (as much as we can) of this logic in the presenter so we can test it?

e.g.:
View would send `state.eventSink(TimelineEvents.OnScrollFinished(firstVisibleIndex))` whenever `isScrollFinished` is true.
Then you can move in the presenter all the remaining logic so it's testable.

julioromano (approved):
LGTM with a comment.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
jmartinesp: Done in https://github.com/vector-im/element-x-android/pull/713/commits/3f23130547704dbfe9f2db84fbb5ca7fb2766a3c.

PR #715, opened 2023-06-28
Create a dummy value by default for `LocalSnackbarDispatcher`

jmartinesp:
This should solve the issue with needing to provide a `LocalSnackbarDispatcher` for previews and screenshot tests.


jmartinesp:
As a side note: we could also remove the need for a `LocalSnackbarDispatcher` by making `SnackbarDispatcher ` an `object`. I didn't do that because it seemed more correct to do it through DI, but it would simplify this issue a lot and remove a few pain points.

csmith (approved):
LGTM, thanks

PR #719, opened 2023-06-29
MatrixRoom API refinement

julioromano:
- `syncUpdateFlow` becomes a `val` and always returns the same instance of the underlying `StateFlow` instead of different `Flow` instances to allow consumers not to remember the `Flow` and not to specify an unneeded initial value.
- `timeline` becomes a `val` as it already always returns the same instance.
- Amends calling code accordingly
- nit: Removes a few unneeded `val`s in `RustMatrixClient`
- Fixes a small bug in `MessagesPresenter` that allowed to sometime show a newly created room's name as "Empty room" (changes `LaunchedEffect(syncUpdateFlow)` to `LaunchedEffect(syncUpdateFlow.value)`)


ganfra (approved):
Thanks, LGTM!

PR #721, opened 2023-06-29
Feature/fga/media upload progress

ganfra:
Branch progress on media upload
- Introduces `Determinate` and `Indeterminate`  `DialogProgressType`
- Starts using `Indeterminate` as long as the `ProgressCallback` is not starting, and then switch to `Determinate`.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
@@ -25,9 +25,8 @@ import dagger.assisted.Assisted
 import dagger.assisted.AssistedFactory
 import dagger.assisted.AssistedInject
 import io.element.android.features.messages.impl.attachments.Attachment
-import io.element.android.libraries.architecture.Async
 import io.element.android.libraries.architecture.Presenter
-import io.element.android.libraries.architecture.runUpdatingState
+import io.element.android.libraries.matrix.api.core.ProgressCallback
 import io.element.android.libraries.mediaupload.api.MediaSender
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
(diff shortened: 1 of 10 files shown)


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/ProgressDialog.kt
```diff
@@ -37,10 +37,16 @@ import io.element.android.libraries.designsystem.preview.PreviewGroup
 import io.element.android.libraries.designsystem.theme.components.CircularProgressIndicator
 import io.element.android.libraries.designsystem.theme.components.Text
 
+sealed interface ProgressDialogType {
```
ElementBot, line 40: - :warning: The file name 'ProgressDialog' does not match the name of the single top-level declaration 'ProgressDialogType'.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
```diff
@@ -81,10 +80,26 @@ class AttachmentsPreviewPresenter @AssistedInject constructor(
...
     private suspend fun sendMedia(
         mediaAttachment: Attachment.Media,
-        sendActionState: MutableState<Async<Unit>>,
+        sendActionState: MutableState<SendActionState>,
     ) {
-        sendActionState.runUpdatingState {
-            mediaSender.sendMedia(mediaAttachment.localMedia.uri, mediaAttachment.localMedia.info.mimeType, mediaAttachment.compressIfPossible)
+        val progressCallback = object : ProgressCallback {
+            override fun onProgress(current: Long, total: Long) {
+                sendActionState.value = SendActionState.Sending.Uploading(current.toFloat() / total)
```
jmartinesp, line 87: This works fine, but maybe we could also add `toFloat()` to `total`?

jmartinesp (approved):
Thanks for the changes, they improve the UX a lot! I have a small nit pick comment, and also I'd suggest to also change the text from `Sending...` when it's indeterminate to `Pre-processing...` maybe, but if it's not in the designs then it's fine.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
ganfra: As you want, it doesn't change anything

PR #729, opened 2023-06-29
Design fixes

bmarty:
Fix color for text in file Event (reported [here](https://github.com/vector-im/element-x-android/pull/724#discussion_r1246887637))
Also improve `ExtraPadding` logic, it's not ideal yet, but a bit better. It's now taking into account the size of the font. The smaller the font is, the higher the number of space is added, to ensure that it will not overlap the timestamp.


csmith (approved):

bmarty:
(will force push on latest develop and build screenshot again)

PR #731, opened 2023-06-29
Feature/fga/timeline back pagination

ganfra:
This PR:
- Fixes a crash when leaving a room (new because of syncUpdateFlow being used correctly).
- Disable pre-fetching in the timeline, as it causes more issues than it solves for now (lets make it better when rust room cache will be smarter).
- Convert a bunch of flows exposed as `fun` to `val`


bmarty (approved):
LGTM, thanks!
UI test failure can be ignored.

PR #740, opened 2023-06-30
Show location replies per the designs

csmith:
Show the pin icon on a shaded background next to the sender's name and the text "Shared location".

Move the pin asset from features/locations to libraries/matrixui to avoid creating dependency loops. Also remove the large transparent space at the bottom of the icon (used to center it on the map) and instead manually adjust the position where necessary.


bmarty (approved):
LGTM.

A remark: I think `pin.xml` and `PinIcon` should be moved to the module `:libraries:designsystem`, since is not directly related to Matrix.

PR #756, opened 2023-07-03
Add 'more reactions' button to message

jonnyandrew:
## Changes

- Add 'more reactions' button to message
- Fix display of existing emoji reactions to match designs
- Refactor emoji reactions to reduce nesting of composables

## Context
Part of
- https://github.com/vector-im/element-x-android/issues/342
- https://github.com/vector-im/element-x-android/issues/288

## Screenshots

| Before | After |
|-|-|
| before-1 | after-1|
| before-2 | after-2 |


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/MessagesReactionButton.kt
```diff
@@ -45,35 +46,50 @@ import io.element.android.libraries.theme.ElementTheme
...
 fun MessagesReactionButton(reaction: AggregatedReaction, modifier: Modifier = Modifier, onClick: () -> Unit) {
-    // First Surface is to render a border with the same background color as the background
-    Surface(
-        modifier = modifier.clickable(onClick = onClick::invoke),
+    val buttonColor = if(reaction.isHighlighted) {
         // TODO Should use compound.bgSubtlePrimary
-        color = ElementTheme.legacyColors.gray300,
-        border = BorderStroke(2.dp, MaterialTheme.colorScheme.background),
-        shape = RoundedCornerShape(corner = CornerSize(14.dp)),
+        ElementTheme.legacyColors.gray400
```
bmarty, line 51: I think now that compuond is available we can use the color `ElementTheme.colors.bgSubtlePrimary`. LegacyColors will be removed at some point.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemReactionsView.kt
```diff
@@ -30,7 +30,8 @@ import io.element.android.libraries.designsystem.preview.ElementPreviewLight
 fun TimelineItemReactionsView(
```
ElementBot, line 30: - :warning: Parameters in a composable function should be ordered following this pattern: params without defaults, modifiers, params with defaults and optionally, a trailing function that might not have a default param.
Current params are: [reactionsState: TimelineItemReactions, modifier: Modifier = Modifier, onReactionClicked: (emoji: String) -> Unit, onMoreReactionsClicked: () -> Unit] but should be [reactionsState: TimelineItemReactions, onReactionClicked: (emoji: String) -> Unit, modifier: Modifier = Modifier, onMoreReactionsClicked: () -> Unit].

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/MessagesReactionButton.kt
jonnyandrew: https://github.com/vector-im/element-x-android/pull/756/commits/8744cacb92f9eb3f332e84ca35873111beef167d

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
```diff
@@ -103,6 +104,9 @@ fun TimelineItemEventRow(
     fun onReactionClicked(emoji: String) =
         onReactionClick(emoji, event)
 
+    fun onMoreReactionsClicked() =
```
jmartinesp, line 107: This function might pass an outdated version of `event` to `onMoreReactionsClick`, but I guess it's not really and issue since we're just using the event id. Still, using a lambda here would make this less error-prone (see https://github.com/vector-im/element-x-android/pull/771#discussion_r1252736583).

jmartinesp (approved):
Just a comment, otherwise LGTM.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
jonnyandrew: Good point, I actually ran into this exact issue just now while working on something else. 

I think the local function captures the outer function parameter the first time it's called but doesn't get updated on recomposition.

Maybe we should ensure local functions do not reference arguments to the parent composable to avoid subtle bugs like this?

bmarty (approved):
Thanks for the update on the colors!

PR #759, opened 2023-07-04
Add missing parameter for the SNAPSHOT version of analytics library

jmartinesp:
## Changes

- Adds missing parameter to analytics `Composer` event.
- Sets the analytics library version to at least a commit hash so we don't rely on `main-SNAPSHOT` anymore, since it will change and cause cache issues.

TODO in the future: use an actual defined version for the analytics library, but [the latest one](https://github.com/matrix-org/matrix-analytics-events/releases/tag/v0.5.0) is quite outdated.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
@@ -135,7 +135,8 @@ class MessageComposerPresenter @Inject constructor(
                         Composer(
                             inThread = false,
                             isEditing = composerMode.value is MessageComposerMode.Edit,
-                            isReply = composerMode.value is MessageComposerMode.Reply
+                            isReply = composerMode.value is MessageComposerMode.Reply,
+                            isLocation = false,
                         )
                     )
                 }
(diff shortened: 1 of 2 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
```diff
@@ -135,7 +135,8 @@ class MessageComposerPresenter @Inject constructor(
                         Composer(
                             inThread = false,
                             isEditing = composerMode.value is MessageComposerMode.Edit,
-                            isReply = composerMode.value is MessageComposerMode.Reply
+                            isReply = composerMode.value is MessageComposerMode.Reply,
+                            isLocation = false,
```
bmarty, line 139: Maybe add a todo for the team working on location?

bmarty (approved):
Thanks!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
jmartinesp: I don't think the composer mode will ever be affected by location. Sending a location is handled in a whole different screen on Android.

PR #767, opened 2023-07-04
Adds new optional parameters to `MatrixRoom.sendLocation()`

julioromano:
For location sharing.


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
```diff
@@ -371,10 +373,20 @@ class RustMatrixRoom(
     //TODO expose inner parameters
```
csmith, line 373: Is this TODO now TODONE? :thinking:

csmith (approved):
LGTM

PR #771, opened 2023-07-04
Fix actions for redacted, not sent and media messages

### LINKED ISSUE #712: Some Message Actions should be filtered for failed local echoes and redacted messages
We should align the behaviour of EXA to EXI.

given this PR:
https://github.com/vector-im/element-x-ios/pull/1184

We are showing for failed local echoes only the following actions:
`.copy, .redact, .viewSource`
(where redact is actually using the `cancelSend` function provided the SDK instead of the redact one for failed echoes).

We also want to show only viewSource for redacted messages.

In both cases the user should not be able to react to these types of messages so the reaction part of the menu should be hidden.

Also we are currently looking into adding the following two functionalities to EXI:
~~https://github.com/vector-im/element-x-ios/issues/1186 (adding resend to actions for failed echoes)~~ paused for now
https://github.com/vector-im/element-x-ios/issues/1185 (only View Source and retry to decrypt should be available for unencrypted messages and reactions should be hidden too)
https://github.com/vector-im/element-x-ios/pull/1207 (edit for local failed echos, which is just a cancel and resend)

jmartinesp:
## Changes

- Redacted messages should only have the 'view source' action.
- Not sent messages can only have the following actions: 
  - Copy.
  - Edit, which will cancel the current one and send another.
  - View source.
  - Redact, which will cancel sending the event.
- Only text based events can be edited, so media and location ones are excluded.
- Reactions should only be available to sent messages.
- The action list shouldn't be displayed if there are no actions available (i.e.: redacted ones in release builds, with no 'view source' action)

## Why

Fixes #712 .


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -161,28 +161,20 @@ fun TimelineItemRow(
...
-            }
-
             if (timelineItem.content is TimelineItemStateContent) {
                 TimelineItemStateEventRow(
                     event = timelineItem,
                     isHighlighted = highlightedItem == timelineItem.identifier(),
-                    onClick = ::onClick,
-                    onLongClick = ::onLongClick,
+                    onClick = { onClick(timelineItem) },
+                    onLongClick = { onLongClick(timelineItem) },
```
csmith, line 169: So the TimelineItem passed in the callback isn't right? Should we change the callback to just be a `() -> Unit` in that case? :thinking:

jmartinesp: I'm not 100% sure on my take [here](https://github.com/vector-im/element-x-android/pull/774#discussion_r1252360667), but I think that's probably what's happening with these function references.

jmartinesp: This already happened in `ActionListView`, and switching the function reference back to a lambda fixed the issue, returning the current value instead.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/TimelineItem.kt
```diff
@@ -69,6 +70,10 @@ sealed interface TimelineItem {
         val showSenderInformation = groupPosition.isNew() && !isMine
 
         val safeSenderName: String = senderDisplayName ?: senderId.value
+
+        val failedToSend: Boolean = sendState is EventSendState.SendingFailed
+
+        val isTextMessage: Boolean = content is TimelineItemTextBasedContent
```
csmith, line 76: Could we use this to replace the `isMediaItem` check here: https://github.com/vector-im/element-x-android/blob/bb6ceb4541463922e46cbe995e99de49014596e4/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt#L231-L233

or are they not quite the same?

csmith (approved):
Nice, LGTM

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/TimelineItem.kt
jmartinesp: Not exactly, that isMediaItem is used to have different paddings and spacing in the message bubbles since their previews leave no borders. If we used `!isTextMessage` here instead, we'd have weird padding and spacing values for redacted, undecrypted or unknown events too.

PR #773, opened 2023-07-04
Tap on locations in timeline to see a larger map

### LINKED ISSUE #741: [Story] View expanded map
Ref: https://github.com/vector-im/element-meta/issues/1678

csmith:
Show a fully-featured MapView, centered on the dropped pin,
which allows panning/zooming. Share button allows opening
in a map application.

Supports showing a description at the top of the screen,
if one is supplied with the event.

Out of scope: showing the local user's location (being
done as a separate story).

Includes some minor tidying: remove duplicate Location,
and make GeoURI parsing a method on that class; fix the
pointer location in MapView (I broke it earlier, whoops!)

Closes #741 

https://github.com/vector-im/element-x-android/assets/189133/ddcde5ca-24c8-41da-b7b8-d5b9f015e083


features/location/impl/src/main/kotlin/io/element/android/features/location/impl/view/AndroidLocationActions.kt
```diff
@@ -0,0 +1,67 @@
...
+        val context = LocalContext.current
+        return DisposableEffect(Unit) {
+            activityContext = context
+            onDispose {
+                activityContext = null
+            }
+        }
+    }
+
+    override suspend fun share(location: Location, label: String?) {
```
jmartinesp, line 51: Does this function really need to be suspend and use coroutines?

features/location/impl/src/main/kotlin/io/element/android/features/location/impl/view/LocationActions.kt
```diff
@@ -0,0 +1,29 @@
...
+
+package io.element.android.features.location.impl.view
+
+import androidx.compose.runtime.Composable
+import io.element.android.features.location.api.Location
+
+interface LocationActions {
+
+    @Composable
+    fun Configure()
```
jmartinesp, line 25: I don't really like this strategy, but I guess it's needed to provide the activity context instead of the application one since we can't open external applications using the app context, right?

features/location/impl/src/main/kotlin/io/element/android/features/location/impl/view/ViewLocationView.kt
```diff
@@ -0,0 +1,125 @@
...
+    onBackPressed: () -> Unit = {},
+) {
+    val mapState = rememberMapState(
+        location = state.location,
+        position = MapState.CameraPosition(state.location.lat, state.location.lon, 15.0),
+    )
+
+    Scaffold(modifier,
+        topBar = {
+            CenterAlignedTopAppBar(
```
jmartinesp, line 64: More iOS-y screens, yay 🥲

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesFlowNode.kt
```diff
@@ -59,6 +62,7 @@ class MessagesFlowNode @AssistedInject constructor(
     @Assisted buildContext: BuildContext,
     @Assisted plugins: List<Plugin>,
     private val sendLocationEntryPoint: SendLocationEntryPoint,
+    private val viewLocationEntryPoint: ViewLocationEntryPoint,
```
jmartinesp, line 65: Nit: could this be `ShowLocationEntryPoint` or `LocationViewerEntryPoint`? Seeing `ViewLocation` here makes me wonder if it has something to do with view layouting for some reason.

jmartinesp (approved):
A couple of questions and suggestions, but I found nothing blocking.

features/location/impl/src/main/kotlin/io/element/android/features/location/impl/view/LocationActions.kt
csmith: Yeah 🫤

(This is the same pattern used for local media actions, so if we do come up with a nicer way of doing it we can fix them all at once)

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesFlowNode.kt
csmith: Yeah, `ViewLocationView` was definitely a bit weird :joy:. Done.

features/location/impl/src/main/kotlin/io/element/android/features/location/impl/view/LocationActions.kt
jmartinesp: About this: since this is a share action, I think we do want to start whatever new activity appears here as a new task, don't we? We could just use the app context and add the missing `FLAG_ACTIVITY_NEW_TASK` in that case. From the docs:

> This flag is generally used by activities that want to present a "launcher" style behavior: they give the user a list of separate things that can be done, which otherwise run completely independently of the activity launching them.

features/location/impl/src/main/kotlin/io/element/android/features/location/impl/view/AndroidLocationActions.kt
csmith: Apparently not! `startActivity` seems to handle making sure it's run on the main thread internally.

features/location/impl/src/main/kotlin/io/element/android/features/location/impl/view/AndroidLocationActions.kt
```diff
@@ -0,0 +1,67 @@
...
+        runCatching {
+            // Ref: https://developer.android.com/guide/components/intents-common#ViewMap
+            val suffix = if (label != null) "(${Uri.encode(label)})" else ""
+            val uri = Uri.parse("geo:0,0?q=${location.lat},${location.lon}$suffix")
```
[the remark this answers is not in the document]
csmith: Yes, probably :)

features/location/impl/src/main/kotlin/io/element/android/features/location/impl/view/LocationActions.kt
csmith: Ooh, yes. Good catch, thanks.

PR #774, opened 2023-07-04
Swipe to reply

### LINKED ISSUE #762: [Timeline] Add swipe to reply gesture
User can swipe a message to the right (to the end?) to be able to reply it. This is a shortcut for:
- long click on a message
- choose "Reply"

The gesture is available only for messages which can be replied (text, image, etc.).

iOS PR: https://github.com/vector-im/element-x-ios/pull/1148
Desing: TBC.

bmarty:
Closes #762


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/ReplySwipeIndicator.kt
```diff
@@ -0,0 +1,72 @@
...
+import io.element.android.libraries.designsystem.preview.ElementPreviewLight
+import io.element.android.libraries.designsystem.theme.components.Icon
+
+/**
+ * A swipe indicator that appears when swiping to reply to a message.
+ *
+ * @param swipeProgress the progress of the swipe, between 0 and X. When swipeProgress >= 1 the swipe will be detected.
+ */
+@Composable
+fun RowScope.ReplySwipeIndicator(swipeProgress: Float) {
```
ElementBot, line 41: - :warning: This @Composable function emits content but doesn't have a modifier parameter.

See https://mrmans0n.github.io/compose-rules/rules/#when-should-i-expose-modifier-parameters for more information.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
```diff
@@ -108,6 +118,72 @@ fun TimelineItemEventRow(
...
+    if (event.groupPosition.isNew()) {
+        Spacer(modifier = modifier.height(16.dp))
+    } else {
+        Spacer(modifier = modifier.height(2.dp))
+    }
+}
+
+@Composable
+private fun TimelineItemEventRowContent(
+    modifier: Modifier,
```
ElementBot, line 176: - :warning: This @Composable function has a modifier parameter but it doesn't have a default value.

See https://mrmans0n.github.io/compose-rules/rules/#modifiers-should-have-default-parameters for more information.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -165,6 +170,10 @@ fun TimelineItemRow(
                 onClick(timelineItem)
             }
 
+            fun onSwipeToReply() {
+                onSwipeToReply(timelineItem)
```
jmartinesp, line 174: This can lead to issues with cached values for `timelineItem`. The `timelineItem` returned here is not the one from the last recomposition, but from the 1st one, when this function is assigned, so if it's later edited or redacted we'll be reporting the wrong value on the callback. I had to revert `onClick` and `onLongClick` on https://github.com/vector-im/element-x-android/pull/771 for this same reason.

bmarty: Ah, good point, thanks for the feedback. I will change that.

bmarty: 218002627d9397ec83534b08da1437b800b0bf8a

csmith (approved):

PR #778, opened 2023-07-05
Highlight user's reactions in message actions menu

jonnyandrew:
## Changes
- Highlight user's reactions in message actions menu
- Fix issue where message `onClick` and `onLongClick` do not pass the correct event to the parent composable's callback

## Context
Part of https://github.com/vector-im/element-x-android/issues/342

## Screenshots
[message-actions-reaction.webm](https://github.com/vector-im/element-x-android/assets/4940864/abacf267-3656-470c-ace2-46c04852709f)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/actionlist/ActionListView.kt
```diff
@@ -320,6 +324,7 @@ private val emojiRippleRadius = 24.dp
 
 @Composable
 internal fun EmojiReactionsRow(
+    highlightedEmojis: List<String>,
```
ElementBot, line 327: - :warning: The Compose Compiler cannot infer the stability of a parameter if a List<String> is used in it, even if the item type is stable.
You should use Kotlinx Immutable Collections instead: `highlightedEmojis: ImmutableList<String>` or create an `@Immutable` wrapper for this class: `@Immutable data class HighlightedEmojisList(val items: List<String>)`

See https://mrmans0n.github.io/compose-rules/rules/#avoid-using-unstable-collections for more information.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -161,28 +161,28 @@ fun TimelineItemRow(
             )
         }
         is TimelineItem.Event -> {
-            fun onClick() {
+            val onClicked = {
                 onClick(timelineItem)
             }
 
-            fun onLongClick() {
+            val onLongClicked = {
```
bmarty, line 168: This is also done in #771 , we may have conflict here...

bmarty (commented):
LGTM, but could be nice to have new screenshots to render this new state on the `EmojiReactionsRow`

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
jonnyandrew: [2ad307d](https://github.com/vector-im/element-x-android/pull/778/commits/2ad307dfc2dccf3974e319832b2440a99f93b3d7)

bmarty (approved):
Perfect, thanks!

PR #779, opened 2023-07-05
Upgrades paparazzi to latest stable 1.3.0 version

julioromano:


bmarty (commented):
We are waiting for 1.3.1, see comment in #500.

PR #781, opened 2023-07-05
Implement waiting list screen (error IO_ELEMENT_X_WAIT_LIST)

### LINKED ISSUE #768: Waiting list screen - Implement `IO_ELEMENT_X_WAIT_LIST` error code on `/login`
As a user, I want to understand why I cannot login and how I can be part of this allow list defined on my homeserver.

iOS PRs:
- https://github.com/vector-im/element-x-ios/pull/1165
- https://github.com/vector-im/element-x-ios/pull/1183
- Latest design: https://github.com/vector-im/element-x-ios/pull/1224

bmarty:
Closes #768 

Only the first screen is used, and the "You're in" screen are displayed (no "setting up your account" screen, which may be handled later, and anyway not in the `:login` module).

Dark and light are identical (so we have some hard-coded colors), only the dialog colors may change.

## Demo:

|Without "You're in" (previous demo)|You're in demo (+ dark)|
|-|-|
|WaitList|WaitListDarkYoureIn|

Note: since we are using a test account here, there is no way to make it successfully login. But at each attempt, if the login is successful, the regular flow will continue: analytics optin then room list. Tested OK locally.

## Screenshot

|Portrait|Landscape|
|-|-|
|image|image|
|image|image|

### App not logged in

- Login using `m1-element` account. I can provide the password if necessary. This account will be rejected. There is no way to go further using this account, but the code can be hacked to use different credentials when attempting login again, so that we can see the "You're in" screen. Note: if the app is killed once the You're in screen is displayed, restarting it shows the room list.
- Login using any another account: the wait screen is not displayed and the room list is displayed (acually the analytics screen first)

### App is logged in

- cold start show the room list


bmarty:
After discussing with @ganfra I will iterate to have the "You're in" screen.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/DefaultLoginUserStory.kt
```diff
@@ -0,0 +1,38 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+import javax.inject.Inject
+
+@SingleIn(AppScope::class)
+@ContributesBinding(AppScope::class)
+class DefaultLoginUserStory @Inject constructor() : LoginUserStory {
+    // True by default, will be set to false when the login user story is started, and set to true again once it's done.
+    private val loginFlowIsDone: MutableStateFlow<Boolean> = MutableStateFlow(true)
+
+    override fun loginFlowIsDone(): Flow<Boolean> = loginFlowIsDone
```
ganfra, line 33: use `val` instead so it's clearer we are always returning the same one.
We have to change this everywhere, but a least new code should be like that.
Also maybe you should expose a `StateFlow`?

ganfra (approved):
One remark, otherwise LGTM, good work!

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/DefaultLoginUserStory.kt
bmarty: Yes, 072b59e06b21cc905d9c27f55c7c38e363ecaba9

PR #782, opened 2023-07-05
Restore functionality of wide/tall previews and screenshots

julioromano:
New preview annotations has been introduced.
Where `@Preview(widthDp = 600)` was used before it now becomes `@ElementWidePreviews`
Where `@Preview(heightDp = 1000)` was used before it now becomes `@ElementTallPreviews`


bmarty (approved):
OK, thanks!

PR #788, opened 2023-07-06
Fix displaying reaction row in message action list

jmartinesp:
## Changes

- Rename `EventSendState` to `LocalEventSendState` so it's clear this only accounts for local send state.
- Make `TimelineItem.Event.sendState` nullable with no default value and rename it to `localSendState`. This value will now only have info when the local send state retrieved from the Rust SDK is not null.

## Why

Fixes the reactions row not appearing in the message action list for most messages.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemEventFactory.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemEventFactory.kt
@@ -71,6 +71,14 @@ class TimelineItemEventFactory @Inject constructor(
             url = senderAvatarUrl,
             size = AvatarSize.TimelineSender
         )
+        // Send state based on local data
+        fun localSendState(): EventSendState {
+            return currentTimelineItem.event.localSendState ?: EventSendState.NotSentYet
+        }
+        // Send state: if the event has an event id, then it's already sent
+        val processedSendState = currentTimelineItem.eventId?.let {
+            currentTimelineItem.event.localSendState ?: EventSendState.Sent(it)
+        } ?: localSendState()
         return TimelineItem.Event(
             id = currentTimelineItem.uniqueId,
             eventId = currentTimelineItem.eventId,
@@ -83,7 +91,7 @@ class TimelineItemEventFactory @Inject constructor(
             sentTime = sentTime,
             groupPosition = groupPosition,
             reactionsState = currentTimelineItem.computeReactionsState(),
-            sendState = currentTimelineItem.event.localSendState ?: EventSendState.NotSentYet,
+            sendState = processedSendState,
             inReplyTo = currentTimelineItem.event.inReplyTo(),
             debugInfo = currentTimelineItem.event.debugInfo,
         )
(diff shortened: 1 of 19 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemEventFactory.kt
```diff
@@ -71,6 +71,14 @@ class TimelineItemEventFactory @Inject constructor(
...
             size = AvatarSize.TimelineSender
         )
+        // Send state based on local data
+        fun localSendState(): EventSendState {
+            return currentTimelineItem.event.localSendState ?: EventSendState.NotSentYet
+        }
+        // Send state: if the event has an event id, then it's already sent
+        val processedSendState = currentTimelineItem.eventId?.let {
+            currentTimelineItem.event.localSendState ?: EventSendState.Sent(it)
+        } ?: localSendState()
```
jonnyandrew, line 81: I think 'send state' in the SDK is only applicable to local items, so wondering if it could be a bit confusing to fill a send state for a non-local timeline item on the Android side?

Perhaps something more like `isRemote` or `canReact`?

Also, FYI, the logic might go out of date after https://github.com/matrix-org/matrix-rust-sdk/issues/2127 so perhaps worth centralising it in the Rust SDK

jmartinesp: `isRemote` sounds good, it's the same thing I wanted to achieve with `sendState` vs `localSendState` here, but it's easier to understand. My only problem with it is we're defaulting `localSendState` to `NotSent` here, so you can have a remote event with 'not sent' local send state, which is a bit contradictory. Maybe I can make it nullable and not default to anything, or make it a sealed interface `SendState { Local(...), Remote(eventId) }`.

> Also, FYI, the logic might go out of date after https://github.com/matrix-org/matrix-rust-sdk/issues/2127 so perhaps worth centralising it in the Rust SDK

I find it a bit weird wanting to apply reactions to local echoes, but good to know. Thanks for the heads up!

jonnyandrew: > Maybe I can make it nullable and not default to anything

That sounds like a good idea; this keeps it better aligned with the SDK and there are some events where 'send state' does not apply (i.e. events not sent by the user)

jonnyandrew: > I find it a bit weird wanting to apply reactions to local echoes

I think the use case here is for sending a message pre-populated with emoji reactions as a prompt e.g:

```
Should we order pizza tonight?
👍 👎 🤷
```

csmith (approved):
LGTM aside from Jonny's comment about naming

jmartinesp:
> LGTM aside from Jonny's comment about naming

It should be fixed in https://github.com/vector-im/element-x-android/pull/788/commits/c2c82f6f22011ba8c78bedc10cc85722d0ecf1ba, the rebase deleted the commit.

PR #794, opened 2023-07-06
Feature/fga/waiting ss room

ganfra:
Add a way to wait for a `MatrixRoom` in the `SlidingSync` so we don't end up showing "unknown room" node when the `RoomList` is just not up-to-date.
The error case will be improved later (waiting for design).


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
```diff
@@ -127,9 +130,16 @@ class RustMatrixClient constructor(
...
 
-    override fun getRoom(roomId: RoomId): MatrixRoom? {
-        val roomListItem = roomListService.roomOrNull(roomId.value) ?: return null
-        val fullRoom = roomListItem.fullRoom()
+    override suspend fun getRoom(roomId: RoomId): MatrixRoom? {
+        var cachedPairOfRoom = pairOfRoom(roomId)
+        if (cachedPairOfRoom == null) {
+            roomSummaryDataSource.allRoomsLoadingState().firstOrNull {
+                it is RoomSummaryDataSource.LoadingState.Loaded
+            }
```
csmith, line 138: Maybe put this in an extension method so it can have a descriptive name? Took me a minute to realise what this was actually doing :sweat_smile: 

Also should it have some kind of timeout or error handling? If the room summaries are never loaded this will block indefinitely, right?

csmith (approved):

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
ganfra: This will `suspend`, not `block` :P
So, I think it's ok to not timeout, if room summaries are not loaded, the app is not usable anyway. So we can just wait for the network to come back, things like that. Let's see how it goes without it, and if we think we should really add it, we'll.

PR #797, opened 2023-07-06
Don't show own user in search results

csmith:
Anywhere we do a global search (starting a DM, creating a room, inviting a user to a room) should filter out the local user.

### DIFF
--- a/libraries/usersearch/impl/src/main/kotlin/io/element/android/libraries/usersearch/impl/MatrixUserRepository.kt
+++ b/libraries/usersearch/impl/src/main/kotlin/io/element/android/libraries/usersearch/impl/MatrixUserRepository.kt
@@ -32,15 +32,13 @@ import javax.inject.Inject
 
 @ContributesBinding(SessionScope::class)
 class MatrixUserRepository @Inject constructor(
-    client: MatrixClient,
+    private val client: MatrixClient,
     private val dataSource: UserListDataSource
 ) : UserRepository {
 
-    private val sessionId = client.sessionId
-
     override suspend fun search(query: String): Flow<List<UserSearchResult>> = flow {
         // Manually add a fake result with the matrixId, if provided and not the local user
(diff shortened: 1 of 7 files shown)


libraries/usersearch/impl/src/main/kotlin/io/element/android/libraries/usersearch/impl/MatrixUserRepository.kt
```diff
@@ -31,24 +32,31 @@ import javax.inject.Inject
...
 ) : UserRepository {
 
+    private val sessionId = client.sessionId
+
     override suspend fun search(query: String): Flow<List<UserSearchResult>> = flow {
-        // Manually add a fake result with the matrixId, if any
-        val isUserId = MatrixPatterns.isUserId(query)
-        if (isUserId) {
+        // Manually add a fake result with the matrixId, if provided and not the local user
+        val isAnotherUsersId =  query != sessionId.value && MatrixPatterns.isUserId(query)
```
ganfra, line 43: Maybe create a MatrixClient.isMe(userId: UserId): Boolean extension method and check for this already existent usage?

ganfra (approved):
Just one remark otherwise LGTM

libraries/usersearch/impl/src/main/kotlin/io/element/android/libraries/usersearch/impl/MatrixUserRepository.kt
bmarty, line 43: `isAnotherUsersId` -> `isAnotherUserId`, and please format the file :)

csmith: It was meant to read like "is another user's id", but I can see the ambiguity without punctuation :laughing:. I don't think `isAnotherUserId` quite conveys what it means, so I've changed it entirely to `shouldQueryProfile`.

PR #798, opened 2023-07-06
Update template for user story

kittykat:
Update the label/description - labels are compulsory and end up in the issue body
Split out signoff from tasklist - that way signoffs do not count towards the total tasks in the issue


bmarty (approved):
LGTM, thanks!

PR #803, opened 2023-07-06
Feature/fga/app nav node fixes

ganfra:
Slightly rework : 
- `RootFlowNode` and `CacheService` to avoid some weird bad node restoration
- `LoggedInFlowNode` root should always be `RoomList`


jmartinesp:
Could you explain a bit more what this does and why the changes were needed? I'm trying to follow the logic, but I have no idea why the previous `cacheIndex` was needed and now I have no idea why using a session id + UUID fixes the issues 😅 .

ganfra:
Yes sorry, so the idea was to remove the `cacheIndex` logic because it was a trick to make the navigation working, and it's not really the responsability of the `CacheService` to do so.
If you use the same `NavTarget` with the same `SessionId` (which happens when you clear cache, as you want to re-create a fresh new `LoggedInFlowNode` with the same params), it won't do anything, because the `backstack.newRoot(navTarget)` method checks the equality of the navTarget.
The other possibility is to create a new Operation on the backstack which force to recreate everything, but I'm not really happy with this.

bmarty:
I have briefly explain it [here](https://github.com/vector-im/element-x-android/pull/648#discussion_r1239763418), but it's true that the code could be a bit more documented (my bad at the origin)

appnav/src/main/kotlin/io/element/android/appnav/RootFlowNode.kt
```diff
@@ -259,9 +269,16 @@ class RootFlowNode @AssistedInject constructor(
     }
 
     private suspend fun attachSession(sessionId: SessionId): LoggedInFlowNode {
-        val cacheIndex = cacheService.cacheIndex().first()
-        return attachChild {
-            backstack.newRoot(NavTarget.LoggedInFlow(sessionId, cacheIndex))
+        //TODO handle multi-session
+        return waitForChildAttached { navTarget ->
+            navTarget is NavTarget.LoggedInFlow && navTarget.sessionId == sessionId
```
csmith, line 274: nit: should this function be renamed? It sounds like it does the attaching, but now it just waits for a session to be attached elsewhere?

csmith (approved):
LGTM, just an optional naming suggestion

appnav/src/main/kotlin/io/element/android/appnav/RootFlowNode.kt
ganfra: I reused the name from appyx, because they already have a `waitForChildAttached` but without `navTarget` checking. 
It does not `attach` the node anymore, just waits to see it in the hierarchy. 
It was not a good thing to `attach` the node because it's already done by listening to the `isAuthenticatedFlow`

PR #805, opened 2023-07-06
Force the default locale to the selected one for screenshot tests

jmartinesp:
This should force the locale used for `DatePicker` to `en_US` on screenshot tests, making them usable again locally.

Right now we have an issue with the first day of week and `DatePicker` that makes verification always fail locally.


bmarty (approved):
OK for me, but maybe require an approval from @csmith if it fixes his issue locally?

csmith (approved):
Thanks! Can confirm they're passing locally for me now :+1:

PR #811, opened 2023-07-07
Display the "verification success" Snackbar only when the state is `VerificationFlowState.Finished`.

### LINKED ISSUE #809: Spurious "verification successful prompt"
### Steps to reproduce

1. I tried to self-verify an EAX instance i hadn't used in a while
2. Hit the failure mode described in https://github.com/vector-im/element-x-ios/issues/1276
3. After failing to self-verify, opened a room
4. Saw an unexpected "verification successful" prompt in grey at the bottom of the screen, which then disappeared

#### What did you expect?

No spurious "verification successful" prompts, if verification has just dismally failed

### Your phone model

_No response_

### Operating system version

_No response_

july 6 nightly

### Homeserver

_No response_

### Will you send logs?

No

No

bmarty:
Apply the same fix for `displayLeftRoomMessage`

Closes #809 

I have detected another issue: we can start the verification from the settings now and when complete, the snackbar will be displayed only when back to the room list. I will handle that separately. #812 

Test: verify the session, cancel it, let it timeout, verify again, play a lot with it. Before the fix the Snackbar was displayed a lot. With the fix it is displayed only when the verification is successful.

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/LoggedInEventProcessor.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/LoggedInEventProcessor.kt
@@ -24,7 +24,7 @@ import io.element.android.libraries.matrix.api.verification.VerificationFlowStat
 import io.element.android.libraries.ui.strings.CommonStrings
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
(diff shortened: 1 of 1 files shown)


jmartinesp (approved):
LGTM, thanks!

PR #816, opened 2023-07-07
LoadingRoomNode : fix windowInsets...

ganfra:


appnav/src/main/kotlin/io/element/android/appnav/room/LoadingRoomNodeView.kt
```diff
@@ -72,7 +72,8 @@ fun LoadingRoomNodeView(
                 modifier = Modifier
                     .fillMaxSize()
                     .padding(padding)
-                    .padding(16.dp), contentAlignment = Alignment.Center
+                    .padding(16.dp),
```
csmith, line 75: I think it's better to call `.consumeWindowInsets(padding)` here than pass the child a modifier with window insets which it then has to ignore?

ganfra: Is this not done automatically by the scaffold? I'm just adding some paddings for the text to not be touching borders

csmith: No, it changed a little while ago so Scaffold doesn't do it any more.

The [documentation for the `content` param](https://developer.android.com/reference/kotlin/androidx/compose/material3/package-summary#Scaffold(androidx.compose.ui.Modifier,kotlin.Function0,kotlin.Function0,kotlin.Function0,kotlin.Function0,androidx.compose.material3.FabPosition,androidx.compose.ui.graphics.Color,androidx.compose.ui.graphics.Color,androidx.compose.foundation.layout.WindowInsets,kotlin.Function1)) now says:

> The lambda receives a [PaddingValues](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/PaddingValues) that should be applied to the content root via Modifier.padding and [Modifier.consumeWindowInsets](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/package-summary#(androidx.compose.ui.Modifier).consumeWindowInsets(androidx.compose.foundation.layout.WindowInsets)) to properly offset top and bottom bars.

csmith: Yeah, https://github.com/vector-im/element-x-android/commit/848f1cb98868f60f4c83888809c76f4749209a22 changed all the existing uses at that point (along with a bunch of other unrelated changes :sweat:)

ganfra:
@csmith so I just added the `consumeWindowInsets`, even If I don't think in this case it was changing anything :P

csmith (approved):

ganfra:
> > @csmith so I just added the `consumeWindowInsets`, even If I don't think in this case it was changing anything :P
> 
> Oh, I was hoping it'd mean you didn't need the windowInset override in the TopBar :(

Nope the TopBar is not a child of content, so I guess it's not propagated like this

PR #817, opened 2023-07-07
Reduce code coverage threshold for presenters

csmith:
Drop from 90% to 85% to allow for smaller presenters where the number of untestable instructions generated by kotlin/compose can reach ~10% of the entire file.


bmarty (approved):
Thanks!

PR #820, opened 2023-07-07
Update rust sdk to 0.1.29 (new back pagination status api)

ganfra:
Mostly handling this https://github.com/matrix-org/matrix-rust-sdk/pull/2228

We don't have `VirtualItems` `Loading` and `TimelineStart` anymore, but a `paginationStatus` API so we can decide when we are showing the loading more indicator in the timeline. 

We are just missing the initial value of the paginationStatus, so it can lead to wrongly showing the indicator where there is nothing more to load. 

Also should fix a weird issue about jumping in the timeline.


jmartinesp:
I'm getting this when joining a room. Also, when I tap on `Accept ` in the invite, it's not removed from the list and the room doesn't appear on the room list.

https://github.com/vector-im/element-x-android/assets/480955/46365e4b-6c85-4c1e-9c02-65f514728319

ganfra:
Not ready yet, sorry.

ganfra:
Let me handle timeline changes in a separate PR.

PR #824, opened 2023-07-07
Remove CenterAlignedTopAppBar

### LINKED ISSUE #655: „skip“ when creating a room sticks to the title if the translated text is long(er)
### Steps to reproduce

1. Set language to German
1. Tap pencil floating button on the lower right
2. Select new room

#### What did you expect?

Can clearly see what my option here are

#### What happened instead?

„Personen einladen“ (invite people) and „Überspringen“ (skip) look like a single sentence and are too close to each other. As a user I may not easily get, that I have the option to skip it.

image

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

No

No

csmith:
Centering the titles is very iOS-y and not aligned with the material
guidelines (which say center aligned top bars are only for the main
root page in the app). They also present issues when we have titles
and textual actions that end up being quite long in other languages,
as they end up merging together.

Fixes #655


ganfra (approved):
Nice, looks better :-)

PR #826, opened 2023-07-07
Improve swipe to reply rendering

### LINKED ISSUE #810: Swipe to reply is way too sensitive
I am triggering it just by trying to scroll the timeline.

bmarty:
Stop using SwipeToDismiss which was not ideal, so reply icon was displayed too late.

Also tried to fix #810 using https://issuetracker.google.com/u/1/issues/269627294 but it was not working as expected.
Looks like the swipe gesture must be < 45° regarding on horizontal line to be taken into account, so I think this is acceptable.


julioromano (approved):

bmarty:
There is a regression detected by the screenshot test, I will check that.

PR #834, opened 2023-07-10
Feature/fga/avoid multiple sync loops

ganfra:
Attempts to avoid multiple sync loops because of https://github.com/matrix-org/matrix-rust-sdk/issues/2247

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt
@@ -24,23 +24,30 @@ import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.stateIn
 import org.matrix.rustcomponents.sdk.RoomListService
 import org.matrix.rustcomponents.sdk.RoomListServiceState
+import timber.log.Timber
+import java.util.concurrent.atomic.AtomicBoolean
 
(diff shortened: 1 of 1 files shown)


jmartinesp (approved):
LGTM!

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt
```diff
@@ -49,6 +56,10 @@ class RustSyncService(
         roomListService
             .stateFlow()
             .map(RoomListServiceState::toSyncState)
+            .onEach { state ->
+                Timber.v("Sync state=$state")
+                isSyncing.set(state == SyncState.Syncing)
+            }
```
bnjbvr, line 62: Is this properly mapping `RoomListServiceState::{Init, SettingUp}` to `Syncing`? If not, this is wrong, and may cause a `true -> false` transition (if the room list service was explicitly started, and then it's moving on to `SettingUp`)

jmartinesp:
```kotlin
internal fun RoomListServiceState.toSyncState(): SyncState {
    return when (this) {
        RoomListServiceState.INIT,
        RoomListServiceState.SETTING_UP -> SyncState.Idle
        RoomListServiceState.RUNNING -> SyncState.Syncing
        RoomListServiceState.ERROR -> SyncState.InError
        RoomListServiceState.TERMINATED -> SyncState.Terminated
    }
}
```

Only `Running` is mapped to `Syncing`, so I *think* it should be fine?

jmartinesp: > you start the sliding sync, so the atomic gets set to true

My bad, I didn't realise we were still using `compareAndSet` there. @ganfra maybe we should re-think this?

PR #840, opened 2023-07-10
Cleanup text

bmarty:
Reduce parameter list of `Text` to enforce the usage of `style` (`TextStyle`), which are provided by Compound. It's style possible to override any parameter by copying the style, but we must do that as less as possible.

Also remove legacy `ElementColors` and `ElementTextStyles`. We must not use that anymore.

I try to get the Text style from Figma as much as possible, hopefully there will be no error (or a limited numbers of error). Screenshot test will help to see any regression. Figma is not always up to date regarding the usage of Compound, in this case, I tried to use the closest value.

CC @amshakal @callumu if you could have look at the screenshot change on this PR (even after it's been merged) it would be awesome.


ganfra (approved):
I can't really review that carefully, but I like the idea.

Just some toolbar titles looks a bit too big?

bmarty:
> Just some toolbar titles looks a bit too big?

I agree but using `fontHeadingSmMedium` has been confirmed by @amshakal 

I have fixed some regression and I am generating again the screenshot. I will check them again before merging.

amshakal:
I was looking at pages like settings and create room, where this heading style makes sense. For the rest i'll have to manually test pages on phone to confirm.

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/configureroom/ConfigureRoomView.kt
```diff
@@ -186,8 +187,7 @@ fun ConfigureRoomToolbar(
         title = {
             Text(
                 text = stringResource(R.string.screen_create_room_title),
-                fontSize = 16.sp,
-                fontWeight = FontWeight.SemiBold,
+                style = ElementTheme.typography.aliasScreenTitle,
```
ganfra, line 190: Are you sure this is a good idea to introduce alias? As when you look at figma you see the compound value

PR #843, opened 2023-07-11
Feature/fga/coroutine dispatcher

ganfra:
Rework how we use coroutine dispatchers in matrix module so we avoid being blocked by `Dispatchers.IO` being full.

Why we need this change? Because we are wrapping every rust sdk method with `withContext` and as some of these methods can block for a long time, we can end-up not having available threads in the `Dispatchers.IO` thread pool.

This is probably a temp solution as rust-sdk methods will become suspendable soon™.

`parallelism` values given to `limitedParallelism` are kind of experimental, not sure what we should use. 
Note that this is the recommended way to use `Dispatchers.IO` for blocking calls.

See documentation of Dispatchers.IO and usage of views :

> Dispatchers.IO has a unique property of elasticity: its views obtained with CoroutineDispatcher.limitedParallelism are not restricted by the Dispatchers.IO parallelism. Conceptually, there is a dispatcher backed by an unlimited pool of threads, and both Dispatchers.IO and views of Dispatchers.IO are actually views of that dispatcher. In practice this means that, despite not abiding by Dispatchers.IO's parallelism restrictions, its views share threads and resources with it.
> In the following example
> ```
> // 100 threads for MySQL connection
> val myMysqlDbDispatcher = Dispatchers.IO.limitedParallelism(100)
> // 60 threads for MongoDB connection
> val myMongoDbDispatcher = Dispatchers.IO.limitedParallelism(60)
> ```
> 
> the system may have up to 64 + 100 + 60 threads dedicated to blocking tasks during peak loads, but during its steady state there is only a small number of threads shared among Dispatchers.IO, myMysqlDbDispatcher and myMongoDbDispatcher.

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
@@ -23,7 +23,6 @@ import io.element.android.libraries.matrix.api.core.ProgressCallback
 import io.element.android.libraries.matrix.api.core.RoomId
 import io.element.android.libraries.matrix.api.core.SessionId
 import io.element.android.libraries.matrix.api.core.UserId
-import io.element.android.libraries.matrix.api.room.location.AssetType
 import io.element.android.libraries.matrix.api.media.AudioInfo
 import io.element.android.libraries.matrix.api.media.FileInfo
 import io.element.android.libraries.matrix.api.media.ImageInfo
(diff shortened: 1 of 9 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
```diff
@@ -320,55 +327,55 @@ class RustMatrixRoom(
...
     override suspend fun setTopic(topic: String): Result<Unit> =
-        withContext(coroutineDispatchers.io) {
+        withContext(roomDispatcher) {
             runCatching {
                 innerRoom.setTopic(topic)
             }
         }
 
-    private suspend fun fetchMembers() = withContext(coroutineDispatchers.io) {
+    private suspend fun fetchMembers() = withContext(roomDispatcher) {
```
bmarty, line 372: You do not use `roomMembersDispatcher` here?

bmarty (approved):
LGTM, just one question. Thanks for the explanation in the description.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
ganfra: We are not spamming it, so I don't think it's necessary, we are just calling it once.

PR #844, opened 2023-07-11
Reduce filename size for screenshot test

bmarty:
For a better devX. Having long names is hiding the useful part of the name (the name of the View), for instance in PR review:

image

Now the names will look like:

image

This PR also contains small cleanup.


tests/uitests/src/test/kotlin/ui/T.kt
```diff
@@ -75,7 +75,7 @@ class ScreenshotTest {
     )
 
     @Test
```
ElementBot, line 77: - :warning: A member is named after the class. This might result in confusion. Either rename the member or change it to a constructor.

tests/uitests/src/test/kotlin/ui/T.kt
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-package io.element.android.tests.uitests
+package ui
```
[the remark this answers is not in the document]
bmarty: Yes, I will comment, but I do not want to trigger the CI again for this. Will do in a later PR (I promise!).

tests/uitests/src/test/kotlin/ui/ComponentTestPreview.kt
```diff
@@ -28,4 +28,8 @@ class ComponentTestPreview(
     override val name: String = showkaseBrowserComponent.componentName
 
     override fun toString(): String = showkaseBrowserComponent.componentKey
+        // Strip package (Preview MUST all have a distinct name)
```
julioromano, line 31: Can we keep this as a last resort? There is noway to detect such clashes at compile time so it will degrade the devx.
Can we just strip the common `io.element.android` from the pacakge name for now?

julioromano (commented):
Some other comments

PR #845, opened 2023-07-11
Try to debug project automation column issues

kittykat:
The error I get from my private test repo is different from this one. Need to check what the first step returns


ganfra (approved):
LGTM I guess?

PR #850, opened 2023-07-12
Ensure CI run all the tests.

bmarty:
There were some failing tests, but the CI does not see it. It seems that koverMergedReport does not run all the tests (?).

Draft. Let's the CI fail to check that this is working. Next commits will fix the failing tests.

### DIFF
--- a/.github/workflows/nightlyReports.yml
+++ b/.github/workflows/nightlyReports.yml
@@ -27,7 +27,7 @@ jobs:
           java-version: '17'
 
       - name: ⚙️ Run unit & screenshot tests, generate kover report
-        run: ./gradlew koverMergedReport $CI_GRADLE_ARG_PROPERTIES -Pci-build=true
+        run: ./gradlew test koverMergedReport $CI_GRADLE_ARG_PROPERTIES -Pci-build=true
 
       - name: ✅ Upload kover report
         if: always()
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,7 +38,7 @@ jobs:
           cache-read-only: ${{ github.ref != 'refs/heads/develop' }}
 
       - name: ⚙️ Run unit & screenshot tests, generate kover report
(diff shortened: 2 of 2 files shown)


libraries/deeplink/src/main/kotlin/io/element/android/libraries/deeplink/DeeplinkParser.kt
```diff
@@ -37,21 +36,15 @@ class DeeplinkParser @Inject constructor() {
...
-        val screenPathComponent = pathBits.elementAtOrNull(1)
-        val roomId = tryOrNull { screenPathComponent?.let(::RoomId) }
 
-        return when {
-            roomId != null -> {
+        return when (val screenPathComponent = pathBits.elementAtOrNull(1)) {
+            null -> DeeplinkData.Root(sessionId)
+            DeepLinkPaths.INVITE_LIST -> DeeplinkData.InviteList(sessionId)
+            else -> {
+                val roomId = screenPathComponent.let(::RoomId)
```
bmarty, line 44: See the commit message:

Rework DeeplinkParser to fix a test (and fix a bug in release mode).
The test was failing in release mode because there is not check on `RoomId` format, so INVITE_LIST value ("invites") is seen as a valid RoomId.
First check for known paths, then try to parse as RoomId. The `tryOrNull` will return null only in debug mode, so I think we can remove it.
Error was:
value of: getFromIntent(...)
expected: InviteList(sessionId=@alice:server.org)
but was : Room(sessionId=@alice:server.org, roomId=invites, threadId=null)
	at io.element.android.libraries.deeplink.DeeplinkParserTest.nominal cases(DeeplinkParserTest.kt:54)

bmarty:
The pb is that the test are much longer to run:

image

But I suggest to merge and check that later.

jmartinesp (approved):
LGTM. I'll take a look at the longer test times tomorrow, no worries.

PR #852, opened 2023-07-12
More log and various fixes

bmarty:
In this PR:

- Fix issue when room list was not visible if the node are destroyed (repro on Samsung S9+). `RootFlowNode.kt` is now saving and restoring the state using the `MatrixClientHolder`, which is not a singleton anymore.
- Fix issue when image are not loaded after clear cache.
- More log on Node lifecycle. Will help to see where is the user looking at the logs (and rageshakes).
- Improve some logs.

Co-authored by @ganfra.


ganfra (approved):
Thanks!

ganfra (approved):
Thanks Maestro!

PR #858, opened 2023-07-12
Ensure pending intent data are unique.

### LINKED ISSUE #855: Unable to dismiss notification
Sometimes I can have lots of notifications, like this:

image

I can dismiss them all, but after a few seconds, they are coming back.

bmarty:
Fixes #855.

`Intent.data` of `PendingIntent` must be unique for the system not to mess up.


ganfra (approved):
Not tested but LGTM!

PR #859, opened 2023-07-12
Create value class for transactionId

bmarty:
There is no validation on the format, but validation is a bonus for userId, roomId, etc.

The main advantage of using value classes instead of Strings everywhere is to detect errors at compilation time (this PR is not fixing any issue).


csmith (approved):
Nice, LGTM

PR #863, opened 2023-07-12
Feature/fga/timeline pagination

ganfra:
Improves timeline pagination : 
- Avoid paginating while the timeline is not ready (aka cached items have been processed)
- Split cache items to emit data faster when opening the room

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/MatrixTimelineDiffProcessor.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/MatrixTimelineDiffProcessor.kt
@@ -17,7 +17,6 @@
 package io.element.android.libraries.matrix.impl.timeline
 
 import io.element.android.libraries.matrix.api.timeline.MatrixTimelineItem
-import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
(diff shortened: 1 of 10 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/MatrixTimelineDiffProcessor.kt
```diff
@@ -30,20 +29,16 @@ internal class MatrixTimelineDiffProcessor(
...
 ) {
 
-    private val initLatch = CompletableDeferred<Unit>()
     private val mutex = Mutex()
 
     suspend fun postItems(items: List<TimelineItem>) {
         updateTimelineItems {
             val mappedItems = items.map { it.asMatrixTimelineItem() }
-            addAll(mappedItems)
+            addAll(0, mappedItems)
```
jmartinesp, line 37: Is this 0 here needed? Wouldn't this callback only be called the 1st time we receive items?

jmartinesp (approved):
LGTM! (And will create a few conflicts with https://github.com/vector-im/element-x-android/pull/839 🥲 )

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/MatrixTimelineDiffProcessor.kt
ganfra: Nope because I'm now splitting the initialItems in small chunks to post items asap if you have still paginated a lot manually in the room : 
https://github.com/vector-im/element-x-android/blob/94cb694a0dcf47af66a742bdb76cd68a57e3c5a4/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/RustMatrixTimeline.kt#L89C2-L90C48

PR #865, opened 2023-07-13
Try to improve 'Tests' CI job times

jmartinesp:
This PR separates the test CI jobs in:

- `gradle test` for debug and release modes (we were missing release mode here).
- `gradle verifyPaparazziDebug koverMergedReport koverMergedVerify` to run the screenshot tests, get coverage and verify %s (verifyPaparazziDebug and kover tasks need to run in the same gradle command to pass the coverage info).

It should decrease test times, since we avoid running the screenshot tests 3 times (twice for `test`, once for `verifyPaparazziDebug`) when we need them just once.

### DIFF
--- a/.github/workflows/nightlyReports.yml
+++ b/.github/workflows/nightlyReports.yml
@@ -27,10 +27,10 @@ jobs:
           java-version: '17'
 
       - name: ⚙️ Run unit tests, debug and release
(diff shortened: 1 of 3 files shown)


ganfra (approved):
Thanks for working on this, LGTM!

PR #867, opened 2023-07-13
Fix: make sure we ignore notifications for open rooms

### LINKED ISSUE #866: Notifications are displayed for messages sent to the current room
### Steps to reproduce

1. Open a room in the app.
2. From another client, send a message.

#### What did you expect?

Since the room is already displayed, no notification should be displayed.

#### What happened instead?

The notification is displayed.

### Your phone model

OnePlus 6T

### Operating system version

Android 11.0

Element X - develop (106d0708462a0f45a0c04102012f8c3cdbe6e1c3)

### Homeserver

_No response_

### Will you send logs?

No

Yes

jmartinesp:
## Changes

To ignore notifications for the currently displayed room, we were previously checking `ProcessLifecycleOwner.get().lifecycle` in `NotifiableMessageEvent`. However, this value seems to always return the `INITIALIZED`, not ever updating.

The new changes add an observer that publishes the current app state to `AppNavigationStateService`.

## Why

Fixes #866.


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
```diff
@@ -87,7 +87,7 @@ class RustMatrixClient constructor(
 
     override val sessionId: UserId = UserId(client.userId())
     private val app = client.app().use { builder ->
-        builder.finish()
+        builder.withEncryptionSync(false, "").finish()
```
ganfra, line 90: Should be part of another PR if we really enable this I think

app/src/main/kotlin/io/element/android/x/ElementXApplication.kt
```diff
@@ -43,6 +49,13 @@ class ElementXApplication : Application(), DaggerComponentOwner {
             initializeComponent(MatrixInitializer::class.java)
             initializeComponent(EmojiInitializer::class.java)
         }
+
+        val appNavigationStateService = appComponent.appNavigationStateService()
+        ProcessLifecycleOwner.get().lifecycle.addObserver(object : LifecycleEventObserver {
```
ganfra, line 54: Can you not listen to this directly in the `AppNavigationStateService`?

services/appnavstate/api/src/main/kotlin/io/element/android/services/appnavstate/api/AppNavigationStateService.kt
```diff
@@ -24,6 +25,9 @@ import kotlinx.coroutines.flow.StateFlow
 
 interface AppNavigationStateService {
     val appNavigationStateFlow: StateFlow<AppNavigationState>
+    val appIsInForeground: StateFlow<Boolean>
```
ganfra, line 28: Maybe encapsulate so we only have one StateFlow?
Something like that : (not sure about the naming)
```
data class AppNavigationState(
    val isInForeground: Boolean,
    val currentNavigationStep: NavigationStep 
)
```

ganfra (commented):
Some remarks

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
jmartinesp: My bad, I thought I had removed that!

csmith (approved):

jmartinesp:
Just to double check @ganfra any further comments after the latest changes?

PR #868, opened 2023-07-13
Feature/fga/better timeline scroll

ganfra:
This PR provides improvements over : 
- The JumpToBottomButtin look and feel
- The auto-scroll when new message come in 
- The id of `TimelineItem.GroupedEvents`  so it's unique and stable


ganfra:
@jmartinesp Can you approve or comments if you want some more changes as we know the issues on the timeline are not related?

jmartinesp (approved):
I wanted to test it manually a bit first, to make sure there are no corner cases we missed, but that won't be possible until the other bug is fixed so I'll approve it and we thoroughly test it later and modify/revert it in the worst case scenario.

Also, a couple of minor comments, feel free to ignore them if you'd rather focus on something else.

jmartinesp:
I just tested it and it works fine. I pushed the contents of the branch after merging the latest `develop` commits and triggered a record screenshots workflow for it which should fix the Test workflow.

PR #876, opened 2023-07-14
Extract MessageComposerContext class from MessageComposerPresenter

julioromano:
When sending "Composer" analytics from screens other than the composer's (e.g. send location from map) we need to know the composer's mode in order to properly fill the analytics event. `MessageComposerContext` hoists this state so that other presenters can also read it.

Related to:
https://github.com/vector-im/element-meta/issues/1674
https://github.com/vector-im/element-meta/issues/1682

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -48,7 +48,7 @@ sqldelight = "1.5.5"
 telephoto = "0.4.0"
 
 # DI
-dagger = "2.46.1"
+dagger = "2.47"
 anvil = "2.4.6"
 
 # Auto service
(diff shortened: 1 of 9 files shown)


gradle/libs.versions.toml
```diff
@@ -48,7 +48,7 @@ sqldelight = "1.5.5"
 telephoto = "0.4.0"
 
 # DI
-dagger = "2.46.1"
```
jmartinesp, line 51: Should this be in a separate PR?

jmartinesp (approved):
LGTM, but there's an unexpected dependency version upgrade. Shouldn't renovate take care of that?

PR #878, opened 2023-07-14
Add fixed automation for moving issues between columns

kittykat:
This should fix the problem we were having earlier


jmartinesp (approved):
Let's see how it goes

PR #882, opened 2023-07-17
Sync Strings

ElementBot:


julioromano (approved):

features/messages/impl/src/main/res/values-fr/translations.xml
```diff
@@ -4,10 +4,27 @@
     <item quantity="one">"%1$d changement dans la conversation"</item>
     <item quantity="other">"%1$d changements dans la conversation"</item>
   </plurals>
+  <plurals name="screen_room_timeline_more_reactions">
+    <item quantity="one"></item>
```
bmarty, line 8: There is a pb here, the value is missing. The CI should detect missing values. :/

features/messages/impl/src/main/res/values-sk/translations.xml
```diff
@@ -5,6 +5,11 @@
     <item quantity="few">"%1$d zmeny miestnosti"</item>
     <item quantity="other">"%1$d zmien miestnosti"</item>
   </plurals>
+  <plurals name="screen_room_timeline_more_reactions">
+    <item quantity="one"></item>
```
bmarty, line 9: There is a pb here, the value is missing. The CI should detect missing values. :/

features/messages/impl/src/main/res/values-fr/translations.xml
julioromano: It does but it's only a warning so I thought it was intentional

bmarty: I have created https://github.com/vector-im/element-x-android/issues/885 to improve that when we will have more time.

PR #883, opened 2023-07-17
Fix sliding sync loop restarts due to expirations

### LINKED ISSUE #880: Sliding sync loop continuously gets expired session errors and restarts
### Steps to reproduce

1. Use a version of the app using Rust SDK v0.1.31.
2. Open a room.
3. Send two messages to that room from another client.

#### What did you expect?

The 2 messages should be received and displayed quite fast.

#### What happened instead?

The first message is displayed quite quickly, but the 2nd one gets 'stuck'. Eventually, the logs display an 'expired session' error for the sync loop and the sync loop gets restarted, and *then* we receive the 2nd message after 5-10s, although sometimes that message just never arrives.

Note that this doesn't happen in versions of the app using the SDK at v0.1.29 (v0.1.30 was skipped).

### Your phone model

OnePlus 6T

### Operating system version

Android 11

Element X (using Rust SDK 0.1.31)

### Homeserver

_No response_

### Will you send logs?

Yes

Yes

jmartinesp:
## Changes

Both `NotifiableEventResolver` and `DefaultNotificationDrawerManager` were creating new Rust SDK Clients while processing notifications instead of reusing the already existing one through `MatrixClientsHolder`.

## Why

Fixes #880.


jmartinesp:
I can't approve the PR since I created it, but with the latest changes it seems to work fine. I'd just add some extra documentation to `RootNavStateFlowFactory` if possible.

jmartinesp:
Also, I don't know why but Kover verification task is failing consistently with this branch in the CI, but it works fine for me locally...

ganfra (approved):
Added some more rework as we discussed. Let's merge it and see how it goes.

bmarty:
I am checking the branch right now FTR (not blocking merging it)

bmarty:
I have got an infinite loading of room:
- open a room (timeline)
- change the phone language
- go back to the app.

Previously (before the latest commits) the app was going back to the room list, which was less bad.

bmarty:
> I have got an infinite loading of room:
> 
>     * open a room (timeline)
> 
>     * change the phone language
> 
>     * go back to the app.
> 
> 
> Previously (before the latest commits) the app was going back to the room list, which was less bad.

OK, this is fixed with last commit https://github.com/vector-im/element-x-android/pull/883/commits/207a20b67c4e5ea7f932391bfd9fcbf2265241c3. Well done @ganfra !

bmarty (approved):

PR #884, opened 2023-07-17
Disable Paparazzi tasks when Kover is running.

jmartinesp:
It allows us to split the test jobs between unit tests, screenshot test and coverage reports.

I moved the `sonar` job to the quality workflow, as it seems more related to quality than tests, and we're not collecting coverage from the tests anyway, but maybe it should have its own workflow?

### DIFF
--- a/tests/uitests/build.gradle.kts
+++ b/tests/uitests/build.gradle.kts
@@ -17,6 +17,8 @@
 import extension.allFeaturesImpl
 import extension.allLibrariesImpl
 import extension.allServicesImpl
+import kotlinx.kover.tasks.KoverReportTask
+import org.gradle.api.internal.provider.DefaultProvider
 
 plugins {
     id("io.element.android-compose-library")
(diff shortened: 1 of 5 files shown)


tests/uitests/build.gradle.kts
```diff
@@ -17,6 +17,8 @@
 import extension.allFeaturesImpl
 import extension.allLibrariesImpl
 import extension.allServicesImpl
+import kotlinx.kover.tasks.KoverReportTask
```
ElementBot, line 20: - :no_entry_sign: Unused import

tests/uitests/build.gradle.kts
```diff
@@ -17,6 +17,8 @@
...
+import org.gradle.api.internal.provider.DefaultProvider
```
ElementBot, line 21: - :no_entry_sign: Unused import

csmith (approved):

PR #886, opened 2023-07-17
Location sharing: don't hardcode API key

csmith:
In an effort to make it easier for forks to (a) use their own API keys (b) change map styles or maybe even providers, move the MapTiler key out of the source code and pass it in via env var or property.

Also refactor the utility classes slightly to keep all the URL related functions together, to reduce the chance of collisions when maintaining such forks.

----

If you don't configure an API key then everything still works, but the tiles/maps won't load (because maptiler won't serve them without an API key):


features/location/api/build.gradle.kts
```diff
@@ -14,14 +14,33 @@
...
+
 android {
     namespace = "io.element.android.features.location.api"
+
+    defaultConfig {
+        resValue(
+            type = "string",
+            name = "maptiler_api_key",
+            value = System.getenv("ELEMENT_ANDROID_MAPTILER_API_KEY")
+                ?: readLocalProperty("services.maptiler.apikey") as? String
```
bmarty, line 40: I did not test it, but can't you just invoke `project.property("services.maptiler.apikey")`(and remove the `fun readLocalProperty`), so that the key can be placed in any project property file, or even passed in the command line when building the project.

csmith: That's what I did originally, but gradle doesn't actually include configuration from `local.properties` (or any equivalent) in its normal property resolution. The only file it sources from that wouldn't be committed is `$GRALDE_USER_HOME/gradle.properties`, and putting project-specific secrets in a global file seems a bit gross (it gets very messy if every project does that, risks having name collisions, leaking secrets to other builds, etc).

We could use a plugin like [gradle-properties-plugin](https://github.com/stevesaliman/gradle-properties-plugin), but that felt a bit overkill for now.

bmarty: OK, thanks. TIL gradle ignores `local.properties`. Related: https://github.com/gradle/gradle/issues/12283

julioromano: @csmith why you used `resValue` and not `buildConfigField` ?

features/location/api/src/main/kotlin/io/element/android/features/location/api/internal/MapsUtils.kt
```diff
@@ -1,91 +0,0 @@
...
-    lat: Double,
-    lon: Double,
-    desiredZoom: Double,
-    desiredWidth: Int,
-    desiredHeight: Int,
-    darkMode: Boolean,
-    doubleScale: Boolean,
-    attributionPlacement: AttributionPlacement,
-): String {
-    require(desiredWidth > 0 && desiredHeight > 0) {
```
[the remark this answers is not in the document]
csmith: Things like the max width/height are maptiler specific, and it didn't seem worth "exposing" that and adding more complexity if someone wants to change map provider.

Most other things just aren't going to happen the way we use this - e.g. we always pass a fixed zoom in for a static map, we're not losing anything by checking it's >=0 and coercing it to be <=22.

And if something _does_ go wrong, we already handle the map failing to load (which is probably a better failure mode than throwing an exception?).

features/location/api/build.gradle.kts
csmith: > why you used resValue and not buildConfigField ?

No compelling reason. Could have used either, really. I dislike BuildConfigs a little because Android Studio seems incapable of finding them half the time, and they can end up being a dumping ground for any random thing related to the app.

PR #887, opened 2023-07-17
Fix composer state

### LINKED ISSUE #870: Message input box is cleared after viewing an image from timeline
### Steps to reproduce

1. Go into a room
2. Type a message
3. Tap an existing photo in timeline
4. Go back from preview to timeline

#### What did you expect?
Message input box contents to stay intact.

#### What happened instead?
Content of the input box disappears.

### Your phone model

Google Pixel 3

### Operating system version

Android 12

c919618bc0152de22c2ac2b36ed911a2c59e06e8 (current develop branch)

### Homeserver

Synapse 1.87.0

### Will you send logs?

Yes

No

bmarty:
Avoid loosing text in composer when navigating to a sub Node.

Fixes #870 

First commit is some cleanup. We may iterate again when markdown will be properly implemented using the RichTextEditor.

## Test

- Send an image to a timeline
- Type some text in the composer
- Open the image
- Close the image viewer

Before, the composer was emptied. Now the entered text is still present.

Note: this PR does not implement draft, i.e. if the user goes back to the room list, the entered text is lost. (tracked by the lucky issue #888)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerView.kt
```diff
@@ -69,7 +69,7 @@ fun MessageComposerView(
             onAddAttachment = ::onAddAttachment,
             onFocusChanged = ::onFocusChanged,
             composerCanSendMessage = state.isSendButtonVisible,
-            composerText = state.text?.charSequence?.toString(),
+            composerText = state.text,
```
bmarty, line 72: `charSequence?.toString()` was used, so any information in the CharSequence was lost anyway.

libraries/textcomposer/src/main/kotlin/io/element/android/libraries/textcomposer/TextComposer.kt
```diff
@@ -96,7 +96,7 @@ fun TextComposer(
     focusRequester: FocusRequester = FocusRequester(),
     onSendMessage: (String) -> Unit = {},
     onResetComposerMode: () -> Unit = {},
-    onComposerTextChange: (CharSequence) -> Unit = {},
+    onComposerTextChange: (String) -> Unit = {},
```
bmarty, line 99: The lambda `BasicTextField.onValueChange` parameter is a String, so this change has no impact.

PR #899, opened 2023-07-18
Make the functions in SystemUtils extensions

julioromano:
- They are now all extensions over `Context` or `Activity` (when `Context` is not enough) (some of them already were).
- Allows for IDE completion.


bmarty (approved):
Thanks for this change, LGTM, as long as the impacted and used methods still works as expected.

PR #901, opened 2023-07-18
Feature/fga/power level

ganfra:
Use the new api from rust sdk to get power level without the users being fetched.
Need PR about updating rust sdk first.


libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/MatrixRoom.kt
```diff
@@ -99,11 +99,11 @@ interface MatrixRoom : Closeable {
...
     suspend fun inviteUserById(id: UserId): Result<Unit>
 
-    suspend fun canInvite(): Result<Boolean>
+    suspend fun canUserInvite(userId: UserId): Result<Boolean>
 
-    suspend fun canSendStateEvent(type: StateEventType): Result<Boolean>
+    suspend fun canUserSendState(userId: UserId, type: StateEventType): Result<Boolean>
 
-    suspend fun canSendEvent(type: MessageEventType): Result<Boolean>
+    suspend fun canUserSendMessage(userId: UserId, type: MessageEventType): Result<Boolean>
```
bmarty, line 106: Not sure that the new parameter `userId` will be useful for those 3 methods, but we will see.

bmarty (approved):
LGTM, thanks. I have added some small cleanup commits.

PR #903, opened 2023-07-18
React to session deletion (from another session)

### LINKED ISSUE #889: Handle correctly session deletion from another client
Currently the app does not react if the session is deleted from another session.

You can still open room and send messages (an error is displayed).

After app cold start, the room list is not displayed, but nothing happen.

User can still sign out to log in again, so they are not stuck.

bmarty:
This is a quick implementation to react on session deleted on server (for instance using another session).

Design may be added later to explain why the session has been destroyed.

FTR we have this design (more or less :) ) on Element Android right now:

|Logout|Soft Logout|
|-|-|
|image|image|

Also we may want to take the isSoftLogout parameter later (design also needed here).

Limited by https://github.com/matrix-org/sliding-sync/issues/212, i.e. the app will react only when user do a request that is not a sliding sync request, like loading an avatar, etc. After a cold start, the room list is not loaded, so the only way to have the logout is to try to create a room. Also, user can still logout manually.

Closes #889

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
@@ -61,6 +61,7 @@ import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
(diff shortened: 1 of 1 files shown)


ganfra (approved):
LGTM!

bmarty:
(Ignoring screenshot test failing, not related to the change here, but waiting for Maestro cloud to be happy before merging (checking logout))

bmarty:
No more Maestro quota :/

PR #907, opened 2023-07-18
RoomList: introduce RoomListDataSource 

ganfra:
This PR introduce a new class `RoomListDataSource` which is injected in `RoomListPresenter` and will keep the mapped item in memory. 
It'll avoid losing the `scrollPosition` when navigating back to the `RoomListNode`.


bmarty (approved):
Room list and room filter tested OK on my side, and room list scroll state is preserved. Thanks!

PR #912, opened 2023-07-19
Reactions layout improvements

langleyd:
## What's in this PR
- Add the custom reaction layout that only shows the expand UI after 2 lines.
- It also enforces that the add more and expand buttons are always on the same line.
- In LTR languages we want an incoming message's reactions to be LRT and outgoing to be RTL.
- For RTL languages it should be the opposite.
- Add more button should have primary text colour.


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemReactionsLayout.kt
```diff
@@ -0,0 +1,207 @@
...
+ * @param itemSpacing The horizontal spacing between items
+ * @param rowSpacing The vertical spacing between rows
+ * @param expanded Whether the layout should display in expanded or collapsed state
+ * @param rowsBeforeCollapsible The number of rows before the collapse/expand button is shown
+ * @param expandButton The expand button
+ * @param addMoreButton The add more button
+ * @param reactions The reaction buttons
+ */
+@Composable
+fun TimelineItemReactionsLayout(
```
ElementBot, line 47: - :warning: Documentation of TimelineItemReactionsLayout is outdated

jmartinesp:
I've added a few commits to take care of lint issues, I hope you don't mind.

jmartinesp (approved):
LGTM, thanks!

PR #913, opened 2023-07-19
Avoid spamming sync flow by checking item origin

ganfra:


jmartinesp (approved):
Your changes look good to me. I also added a fix for the loop for fetching event details.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/TimelineDiffExt.kt
```diff
@@ -0,0 +1,54 @@
...
+            pushFront()?.eventOrigin()
+        }
+        TimelineChange.SET -> {
+            set()?.item?.eventOrigin()
+        }
+        TimelineChange.INSERT -> {
+            insert()?.item?.eventOrigin()
+        }
+        TimelineChange.RESET -> {
+            reset()?.first()?.eventOrigin()
```
bmarty, line 46: firstOrNull?

PR #916, opened 2023-07-19
Location expanded view: show own location

julioromano:
If the location permission is granted:
- Shows the user's own location
- Shows a button to center the map on it

Part of:


csmith (approved):
Nice, LGTM

PR #923, opened 2023-07-19
Add screenshots with ~1.5 lines long desc to expanded location view

julioromano:
This will help in catching alignment regressions.

Related to:

### DIFF
--- a/features/location/impl/src/main/kotlin/io/element/android/features/location/impl/show/ShowLocationStateProvider.kt
+++ b/features/location/impl/src/main/kotlin/io/element/android/features/location/impl/show/ShowLocationStateProvider.kt
@@ -52,6 +52,13 @@ class ShowLocationStateProvider : PreviewParameterProvider<ShowLocationState> {
             ),
             ShowLocationState(
                 Location(1.23, 2.34, 4f),
+                description = "For some reason I decided to to write a small essay that wraps at just two lines!",
+                hasLocationPermission = false,
+                isTrackMyLocation = false,
+                eventSink = {},
+            ),
+            ShowLocationState(
+                Location(1.23, 2.34, 4f),
                 description = "For some reason I decided to write a small essay in the location description. " +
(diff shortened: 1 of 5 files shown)


csmith:
Good catch, but I'm not sure justifying the text is correct here. We want it centered, and if you have a description that wraps one or two words onto the second line it looks pretty bad:

image

I wonder why it's not centered properly - something to do with the overflow ellipses? :thinking:

features/location/impl/src/main/kotlin/io/element/android/features/location/impl/show/ShowLocationView.kt
```diff
@@ -138,12 +137,12 @@ fun ShowLocationView(
             state.description?.let {
                 Text(
                     text = it,
-                    textAlign = TextAlign.Center,
+                    textAlign = TextAlign.Justify,
```
jmartinesp, line 140: I'd just use `TextAlign.Start` here, justify looks a bit weird, as @csmith mentioned.

julioromano: TextAlign.Start doesn't change things much:
Screenshot 2023-07-20 at 09 33 46
Screenshot 2023-07-20 at 09 33 41
Screenshot 2023-07-20 at 09 33 38

jmartinesp: Oh sorry, I misunderstood it, you *always* want it centered, even if the text has several lines and overflows. It seems a bit weird to me, but in that case we should go back to center aligning the text and trying to figure out why we have that extra indentation instead.

julioromano:
Ok I double checked with Callum so I left it as is, but I still added one screenshot with the second line half filled so that we can catch regressions on this.

csmith (approved):

PR #926, opened 2023-07-19
Extract maptiler's map ids

julioromano:
Maptiler custom map ids are only useable by the account that create them. So if we hardcode them forkers won't be able to use the maps even if the bring in their own api key (because they can't access our maps with their api key).

Requires to set our map ids in `local.properties` for local development:
```
services.maptiler.lightMapId=9bc819c8-e627-474a-a348-ec144fe3d810
services.maptiler.darkMapId=dea61faf-292b-4774-9660-58fcef89a7f3
```


features/location/api/build.gradle.kts
```diff
@@ -37,9 +37,23 @@ android {
...
+            value = System.getenv("ELEMENT_ANDROID_MAPTILER_LIGHT_MAP_ID")
+                ?: readLocalProperty("services.maptiler.lightMapId")
+                ?: "basic-v2" // fall back to maptiler's default map.
+        )
+        resValue(
+            type = "string",
+            name = "maptiler_dark_map_id",
+            value = System.getenv("ELEMENT_ANDROID_MAPTILER_DARK_MAP_ID")
+                ?: readLocalProperty("services.maptiler.darkMapId")
+                ?: "basic-v2" // fall back to maptiler's default map.
```
csmith, line 55: basic-v2-dark?

csmith (commented):
We'll need to pass our custom styles in through CI now, too, right? I guess they don't need to be secrets, but they'll still need to be set in ENV vars in the workflows

features/location/api/build.gradle.kts
julioromano: OH indeed it works but it was not shown in maptiler's dashboard. How did you figure it out?

csmith: It's a bit buried: https://www.maptiler.com/maps/dark/ has a "Use Basic Dark" link and the style ID is in the URL. (I think I'd seen it before, otherwise I wouldn't have known to gone hunting there!)

csmith (approved):
LGTM, just needs env vars adding for the CI builds

julioromano:
> We'll need to pass our custom styles in through CI now, too, right? I guess they don't need to be secrets, but they'll still need to be set in ENV vars in the workflows

Done!

.github/workflows/release.yml
```diff
@@ -27,6 +27,10 @@ jobs:
       - name: Configure gradle
         uses: gradle/gradle-build-action@v2.6.1
       - name: Create app bundle
+        env:
+          ELEMENT_ANDROID_MAPTILER_API_KEY: ${{ secrets.MAPTILER_KEY }}
```
csmith, line 31: Oh, did I miss one? Nice catch!

csmith (approved):

.github/workflows/release.yml
julioromano: No it's been added very recently

julioromano:
Maptiler custom map ids are only useable by the account that create them. So if we hardcode them forkers won't be able to use the maps even if the bring in their own api key (because they can't access our maps with their api key).

Requires to set our map ids in `local.properties` for local development:
```
services.maptiler.lightMapId=9bc819c8-e627-474a-a348-ec144fe3d810
services.maptiler.darkMapId=dea61faf-292b-4774-9660-58fcef89a7f3
```

PR #930, opened 2023-07-20
Correct location event body

julioromano:
- Now sending the correct body format as agreed with design. This won't be show in EX clients though.

Related to:
- https://github.com/vector-im/element-meta/issues/1682


csmith (approved):
Nice

PR #933, opened 2023-07-20
Pixel perfect static images

julioromano:
1. On devices less than xhdpi request a 1x image from MapTiler (such devices are generally old, slower and with little memory so avoiding to get the 2x image only to have to shrink it later could help).
2. Coerce too big width/height combos within the API limits keeping the aspect ratio (this will allow requests on big horizontal displays to succeed).
3. Don't crash when given weird width/height combos (i.e. zero or negative).
4. Introduce interfaces to hide this whole logic and make it easier for forks to implement their own.

Related to:


features/location/api/src/test/kotlin/io/element/android/features/location/api/internal/MapUrlsKt.kt
```diff
@@ -0,0 +1,64 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.features.location.api.internal
+
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+
+internal class MapUrlsKtTest {
```
ElementBot, line 22: - :warning: The file name 'MapUrlsKt' does not match the name of the single top-level declaration 'MapUrlsKtTest'.

csmith:
I don't think we want to do this.

Changing the size of the map we request affects the information density returned in the image. On an `xxxhdpi` display that would mean maptiler gives us a map suitable for display on something the size of a small computer monitor, and we display it in the space of a few postage stamps. It would be crisp, but it achieves that by rendering more area around the location we care about, effectively zooming out.

Sizing it to the physical pixel dimensions and then requesting the `@2x` version also means we end up with an image twice as large as needed in both dimensions. e.g. if we're filling a 300px x 150px area on an `xdpi` device we'd now get a 600px x 300px image and that would have to be scaled down by the platform to actually display.

iOS has the same behaviour as the old code: they pass the size _in points_ to MapTiler, which will be 2x or 3x smaller than the number of pixels depending on the device. (Presumably, on 3x devices the map isn't "pixel perfect" either)

julioromano:
> Sizing it to the physical pixel dimensions and then requesting the @2x version also means we end up with an image twice as large as needed in both dimensions. e.g. if we're filling a 300px x 150px area on an xdpi device we'd now get a 600px x 300px image and that would have to be scaled down by the platform to actually display.

That's not what this code does: Requested dimension is halved for @2x images thereby yielding the same pixel size static image but with twice the resolution (i.e. zoomed in).

csmith:
> That's not what this code does: Requested dimension is halved for @2x images thereby yielding the same pixel size static image but with twice the resolution (i.e. zoomed in).

Ah, I missed that. As it happens, that's exactly what I did when I first implemented it (halving the pixel count and requesting `@2x`), and I had to switch to the current implementation to get maps that looked OK/consistent.

julioromano:
To give you an idea here's how the 3 possible choices look:

current
2xHalved
2xNotHalved

csmith:
You can see the zoom out in the second image as well, and that zoom will change based on the density of the device, which we don't want. If you find a device with a density double your current one, it'll look like the bottom image, right?

julioromano:
> If you find a device with a density double your current one, it'll look like the bottom image, right?

Yes, and on the same device the current implementation will look twice as more blurred.
Anyways ~1000dpi devices are not a thing nowadays. They currently top out at ~560dpi. So it's not a thing we should worry about soon.

I don't think there's a right choice here: both choices have pros and cons.
**Current)** Pixel sharp. Text and map will shrink effectively zooming out as the device dpi increases further from @2x (i.e. ~320dpi).
**Proposed)** Blurred. Text and map will be of constant size regardless of the device. Blur will increase as the device dpi increases further from @2x (i.e. ~320dpi).

So either we keep the sharpness constant at the expense of a decreased content size.
OR
We keep the content size constant at the expense of a decreased sharpness (i.e. blur)

csmith:
IMO if the lack of crispness is an issue we should look at alternatives to maptiler that can provide images in the right density rather than just 1x and 2x (or at least provide controls for the map area separate to the dimensions of the static map). [But obviously that's probably out of scope for this initial implementation :sweat: ]

bmarty:
Just putting again the screenshots from [previous comment](https://github.com/vector-im/element-x-android/pull/933#issuecomment-1643967469) but in a table so that we can compare easily

|current|2xHalved|2xNotHalved|
|-|-|-|
|current|2xHalved|2xNotHalved

julioromano:
Thanks! The table helps to evaluate and compare the various content sizes, but to evaluate sharpness the images should be looked at 100% size (or even better by running the code on a screen of a real device).

bmarty (approved):
I have no particular remark on the code, but maybe what can be strange is that the event will be rendered differently between 2 devices. Also I think user can still open the map fullscreen by clicking on it and zoom as they want, so I think it's fine.

csmith:
Just to illustrate the pathological case, this is what a [Sony Xperia Z5 Premium](https://www.gsmarena.com/sony_xperia_z5_premium-7536.php) would look like:

<table>
<tr><td>Current (using DPI and 2x)<br>

</td><td>New (using pixels/2 and 2x)<br>

</td></tr></table>

And this is what it looks like in eleweb:

image

julioromano:
> IMO if the lack of crispness is an issue we should look at alternatives to maptiler that can provide images in the right density rather than just 1x and 2x

@csmith You hit the nail on the head. To properly (i.e. with a correct content size) rasterize a vector image (such as a map) we need 2 things:
1) width/height dimensions in pixels.
2) screen density (in DPI or as a scaling factor).

We can easily get this information from the Android OS but MapTiler's static map API doesn't take it as an input. It effectively only allows 1 and 2 as scaling factors.

I've changed the PR to basically keep the existing behavior but with a few tweaks:
1) On devices less than `xhdpi` request a 1x image from MapTiler (such devices are generally old, slower and with little memory so avoiding to get the 2x image only to have to shrink it later could help).
2) Coerce too big width/height combos within the API limits keeping the aspect ratio (this will allow requests on big horizontal displays to succeed).
3) Don't crash when given weird width/height combos (i.e. zero or negative).
2) Introduce interfaces to hide this whole logic and make it easier for forks to implement their own.

julioromano:
> 

Done!

PR #934, opened 2023-07-20
Clear all notifications of the session when the user signs out.

bmarty:
It seems that this part was missing :/


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotificationDrawerManager.kt
```diff
@@ -23,17 +23,18 @@ import io.element.android.libraries.core.data.tryOrNull
 import io.element.android.libraries.core.meta.BuildMeta
 import io.element.android.libraries.di.AppScope
 import io.element.android.libraries.di.SingleIn
+import io.element.android.libraries.matrix.api.MatrixClientProvider
```
ganfra, line 26: Seems not needed?

ganfra (approved):
One remark otherwise LGTM

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotificationDrawerManager.kt
bmarty: Yes, it's used, but the addition is caused by import reordering. It was at line 31 before.

PR #935, opened 2023-07-20
Code quality: warnings are errors and enable more Detekt rules.

bmarty:
First commit should make the CI fails. -> OK, got `e: warnings found and -Werror specified` in the log and the build is failing.
In the next commits I will handle the existing warnings.
Also took the opportunity to enable some `Detekt` rules and have a cleaner codebase.


bmarty:
@julioromano as a first and quick solution to detect warnings, I have added a script which can be run before pushing the branch or creating a PR. See in ce22aecbf2cc14303b18d0754e39d8b2111a904d.
The advantage is that it can also be used a pre-hook push (or commit, but can be long to run), if developers wants to do it.
Let me know what you think.

julioromano:
> @julioromano as a first and quick solution to detect warnings, I have added a script which can be run before pushing the branch or creating a PR. See in [ce22aec](https://github.com/vector-im/element-x-android/commit/ce22aecbf2cc14303b18d0754e39d8b2111a904d). The advantage is that it can also be used a pre-hook push (or commit, but can be long to run), if developers wants to do it. Let me know what you think.

Thank you, I'll try it that first.

bmarty:
(I will quickly fix the 2 reported error on develop)

PR #945, opened 2023-07-21
Upload APK to Firebase App Distribution for now

bmarty:
Revert change from #941, upload APK to Firebase App Distribution for now.

We need a matching and released PlayStore application to be able to upload an AAB, and we do not have that for now.

We will revisit this later.


.github/workflows/nightly.yml
```diff
@@ -45,11 +46,11 @@ jobs:
...
         continue-on-error: true # don't block anything by this upload failing (for now)
         run: |
-          ./gradlew assembleNightly $CI_GRADLE_ARG_PROPERTIES
+#          ./gradlew assembleNightly $CI_GRADLE_ARG_PROPERTIES
           curl -u "$BROWSERSTACK_USERNAME:$BROWSERSTACK_PASSWORD" -X POST "https://api-cloud.browserstack.com/app-automate/upload" -F "file=@app/build/outputs/apk/nightly/app-universal-nightly.apk" -F "custom_id=element-x-android-nightly"
         env:
-          ELEMENT_ANDROID_MAPTILER_API_KEY: ${{ secrets.MAPTILER_KEY }}
-          ELEMENT_ANDROID_MAPTILER_LIGHT_MAP_ID: ${{ secrets.MAPTILER_LIGHT_MAP_ID }}
-          ELEMENT_ANDROID_MAPTILER_DARK_MAP_ID: ${{ secrets.MAPTILER_DARK_MAP_ID }}
+#          ELEMENT_ANDROID_MAPTILER_API_KEY: ${{ secrets.MAPTILER_KEY }}
```
julioromano, line 52: Don't we need these also for apks?

julioromano (approved):
Pre emptive LGTM with a comment

.github/workflows/nightly.yml
bmarty: No, the APK is built by the previous step and the env var are set there (line 37).

PR #947, opened 2023-07-21
Sentry

### LINKED ISSUE #618: Integrate Sentry
Same config than Element Android for now.

bmarty:
Still waiting for sentry dns, but in the mean time this PR can be reviewed.

It also includes some fixes around analytics:
- be able (for fork) to fully disable analytics.
- cleanup of the code.
- add a "Crash me" item in the developer settings.

Screenshot recording on progress.

Closes #618

### DIFF
--- a/plugins/src/main/kotlin/extension/DependencyHandleScope.kt
+++ b/plugins/src/main/kotlin/extension/DependencyHandleScope.kt
@@ -103,6 +103,8 @@ fun DependencyHandlerScope.allLibrariesImpl() {
 }
 
 fun DependencyHandlerScope.allServicesImpl() {
+    // For analytics configuration, either use noop, or use the impl, with at least one analyticsproviders implementation
+    // implementation(project(":services:analytics:noop"))
     implementation(project(":services:analytics:impl"))
     implementation(project(":services:analyticsproviders:posthog"))
     implementation(project(":services:apperror:impl"))
--- a/plugins/src/main/kotlin/extension/DependencyHandleScope.kt
+++ b/plugins/src/main/kotlin/extension/DependencyHandleScope.kt
@@ -107,6 +107,8 @@ fun DependencyHandlerScope.allServicesImpl() {
     // implementation(project(":services:analytics:noop"))
     implementation(project(":services:analytics:impl"))
     implementation(project(":services:analyticsproviders:posthog"))
+    implementation(project(":services:analyticsproviders:sentry"))
+
     implementation(project(":services:apperror:impl"))
     implementation(project(":services:appnavstate:impl"))
     implementation(project(":services:toolbox:impl"))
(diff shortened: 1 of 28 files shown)


services/analytics/noop/src/main/kotlin/io/element/android/services/analytics/noop/NoopAnalyticsService.kt
```diff
@@ -0,0 +1,48 @@
...
+import io.element.android.services.analyticsproviders.api.AnalyticsProvider
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flowOf
+import javax.inject.Inject
+
+@SingleIn(AppScope::class)
+@ContributesBinding(AppScope::class)
+class NoopAnalyticsService @Inject constructor(
+) : AnalyticsService {
+    override fun getAvailableAnalyticsProviders(): Set<AnalyticsProvider> = emptySet()
```
bmarty, line 35: This will ensure the analytics setting is hidden

services/analytics/noop/src/main/kotlin/io/element/android/services/analytics/noop/NoopAnalyticsService.kt
```diff
@@ -0,0 +1,48 @@
...
+    override fun getUserConsent(): Flow<Boolean> = flowOf(false)
+    override suspend fun setUserConsent(userConsent: Boolean) = Unit
+    override fun didAskUserConsent(): Flow<Boolean> = flowOf(true)
```
bmarty, line 38: returning true will ensure that the analytics opt-in screen is not displayed.

plugins/src/main/kotlin/extension/DependencyHandleScope.kt
```diff
@@ -103,8 +103,12 @@ fun DependencyHandlerScope.allLibrariesImpl() {
 }
 
 fun DependencyHandlerScope.allServicesImpl() {
+    // For analytics configuration, either use noop, or use the impl, with at least one analyticsproviders implementation
+    // implementation(project(":services:analytics:noop"))
```
jmartinesp, line 107: We could try using some config value like:

```kotlin
if (project.property("disable-analytics") == "true") {
    implementation(project(":services:analytics:noop"))
} else {
    implementation(project(":services:analytics:impl"))
}
```

Or some other custom value we could add to a config module maybe? We don't really need it now though, it's just a suggestion for the future.

jmartinesp (approved):
LGTM, thanks!

plugins/src/main/kotlin/extension/DependencyHandleScope.kt
bmarty: Yes, good point, I would love to have a common place where all the possible tweaks, configurations and settings are centralized... At least the project is quite well architectured to be able to do whatever we want in the future regarding that.

bmarty:
Crash reported on the new project, only once analytics has been enabled.

image

bmarty:
Screenshot test has been cancelled, I do not know why:

> Error: The operation was canceled.

I do not want to block this PR from being merged because of that. There is no change in the UI iun this PR.

PR #963, opened 2023-07-26
Show location request permission in expanded location view

julioromano:
When showing the map after clicking on a shared location,
the crosshair button is now always visible and it triggers the permission request flow if there are no location permissions.

Related to:


csmith (approved):

bmarty:
(just rebased and triggered SC recording)

bmarty (approved):
Let's merge this now.

PR #967, opened 2023-07-26
Update detekt config.

bmarty:
Enable mode detekt rules, which are disabled by default.

### DIFF
--- a/libraries/maplibre-compose/src/main/kotlin/io/element/android/libraries/maplibre/compose/MapboxMap.kt
+++ b/libraries/maplibre-compose/src/main/kotlin/io/element/android/libraries/maplibre/compose/MapboxMap.kt
@@ -236,7 +236,7 @@ private fun MapView.lifecycleObserver(previousState: MutableState<Lifecycle.Even
             Lifecycle.Event.ON_DESTROY -> {
                 //handled in onDispose
             }
-            Lifecycle.Event.ON_ANY -> Unit
+            Lifecycle.Event.ON_ANY -> error("ON_ANY should never be used")
         }
         previousState.value = event
     }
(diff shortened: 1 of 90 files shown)


libraries/maplibre-compose/src/main/kotlin/io/element/android/libraries/maplibre/compose/MapboxMap.kt
```diff
@@ -236,7 +236,7 @@ private fun MapView.lifecycleObserver(previousState: MutableState<Lifecycle.Even
             Lifecycle.Event.ON_DESTROY -> {
                 //handled in onDispose
             }
-            else -> throw IllegalStateException()
+            Lifecycle.Event.ON_ANY -> Unit
```
bmarty, line 239: I am wondering if this case could happen...

julioromano: What triggered this change by detekt?

bmarty: I tried to group the change in the detekt config and the fixes per commit, so it's in 73347deaafd6f851ecc20a72bf2be57b94741891: `UseCheckOrError`.

But you can confirm if the change is fine. I do not think we want to crash the app there.

julioromano: This code comes from here: https://github.com/googlemaps/android-maps-compose/blob/main/maps-compose/src/main/java/com/google/maps/android/compose/GoogleMap.kt#L214

It's fine to throw because `ON_ANY` is not a real event and it is never sent it is impossible to reach (see https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:lifecycle/lifecycle-common/src/main/java/androidx/lifecycle/DefaultLifecycleObserverAdapter.kt;l=31?q=DefaultLifecycleObserverAdapter ).
So it won't crash. If detekt complains we can change `throw IllegalStateException()` with `error()`.

bmarty: 2da6ceeb5f2b1fad27c2e842cf907da840883f94

julioromano (approved):

PR #986, opened 2023-07-27
Big font size improvements

### LINKED ISSUE #984: Large font size causes layout issues
### Steps to reproduce

Room list and timeline

2. What do you click?

Set the system font size to large

#### What did you expect?

- Room list displays as per design
- Timeline message text input displays as per design

#### What happened instead?

- Room list "All chats" title displays incorrectly: it is only partially visible.

image

- Timeline message text input displays incorrectly: both the plus '+' button and send buttons appear too small relative to the text field

image

---

Could we add a screenshot test for each screen at larger font scale?

### Your phone model

Google Pixel 6

### Operating system version

Android 13

0.1.2-nightly

### Homeserver

_No response_

### Will you send logs?

No

Yes

bmarty:
Fixes #984 and other issue that I have seen when setting a bigger font on the device.


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/text/UnitConverters.kt
```diff
@@ -52,3 +52,19 @@ fun Dp.toPx(): Float = with(LocalDensity.current) { toPx() }
  */
 @Composable
 fun Dp.roundToPx(): Int = with(LocalDensity.current) { roundToPx() }
+
+/**
+ * Return the maximum value between the receiver value and the value with fonScale applied.
```
csmith, line 57:
```suggestion
 * Return the maximum value between the receiver value and the value with fontScale applied.
```

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/text/UnitConverters.kt
```diff
@@ -52,3 +52,19 @@ fun Dp.toPx(): Float = with(LocalDensity.current) { toPx() }
...
+ */
+@Composable
+fun Dp.scaleMax(): Dp = with(LocalDensity.current) {
+    return this@scaleMax * fontScale.coerceAtMost(1f)
+}
+
+/**
+ * Return the minimum value between the receiver value and the value with fonScale applied.
```
csmith, line 65:
```suggestion
 * Return the minimum value between the receiver value and the value with fontScale applied.
```

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/text/UnitConverters.kt
```diff
@@ -52,3 +52,19 @@ fun Dp.toPx(): Float = with(LocalDensity.current) { toPx() }
...
+ */
+@Composable
+fun Dp.scaleMin(): Dp = with(LocalDensity.current) {
+    return this@scaleMin * fontScale.coerceAtLeast(1f)
```
csmith, line 69: I might be missing something, but these feel backwards. If the font scale is 1.5, then `scaleMin` will return `1.5x` but `scaleMax` will return `1.0x`? Feels very weird to have min > max :thinking:

csmith (commented):
I guess the screenshot workflow needs running :)

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/text/UnitConverters.kt
jonnyandrew: I also find the API a bit confusing but I think it is a confusing concept.

How about combining the `scale()` function so that it covers the most common use case with default parameters?

```kotlin
/**
 * Apply the system font scale to a density pixel value.
 * 
 * Specify min or max to limit the scaling applied. By default, the 
 * applied scale is limited to a minimum of 1.
 */
fun Dp.scale(
    min: Float = 1f
    max: Float = Float.MAX_VALUE
): Dp = with(LocalDensity.current) {
    this@scaleMin * fontScale.coerceAtLeast(min).coerceAtMost(max)
}

**Review by jonnyandrew — COMMENTED** [review]
Looks good and thanks for fixing! I have a suggestion about the new scale API

File: libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/text/UnitConverters.kt (hunk 3 above)
Comment (bmarty) [reply libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/text/UnitConverters.kt:69]: Thanks for the suggestion. I can definitely improve this. The "problem" with fun `Dp.scale` - IIUC - is that you can use it like this:

```
16.dp.scale()
```

to ensure that the final size will be minimum `16.dp`, or like this:

```
16.dp.scale(0f, 1f)
```

to ensure that the final size will be maximum `16.dp`, and I do not like this oddity :).

I will try to find something better.

File: libraries/textcomposer/src/main/kotlin/io/element/android/libraries/textcomposer/TextComposer.kt [hunk 4]
```diff
@@ -165,7 +169,7 @@ fun TextComposer(
                             singleLine = false,
                             visualTransformation = VisualTransformation.None,
                             shape = roundedCorners,
-                            contentPadding = PaddingValues(top = 10.dp, bottom = 10.dp, start = 12.dp, end = 42.dp),
+                            contentPadding = PaddingValues(top = 10.dp.applyScaleUp(), bottom = 10.dp.applyScaleUp(), start = 12.dp.applyScaleUp(), end = 42.dp.applyScaleUp()),
```
ElementBot, line 172: - :warning: Line detected, which is longer than the defined maximum line length in the code style.

ElementBot, line 172: - :warning: Line detected, which is longer than the defined maximum line length in the code style.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/text/UnitConverters.kt
bmarty: [5ace169](https://github.com/vector-im/element-x-android/pull/986/commits/5ace169440daf61320a14e7eb3ae7ffa92c34e38), let me know what you think!

jonnyandrew: I am happy with this although (for me) it is a bit verbose for a utility which will likely be littered all over the codebase.

At the risk of bike shedding, one more suggestion:

```kotlin
enum class DpScaleMode {
    Exact, UpOnly, DownOnly
}

fun Dp.scale(
    mode: DpScaleMode = DpScaleMode.UpOnly,
): Dp ...
```

At the end of the day I think it is a style thing, so am happy to go with your suggestion!

jonnyandrew (approved):

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/text/UnitConverters.kt
bmarty: (I do not think we will use those 2 extensions a lot, only on the most important part of the app I guess)

bmarty:
Thanks for the review. I have fixed the issue with the screenshot, and I am recording them again.
Once the CI will be happy, I will push a commit to fix the typo in the comment and merge the PR.

PR #990, opened 2023-07-27
[Compound] Platform components (Lists)

jmartinesp:
## Changes

Creates `ListItem`,  `ListSectionHeader` and `ListSupportingText` composables. These new components will replace our current list items in a new PR, since this one is already large enough.

## Why

Closes https://github.com/vector-im/compound/issues/169


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/ListItem.kt
```diff
@@ -0,0 +1,419 @@
...
+ * @param trailingContent The content to be displayed after the headline content.
+ * @param colors The colors to be used for the list item.
+ * @param tonalElevation The tonal elevation to be used for the list item.
+ * @param shadowElevation The shadow elevation to be used for the list item.
+ * @param isPrimaryAction Whether the list item is a primary action. This will change the color of the leading content to use primary tokens.
+ * @param isDestructive Whether the list item is a destructive action. This will change the color of the headline content to use critical tokens.
+ * @param enabled Whether the list item is enabled. When disabled, will change the color of the headline content and the leading content to use disabled tokens.
+ * @param onClick The callback to be called when the list item is clicked.
+ *
+ * **Note:** [isPrimaryAction], [isDestructive] and [enabled] should only be used in single line [ListItem] components, and only one should be customised at a time.
```
ElementBot, line 59: - :warning: Line detected, which is longer than the defined maximum line length in the code style.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/ListItem.kt
```diff
@@ -0,0 +1,419 @@
...
+    onClick: () -> Unit = { },
+) {
+    val headlineColor = when {
+        isDestructive -> ElementTheme.colors.textCriticalPrimary
+        !enabled -> ElementTheme.colors.textDisabled
+        else -> ElementTheme.colors.textPrimary
+    }
+    val leadingContentColor = when {
+        isPrimaryAction -> ElementTheme.colors.iconPrimary
+        isDestructive -> ElementTheme.colors.iconCriticalPrimary
```
bmarty, line 85: Maybe test `isDestructive` first?

After thinking about it, I understand that parameters `isPrimaryAction` and `isDestructive` are probably exclusive (i.e. they cannot be both equal to `true`), but maybe enforce this by introducing an enum instead? If I am wrong, maybe add a preview with both parameter to `true`?

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/ListItem.kt
```diff
@@ -0,0 +1,419 @@
...
+    supportingContent: @Composable (() -> Unit)? = null,
+    leadingContent: @Composable (() -> Unit)? = null,
+    trailingContent: @Composable (() -> Unit)? = null,
+    colors: ListItemColors = ListItemDefaults.colors(),
+    tonalElevation: Dp = ListItemDefaults.Elevation,
+    shadowElevation: Dp = ListItemDefaults.Elevation,
+    isPrimaryAction: Boolean = false,
+    isDestructive: Boolean = false,
+    enabled: Boolean = true,
+    onClick: () -> Unit = { },
```
bmarty, line 76: There is always a pb with having a default empty lambda, which is when the lambda stays empty, the click effect is still visible (since we invoke `modifier.clickable(enabled = enabled, onClick = onClick)` below). I think we may have `ListItem` with no click action, but if not, maybe remove the default parameter value. WDYT?

bmarty (approved):
LGTM, 2 minor remarks.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/ListItem.kt
```diff
@@ -0,0 +1,419 @@
...
+ * @param trailingContent The content to be displayed after the headline content.
+ * @param colors The colors to be used for the list item.
+ * @param tonalElevation The tonal elevation to be used for the list item.
+ * @param shadowElevation The shadow elevation to be used for the list item.
+ * @param isPrimaryAction Whether the list item is a primary action. This will change the color of the leading content to use primary tokens.
+ * @param isDestructive Whether the list item is a destructive action. This will change the color of the headline content to use critical tokens.
+ * @param enabled Whether the list item is enabled. When disabled, will change the color of the headline content and the leading content to use disabled tokens.
+ * @param onClick The callback to be called when the list item is clicked.
+ *
+ * **Note:** [isPrimaryAction], [isDestructive] and [enabled] should only be used in single line [ListItem] components,
```
julioromano: There is no notion of "single line ListItem" as of yet. So this doc line leaves a lot to the dev's interpretation.

Perhaps we should remove `isPrimaryAction` and `isDestructive` from here, create another composable which only has a single line of content and also an enum param which is something like "StandardAction, PrimaryAction, DesctructiveAction".
This new composable could still be called `ListItem` (because the param list will be a lot different) or it could be called `SingleLineListItem`.

julioromano (commented):
Some API improvements suggestions

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/ListItem.kt
```diff
@@ -0,0 +1,419 @@
...
+ * and only one should be customised at a time.
+ */
+@Suppress("LongParameterList")
+@Composable
+fun ListItem(
+    headlineContent: @Composable () -> Unit,
+    modifier: Modifier = Modifier,
+    overlineContent: @Composable (() -> Unit)? = null,
```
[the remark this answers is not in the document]
jmartinesp: Done in https://github.com/vector-im/element-x-android/pull/990/commits/53ebf11b6a791e315bac375a9f24bb92f22e5b04

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/ListItem.kt
jmartinesp: Actually, this sounds better than what was implemented on https://github.com/vector-im/element-x-android/pull/990/commits/53ebf11b6a791e315bac375a9f24bb92f22e5b04.

I still need to create a couple of PRs to replace usages of `ListItems` through the app and add Lists to dialogs. Do you mind if I add this change in one of this other PRs, or even a dedicated one for it?

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/ListItem.kt
jmartinesp: I implemented a `style` parameter to handle this in an enum way. Thanks!

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/ListItem.kt
jmartinesp: Done in https://github.com/vector-im/element-x-android/pull/990/commits/53ebf11b6a791e315bac375a9f24bb92f22e5b04.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/ListItem.kt
```diff
@@ -117,25 +119,40 @@ fun ListItem(
...
+                LocalTextStyle provides ElementTheme.typography.fontBodyMdRegular,
             ) {
                 content()
             }
         }
     }
+
+    val setUpModifier = modifier
+        .then(if (isCompact) Modifier else Modifier.padding(vertical = 4.dp))
+        .clickable(enabled = onClick != null, onClick = onClick ?: {})
```
bmarty, line 132: Maybe also take into account the `enabled` parameter? I do not know if we want to have a disabled ListItem still clickable. Maybe...

```
.clickable(enabled = enabled && onClick != null, onClick = onClick ?: {})
```

jmartinesp: Yes, you're right. It depends on the implementation, but here `enabled` is up to us to apply, so it should be taken into account.

jmartinesp:
I'll actually remove the `isCompact` parameter since it's not working as expected. Even when applied, the `minHeight` for the `ListItem` will always be `56.dp`, while the Figma designs expects it to be `48.dp`, which is not possible AFAICT.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/ListItem.kt
julioromano: Not a problem, but we can also leave it as is if it's not a problem to use "destructive" or "primary" style when there is also a supporting content.

PR #992, opened 2023-07-27
Feature/fga/update rust sdk 0.1.37

ganfra:
Closes https://github.com/vector-im/element-x-android/pull/968

This should fix https://github.com/vector-im/element-x-android/issues/981

Also as fetchMembers can lock the timeline for some time, only call it after timeline is ready so opening is fast.

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/item/event/TimelineEventContentMapper.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/item/event/TimelineEventContentMapper.kt
@@ -96,6 +96,7 @@ class TimelineEventContentMapper(private val eventMessageMapper: EventMessageMap
                     data = kind.msg.map()
                 )
             }
+            else -> UnknownContent
         }
     }
 }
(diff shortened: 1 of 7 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/item/event/TimelineEventContentMapper.kt
```diff
@@ -96,6 +96,7 @@ class TimelineEventContentMapper(private val eventMessageMapper: EventMessageMap
                     data = kind.msg.map()
                 )
             }
+            else -> UnknownContent
```
jmartinesp, line 99: Maybe we should add the 2 poll contents here explicitly with a TODO/FIXME?

ganfra: I guess @julioromano has already done it locally...

bmarty (approved):
LGTM.
After opening and closing many times the same room, the app is becoming very slow... But I did not have any deadlock, so not blocking the PR for this.

bmarty:
Maestro is failing with:

> java.lang.IllegalStateException: Room object has already been destroyed
07-27 21:09:18.370 E/AndroidRuntime( 6732): 	at org.matrix.rustcomponents.sdk.Room.id(matrix_sdk_ffi.kt:16579)

image

After the leave room dialog is cancelled :/

Not blocking the new SDK to be integrated though.

PR #997, opened 2023-07-28
Feature/fga/remove from main thread

ganfra:
Remove some code from main thread

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
@@ -122,17 +122,12 @@ class MessagesPresenter @AssistedInject constructor(
         }
 
         val inviteProgress = remember { mutableStateOf<Async<Unit>>(Async.Uninitialized) }
-
-        val showReinvitePrompt by remember(
-            hasDismissedInviteDialog,
-            composerState.hasFocus,
-            syncUpdateFlow,
-        ) {
-            derivedStateOf {
-                !hasDismissedInviteDialog && composerState.hasFocus && room.isDirect && room.activeMemberCount == 1L
+        var showReinvitePrompt by remember { mutableStateOf(false) }
+        LaunchedEffect(hasDismissedInviteDialog, composerState.hasFocus, syncUpdateFlow) {
+            withContext(dispatchers.io) {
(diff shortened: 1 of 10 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
```diff
@@ -122,17 +122,12 @@ class MessagesPresenter @AssistedInject constructor(
...
         val inviteProgress = remember { mutableStateOf<Async<Unit>>(Async.Uninitialized) }
-
-        val showReinvitePrompt by remember(
-            hasDismissedInviteDialog,
-            composerState.hasFocus,
-            syncUpdateFlow,
-        ) {
-            derivedStateOf {
-                !hasDismissedInviteDialog && composerState.hasFocus && room.isDirect && room.activeMemberCount == 1L
+        var showReinvitePrompt by remember { mutableStateOf(false) }
```
jmartinesp, line 125: Is this change needed? The check seems really fast to me, and using the `LaunchedEffect` will trigger a recomposition, right? The `room.activeMemberCount` does come from Rust, but the originating code is just a sum of 2 numbers and it's sync AFAIK, so it should be fast.

ganfra: It's hitting the db underneath if I'm not mistaken

jmartinesp: Oh, you're right, it's even async under the hood. I thought we mapped that instead of calling the underlying function each time. This change makes sense then, thanks!

jmartinesp (approved):
LGTM! There is a warning about unused imports.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
ganfra: Ok, I'm mistaken, let's revert this one then.

ganfra: Actually, I think almost all methods are async in the rust layer, but exposed as sync...
Using `val` in our layer is then misleading even more, we should probably use `suspend fun` at some point.

PR #1016, opened 2023-08-01
Fix sending read receipts when entering a room

jmartinesp:
The callback in `TimelineScrollHelper` was only called when the timeline items were empty, so no initial read receipt was sent anymore.

Instead, we're sending a read receipt when the initial items for the timeline are loaded.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
@@ -92,14 +92,14 @@ class TimelinePresenter @Inject constructor(
 
         LaunchedEffect(timelineItems.size) {
             // We just loaded the initial items, we should send a read receipt
-            if (prevMostRecentItemId.value == null && timelineItems.isNotEmpty()) {
-                appScope.sendReadReceiptIfNeeded(
-                    firstVisibleIndex = 0,
-                    timelineItems = timelineItems,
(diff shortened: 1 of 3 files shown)


jmartinesp:
I'm not too happy with this solution, but I can't really think of anything better right now.

jmartinesp:
The implementation at https://github.com/vector-im/element-x-android/pull/1016/commits/ad4bc79030d1d7bada1f501b67360a55127182b8, while simpler, makes it impossible to unit test the presenter properly, so I'll have to revert it...

jmartinesp:
@ganfra In the end I've done a simpler implementation than the initial one, but keeping the same logic. Moving sending the read receipt for the initial items to the presenter made sense, but it made most of the logic associated with `OnScrollFinished` untestable.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
```diff
@@ -91,6 +91,16 @@ class TimelinePresenter @Inject constructor(
         }
 
         LaunchedEffect(timelineItems.size) {
+            // We just loaded the initial items, we should send a read receipt
```
ganfra, line 94: Remove the code?

ganfra (approved):
One remark, otherwise makes sense.
But I also encounter another issue, maybe it should handled by the sdk, but if the latest events are hidden (like reactions or edit events), the has unread won't be updated properly neither.

PR #1020, opened 2023-08-02
Reactions ux updates

langleyd:
- Fix ordering of reaction count/key label on outgoing messages
- Enforce all reaction buttons have the same height in reaction layout
- Fix emoji circles on action list
- Fix shape of reaction summary button when pressed


langleyd:
Yea I think performance should be ok, it’s all linear computation. We could make the the height static but that would be at the expense of not supporting dynamic font sizes. Or we could probably make it marginally more performant if we used less functional code but at the expense of readability. Happy to take another look if there are any ideas or we find it’s not performant though.

PR #1022, opened 2023-08-02
Feature/fga/matrix client versions

ganfra:
Use the same "hack" than iOS to speed up startup.
https://github.com/vector-im/element-x-ios/blob/develop/ElementX/Sources/Services/UserSession/UserSessionStore.swift#L118


jmartinesp (approved):
LGTM!

PR #1029, opened 2023-08-07
Sync Strings

ElementBot:


julioromano:
There is a string with an error. I've fixed it in localazy but now we have to wait ofr this action to rerun.

PR #1036, opened 2023-08-08
Feature/fga/rust sdk tracing

ganfra:
Use the new rust api for tracing : 
- Provides a `Timber.Tree` delegating to `log_event` 
- Use the new `TracingConfiguration` so we can directly log to files from the rust
- Clean files older than 24hours, waiting for a proper rotation from the rust

Tested on a local version of the rust sdk, waiting for https://github.com/matrix-org/matrix-rust-sdk/pull/2384


app/src/main/kotlin/io/element/android/x/initializer/TracingInitializer.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    override fun create(context: Context) {
+        val appBindings = context.bindings<AppBindings>()
+        val tracingService = appBindings.tracingService()
+        val bugReporter = appBindings.bugReporter()
+        Timber.plant(tracingService.createTracingTree())
+        val tracingConfiguration = if (BuildConfig.DEBUG) {
+            TracingConfiguration(
+                filterConfiguration = TracingFilterConfigurations.debug,
+                writesToStdout = false,
```
jmartinesp, line 39: Why is this `false` in both cases? I expected it to be true so the lines are displayed in the logcat output and not just written to the log files.

jmartinesp: I guess it's a special case for iOS? I remember they mentioned something about using `stderr` and it seems like setting this to `true` pipes the output to `stderr`.

ganfra: Maybe ? We don't need it on our side. What do you think about not logging to logcat in release ?

jmartinesp: As long as we have those log messages written to our log files I guess it's ok.

jmartinesp (approved):
LGTM!

PR #1041, opened 2023-08-09
Enable filter push notifications by push rules

### LINKED ISSUE #640: Silent unexpected notifications
Make sure we do not display notifications for pushed events that do not match a push rule.

User story: https://github.com/vector-im/element-meta/issues/1794.

jmartinesp:
Closes #640 .

### DIFF
--- a/changelog.d/640.bugfix
+++ b/changelog.d/640.bugfix
@@ -1,0 +1,1 @@
+Filter push notifications using push rules.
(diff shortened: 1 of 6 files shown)


ganfra (approved):
Makes sense, thanks!
Also I think it closes https://github.com/vector-im/element-x-android/issues/1027

PR #1042, opened 2023-08-09
Feature/fga/sync states

ganfra:
Rework a bit how we manage sync : 
- Replace `RoomSummaryDataSource` with `RoomListService` to better align with the sdk api.
- It allows to have a clearer diff between `SyncService.State`, `RoomListService.State` and `RoomList.LoadedState`.
- Always relaunch sync if SyncService.State is not Running.
- Show Sync spinner only after 1500ms in the same state and only if NetworkState is Connected.


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentMessageFactory.kt
```diff
@@ -132,7 +132,7 @@ class TimelineItemContentMessageFactory @Inject constructor(
     }
 
     private fun aspectRatioOf(width: Long?, height: Long?): Float? {
-        return if (height != null && width != null) {
+        return if (height != null && width != null && height > 0 && width > 0) {
```
ElementBot, line 135: - :warning: This condition is too complex (4). Defined complexity threshold for conditions is set to '4'

ElementBot, line 135: - :warning: This condition is too complex (4). Defined complexity threshold for conditions is set to '4'

ganfra: Should be part of another PR, so I removed it.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/roomlist/RoomList.kt
```diff
@@ -23,25 +23,34 @@ import kotlinx.coroutines.withTimeout
 import timber.log.Timber
 import kotlin.time.Duration
 
-interface RoomSummaryDataSource {
-
+/**
+ * Holds some flows related to a specific set of rooms.
+ * Can be retrieved from [RoomListService] methods.
+ */
+interface RoomList {
```
jmartinesp, line 30: `RoomListState`? Otherwise it looks like something we can interact with, instead of just holding state.

appnav/src/main/kotlin/io/element/android/appnav/loggedin/LoggedInPresenter.kt
```diff
@@ -53,18 +64,27 @@ class LoggedInPresenter @Inject constructor(
...
-        // fun handleEvents(event: LoggedInEvents) {
-        //     when (event) {
-        //     }
-        // }
-
+        var showSyncSpinner by remember {
+            mutableStateOf(false)
+        }
+        LaunchedEffect(roomListState, networkStatus) {
+            showSyncSpinner = if (networkStatus == NetworkStatus.Offline) {
```
jmartinesp, line 74: Maybe a `when` would be more readable here?

jmartinesp (approved):
LGTM, just a couple of suggestions.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/roomlist/RoomList.kt
ganfra: Mmmm, we already have a State inside, so might be a bit weird...

PR #1043, opened 2023-08-09
[Compound] Implement dialogs

jmartinesp:
Changes:

- Makes `TextButton` padding smaller when it doesn't have an icon.
- Move `AlertDialogContent` to `theme/components` to make it part of the 'platform components' instead of having it as a custom component.
- Update designs to match Figma.

Closes https://github.com/vector-im/compound/issues/202.

**Note:** this PR only covers simple dialogs, and doesn't try to implement dialogs containing lists since the work on lists is paused. Also, this PR does not handle 'disabled' state for buttons since I couldn't verify the behavior needed, so this will be added in a separate PR in the future.


julioromano (approved):
LGTM!

PR #1049, opened 2023-08-10
[Compound] implement IconButton changes

jmartinesp:
## Changes:

- Ensure we use `LocalContentColor` for tinting.
- Use Compound's disabled color token.
- Update previews to match Figma designs.


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/IconButton.kt
```diff
@@ -36,11 +42,15 @@ fun IconButton(
     interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
     content: @Composable () -> Unit
 ) {
+    val colors = IconButtonDefaults.iconButtonColors(
+        contentColor = LocalContentColor.current,
```
ganfra, line 46: Why are we using `LocalContentColor` here and `ElementTheme.colors` for disabled?

ganfra (approved):
One question, otherwise LGTM

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/IconButton.kt
jmartinesp: The way it was designed in [Figma](https://www.figma.com/file/G1xy0HDZKJf5TCRFmKb5d5/Compound-Android-Components?type=design&node-id=1182%3A48861&mode=design&t=bMycIvAOH66Apkuv-1), the icon button could use either primary or secondary colors, so it made more sense to just allow the parent component to decide which color should be used for the icon.

As for disabled, it's always the same color for icons, so we can just hardcode it.

ganfra: Ok makes sense!

PR #1053, opened 2023-08-11
Sync: enable the encryption loop

ganfra:
Just enable the encryption loop.
Maybe wait for the next rust sdk (0.1.42) before merging


jmartinesp (approved):
LGTM. I also tested it for some time and I couldn't find any issues with it.

PR #1058, opened 2023-08-14
Media upload cancellation

### LINKED ISSUE #769: Cancel upload media
We should be able to cancel sending media.
We have the API for this from the rust-sdk, but we need to branch it with coroutine cancellation mechanism.

jmartinesp:
## Changes

- Adds `MediaUploadHandler` as an abstraction over `SendAttachmentJoinHandler`.
- Adds a cancel option to the send media dialogs.
- Ties media upload to the coroutine that runs the code, so if the coroutine is cancelled, so is the upload.
- Makes sure we remove any temporary files from cache both when either finish uploading or it fails.

## Why

Closes #769 .

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
@@ -16,7 +16,6 @@
 
 package io.element.android.features.messages.impl.attachments.preview
 
-import android.net.Uri
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.mutableStateOf
@@ -31,7 +30,10 @@ import io.element.android.libraries.matrix.api.core.ProgressCallback
 import io.element.android.libraries.mediaupload.api.MediaSender
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
+import kotlin.coroutines.coroutineContext
 
 class AttachmentsPreviewPresenter @AssistedInject constructor(
     @Assisted private val attachment: Attachment,
@@ -52,13 +54,16 @@ class AttachmentsPreviewPresenter @AssistedInject constructor(
             mutableStateOf<SendActionState>(SendActionState.Idle)
         }
 
-        val ongoingSendAttachment = remember { mutableStateOf<Uri?>(null) }
+        val ongoingSendAttachmentJob = remember { mutableStateOf<Job?>(null) }
 
         fun handleEvents(attachmentsPreviewEvents: AttachmentsPreviewEvents) {
             when (attachmentsPreviewEvents) {
-                AttachmentsPreviewEvents.SendAttachment -> coroutineScope.sendAttachment(attachment, sendActionState, ongoingSendAttachment)
+                AttachmentsPreviewEvents.SendAttachment -> ongoingSendAttachmentJob.value = coroutineScope.sendAttachment(attachment, sendActionState)
                 AttachmentsPreviewEvents.ClearSendState -> {
-                    ongoingSendAttachment.value?.let { mediaSender.cancel(it) }
+                    ongoingSendAttachmentJob.value?.let {
+                        it.cancel()
+                        ongoingSendAttachmentJob.value = null
+                    }
                     sendActionState.value = SendActionState.Idle
                 }
             }
@@ -74,14 +79,12 @@ class AttachmentsPreviewPresenter @AssistedInject constructor(
     private fun CoroutineScope.sendAttachment(
         attachment: Attachment,
         sendActionState: MutableState<SendActionState>,
-        ongoingSendAttachment: MutableState<Uri?>,
     ) = launch {
         when (attachment) {
             is Attachment.Media -> {
                 sendMedia(
                     mediaAttachment = attachment,
                     sendActionState = sendActionState,
-                    ongoingSendAttachment = ongoingSendAttachment,
                 )
             }
         }
@@ -90,12 +93,13 @@ class AttachmentsPreviewPresenter @AssistedInject constructor(
     private suspend fun sendMedia(
         mediaAttachment: Attachment.Media,
         sendActionState: MutableState<SendActionState>,
-        ongoingSendAttachment: MutableState<Uri?>,
     ) = runCatching {
-        ongoingSendAttachment.value = mediaAttachment.localMedia.uri
+        val context = coroutineContext
         val progressCallback = object : ProgressCallback {
             override fun onProgress(current: Long, total: Long) {
-                sendActionState.value = SendActionState.Sending.Uploading(current.toFloat() / total.toFloat())
+                if (context.isActive) {
+                    sendActionState.value = SendActionState.Sending.Uploading(current.toFloat() / total.toFloat())
+                }
             }
         }
         sendActionState.value = SendActionState.Sending.Processing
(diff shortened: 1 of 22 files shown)


tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.roomdetails.impl.members_null_DefaultGroup_RoomMemberListDarkPreview_0_null_0,NEXUS_5,1.0,en].png
Comment (jmartinesp) [inline tests/uitests/src/test/snapshots/images/io.element.android.tests.uitests_ScreenshotTest_preview_tests[io.element.android.features.roomdetails.impl.members_null_DefaultGroup_RoomMemberListDarkPreview_0_null_0,NEXUS_5,1.0,en].png:1]: I actually have no idea why this screenshot was updated 🤔 .

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerEvents.kt
```diff
@@ -16,6 +16,7 @@
 
 package io.element.android.features.messages.impl.messagecomposer
 
+import android.net.Uri
```
ElementBot, line 19: - :warning: The import 'android.net.Uri' is unused.

features/messages/impl/src/test/kotlin/io/element/android/features/messages/attachments/AttachmentsPreviewPresenterTest.kt
```diff
@@ -29,6 +29,8 @@ import io.element.android.features.messages.impl.attachments.preview.Attachments
 import io.element.android.features.messages.impl.attachments.preview.AttachmentsPreviewPresenter
 import io.element.android.features.messages.impl.attachments.preview.SendActionState
 import io.element.android.features.messages.impl.media.local.LocalMedia
+import io.element.android.features.messages.impl.messagecomposer.AttachmentsState
```
ElementBot, line 32: - :warning: The import 'io.element.android.features.messages.impl.messagecomposer.AttachmentsState' is unused.

features/messages/impl/src/test/kotlin/io/element/android/features/messages/attachments/AttachmentsPreviewPresenterTest.kt
```diff
@@ -29,6 +29,8 @@ import io.element.android.features.messages.impl.attachments.preview.Attachments
...
+import io.element.android.features.messages.impl.messagecomposer.MessageComposerEvents
```
ElementBot, line 33: - :warning: The import 'io.element.android.features.messages.impl.messagecomposer.MessageComposerEvents' is unused.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
```diff
@@ -93,13 +108,17 @@ class AttachmentsPreviewPresenter @AssistedInject constructor(
             mimeType = mediaAttachment.localMedia.info.mimeType,
             compressIfPossible = mediaAttachment.compressIfPossible,
             progressCallback = progressCallback
-        ).fold(
-            onSuccess = {
-                sendActionState.value = SendActionState.Done
-            },
-            onFailure = {
-                sendActionState.value = SendActionState.Failure(it)
+        ).getOrThrow()
```
bmarty, line 111: I am not sure. Is this `.getOrThrow()` necessary?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
```diff
@@ -212,13 +228,13 @@ class MessageComposerPresenter @Inject constructor(
     private fun CoroutineScope.sendAttachment(
         attachment: Attachment,
         attachmentState: MutableState<AttachmentsState>,
-    ) = launch {
-        when (attachment) {
-            is Attachment.Media -> {
+    ) = when (attachment) {
```
bmarty, line 231: There is some code duplication in the class `AttachmentsPreviewPresenter`, is there a (easy) way to reduce this?

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/diff/DiffCacheUpdater.kt
```diff
@@ -58,8 +58,7 @@ class DiffCacheUpdater<ListItem, CachedItem>(
         }
     }
 
-    fun updateWith(newOriginalList: List<ListItem>) = synchronized(lock) {
-        val timeToDiff = measureTimeMillis {
+    fun updateWith(newOriginalList: List<ListItem>) = synchronized(lock) { val timeToDiff = measureTimeMillis {
```
bmarty, line 61: I think this is an unexpected change.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/media/MediaUploadHandlerImpl.kt
```diff
@@ -0,0 +1,43 @@
...
+class MediaUploadHandlerImpl(
+    private val filesToUpload: List<File>,
+    private val sendAttachmentJoinHandle: SendAttachmentJoinHandle,
+) : MediaUploadHandler {
+    override suspend fun await(): Result<Unit> =
+        runCatching {
+            sendAttachmentJoinHandle.join()
+            cleanUpFiles()
+        }
+        .onFailure { cleanUpFiles() }
```
bmarty, line 33: I would avoid include the `cleanUpFiles()` call in the `runCatching`block.

I think it's better to write:

```kotlin
        runCatching {
            sendAttachmentJoinHandle.join()
        }
        .also { cleanUpFiles() }
```

WDYT?

libraries/mediaupload/api/src/test/kotlin/io/element/android/libraries/mediaupload/api/MediaSenderTests.kt
```diff
@@ -0,0 +1,116 @@
...
+        val sender = aMediaSender(preProcessor)
+
+        val uri = Uri.parse("content://image.jpg")
+        sender.sendMedia(uri = uri, mimeType = "image/jpeg", compressIfPossible = true)
+
+        assertThat(preProcessor.processCallCount).isEqualTo(1)
+    }
+
+    @Test
+    fun `given an attachment when sending it the MatrxiRoom will call sendMedia`() = runTest {
```
bmarty, line 48:
```suggestion
    fun `given an attachment when sending it the MatrixRoom will call sendMedia`() = runTest {
```

bmarty (commented):
Nice work.
I tried to send an image and cancel the upload. I was still on the image preview, like if the image had not been sent. But the image was sent in the timeline.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
jmartinesp: We need it to catch the `CancellationException` that won't come inside the `Result`. I tried several ways to get it wrapped inside this Result, but none of them worked.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
jmartinesp: The 2 functions are quite similar, but they pass different data to their `sendMedia` counterparts. We could try wrapping those in  some use case or helper class, but I'm not sure it's worth it for just 2 occurrences. Also, I think we want to remove the upload progress from the `AttachmentsPreviewView` in the future as display it on the the timeline item instead, so part of this code will probably disappear.

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/diff/DiffCacheUpdater.kt
jmartinesp: Oops, nice catch!

bmarty (approved):
Thanks for the update and for your answers!

PR #1061, opened 2023-08-16
Add workaround for blur in `ElementLogoAtom` for OS < 9.

jmartinesp:
## Changes

- Creates `canUseBlurModifier()` and `canUseBlurMaskFilter()` functions to check whether the blur modifier and filters will work or not.
- Modifies `ElementLogoAtom` to use a compat blur version, in case we make it work in the future in older versions, and also create its shadow based on the result of `canUseBlurMaskFilter()`.

## Why

While the blurred shadow worked well in most devices, there are some where it `BlurMaskFilter` wasn't being applied and the result was terrible:

<table>
<tr>
 <td>Before (mask filter disabled)</td><td>After (with workaround)</td>
<tr>
 <td>

 </td>
<td>
image
</td>
</table>


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/modifiers/Blur.kt
```diff
@@ -0,0 +1,103 @@
...
+                )
+            }
+        }
+    }
+)
+
+fun Modifier.blurCompat(
+    radius: Dp,
+    edgeTreatment: BlurredEdgeTreatment = BlurredEdgeTreatment.Rectangle
+): Modifier = composed {
```
ElementBot, line 97: - :warning: Unnecessary use of Modifier.composed

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/atoms/ElementLogoAtom.kt
```diff
@@ -66,22 +57,35 @@ fun ElementLogoAtom(
...
-            Modifier
-                .size(size.outerSize)
-                .shapeShadow(
-                    color = shadowColor,
-                    cornerRadius = size.cornerRadius,
-                    blurRadius = size.shadowRadius,
-                    offsetY = 8.dp,
-                )
-        )
+        if (canUseBlurMaskFilter()) {
```
bmarty, line 60: Maybe pass this as a parameter, defaulting `canUseBlurMaskFilter()` to be able to preview the composable when this value is false.

bmarty (approved):
LGTM. It's a shame that this change update the screenshot test, maybe because we are using default `NEXUS_5`?

PR #1064, opened 2023-08-16
Add feature flag for polls

Florian14:
Add a feature flag in the developer settings for Polls feature

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentFactory.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentFactory.kt
@@ -18,6 +18,8 @@ package io.element.android.features.messages.impl.timeline.factories.event
 
 import io.element.android.features.messages.impl.timeline.model.event.TimelineItemEventContent
 import io.element.android.features.messages.impl.timeline.model.event.TimelineItemUnknownContent
+import io.element.android.libraries.featureflag.api.FeatureFlagService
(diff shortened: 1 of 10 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentFactory.kt
```diff
@@ -57,8 +60,10 @@ class TimelineItemContentFactory @Inject constructor(
             is RoomMembershipContent -> roomMembershipFactory.create(eventTimelineItem)
             is StateContent -> stateFactory.create(eventTimelineItem)
             is StickerContent -> stickerFactory.create(itemContent)
-            is PollContent -> pollFactory.create(itemContent)
-            is PollEndContent -> pollEndFactory.create(itemContent)
+            is PollContent ->
+                if (featureFlagService.isFeatureEnabled(FeatureFlags.Polls)) pollFactory.create(itemContent) else TimelineItemUnknownContent
```
ganfra, line 64: I'd have put the check in the PollFactory

ganfra (approved):
One remark otherwise LGTM

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentFactory.kt
Florian14: I was not sure but yeah, let's keep this factory lighter

PR #1068, opened 2023-08-17
Fix snackbar padding

jmartinesp:
It turns out `Snackbar(snackbarData = ...)` adds some default padding that `Snackbar(action, dismissAction, ...)` doesn't, and the resulting Snackbar was touching the screen borders.


bmarty (approved):
OK, thanks!

PR #1069, opened 2023-08-17
Feature/bma/fix maestro

### LINKED ISSUE #1066: element.io now supports slidinc sync, so maestro test is failing.
We should use another server to test this user story.

https://github.com/vector-im/element-x-android/blob/develop/.maestro/tests/account/changeServer.yaml#L12

bmarty:
Fix #1066 

Locally, maestro test is failing when leaving a room. Let's see how it's behave on maestro cloud. Action triggered manually: https://github.com/vector-im/element-x-android/actions/runs/5892168780

This PR also contains some minor improvement in the maestro tests.

### DIFF
--- a/.maestro/allTests.yaml
+++ b/.maestro/allTests.yaml
@@ -1,5 +1,7 @@
 appId: ${APP_ID}
 ---
+## Check that all env variables required in the whole test suite are declared (to fail faster)
+- runScript: ./scripts/checkEnv.js
 - runFlow: tests/init.yaml
 - runFlow: tests/account/login.yaml
 - runFlow: tests/settings/settings.yaml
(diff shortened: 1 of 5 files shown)


jmartinesp:
It seems like [it failed](https://github.com/vector-im/element-x-android/actions/runs/5892168780/job/15980828010) ☹️ .

bmarty:
Yes, this is the issue about the leave room breaking the sync. I think it does not prevent this PR from being merged, at least the test flow is going further...

jmartinesp (commented):
In that case, LGTM. Let's hope there is a fix in the Rust SDK for the leave room issue soon 🤞

jmartinesp (approved):
I guess I pressed the wrong button before because it only added a comment. LGTM (again).

PR #1072, opened 2023-08-17
some reformating 

mittwerk:


bmarty:
Hello @mittwerk , this PR cannot be reviewed, it's too big. Also you are formatting some files which are coming from external tools, like translations from `Localazy` for instance.
If you want to help, maybe create a new PR with formatting of Kotlin code for instance?
Thanks!

PR #1087, opened 2023-08-18
org.gradle.jvmargs=-Xmx4g

julioromano:
I have to do it locally every time i checkout a branch. Perhaps it's best if we push this to origin.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -22,7 +22,7 @@
 # http://www.gradle.org/docs/current/userguide/build_environment.html
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8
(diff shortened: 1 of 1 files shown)


jmartinesp:
I think it's a good idea to bump the default value since it's too low for some tasks now (lint, tests sometimes), but you could just add this line to your `~/.gradle/gradle.properties` and it will override the default one in the project too if I'm not mistaken.

julioromano:
> I think it's a good idea to bump the default value since it's too low for some tasks now (lint, tests sometimes), but you could just add this line to your `~/.gradle/gradle.properties` and it will override the default one in the project too if I'm not mistaken.

Is there any downside in bumping the one in origin too? The tests ran pretty fast in this PR.

jmartinesp:
> Is there any downside in bumping the one in origin too? The tests ran pretty fast in this PR.

No, I'm ok with increasing the max heap size. Let's see it @ganfra agrees too.

ganfra (approved):
Ok for me!

PR #1088, opened 2023-08-18
Fix TextButtons being displayed in black

### LINKED ISSUE #1077: In dark theme, change-server and opt-out-of-analytics buttons are black-on-black
### Steps to reproduce

1. Launch app on dark theme
2. Fail to find button to change server
3. Realise it's black on black

#### What did you expect?

Legible buttons

#### What happened instead?

https://www.google.co.uk/books/edition/The_Restaurant_at_the_End_of_the_Univers/r--6ujORIVgC?hl=en&gbpv=1&pg=PA121

### Your phone model

_No response_

### Operating system version

_No response_

0.1.2 (40001020) playstore

### Homeserver

_No response_

### Will you send logs?

No

No

jmartinesp:
Set the default `LocalContentColor` to `onSurface`, since it's `Color.Black` by default.

Also modified `WaitListView` to use the right colors, since it's a special case where theme is hardcoded.

Other possible solutions are:
- Set it to `Color.Unspecified` and let the components use a fallback color.
- Ensure there is always a `Surface` component providing this value at the root of each screen.

Fixes #1077.


tests/uitests/src/test/snapshots/images/ui_S_t[f.login.impl.screens.waitlistscreen_null_WaitListViewDark_0_null_4,NEXUS_5,1.0,en].png
Comment (bmarty) [inline tests/uitests/src/test/snapshots/images/ui_S_t[f.login.impl.screens.waitlistscreen_null_WaitListViewDark_0_null_4,NEXUS_5,1.0,en].png:1]: The text is "less" white than before, I do not think this is OK.

image

libraries/theme/src/main/kotlin/io/element/android/libraries/theme/ElementTheme.kt
```diff
@@ -115,6 +116,7 @@ fun ElementTheme(
     }
     CompositionLocalProvider(
         LocalCompoundColors provides currentCompoundColor,
+        LocalContentColor provides colorScheme.onSurface,
```
bmarty, line 119: Should it be `colorScheme.primary` instead?

bmarty (commented):
Thanks, a remark.

tests/uitests/src/test/snapshots/images/ui_S_t[f.login.impl.screens.waitlistscreen_null_WaitListViewDark_0_null_4,NEXUS_5,1.0,en].png
Comment (jmartinesp) [reply tests/uitests/src/test/snapshots/images/ui_S_t[f.login.impl.screens.waitlistscreen_null_WaitListViewDark_0_null_4,NEXUS_5,1.0,en].png:1]: It's actually matching the design in Figma (if I'm looking at the right one). The title uses `textPrimary` and the subtitle uses pure white for some reason.

libraries/theme/src/main/kotlin/io/element/android/libraries/theme/ElementTheme.kt
jmartinesp: Both semantic tokens point to the same color token, and I think `onSurface` matches the default behavior better, as this `LocalContentColor` is usually provided by a parent `Surface` component, but I can change it if needed.

bmarty:
(just to make this public: subtitle as to use the primary color as well. Figma has been updated)

jmartinesp:
> (just to make this public: subtitle as to use the primary color as well. Figma has been updated)

This should be fixed now.

bmarty (approved):
Thanks!

PR #1091, opened 2023-08-18
Fix rendering of inline elements in list items in messages

### LINKED ISSUE #1090: Weird formatting in bullet lists with long points
### Steps to reproduce

Go to this message in the current Element X Play store build: https://matrix.to/#/!QQpfJfZvqxbCfeDgCj:matrix.org/$tVPoJHsPTfA8vjv4xm07y5lPrgwCU7UqOSXLdVsH02s?via=seymour.family&via=matrix.org&via=element.io

#### What did you expect?
Bullet lists to be properly formatted

#### What happened instead?
Bullet lists are cut off before the end of each long point

### Your phone model

Pixel 7

### Operating system version

Android 13

EAX Play Store build 0.1.2

### Homeserver

Synapse 1.90 and SS proxy 0.99.5

### Will you send logs?

Yes

No

jmartinesp:
When inline elements such as links were used inside list items, they were being added as block elements, and sometimes not   even added at all, only as a line break with the list item marker.

Fixes #1090.

### DIFF
--- a/changelog.d/1090.bugfix
+++ b/changelog.d/1090.bugfix
@@ -1,0 +1,1 @@
+Fix rendering of inline elements in list items.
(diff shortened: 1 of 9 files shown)


bmarty (approved):
LGTM, thanks!

PR #1092, opened 2023-08-18
Fix excludes so screenshot tests are only run once

jmartinesp:
This is reproducible locally and also in the CI (some PRs are not passing because of OOM errors).


bmarty (approved):
Oh thanks, I forgot that in #844.

PR #1102, opened 2023-08-21
Linkify raw links in HTML message contents

### LINKED ISSUE #1079: URLs in formatted text messages aren't linkified
### Steps to reproduce

1. Receive a text message with a plain URL in it - e.g.

Screenshot 2023-08-17 at 22 48 33

For whatever reason, the sender's client didn't explicitly hyperlink it in the HTML:

```json
{
  "content": {
    "body": "https://github.com/vector-im/element-x-android/pull/1055#issuecomment-1675033424",
    "format": "org.matrix.custom.html",
    "formatted_body": "https://github.com/vector-im/element-x-android/pull/1055#issuecomment-1675033424",
    "msgtype": "m.text"
  },
  "origin_server_ts": 1692006809060,
  "room_id": "!aCvOWKNDeXXpPgquKq:matrix.org",
  "sender": "@onestacked:onestacked.duckdns.org",
  "type": "m.room.message",
  "unsigned": {},
  "event_id": "$0_vVJyrFYvbYQww67TO7fhjfenSwXcdholTtyXGYmow",
  "user_id": "@onestacked:onestacked.duckdns.org"
}
```

#### What did you expect?

Things which look like URLs should be hyperlinked by the receiving client automatically, otherwise it's just frustrating for the user who wonders why they can't click on them.

#### What happened instead?

The url isn't linkified.

### Your phone model

S10

### Operating system version

12

0.1.2 (40001020) playstore

### Homeserver

_No response_

### Will you send logs?

No

No

jmartinesp:
## Changes

- Creates a `ClickableLinkText` for `String`.
- Adds a `linkify` parameter to the original function, which is `true` by default.
- Does the linkify logic inside that component, if `linkify` is true.
- Makes sure we don't linkify urls that are already inside url annotations or domains that are parts of user mentions or room aliases.

## Why

Fixes #1079.

## Screenshots

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>
 <td>
</table>

### DIFF
--- a/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/ClickableLinkText.kt
+++ b/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/ClickableLinkText.kt
@@ -61,7 +61,7 @@ fun ClickableLinkText(
     inlineContent: ImmutableMap<String, InlineTextContent> = persistentMapOf(),
 ) {
     ClickableLinkText(
-        text = AnnotatedString(text),
+        annotatedString = AnnotatedString(text),
         interactionSource = interactionSource,
         modifier = modifier,
         linkify = linkify,
@@ -76,7 +76,7 @@ fun ClickableLinkText(
 @OptIn(ExperimentalTextApi::class)
 @Composable
 fun ClickableLinkText(
-    text: AnnotatedString,
+    annotatedString: AnnotatedString,
     interactionSource: MutableInteractionSource,
     modifier: Modifier = Modifier,
     linkify: Boolean = true,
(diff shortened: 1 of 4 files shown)


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/ClickableLinkText.kt
```diff
@@ -32,27 +35,62 @@ import androidx.compose.ui.platform.LocalUriHandler
...
     interactionSource: MutableInteractionSource,
     modifier: Modifier = Modifier,
+    linkify: Boolean = true,
     linkAnnotationTag: String = "",
     onClick: () -> Unit = {},
     onLongClick: () -> Unit = {},
     style: TextStyle = LocalTextStyle.current,
     inlineContent: ImmutableMap<String, InlineTextContent> = persistentMapOf(),
 ) {
+    val processedText = if (linkify) {
```
ganfra, line 89: remember?

bmarty (approved):
LGTM, thanks!

PR #1108, opened 2023-08-21
Fix memory leak on room members

ganfra:
We are using parallelMap to speed up a bit the processing, but this can be cancelled so all members are not properly destroyed.


bmarty (approved):
Good catch, thanks!

PR #1112, opened 2023-08-22
Fix button theme

bmarty:
Fix a small mistake on padding on Text Button. The most visible one is the missing of padding between the Progress and the text, but there is also a minor change on the content padding, which is less visible on the screenshot.

Should be aligned with https://www.figma.com/file/G1xy0HDZKJf5TCRFmKb5d5/Compound-Android-Components?node-id=369%3A63375&mode=dev now.

### DIFF
--- a/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/Button.kt
+++ b/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/Button.kt
@@ -136,29 +136,33 @@ internal fun ButtonInternal(
     val contentPadding = when (size) {
         ButtonSize.Medium -> {
             when (style) {
+                ButtonStyle.Filled,
+                ButtonStyle.Outlined -> PaddingValues(horizontal = 16.dp, vertical = 10.dp)
                 ButtonStyle.Text -> PaddingValues(horizontal = 12.dp, vertical = 10.dp)
-                else -> PaddingValues(horizontal = 16.dp, vertical = 10.dp)
             }
         }
         ButtonSize.Large -> {
             when (style) {
+                ButtonStyle.Filled,
+                ButtonStyle.Outlined -> PaddingValues(horizontal = 24.dp, vertical = 13.dp)
                 ButtonStyle.Text -> PaddingValues(horizontal = 16.dp, vertical = 13.dp)
-                else -> PaddingValues(horizontal = 24.dp, vertical = 13.dp)
             }
         }
     }
 
     val shape = when (style) {
-        ButtonStyle.Filled, ButtonStyle.Outlined -> RoundedCornerShape(percent = 50)
+        ButtonStyle.Filled,
+        ButtonStyle.Outlined -> RoundedCornerShape(percent = 50)
         ButtonStyle.Text -> RectangleShape
     }
 
     val border = when (style) {
-        ButtonStyle.Filled, ButtonStyle.Text -> null
+        ButtonStyle.Filled -> null
         ButtonStyle.Outlined -> BorderStroke(
             width = 1.dp,
             color = ElementTheme.colors.borderInteractiveSecondary
         )
+        ButtonStyle.Text -> null
     }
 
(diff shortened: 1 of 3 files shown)


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/Button.kt
```diff
@@ -133,44 +133,59 @@ internal fun ButtonInternal(
...
+        ButtonSize.Medium -> when (style) {
+            ButtonStyle.Filled,
+            ButtonStyle.Outlined -> if (hasNoStartDrawable)
+                Triple(24.dp, 0.dp, 24.dp)
+            else
+                Triple(16.dp, 8.dp, 24.dp)
+            ButtonStyle.Text -> if (hasNoStartDrawable)
+                Triple(12.dp, 0.dp, 12.dp)
+            else
+                Triple(12.dp, 8.dp, 16.dp)
```
bmarty, line 148: 16 instead of 12 for end padding

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/Button.kt
```diff
@@ -133,44 +133,59 @@ internal fun ButtonInternal(
...
 
-    val contentPadding = when (size) {
-        ButtonSize.Medium -> {
-            when (style) {
-                ButtonStyle.Text -> PaddingValues(horizontal = 12.dp, vertical = 10.dp)
-                else -> PaddingValues(horizontal = 16.dp, vertical = 10.dp)
-            }
+    val hasNoStartDrawable = !showProgress && leadingIcon == null
+
+    val (paddingStart, paddingCenter, paddingEnd) = when (size) {
```
jmartinesp, line 138: `paddingCenter` is always 0 with no start drawable and 8 with one, couldn't we extract that from the `when` to simplify it? That way you could just return a `PaddingValues` with start and end values.

jmartinesp (commented):
Thanks for the changes! I never realised the progress had been misplaced, nice catch.

I have a couple of suggestions to try to simplify the code.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/Button.kt
```diff
@@ -133,44 +133,59 @@ internal fun ButtonInternal(
         ButtonSize.Large -> 48.dp
     }
 
-    val contentPadding = when (size) {
-        ButtonSize.Medium -> {
-            when (style) {
-                ButtonStyle.Text -> PaddingValues(horizontal = 12.dp, vertical = 10.dp)
-                else -> PaddingValues(horizontal = 16.dp, vertical = 10.dp)
-            }
+    val hasNoStartDrawable = !showProgress && leadingIcon == null
```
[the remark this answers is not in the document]
bmarty: I was expected this remark :). I choose this name to have the code ordered as in Figma.

|Code|Figma|
|-|-|
|image|image|

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/Button.kt
jmartinesp: If you look at them side by side it is easier to understand and follow the code, but to someone who just came to the code it'll probably look a lot more convoluted (at least that was my case). Also, if the structure of the designs change in the future (something added, something moved) we'd have to change the code again to match it.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/Button.kt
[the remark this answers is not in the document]
jmartinesp: Same as below: I know it was easier to find and fix the issue this way, but in my opinion once it's done we should try to simplify the code. I think it should be enough to try to make the previews match the designs.

PR #1031, opened 2023-08-07
"View only" polls in the timeline

Florian14:
This PR renders the polls created with other clients.

**Notes**
- The design isn't final yet
- The poll component isn't interactive (votes cannot be submitted)
- Actions on the poll events need to be implemented/refined
- ⚠️ More work is required in the `matrix-rust-sdk` to be able to see the polls in the timeline. The screenshots are real, but they have been taken while using a custom development branch for the matrix-rust-sdk
- The timestamp in the bubble overlaps the total votes label, this issue should be addressed in a dedicated PR as this is not specific to polls

**Screenshots**

|Disclosed|Undisclosed|
|-|-|
|image|image|

PR #1113, opened 2023-08-22
[Poll] Render ended poll and improve UI in the timeline

Florian14:
Successor of #1031 and continue https://github.com/vector-im/element-meta/issues/1830

# Content

- Improve the UI to match the final [design](https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=8770-8521&mode=design&t=rQNwjhQM0kPgg54n-4)
- Render ended poll on the start event

Note:
- Poll events are still not produced with the current matrix sdk, screenshots have been taken with a custom sdk.
- Timestamp overlapping the poll tile will be fixed in a separate PR.
- I will write some tests to ensure a certain poll event produces the correct UI models.

# Screenshots

[poll_timeline_demo.webm](https://github.com/vector-im/element-x-android/assets/11990514/5d1bb62b-2ac9-4e57-b6bf-5df0a4178197)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentPollFactory.kt
```diff
@@ -36,18 +36,27 @@ class TimelineItemContentPollFactory @Inject constructor(
...
         // Todo Move this computation to the matrix rust sdk
-        val showResults = content.kind == PollKind.Disclosed && matrixClient.sessionId in content.votes.flatMap { it.value }
         val pollVotesCount = content.votes.flatMap { it.value }.size
         val userVotes = content.votes.filter { matrixClient.sessionId in it.value }.keys
+        val isEndedPoll = content.endTime != null
+        val winnerIds = content.answers.map { it.id }
+            .groupBy { content.votes[it]?.size ?: 0 } // Group by votes count
+            .maxBy { it.key } // Keep max voted answers
+            .takeIf { it.key > 0 } // Ignore if no option has been voted
+            ?.value.orEmpty()
```
ElementBot, line 46: - :warning: Chained call `orEmpty()` should be wrapped to a new line since preceding calls were.

features/poll/api/src/main/kotlin/io/element/android/features/poll/api/PollContentView.kt
```diff
@@ -40,79 +42,121 @@ import io.element.android.libraries.ui.strings.CommonStrings
...
+    Text(
+        modifier = Modifier.align(Alignment.End),
+        style = ElementTheme.typography.fontBodyXsRegular,
+        color = ElementTheme.colors.textSecondary,
+        text = stringResource(CommonStrings.common_poll_total_votes, votesCount),
+    )
+}
+
+@Composable
+fun ColumnScope.UndisclosedPollBottomNotice() {
```
ElementBot, line 117: - :warning: This @Composable function emits content but doesn't have a modifier parameter.

See https://mrmans0n.github.io/compose-rules/rules/#when-should-i-expose-modifier-parameters for more information.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/RadioButton.kt
```diff
@@ -63,18 +63,17 @@ internal fun compoundRadioButtonColors(): RadioButtonColors {
...
 @Composable
-private fun ContentToPreview() {
+private fun ContentToPreview(defaultCheck: Boolean) {
+    var checked by remember { mutableStateOf(defaultCheck) }
     Column {
-        Row(horizontalArrangement = Arrangement.spacedBy(6.dp)) {
-            RadioButton(selected = false, onClick = {})
-            RadioButton(selected = false, enabled = false, onClick = {})
-        }
-        Row(horizontalArrangement = Arrangement.spacedBy(6.dp)) {
```
jmartinesp, line 75: Actually, these were used so the designers could take a quick look in Showkase and verify the buttons look good in all possible states by looking at a single preview. Could we go back to displaying the 4 states at the same time?

If we want the interaction to work, we could use `selected = checked` in the first 2 components and `selected = !checked` in the next two?

jmartinesp (approved):
LGTM, I just have a comment about the previews modified.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/RadioButton.kt
Florian14: Done, I have an issue with the interactive mode on the `RadioButton` preview but the interaction works fine in Showkase

Florian14: (I just triggered the screenshots action)

PR #1115, opened 2023-08-22
Make links in messages clickable again

### LINKED ISSUE #1111: Links are not clickable anymore (newest nightly)
### Steps to reproduce

1. Installed newest nightly build via Firebase App Distribution/ App Tester
2. Open a chat and click on a link

#### What did you expect?

Link is opened in my browser

#### What happened instead?

The link is styled as expected but nothing happens on click. The message bubble just lights up like every other that is clicked.

May be related to https://github.com/vector-im/element-x-android/issues/1079 that I found in the change log (?)

### Your phone model

Pixel 6

### Operating system version

Android 13

0.1.4-nightly (40001040) from this morning (Aug 22, 2023 at 6:29 AM)

### Homeserver

_No response_

### Will you send logs?

No

No

jmartinesp:
## Changes

- Creates a `LINK_TAG` constant to use as default in `ClickableLinkText`.
- Renames `processedText` to `annotatedString` to make links work again in linkified texts (suppressing shadowing, which is justified here IMO since we don't care about the original value anymore).

## Why

Fixes #1111.

### DIFF
--- a/changelog.d/1111.bugfix
+++ b/changelog.d/1111.bugfix
@@ -1,0 +1,1 @@
+Make links in messages clickable again.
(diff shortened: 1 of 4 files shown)


jmartinesp:
In the near future we should be using `UrlAnnotation` instead of `StringAnnotations` for this, but they're still an experimental API, so we may move to using them once they are stable.

bmarty (approved):
LGTM, thanks!

PR #1127, opened 2023-08-23
Enable OIDC support

### LINKED ISSUE #1114: Finish OIDC integration
OIDC has been merged into the SDK

A couple of changes worth pointing out:

- [x] Client.accountURL contains a URL to be presented from an Account and devices button in the Settings screen.
- [x] Client.logout now returns an optional URL. When the URL is there, it should be presented to the user after logout for RP initiated logout on their account page.

bmarty:
- upgrade to SDK 0.1.45

## Oidc

- enable OIDC (remove the TODO and update the code)
- add setting entry point to configure account (with Localazy English sync to get `screen_settings_oidc_account`)
- require a database migration to add `oidcData` field from the SDK.
- Open URL on signout if provided (no url is provided on element.io)

Closes #1114 

Tested with my element.io account, and it's working fine.

## Refresh token

With this new version of the SDK, we now have didRefreshTokens() callback.
- update session db in case of didRefreshTokens() invocation

Not tested, but added a unit test about it.

### DIFF
--- a/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/confirmaccountprovider/ConfirmAccountProviderPresenter.kt
+++ b/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/confirmaccountprovider/ConfirmAccountProviderPresenter.kt
@@ -17,6 +17,7 @@
 package io.element.android.features.login.impl.screens.confirmaccountprovider
 
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
@@ -26,8 +27,11 @@ import androidx.compose.runtime.rememberCoroutineScope
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedFactory
 import dagger.assisted.AssistedInject
+import io.element.android.features.login.api.oidc.OidcAction
+import io.element.android.features.login.impl.DefaultLoginUserStory
 import io.element.android.features.login.impl.accountprovider.AccountProviderDataSource
 import io.element.android.features.login.impl.error.ChangeServerError
+import io.element.android.features.login.impl.oidc.customtab.DefaultOidcActionFlow
 import io.element.android.libraries.architecture.Async
 import io.element.android.libraries.architecture.Presenter
 import io.element.android.libraries.architecture.runCatchingUpdatingState
@@ -40,7 +44,9 @@ import java.net.URL
 class ConfirmAccountProviderPresenter @AssistedInject constructor(
     @Assisted private val params: Params,
     private val accountProviderDataSource: AccountProviderDataSource,
-    private val authenticationService: MatrixAuthenticationService
+    private val authenticationService: MatrixAuthenticationService,
+    private val defaultOidcActionFlow: DefaultOidcActionFlow,
+    private val defaultLoginUserStory: DefaultLoginUserStory,
 ) : Presenter<ConfirmAccountProviderState> {
 
     data class Params(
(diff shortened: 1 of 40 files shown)


features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/root/PreferencesRootNode.kt
```diff
@@ -73,7 +84,16 @@ class PreferencesRootNode @AssistedInject constructor(
...
+            onOpenDeveloperSettings = this::onOpenDeveloperSettings,
+            onSuccessLogout = { onSuccessLogout(activity, it) },
+            onManageAccountClicked = { onManageAccountClicked(activity, state.accountManagementUrl) },
         )
     }
+
+    private fun onSuccessLogout(activity: Activity, url: String?) {
+        Timber.d("Success logout with result url: $url")
+        url?.let {
+            activity.openUrlInChromeCustomTab(null, false, it)
```
jmartinesp, line 96: What is this new URL used for? Could we add some docs? 

Also, on lambdas you can add names to the parameters: they don't really do anything AFAIK, but are useful for self--documented code. Maybe we could use something like `onSuccessLogout: (fooUrl: String?) -> Unit`?

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/root/PreferencesRootPresenter.kt
```diff
@@ -84,4 +93,8 @@ class PreferencesRootPresenter @Inject constructor(
     private fun CoroutineScope.initialLoad(matrixUser: MutableState<MatrixUser?>) = launch {
         matrixUser.value = matrixClient.getCurrentUser()
     }
+
+    private fun CoroutineScope.initAccountManagementUrl(accountManagementUrl: MutableState<String?>) = launch {
```
jmartinesp, line 97: I think this `launch` could be just replaced with making this function `suspend`.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/MatrixClient.kt
```diff
@@ -55,9 +55,15 @@ interface MatrixClient : Closeable {
      * Will close the client and delete the cache data.
      */
     suspend fun clearCache()
-    suspend fun logout()
+
+    /**
+     * Logout the user.
+     * Returns an optional URL. When the URL is there, it should be presented to the user after logout for RP initiated logout on their account page.
```
jmartinesp, line 61: Oh I see, this is what the URL was used for. The `RP` here is the client? I'm not too familiar with OpenID technical details.

jmartinesp (approved):
LGTM, I just have some comments, feel free to ignore them if you're in a hurry. 

Some extra docs or comments would be helpful to better understand the OIDC flow though.

jmartinesp:
I also had a weird issue where I pressed the login button in the webview twice by mistake, and I got stuck in a 'you're already logged in' screen. Since I couldn't do anything else, I had to press back and that got me to the 'you're about to login using...' screen, but the continue button was disabled, so I had to go back once again and restart the flow from scratch. 

It's definitely not a blocker, but I wonder if there's something we could do in the future to improve that.

bmarty:
Thanks for the review. I will handle the remarks in another PR, mainly to have the latest SDK on develop.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/confirmaccountprovider/ConfirmAccountProviderPresenter.kt
```diff
@@ -97,4 +111,33 @@ class ConfirmAccountProviderPresenter @AssistedInject constructor(
             }.getOrThrow()
         }.runCatchingUpdatingState(loginFlowAction, errorTransform = ChangeServerError::from)
     }
+
+    private suspend fun onOidcAction(
+        oidcAction: OidcAction?,
```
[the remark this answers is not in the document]
bmarty: OK

PR #1128, opened 2023-08-23
Cleanup OIDC

bmarty:
Take into account comment from #1127 PR
Add a workaround to prevent user from getting stuck on a disabled "Continue" button.

(manually assigned to @jmartinesp for the follow up on #1127)

### DIFF
--- a/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/LoginFlowNode.kt
+++ b/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/LoginFlowNode.kt
@@ -22,7 +22,9 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.DisposableEffect
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
+import androidx.lifecycle.lifecycleScope
 import com.bumble.appyx.core.composable.Children
+import com.bumble.appyx.core.lifecycle.subscribe
 import com.bumble.appyx.core.modality.BuildContext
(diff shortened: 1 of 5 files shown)


features/login/impl/src/main/kotlin/io/element/android/features/login/impl/LoginFlowNode.kt
```diff
@@ -78,9 +85,26 @@ class LoginFlowNode @AssistedInject constructor(
...
+        lifecycle.subscribe(
+            onResume = {
+                if (customChromeTabStarted) {
+                    customChromeTabStarted = false
+                    // Workaround to detect that the Custom Chrome Tab has been closed
+                    // If there is no coming OidcAction (that would end this Node),
+                    // consider that the user has cancelled the login
+                    // by pressing back or by closing the Custom Chrome Tab.
+                    lifecycleScope.launch {
+                        delay(5000)
```
jmartinesp, line 102: Why the 5s delay here?

bmarty: This is sort of a compromise... If we wait more, the user will see a disabled button for a longer time.
If we wait less, we take the risk to cancel the login operation (the request to the homeserver with the Oidc token) before it finishes. The request does not take a long time, so maybe 2s is enough, but I did not want to take the risk.

The solution is not ideal, maybe we could try to detect that no OidcAction has been triggered (this should happen quite faster) and if this is the case we can cancel faster. I will think again about it.

jmartinesp (approved):
LGTM, thanks!

PR #1138, opened 2023-08-24
Prevent verification while initial sync is in progress

### LINKED ISSUE #1131: Verification attempts during initial sync on the SS proxy fail completely
### Steps to reproduce

Rick encountered:

1. Log in on a big account for the very first time
2. Sit spinning on sync while the SS proxy initial-syncs.
3. Try to verify the new login (by initiating from EAX, i think?)
4. Verif req goes through to another client (e.g. EW)
5. EAX then hangs forever trying to complete the verif as the verif response is presumably stuck behind the initial sync.

#### What did you expect?

Detect initial syncs from the proxy and warn the user explicitly and/or lock the app or show a beautiful animation during the very first initial sync?

Don't ask/let the user try to verify if they're still syncing, as we know it won't work.

### Your phone model

_No response_

### Operating system version

_No response_

playstore 0.1.2 (40001020)

### Homeserver

_No response_

### Will you send logs?

Yes

No

jmartinesp:
## Changes

This PR fixes 2 bugs:

1. The verification prompt won't be displayed after logging in: this is kind of a workaround, but I switched from listening to `RustSyncService.state()` to `RoomListService.state()`. For some reason, when RustSyncService emits a `Running` state, the verification controller is not ready to be used yet and the `RustVerificationService` can't be configured. In fact, it won't be retried until it changes sync state (error, stopped, then running again). This doesn't happen with `RoomListService.State`.
2. The 2 options to verify aren't displayed if the sync service isn't running, as they need it to work.

## Why

Fixes #1131 .

### DIFF
--- a/features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/root/PreferencesRootPresenter.kt
+++ b/features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/root/PreferencesRootPresenter.kt
@@ -60,8 +60,8 @@ class PreferencesRootPresenter @Inject constructor(
         val snackbarMessage by snackbarDispatcher.collectSnackbarMessageAsState()
         val hasAnalyticsProviders = remember { analyticsService.getAvailableAnalyticsProviders().isNotEmpty() }
 
-        // Session verification status (unknown, not verified, verified)
-        val canVerifySession by sessionVerificationService.canVerifySessionFlow.collectAsState(false)
+        // We should display the 'complete verification' option if the current session can be verified
+        val showCompleteVerification by sessionVerificationService.canVerifySessionFlow.collectAsState(false)
 
         val accountManagementUrl: MutableState<String?> = remember {
             mutableStateOf(null)
@@ -77,7 +77,7 @@ class PreferencesRootPresenter @Inject constructor(
             logoutState = logoutState,
             myUser = matrixUser.value,
             version = versionFormatter.get(),
-            showCompleteVerification = canVerifySession,
+            showCompleteVerification = showCompleteVerification,
             accountManagementUrl = accountManagementUrl.value,
             showAnalyticsSettings = hasAnalyticsProviders,
             showDeveloperSettings = showDeveloperSettings,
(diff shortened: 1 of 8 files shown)


features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/root/PreferencesRootPresenter.kt
```diff
@@ -82,7 +77,7 @@ class PreferencesRootPresenter @Inject constructor(
             logoutState = logoutState,
             myUser = matrixUser.value,
             version = versionFormatter.get(),
-            showCompleteVerification = sessionIsNotVerified,
+            showCompleteVerification = canVerifySession,
```
bmarty, line 80: About the naming, using `can` does not enforce the fact that the user `should` verify the session. IIUC, when true, the user `should` and `can` verify the session. Not sure, there is a modal verb to reflect this, so maybe just name the boolean `showCompleteVerification`?

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/verification/SessionVerificationService.kt
```diff
@@ -37,6 +38,11 @@ interface SessionVerificationService {
      */
     val sessionVerifiedStatus: StateFlow<SessionVerifiedStatus>
 
+    /**
+     * Returns whether the current session needs to be verified.
```
bmarty, line 42: Maybe:

```suggestion
     * Returns whether the current session needs to be verified and the SDK is ready to start the verification.
```

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
```diff
@@ -149,13 +148,11 @@ class RustMatrixClient constructor(
...
     init {
         client.setDelegate(clientDelegate)
-        rustSyncService.syncState
-            .onEach { syncState ->
-                if (syncState == SyncState.Running) {
-                    onSlidingSyncUpdate()
-                }
+        roomListService.state.onEach { state ->
+            if (state == RoomListService.State.Running) {
+                onSlidingSyncUpdate()
```
bmarty, line 153: Maybe rename this private fun, since it's not called anymore when the sync service is updated (and it was already not really the case: the existing name was not ideal). I suggest to rename to `onRoomListServiceRunning`.

It was confusing me during the review, but I see that this fun only get the verification controller.

```kotlin
    private fun onSlidingSyncUpdate() {
        if (!verificationService.isReady.value) {
            try {
                verificationService.verificationController = client.getSessionVerificationController()
            } catch (e: Throwable) {
                Timber.e(e, "Could not start verification service. Will try again on the next sliding sync update.")
            }
        }
    }
```

Also for clarity, I would replace 
` if (!verificationService.isReady.value) {`
 by
` if (verificationService.verificationController == null) {`
I think this is equivalent.

Maybe update the Timber log too.

bmarty (approved):
Thanks for the fix, some non-blocking remarks

PR #1146, opened 2023-08-24
ListItem: Don't apply clickable modifier if onClick is null.

julioromano:
If it is applied but disabled (like before this PR) it will hijack all touch input from the view and no other component in it can respond to clicks.

### DIFF
--- a/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/ListItem.kt
+++ b/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/ListItem.kt
@@ -119,7 +119,7 @@ fun ListItem(
 
     androidx.compose.material3.ListItem(
         headlineContent = decoratedHeadlineContent,
-        modifier = Modifier.clickable(enabled = enabled && onClick != null, onClick = onClick ?: {}).then(modifier),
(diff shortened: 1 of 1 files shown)


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/ListItem.kt
```diff
@@ -119,7 +119,7 @@ fun ListItem(
 
     androidx.compose.material3.ListItem(
         headlineContent = decoratedHeadlineContent,
-        modifier = Modifier.clickable(enabled = enabled && onClick != null, onClick = onClick ?: {}).then(modifier),
+        modifier = if (onClick != null) Modifier.clickable(enabled = enabled, onClick = onClick).then(modifier) else modifier,
```
bmarty, line 122: Not new in this PR, but I am wondering if this should not be 
```kotlin
modifier = modifier.then(if (onClick != null) Modifier.clickable(enabled = enabled, onClick = onClick) else modifier,
```
instead. WDYT?

jmartinesp: That breaks the ripple effect if we set any padding.

PR #1150, opened 2023-08-27
first pull request

mittwerk:
## Type of change
- [ ] Technical

## Content

make from interface - fun interface & simplifying one sentence 
## Motivation and context
Try to be useful and begin to help

### DIFF
--- a/libraries/encrypted-db/src/main/kotlin/io/element/encrypteddb/passphrase/PassphraseProvider.kt
+++ b/libraries/encrypted-db/src/main/kotlin/io/element/encrypteddb/passphrase/PassphraseProvider.kt
@@ -19,7 +19,7 @@ package io.element.encrypteddb.passphrase
 /**
  * An abstraction to implement secure providers for SQLCipher passphrases.
  */
-interface PassphraseProvider {
+fun interface PassphraseProvider {
     /**
      * Returns a passphrase for SQLCipher in [ByteArray] format.
      */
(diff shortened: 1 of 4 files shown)


libraries/encrypted-db/src/main/kotlin/io/element/encrypteddb/passphrase/PassphraseProvider.kt
```diff
@@ -19,7 +19,7 @@ package io.element.encrypteddb.passphrase
 /**
  * An abstraction to implement secure providers for SQLCipher passphrases.
  */
-interface PassphraseProvider {
+fun interface PassphraseProvider {
```
bmarty, line 22: I do not think we want to have a `fun interface` here and for the other 2 places you applied the same change. `fun interface` are more useful when we have unnamed object that implement them. This is not the case here, and the interfaces may be extended at some point.

libraries/eventformatter/api/src/main/kotlin/io/element/android/libraries/eventformatter/api/TimelineEventFormatter.kt
```diff
@@ -18,6 +18,6 @@ package io.element.android.libraries.eventformatter.api
 
 import io.element.android.libraries.matrix.api.timeline.item.event.EventTimelineItem
 
-interface TimelineEventFormatter {
+fun interface TimelineEventFormatter {
```
bmarty, line 21: For this one, eventually you could have simplified [this code](https://github.com/vector-im/element-x-android/blob/develop/features/messages/impl/src/test/kotlin/io/element/android/features/messages/fixtures/timelineItemsFactory.kt#L75).

But it's probably better to update it and write a class like:

```kotlin
class NoOpTimelineEventFormatter() : TimelineEventFormatter {
    override fun format(event: EventTimelineItem): CharSequence = ""
}
```

instead.

bmarty (commented):
Thanks for the PR. See my remarks.

Also if you want the PR to be mergeable, ensure that you add a sign-off to it.

PR #1148, opened 2023-08-25
Sync strings and record screenshot

bmarty:
Not sure why the CI was not complaining about the diff for the poll screenshot :/

PR #1152, opened 2023-08-28
Sync Strings

ElementBot:

### DIFF
--- a/features/messages/impl/src/main/res/values-de/translations.xml
+++ b/features/messages/impl/src/main/res/values-de/translations.xml
@@ -4,6 +4,10 @@
     <item quantity="one">"%1$d Raumänderung"</item>
     <item quantity="other">"%1$d Raumänderungen"</item>
   </plurals>
+  <plurals name="screen_room_timeline_more_reactions">
+    <item quantity="one"></item>
+    <item quantity="other">"%1$d weitere"</item>
+  </plurals>
   <string name="screen_room_attachment_source_camera">"Kamera"</string>
   <string name="screen_room_attachment_source_camera_photo">"Foto aufnehmen"</string>
   <string name="screen_room_attachment_source_camera_video">"Video aufnehmen"</string>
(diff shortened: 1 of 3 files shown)


features/messages/impl/src/main/res/values-de/translations.xml
```diff
@@ -4,6 +4,10 @@
     <item quantity="one">"%1$d Raumänderung"</item>
     <item quantity="other">"%1$d Raumänderungen"</item>
   </plurals>
+  <plurals name="screen_room_timeline_more_reactions">
+    <item quantity="one"></item>
```
Florian14, line 8: I don't understand why this empty case has been imported, there are other plurals with only the `other` case, and I don't see any difference in Localazy, wdyt @bmarty?

libraries/ui-strings/src/main/res/values/localazy.xml
```diff
@@ -173,6 +173,14 @@
   <string name="room_timeline_beginning_of_room_no_name">"This is the beginning of this conversation."</string>
   <string name="room_timeline_read_marker_title">"New"</string>
   <string name="screen_analytics_settings_share_data">"Share analytics data"</string>
+  <string name="screen_create_poll_add_option_btn">"Add option"</string>
```
julioromano, line 176: @bmarty These should not be here but rather in the poll feature module.
The problem is that the updated filters in the `config.json` exists only in the `julioromano/create_poll_ui` branche at the moment.
What do we do in such situations?

features/messages/impl/src/main/res/values-de/translations.xml
bmarty: I [asked for people](https://matrix.to/#/!kCCQTCfnABLKGGvQjo:matrix.org/$UUl0w-4Qi5KNKSi8prIZUOOZilmouDeM3lQ7sML4srs?via=matrix.org&via=element.io&via=lant.uk) to complete the translation, but it's not done yet.

libraries/ui-strings/src/main/res/values/localazy.xml
bmarty: That's the issue of having an external tool that does not care about our branch.
I guess for this one it's not a big deal, the string is not used yet (on develop). When you'll merge your PR it will vanish.

bmarty:
The strings have been updated recently (in #1148), we can skip this sync for the reasons above.

PR #1162, opened 2023-08-28
Emojibase integration

langleyd:
Resolves https://github.com/vector-im/element-x-android/issues/1118
- Uses element translations for categories
- Uses Material emoji category icons


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/customreaction/DefaultEmojibaseProvider.kt
```diff
@@ -14,16 +14,14 @@
...
-import com.vanniktech.emoji.EmojiManager
-import com.vanniktech.emoji.google.GoogleEmojiProvider
+import android.content.Context
+import io.element.android.emojibasebindings.EmojibaseDatasource
+import io.element.android.emojibasebindings.EmojibaseStore
 
-class EmojiInitializer : Initializer<Unit> {
-    override fun create(context: android.content.Context) {
-        EmojiManager.install(GoogleEmojiProvider())
+class DefaultEmojibaseProvider(val context: Context) :EmojibaseProvider {
```
jmartinesp, line 23: Nit: it should be `Class(...): Parent {`

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/customreaction/DefaultEmojibaseProvider.kt
```diff
@@ -14,16 +14,14 @@
...
+    override fun loadEmojibase(): EmojibaseStore {
+        return EmojibaseDatasource().load(context)
```
jmartinesp, line 25: Should this be lazy loaded into memory?

jmartinesp (approved):
Thanks for you work on this, I can confirm it works fine! The emoji picker feels a bit sluggish when navigating to different categories, but that was already the case with the previous library IIRC.

I have a couple of minor comments, feel free to ignore them if you're in a hurry.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/customreaction/DefaultEmojibaseProvider.kt
langleyd: I decided to load it each time as it was relatively quick vs having it held in memory when it's not being used.

You think it'd be better to hold in memory? So that subsequent loads are quicker?

jmartinesp: It depends on how much memory it uses while loaded, I guess, but it's true that it loads everything quite fast. 

Making this `DefaultEmojibaseProvider` a singleton and having `loadEmojiBase()` return some lazy loaded value would be more efficient as we'd only have to do it once, but if wastes too much memory it might not be worth it.

langleyd: Yea, on reflection, airing on the side of snappier UI is probably what we are going for now. A slight delay each time you go to leave a reaction might be annoying. Plus, this would bring it in line with iOS.

PR #1180, opened 2023-08-29
Poll: Maestro tests

julioromano:
Adds maestro test for polls.


bmarty (approved):
Nice, thanks. We could add more test by adding and deleting an option for instance, but this is a good base.

PR #1181, opened 2023-08-29
Poll vote

julioromano:
- Adds sendPollVote rust room API (still not operational, need to wait for a rust sdk release)
- Adds an optional EventId in TimelineItemPollContent
- Wires the poll answer click listeners all the way to the TimelinePresenter in order to call the new room API
- Shows question as message summary in long press menu

Closes https://github.com/vector-im/element-meta/issues/2025


jmartinesp (approved):
LGTM after testing this with the latest Rust SDK. It's a shame we need to pass the `onPollAnswerSelected` lambda to so many components, but we can figure a way to improve this in the future. 

Congrats for getting this working! And don't forget to remove the TODOs!

Issue #1185, opened 2023-08-30
OIDC iteration


bmarty:
- [ ] The latest SDK has a `logoUri` property which uses [this URL](https://github.com/vector-im/element-x-ios/blob/905d975a7d9a320c2b9b942f877414574df72b94/ElementX/Sources/Application/AppSettings.swift#L107).
- [ ] The `tosUri` 404s (ooops), it should be using [this URL](https://github.com/vector-im/element-x-ios/blob/905d975a7d9a320c2b9b942f877414574df72b94/ElementX/Sources/Application/AppSettings.swift#L111).

PR #1186, opened 2023-08-30
Upgrade rust sdk to v48

julioromano:
- Sends content instead of string in message reply and edit
- Adds poll response and end APIs
- Adds logoUri to OidcConfiguration


bmarty:
For the compilation error, you can give `https://element.io/mobile-icon.png` to the `OidcConfiguration` change for the parameter `logoUri` (or `null`) and I will iterate later (when handling #1185)

jmartinesp (approved):
Once the issue with `logoUri` is fixed, LGTM!

PR #1194, opened 2023-08-31
PollCreate: Focus on question field when opening screen.

julioromano:


features/poll/impl/src/main/kotlin/io/element/android/features/poll/impl/create/CreatePollView.kt
```diff
@@ -68,6 +75,10 @@ fun CreatePollView(
         onSubmitClicked = { state.eventSink(CreatePollEvents.NavBack) },
         onDismiss = { state.eventSink(CreatePollEvents.HideConfirmation) }
     )
+    val questionFocusRequester = remember { FocusRequester() }
```
bmarty, line 78: As a remark, there are other usage of `FocusRequester()` in teh codebase, by we never used `remember` for them. I guess we should do.

bmarty (approved):
Tested OK, thanks!

PR #1196, opened 2023-08-31
Enable polls

julioromano:
Enable the Polls feature. Allows to create, view, vote and end polls.


jmartinesp (approved):
LGTM, thanks!

bmarty (commented):
I think a changelog file could be included, since this small PR actually enable a new feature in the app. So we have to let the users know.

PR #1215, opened 2023-09-04
Sync Strings

ElementBot:


features/messages/impl/src/main/res/values-de/translations.xml
```diff
@@ -4,6 +4,10 @@
     <item quantity="one">"%1$d Raumänderung"</item>
     <item quantity="other">"%1$d Raumänderungen"</item>
   </plurals>
+  <plurals name="screen_room_timeline_more_reactions">
+    <item quantity="one"></item>
+    <item quantity="other">"%1$d weitere"</item>
```
jmartinesp, line 9: I believe this translation is broken, but it's not used at all in the project. I'm not sure if we should just remove it from Localazy.

bmarty: Let's remove it yes, if it's confirmed that we should not use it.

PR #1219, opened 2023-09-04
Feature/bma/notification decryption

### LINKED ISSUE #1178: Some notifications don't decrypt
### Steps to reproduce

1. Rolling down the notification drawer of Android
2. Looking at EX notifications

#### What did you expect?
See a line of text with the notification

#### What happened instead?
Some notifications only contain the line "message"

Screenshot_20230829-190635

### Your phone model

Pixel 5

### Operating system version

Android 13

0.1.6-nightly

### Homeserver

element.io

### Will you send logs?

No

No

bmarty:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Ensure that the sync service is started when handling a Push.

Add to implement a `reason`, to ensure that we call `syncService.stop()` only when no code is using it.

## Motivation and context

Fixes #1178 

## Screenshots / GIFs

NA

## Tests

From https://github.com/vector-im/element-x-android/issues/1178#issuecomment-1702485220:

- Alice and Bob in room. Bob is using EAX
- bob kills EAX
- Alice rotates their megolm keys by typing '/discardsession' in this room
- Alice sends a message
- Bob should see the notification decrypted on their android device

With this change, Bob can see the message in the notification.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 13

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt
@@ -16,6 +16,7 @@
 
 package io.element.android.libraries.matrix.impl.sync
 
+import io.element.android.libraries.matrix.api.sync.StartSyncReason
 import io.element.android.libraries.matrix.api.sync.SyncService
 import io.element.android.libraries.matrix.api.sync.SyncState
 import kotlinx.coroutines.CoroutineScope
@@ -25,6 +26,8 @@ import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import org.matrix.rustcomponents.sdk.SyncServiceInterface
 import org.matrix.rustcomponents.sdk.SyncServiceState
 import timber.log.Timber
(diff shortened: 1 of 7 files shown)


jmartinesp:
We display notifications too when the app is in foreground for other rooms than the one the user is in. Has this been taken into account in the Rust SDK?

jmartinesp:
It seems like the sample project needs some fixes.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt
```diff
@@ -33,19 +36,36 @@ class RustSyncService(
     private val innerSyncService: SyncServiceInterface,
     sessionCoroutineScope: CoroutineScope
 ) : SyncService {
+    private val mutex = Mutex()
+    private val startSyncReasonSet = mutableSetOf<StartSyncReason>()
```
bnjbvr, line 40: I don't know enough about Kotlin, but can the mutex hold the startSyncReasonSet itself, like we'd do in Rust `Mutex<HashSet<StartSyncReason>>`? That would make it more obvious that one needs the mutex to read/write to the set.

bnjbvr (approved):
Makes sense :+1:

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt
jmartinesp: I don't think we can do something like that, `Mutex` in Kotlin can't hold values.

jmartinesp (approved):
Assuming that 'killing the app' means just removing it from recents (which should keep the notification receiver alive) it seems to be working fine. Once we fix the sample project issue it should be ok to merge.

Issue #1218, opened 2023-09-04
[Design] Iterate on user display name and default avatar color


bmarty:
Waiting for design.

Algorithm to compute user's color: https://github.com/vector-im/compound-web/blob/main/src/components/Avatar/useIdColorHash.ts

```js
export function useIdColorHash(id: string): number {
  const MIN = 1;
  const MAX = 8;
  // Sum up the values of all the char codes in the string
  const charCodeSum = id.split("").reduce((sum, char) => {
    return sum + char.charCodeAt(0);
  }, 0);
  return (charCodeSum % MAX) + MIN;
}
```

jmartinesp:
The designs for Avatars are in [this compound issue](https://github.com/vector-im/compound/issues/137). The cluster avatars can't be implemented yet since we lack the info, but the chosen colors and the algorithm used to pick combinations are in the issue if I'm not mistaken.

bmarty:
Done in #1224. Follow up in #1227

PR #1224, opened 2023-09-05
Iterate on display name and avatar color

bmarty:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Apply color on display name and default Avatar to sender.

## Motivation and context

Part of #1218

## Screenshots / GIFs

See the recorded screenshot.

## Tests

- Open a timeline, in light and dark mode, and observe that the sender name color is similar to what we have on EleWeb.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/avatar/Avatar.kt
```diff
@@ -85,12 +88,13 @@ private fun ImageAvatar(
 @Composable
 private fun InitialsAvatar(
     avatarData: AvatarData,
+    avatarColors: AvatarColors?,
     modifier: Modifier = Modifier,
 ) {
     // Use temporary color for default avatar background
     val avatarColor = ElementTheme.colors.bgActionPrimaryDisabled
```
jmartinesp, line 95: Maybe rename this to `fallbackAvatarColor` and modify the comment above?

libraries/designsystem/src/test/kotlin/io/element/android/libraries/designsystem/colors/AvatarColorsTest.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.google.common.truth.Truth.assertThat
+import io.element.android.libraries.theme.colors.avatarColorsDark
+import io.element.android.libraries.theme.colors.avatarColorsLight
+import org.junit.Test
+
+class AvatarColorsTest {
+
+    @Test
+    fun `ensure list size`() {
+        // avatarColorsDark and avatarColorsLight size must not be modified.
```
jmartinesp, line 28: I'd rather make the `String.toHash()` function composable and use the underlying `size` property for the modulo, but that works too.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
```diff
@@ -344,13 +346,13 @@ private fun MessageSenderInformation(
...
-            Avatar(senderAvatar)
+            Avatar(senderAvatar, avatarColors = avatarColors)
             Spacer(modifier = Modifier.width(4.dp))
             Text(
                 text = sender,
                 maxLines = 1,
                 overflow = TextOverflow.Ellipsis,
-                color = MaterialTheme.colorScheme.primary,
+                color = avatarColors.foreground,
                 style = ElementTheme.typography.fontBodyMdMedium,
```
jmartinesp, line 356: Is this font right? I think it should be bold instead looking at the designs.

jmartinesp (commented):
I think we should double check the font weight since it seems bolder in the designs, other than that, congrats, it looks great!

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/avatar/Avatar.kt
bmarty: Actually, I'm not sure if I should just remove the parameter `avatarColors`, since it is only used by the `InitialsAvatar` composable... So it could be computed only in this case. But since it's already computed for the sender display name, I did not want to compute it twice so I have added a parameter... What would you do?

libraries/designsystem/src/test/kotlin/io/element/android/libraries/designsystem/colors/AvatarColorsTest.kt
bmarty: I have updated the code, `String.toHash()` does not need to be `@Composable`

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
bmarty: This PR just deal with the colors for now... I did not have access to Figma [file](https://www.figma.com/file/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?type=design&node-id=522-742&mode=design&t=dgnm5jgDarXUZnEC-0), but it's OK now, I will iterate.

jmartinesp: We don't have the `Inter` font (I guess that's web-only), but we can set Roboto to Bold. Since the font size should probably be calculated given the size of the avatar component, that could be enough. It would be great to have some actual desgins for Android though...

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/avatar/Avatar.kt
bmarty: Actually we cannot compute the colors here, since InitialAvatar can also used for the room avatar, and this PR is only about user avatar.

jmartinesp: Are you sure? On the latest EW nightly I see plenty of rooms with no avatars defaulting to these kind of avatar colors:

image

I'll ask in the Compound room in any case.

jmartinesp: That said, I doesn't need to be done in this PR.

jmartinesp (approved):
LGTM!

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/avatar/Avatar.kt
bmarty: I see, I will merge and apply the changes to rooms in a dedicated PR.

PR #1226, opened 2023-09-05
Fix test flakiness by warming up molecule tests

jonnyandrew:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Reduce test flakiness by warming up molecule tests. Add a test rule that pre-runs a no-op molecule test and apply it to any presenter tests that use molecule.

## Motivation and context

Through profiling, I found that the first test of every test class takes a significant time to run. It turns out that molecule takes a significant amount of time to initialise when running in the first test of any given test class. This can create flakiness as tests will randomly exceed the default timeout of `runTest` (10 seconds).

To remove this source of flakiness, this solution ensures that the first call to initialise `moleculeFlow` occurs in a way that isn't impacted by the short default timeout.

image

Related issues:
- https://github.com/Kotlin/kotlinx.coroutines/issues/3800
- https://github.com/vector-im/element-x-android/pull/479 (similar solution)

## Screenshots / GIFs

N/A

## Tests

N/A

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/appnav/src/test/kotlin/io/element/android/appnav/RootPresenterTest.kt
+++ b/appnav/src/test/kotlin/io/element/android/appnav/RootPresenterTest.kt
@@ -31,10 +31,16 @@ import io.element.android.features.rageshake.test.screenshot.FakeScreenshotHolde
 import io.element.android.services.apperror.api.AppErrorState
 import io.element.android.services.apperror.api.AppErrorStateService
 import io.element.android.services.apperror.impl.DefaultAppErrorStateService
+import io.element.android.tests.testutils.WarmUpRule
 import kotlinx.coroutines.test.runTest
+import org.junit.Rule
 import org.junit.Test
 
 class RootPresenterTest {
+    @Rule
+    @JvmField
+    val warmUpRule = WarmUpRule()
+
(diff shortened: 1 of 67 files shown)


jmartinesp (approved):
LGTM, it's great that it doesn't affect CI times either!

bmarty (approved):
Thanks for looking at this.
In the spirit of point 3 of https://github.com/vector-im/element-x-android/blob/develop/docs/continuous_integration.md#rules :)

julioromano:
Isn't this actually running the warm up before every test of the test class?
It seems from the wording of the PR that it'd be enough to run the warmup once for all the tests of a given test class, did I get it right?
If this is the case then would it be worth using `@ClassRule` instead of `@Rule`  ?

julioromano:
> Yes it _should_ work as a class rule although I haven't tested this. Once warmed up the rule adds negligible overhead but perhaps we could still incrementally update these?

Followed up with: https://github.com/vector-im/element-x-android/pull/1310

PR #1246, opened 2023-09-07
Feature/bma/enable polls

### LINKED ISSUE #1241: v0.16 - Polls are not enabled
### Steps to reproduce

1. Have someone or yourself in another client start a poll
2. Element X shows "Unsupported Event"

#### What did you expect?

A Poll

#### What happened instead?

Unsupported event.

In 

https://matrix.to/#/!aCvOWKNDeXXpPgquKq:matrix.org/$qTdCinvFCy32_iNPSPv4Amo0wCTi_QdnvhifL2Zn9QQ?via=matrix.org&via=envs.net&via=element.io

@jorgem:element.io Informed that Polls are behind a dev option screen, which is not accessible in Release.

Screenshot_20230906-225806_Element X
Screenshot_20230906-225748_Element

### Your phone model

OP7

### Operating system version

Android 13

Element X v0.1.3

### Homeserver

aguiarvieira.pt

### Will you send logs?

No

No

bmarty:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Small rework of the feature flag module, to improve code clarity and avoid mistake.
Enable polls in release build.

## Motivation and context

Fixes #1241  

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
+++ b/libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
@@ -18,7 +18,7 @@ package io.element.android.libraries.featureflag.api
 
 /**
  * To enable or disable a FeatureFlags, change the `defaultValue` value.
- * Warning: to enable a flog for the release app, you MUST update the file
+ * Warning: to enable a flag for the release app, you MUST update the file
  * [io.element.android.libraries.featureflag.impl.StaticFeatureFlagProvider]
  */
 enum class FeatureFlags(
--- a/libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
+++ b/libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
@@ -18,7 +18,8 @@ package io.element.android.libraries.featureflag.api
 
 /**
  * To enable or disable a FeatureFlags, change the `defaultValue` value.
- * It will impact all the build types.
+ * Warning: to enable a flog for the release app, you MUST update the file
+ * [io.element.android.libraries.featureflag.impl.StaticFeatureFlagProvider]
  */
 enum class FeatureFlags(
     override val key: String,
@@ -40,6 +41,7 @@ enum class FeatureFlags(
     NotificationSettings(
         key = "feature.notificationsettings",
         title = "Show notification settings",
+        // Do not forget to edit StaticFeatureFlagProvider when enabling the feature.
         defaultValue = false,
     ),
 }
--- a/libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
+++ b/libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
@@ -16,21 +16,26 @@
 
 package io.element.android.libraries.featureflag.api
 
+/**
+ * To enable or disable a FeatureFlags, change the `defaultValue` value.
+ * It will impact all the build types.
+ */
 enum class FeatureFlags(
     override val key: String,
     override val title: String,
     override val description: String? = null,
-    override val defaultValue: Boolean = true
+    override val defaultValue: Boolean
 ) : Feature {
     LocationSharing(
         key = "feature.locationsharing",
         title = "Allow user to share location",
+        defaultValue = true,
     ),
     Polls(
         key = "feature.polls",
         title = "Polls",
         description = "Create poll and render poll events in the timeline",
-        defaultValue = false,
+        defaultValue = true,
     ),
     NotificationSettings(
         key = "feature.notificationsettings",
--- a/libraries/featureflag/impl/src/main/kotlin/io/element/android/libraries/featureflag/impl/StaticFeatureFlagProvider.kt
+++ b/libraries/featureflag/impl/src/main/kotlin/io/element/android/libraries/featureflag/impl/StaticFeatureFlagProvider.kt
@@ -22,7 +22,7 @@ import javax.inject.Inject
 
 /**
  * This provider is used for release build.
- * [isFeatureEnabled] just returns the default value of the FeatureFlags.
+ * This is the place to enable or disable feature for the release build.
  */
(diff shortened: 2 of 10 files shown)


libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
```diff
@@ -16,21 +16,26 @@
...
+/**
+ * To enable or disable a FeatureFlags, change the `defaultValue` value.
+ * It will impact all the build types.
+ */
 enum class FeatureFlags(
     override val key: String,
     override val title: String,
     override val description: String? = null,
-    override val defaultValue: Boolean = true
+    override val defaultValue: Boolean
```
ganfra, line 27: I'd keep the defaultValue to true, so it's enabled by default for debug

libraries/featureflag/impl/src/main/kotlin/io/element/android/libraries/featureflag/impl/StaticFeatureFlagProvider.kt
```diff
@@ -20,19 +20,18 @@ import io.element.android.libraries.featureflag.api.Feature
...
+    override val priority = LOW_PRIORITY
 
     override suspend fun isFeatureEnabled(feature: Feature): Boolean {
         return if (feature is FeatureFlags) {
-            when (feature) {
-                FeatureFlags.LocationSharing -> true
-                FeatureFlags.Polls -> false
-                FeatureFlags.NotificationSettings -> false
-            }
+            feature.defaultValue
```
ganfra, line 34: IMO StaticFeatureFlag should really be clear about which feature is enabled or not.
Using a `when` makes it exhaustive, so it's the responsability of the developer to enable it or not before the release. Maybe we should just add a comment at the beginning of the when to make it clear

libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
ganfra, line 27: I'd keep the defaultValue to true as it's the "debug" value.

bmarty: I do not agree, for debug, you can enable it in the develop option.

libraries/featureflag/impl/src/main/kotlin/io/element/android/libraries/featureflag/impl/StaticFeatureFlagProvider.kt
bmarty: Developers forget to update the value here, they just do it in `FeatureFlags`, like for instance in this PR: https://github.com/vector-im/element-x-android/pull/1196.

With this change, I think there is no more possible mistake 🤞

libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
bmarty: But I admit it can be an issue if the flags is impacting the FTUE flow...

libraries/featureflag/impl/src/main/kotlin/io/element/android/libraries/featureflag/impl/StaticFeatureFlagProvider.kt
ganfra: We should just educate developers, it should be clear which feature or not is enabled in release...

libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
ganfra: Yes of course, but defaultValue was more intended to be use as a fallback value. 
We should probably add more comment then, adding a reference to the StaticFeatureFlagProvider.

bmarty: Developers (or people enabling flags) do not have to know about `StaticFeatureFlagProvider` IMO.

libraries/featureflag/impl/src/main/kotlin/io/element/android/libraries/featureflag/impl/StaticFeatureFlagProvider.kt
bmarty: I see. I will try to improve this then.

Ideally we would have a centralized configuration file, that can also be used for gradle Kotlin files.

libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
ganfra: And IMO, it's the only file they have to know :D
We could add a ReleaseFeatureFlags file somewhere, and StaticFeatureFlagProvider would just read from it?

libraries/featureflag/impl/src/main/kotlin/io/element/android/libraries/featureflag/impl/StaticFeatureFlagProvider.kt
ganfra: Yes but we should makes sure it's exhaustive...

libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
bmarty: Maybe it's a naming issue with `defaultValue`.
We can have a flag with `defaultValue = false`, and the feature enabled in `StaticFeatureFlagProvider`. To confusing IMO.

bmarty:
Updated with 762a03259b9d24f65e7cff8eadad98532691776e.

ganfra (approved):

PR #1251, opened 2023-09-07
Use roomInfo instead of individually access roomListItem properties

ganfra:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [X] Technical
- [ ] Other :

## Content

Use `RoomInfo` instead of accessing all properties of `RoomListItem`. It should speed up mapping.


bmarty (approved):
LGTM, maybe add a changelog?

PR #1262, opened 2023-09-08
Take screen density into account when requesting thumbnails

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Have `CoilMediaFetcher` calculate the right size to ask for thumbnails using the screen density.

## Motivation and context

Fixes issues seen with blurry avatars on some homeservers.

## Screenshots / GIFs

|Before|After|
|-|-|
|image|image|

## Tests

- Log in to `element.io`.
- Check avatar images.

They should look a lot shaper.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 11

## Checklist

### DIFF
--- a/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/media/ImageLoaderFactories.kt
+++ b/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/media/ImageLoaderFactories.kt
@@ -46,8 +46,8 @@ class LoggedInImageLoaderFactory @Inject constructor(
                 }
                 add(AvatarDataKeyer())
                 add(MediaRequestDataKeyer())
-                add(CoilMediaFetcher.AvatarFactory(matrixClient))
-                add(CoilMediaFetcher.MediaRequestDataFactory(matrixClient))
+                add(CoilMediaFetcher.AvatarFactory(context, matrixClient))
+                add(CoilMediaFetcher.MediaRequestDataFactory(context, matrixClient))
             }
             .build()
     }
--- a/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/media/CoilMediaFetcher.kt
+++ b/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/media/CoilMediaFetcher.kt
@@ -16,6 +16,7 @@
 
 package io.element.android.libraries.matrix.ui.media
 
+import android.content.Context
 import coil.ImageLoader
 import coil.decode.DataSource
 import coil.decode.ImageSource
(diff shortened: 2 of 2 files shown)


bmarty (commented):
Thanks for looking at this.

Don't you think the fix should be located at the call site?

For instance here: https://github.com/vector-im/element-x-android/blob/develop/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/media/AvatatarDataExt.kt#L26

We should pass a pixel size instead of a DP size. Maybe more complicated though...

jmartinesp:
Yes, that was my first idea, but to add it there we'd either have to make that function `@Composable` or pass either a context or a density value. And if instead of storing the dp size in `AvatarSize` we wanted the pixel one we'd have to do it when those are created, and given it's an enum class that would be quite convoluted to make it work.

Placing this calculation in the fetcher seemed like the best compromise.

libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/media/ImageLoaderFactories.kt
```diff
@@ -46,8 +46,8 @@ class LoggedInImageLoaderFactory @Inject constructor(
                 }
                 add(AvatarDataKeyer())
                 add(MediaRequestDataKeyer())
-                add(CoilMediaFetcher.AvatarFactory(matrixClient))
-                add(CoilMediaFetcher.MediaRequestDataFactory(matrixClient))
+                add(CoilMediaFetcher.AvatarFactory(context, matrixClient))
+                add(CoilMediaFetcher.MediaRequestDataFactory(context, matrixClient))
```
bmarty, line 50: Not a blocker, but instead of passing the whole `context`, can we just pass the screen density here?
Not sure that this can change over time, but maybe (with battery saver mode which simulate low density screen).

Or maybe just `context.resources.displayMetrics`.

I am definitely thinking out loud ;)

bmarty (approved):
OK, thanks for explanation!

libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/media/ImageLoaderFactories.kt
jmartinesp: I'd have to double check, because the class has an `@Inject` annotation. So I think we need to either remove it (not sure if it's possible) or we manually provide the density value it in a module function.

jmartinesp: Actually, I'm wondering about what would happen if we passed it here and then changed the settings for display element scaling in the OS. Maybe we should pass a lambda or a wrapper `DensityProvider` class that hides the context?

jmartinesp: Done in https://github.com/vector-im/element-x-android/pull/1262/commits/e7e41b058938e0de08a4beb5b10250cde8ca0858.

bmarty (approved):
OK, thanks!

PR #1267, opened 2023-09-11
Sync Strings

ElementBot:


bmarty:
Skip this one and do it again when German/French are back.

PR #1273, opened 2023-09-11
TimelineEvent : fix memory leak

ganfra:
I spotted this potential memory leak while digging performance issues.

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/item/event/TimelineEventContentMapper.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/item/event/TimelineEventContentMapper.kt
@@ -42,7 +42,7 @@ import org.matrix.rustcomponents.sdk.OtherState as RustOtherState
 class TimelineEventContentMapper(private val eventMessageMapper: EventMessageMapper = EventMessageMapper()) {
 
     fun map(content: TimelineItemContent): EventContent {
-        return content.use { _ ->
+        return content.use {
             content.kind().use { kind ->
                 map(content, kind)
             }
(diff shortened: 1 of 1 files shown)


jmartinesp:
Those are actually all data classes, can they link something? They don't have an underlying pointer AFAIK.

They're mapped in:

```kotlin
public object FfiConverterTypeTimelineItemContentKind : FfiConverterRustBuffer<TimelineItemContentKind>{
    override fun read(buf: ByteBuffer): TimelineItemContentKind {
        return when(buf.getInt()) {
            1 -> TimelineItemContentKind.Message
            2 -> TimelineItemContentKind.RedactedMessage
            3 -> TimelineItemContentKind.Sticker(
                FfiConverterString.read(buf),
                FfiConverterTypeImageInfo.read(buf),
                FfiConverterString.read(buf),
            )
            ...
```

ganfra:
Actually I'm not sure, as they implement Disposable interface, but yeah you're right, I don't know...

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/item/event/TimelineEventContentMapper.kt
```diff
@@ -34,84 +34,91 @@ import io.element.android.libraries.matrix.impl.media.map
...
 
-    fun map(content: TimelineItemContent): EventContent = content.use {
-        when (val kind = it.kind()) {
-            is TimelineItemContentKind.FailedToParseMessageLike -> {
-                FailedToParseMessageLikeContent(
-                    eventType = kind.eventType,
-                    error = kind.error
-                )
+    fun map(content: TimelineItemContent): EventContent {
+        return content.use { _ ->
```
jmartinesp, line 45: It's a bit weird to ignore the parameter with `_` and then user the original value. I mean, it's valid, but a bit confusing.

jmartinesp (approved):
LGTM, thanks.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/item/event/TimelineEventContentMapper.kt
ganfra: Yeah but I don't like to use `it` and I can't find suitable name not confusing neither...

bmarty:
```suggestion
        return content.use {
```

works as well.

ganfra: You're right, less confusing.

PR #1283, opened 2023-09-12
Element Call SPA integration

### LINKED ISSUE #1300: Element Call SPA integration
For technical reasons on Android, we need an app to host the Element Call SPA so that calls can continue when the app is in background or the device is locked (there is still no guarantee that the system does not decide to kill the app).

For product reasons, we want users to open EC links within the EX app.

We need to be able to:

- Open an EC SPA link into a webview in EX.
- Keep the call running on background or when the device is locked, but without camera input.
- Handle call.element.io universal links.
- Handle custom scheme `element://call?url=some-encoded-url` for customers hosting their own EC instance.

jmartinesp:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Enable Element Call integration with Element X by embedding SPA urls:

- Urls with `https://call.element.io/*` format can be used.
- Urls using a custom scheme with the format `element://call?url=some-encoded-url` can be opened too.
- The call can be used with the app in background or the device locked by using a foreground service.
- The call can be put to background by tapping on the home button and can be returned to by either tapping on the foreground service notification or using the recents menu (I couldn't customize that entry though, neither `android:label` nor `setTaskDescription` seemed to work for some reason).
- Calls can be joined even if the user is not logged in the app.

Note that advanced features like using the logged in user's data for the call or hanging up and closing the call activity are out of scope for this implementation.

## Motivation and context

Closes #1300.

## Screenshots / GIFs

https://github.com/vector-im/element-x-android/assets/480955/cbef2a89-80d4-437b-8182-965f3492e67e

## Tests

- Open an element call URL (either an `https://call.element.io/some-call` or `element://call?url=some-encoded-url` one).
- An EC activity should be displayed.
- It should ask permissions for camera and microphone.
- The call should actually work.
- The call should survive common configuration changes (orientation, font scale, dark/light mode, etc.).
- Going to the home screen or other app, or locking the device should not stop the call, and it should display a permanent notification.
- Tapping on the notification or going back to the call using the recents menu should work.
- Either tapping on the close icon or hitting 'back' should actually finish the activity and remove the entry from the recents menu.

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): 11, 14.

## Checklist


jmartinesp:
> Patch coverage: **`17.03%`** and project coverage change: **`-0.25%`** ⚠️

Sorry for the low coverage, that's as low as I could make it without writing instrumented tests 😢 .

features/call/src/main/res/values/do_not_translate.xml
```diff
@@ -0,0 +1,20 @@
...
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<resources>
+    <string name="element_call">Element Call</string>
```
bmarty, line 19: Nit: I recommend to add `translatable="false"` attribute, to avoid lint warning.

features/call/src/main/kotlin/io/element/android/features/call/ElementCallActivity.kt
```diff
@@ -0,0 +1,265 @@
...
+            else -> null
+        }
+    }
+}
+
+typealias RequestPermissionCallback = (Array<String>) -> Unit
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+private fun CallContents(
```
bmarty, line 218: Maybe extract from this file, with the Preview.

features/call/src/main/kotlin/io/element/android/features/call/CallForegroundService.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    override fun onCreate() {
+        super.onCreate()
+
+        notificationManagerCompat = NotificationManagerCompat.from(this)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val foregroundServiceChannel = NotificationChannelCompat.Builder(
+                "call_foreground_service_channel",
+                NotificationManagerCompat.IMPORTANCE_LOW,
+            ).setName(getString(R.string.call_foreground_service_channel_title_android)).build()
```
bmarty, line 59: We already have this https://github.com/vector-im/element-x-android/blob/develop/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/channels/NotificationChannels.kt#L126 which is safer, (if getString returns empty string, it will crash the app), so maybe just remove this?

features/call/src/main/kotlin/io/element/android/features/call/CallForegroundService.kt
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+
+    private lateinit var notificationManagerCompat: NotificationManagerCompat
+
+    override fun onCreate() {
+        super.onCreate()
+
+        notificationManagerCompat = NotificationManagerCompat.from(this)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
bmarty, line 55: This check is for the notification channel, but it also wrap the `startForeground` call. Is it OK?

features/call/src/main/kotlin/io/element/android/features/call/CallIntentDataParser.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    fun parse(data: String?): String? {
+        val parsedUrl = data?.let { Uri.parse(data) } ?: return null
+        val scheme = parsedUrl.scheme
+        return when {
+            scheme in validHttpSchemes && parsedUrl.host == "call.element.io" -> data
+            scheme == "element" && parsedUrl.host == "call" -> {
+                parsedUrl.getQueryParameter("url")
+                    ?.let { URLDecoder.decode(it, "utf-8") }
+                    ?.takeIf { Uri.parse(it)?.scheme in validHttpSchemes }
```
bmarty, line 34: You do not check the host here as per line 30? This is maybe OK, but a comment would be welcome if this is the case, and a test case could be added to cover the case.

features/call/src/main/kotlin/io/element/android/features/call/CallWebView.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    with(settings) {
+        javaScriptEnabled = true
+        allowContentAccess = true
+        allowFileAccess = true
+        domStorageEnabled = true
+        mediaPlaybackRequiresUserGesture = false
+        databaseEnabled = true
+        loadsImagesAutomatically = true
+        userAgentString = "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/116.0.0.0 Safari/537.36"
```
bmarty, line 69: I guess this is fine, but we do not want to use `UserAgentProvider` here?

features/call/src/main/kotlin/io/element/android/features/call/ElementCallActivity.kt
```diff
@@ -0,0 +1,265 @@
...
+        if (savedInstanceState == null) {
+            updateUiMode(resources.configuration)
+        }
+
+        audioManager = getSystemService(AUDIO_SERVICE) as AudioManager
+        requestAudioFocus()
+
+        setContent {
+            CallContents(
+                url = urlState.value!!,
```
bmarty, line 83: Could be nice to avoid using `!!`. This is fine thanks to line 71, but still :)

bmarty (commented):
Nice work, and it's working pretty well. Some comments and feedback:

- When clicking on an Element Call url from a message, it does not open internally, but opens the web browser when testing on emu API 33 (we need to update https://element.io/.well-known/assetlinks.json I think). On API 29 it works as expected (hence I have to choose between all my app that support that link, and I think this could be avoided)

image

image

- The video captured on the device has a correct zoom on the video feedback, but on Element call web, it looks zoomed too much (maybe out of scope of this PR):

|EXA|Web|
|-|-|
|image|image|

- Analytics is on by default, some users may not like it. Can it take the current analytics setting of EXA? If this is not possible, I think it will be better to have analytics opt in screen (but out of scope)
image

- Rendering issue at the end of the call:

image

- Notification rendering (FTR):

image

bmarty:
One last thing, when the Activity is closed, the call does not end.

features/call/src/main/kotlin/io/element/android/features/call/CallForegroundService.kt
jmartinesp: It's not really the same, I think. That channel is used for alerting of an incoming call, this one is for containing the notification for the foreground service that keeps the call alive when the activity is paused. However, I can probably create a new channel here?

features/call/src/main/kotlin/io/element/android/features/call/CallIntentDataParser.kt
jmartinesp: This is for opening arbitrary urls through `element://call?url=https://some-custom-ec-instance.com/actual-call` so the best we could do is check it's not null or blank, I think.

features/call/src/main/kotlin/io/element/android/features/call/CallForegroundService.kt
bmarty: Ah yes, you're right about the channel diff. They may want to keep different settings.
Not sure about moving the code then, but please include the call to `ifEmpty` like [here](https://github.com/vector-im/element-x-android/blob/develop/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/channels/NotificationChannels.kt#L88).

features/call/src/main/kotlin/io/element/android/features/call/CallIntentDataParser.kt
bmarty: That's what I was suspected. Thanks!

bmarty:
> > One last thing, when the Activity is closed, the call does not end.
> 
> Closed with a back press or tapping in the close icon, you mean?

Tapping the close button at top start.

jmartinesp:
> Tapping the close button at top start.

Ok, it does die after a few seconds, but I just found a way to make it disappear instantly with `onRelease` and `webView.destroy()`.

jmartinesp:
@bmarty: All the issues above should be fixed now.

bmarty (approved):
LGTM, thanks for the update!

PR #1288, opened 2023-09-12
Send composer analytics when sending a message

julioromano:
Composer event are not sent to posthog when sending messages.

After investigation I've found that in `MessageComposerPresenter` the `analyticsService.capture()` call is invoked in the `SetMode` event handler rather than in the `SendMessage` event handler.

This change should result in the desired behavior.


bmarty (approved):
Thanks. I have checked on EA and we do the same thing.

Naming the analytics event `Composer` is a bit weird though, something like `SendMessage` should be more appropriate / more obvious, but this is out of scope of this PR.

PR #1294, opened 2023-09-13
Poll Creation: Switch focus to newly added option field when clicking "Add option".

julioromano:
Poll Creation: Switch focus to newly added option field when clicking "Add option".


jmartinesp (changes requested):
This actually makes the latest option also the one with focus when you start creating a poll. Maybe the `answerFocusRequester` should be added only to the latest item and called after tapping on add option?

julioromano:
Should have fixed it, I also added a programmatic scroll so that the "Add option" button is still visible after adding an option

jmartinesp (approved):
Thanks for taking the time to fix it. After both our changes I think it's working perfectly now. LGTM!

PR #1296, opened 2023-09-13
Test analytics when sending messages

julioromano:
Covers changes in https://github.com/vector-im/element-x-android/pull/1288

### DIFF
--- a/features/messages/impl/src/test/kotlin/io/element/android/features/messages/textcomposer/MessageComposerPresenterTest.kt
+++ b/features/messages/impl/src/test/kotlin/io/element/android/features/messages/textcomposer/MessageComposerPresenterTest.kt
@@ -24,6 +24,7 @@ import app.cash.molecule.moleculeFlow
 import app.cash.turbine.ReceiveTurbine
 import app.cash.turbine.test
 import com.google.common.truth.Truth.assertThat
+import im.vector.app.features.analytics.plan.Composer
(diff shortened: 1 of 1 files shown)


julioromano:
@bmarty odd that codecov says there are no additional lines covered with this PR 🤷

features/messages/impl/src/test/kotlin/io/element/android/features/messages/textcomposer/MessageComposerPresenterTest.kt
```diff
@@ -304,6 +331,14 @@ class MessageComposerPresenterTest {
             assertThat(messageSentState.richTextEditorState.messageHtml).isEqualTo("")
             assertThat(messageSentState.canSendMessage).isFalse()
             assertThat(fakeMatrixRoom.replyMessageParameter).isEqualTo(A_REPLY to A_REPLY)
+            assertThat(analyticsService.capturedEvents).containsExactly(
+                Composer(
+                    inThread = false,
+                    isEditing = false,
+                    isReply = true,
+                    messageType = Composer.MessageType.Text,
```
bmarty, line 339: Just a remark, when you reply with EXA to a message which is in thread, the message will be sent to the thread. This is handled by the SDK. Not sure that we want to reflect this in the analytics. I guess we will use `inThread = true` only when we will properly support threads.

bmarty (approved):
Thanks!

Issue #533, opened 2023-06-06
Add in Notifications settings a toggle for room mentions and another one for VoIP calls


giomfo:
## Description ##
Add in Notifications settings 
- a toggle for room mentions related to the new Override rule `.m.rule.is_room_mention` and the deprecated `.m.rule.roomnotif` rule. The toggle is ON as soon as at least one of them is enabled. Both are updated in case of change.
- a toggle for VoIP calls related to the underride rule `.m.rule.call`

## Suggested Design ##
<img width="274" alt="image" src="https://github.com/vector-im/element-x-android/assets/8969772/8a07a846-0a79-427b-9554-87e8c454847d">

langleyd:
This was done in https://github.com/vector-im/element-x-android/pull/1304, but it was decided we would hide the calls setting as calling was not yet supported in the app. We can add it back easily once it is supported.

langleyd:
Re-opening as  I think these tasks are being used for signoff proccess.

bmarty:
Hello @giomfo can this old issue be closed?

giomfo:
> This was done in [#1304](https://github.com/element-hq/element-x-android/pull/1304), but it was decided we would hide the calls setting as calling was not yet supported in the app. We can add it back easily once it is supported.

@mxandreas @manuroe The toggle to enable/disable the voip call notifications is missing in EX Notification Settings. It has been implemented but it is hidden on purpose until now. Should we add it or not?
Please reopen https://github.com/element-hq/element-x-ios/issues/1025 according to your answer here, to handle the same question on EX-iOS

Issue #532, opened 2023-06-06
Display in Notifications settings the default mode for the group and the direct chats


giomfo:
## Description ##
Add in the notifications settings a section "Notify me for" to provide the default setting for group chats and direct chats.

The default value for group chats depends on 2 Underride rules: `.m.rule.message` and `.m.rule.encrypted`. We will consider only their `enabled` attribute:
- if both are enabled -> "All Messages"
- if both are disabled -> "Mentions and Keywords"
- if only one is enabled -> display a banner to let user know the configuration is not supported. A button is available to fix the pb. In that case, we enabled the second one to fix the pb.

The default value for group chats depends on 2 Underride rules: `.m.rule.encrypted_room_one_to_one` and `.m.rule.room_one_to_one`. We will consider only their `enabled` attribute:
- if both are enabled -> "All Messages"
- if both are disabled -> "Mentions and Keywords"
- if only one is enabled -> display a banner to let user know the configuration is not supported. A button is available to fix the pb. In that case, we enabled the second one to fix the pb.

## Suggested Design ##
<img width="275" alt="image" src="https://github.com/vector-im/element-x-android/assets/8969772/0233fe5e-95ac-42e8-8192-5815f7eecb22">

giomfo:
@yostyle Do you agree with @amshakal that the [option4](https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=9686%3A59303&mode=design&t=EY5fpl0LYqCRUEmW-1) is the best? 

FYI [here](https://github.com/vector-im/element-x-ios/issues/1024#issuecomment-1681821268) is my last wording suggestion

giomfo:
#1037 implement this issue without handling the potential unsupported configuration. 
FYI [option3](https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=9686-52782&mode=design&t=wPiysRyKeopipZXH-4) has been the one selected for iOS (see with @nimau for more details)

Issue #531, opened 2023-06-05
Add in Notifications settings the option to enable/disable notification on this device


giomfo:
## Description ##
Add a toggle at the top of the Notifications settings to enable/disable the notifications at the application level (The push service registration requires this toggle to be ON).
- Notifications are enabled by default in EX (the toggle is enabled by default).
- This toggle value is unchanged if the user rejects the system permission demand in the application after the login 

Here is the suggested design with a wrong label: 
<img width="271" alt="image" src="https://github.com/vector-im/element-x-android/assets/8969772/5ba72cb7-5790-4ac8-9186-0e1ba3a721fb">

We let the system prompt automatically the user to get the permission if it is missing when the application attempts to register to the push service. By default if this permission is missing or lost whereas the toggle is enabled, an item is displayed at the top of the Notification settings screen:
- the users are taken to the system settings when they click on this item.
- the toggle defined at the application level is still ON:

<img width="393" alt="image" src="https://github.com/vector-im/element-x-android/assets/8969772/6a6a5358-79ac-48f9-9944-18e065361a2b">

- if the toggle is switched off, the warning is hidden

Issue #510, opened 2023-06-01
Add a sub-screen "Notifications" in the existing application Settings


giomfo:
Add a section "Notifications' in the global settings

<img width="268" alt="image" src="https://github.com/vector-im/element-x-android/assets/8969772/202b13d1-27f3-4a96-afde-85d6631e1ad9">

PR #1304, opened 2023-09-13
Add global notification settings

langleyd:
## Content

- Adds global notification settings 
- Handles the correction of mismatched settings

## Motivation and context

#510 
#531
#532 
#533 

## What's missing
The settings doesn't yet support viewing a list rooms with  custom room notification settings. This will come in a separate PR.


features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/edit/EditDefaultNotificationSettingView.kt
```diff
@@ -0,0 +1,70 @@
...
+    } else {
+        CommonStrings.screen_notification_settings_group_chats
+    }
+    PreferenceView(
+        modifier = modifier,
+        onBackPressed = onBackPressed,
+        title = stringResource(id = title)
+    ) {
+
+        val validModes = listOf(RoomNotificationMode.ALL_MESSAGES, RoomNotificationMode.MENTIONS_AND_KEYWORDS_ONLY)
```
jmartinesp, line 47: No silent mode here?

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/edit/EditDefaultNotificationSettingState.kt
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.features.preferences.impl.notifications.edit
+
+import io.element.android.libraries.matrix.api.room.RoomNotificationMode
+
+data class EditDefaultNotificationSettingState(
+    val isOneToOne: Boolean,
```
jmartinesp, line 22: We usually use `isDm` through the project, as EX has its own 'DM definition' of being a direct room with 2 users, but I'm wondering if `isOneToOne` might be a better term.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsView.kt
```diff
@@ -0,0 +1,266 @@
...
+) {
+        val context = LocalContext.current
+        if (systemSettings.appNotificationsEnabled && !systemSettings.systemNotificationsEnabled) {
+            PreferenceText(
+                icon = Icons.Filled.NotificationsOff,
+                title = stringResource(id = CommonStrings.screen_notification_settings_system_notifications_turned_off),
+                subtitle = stringResource(id = CommonStrings.screen_notification_settings_system_notifications_action_required,
+                    stringResource(id = CommonStrings.screen_notification_settings_system_notifications_action_required_content_link)),
+                onClick = {
+                    val intent = Intent(Settings.ACTION_APPLICATION_DETAILS_SETTINGS)
```
jmartinesp, line 116: We have `Context.startNotificationSettingsIntent` in the `androidutils` library module, but it uses an `ActivityResultLauncher`. Maybe we could create a version without that and re-use it here?

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/SystemNotificationsEnabledProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import android.content.Context
+import androidx.core.app.NotificationManagerCompat
+import com.squareup.anvil.annotations.ContributesBinding
+import io.element.android.libraries.di.AppScope
+import io.element.android.libraries.di.ApplicationContext
+import io.element.android.libraries.di.SingleIn
+import javax.inject.Inject
+
+interface SystemNotificationsEnabledProvider {
```
jmartinesp, line 27: Is this actually needed? It can be useful if we plan to use it elsewhere in the future, but I think you could have gotten the same result by using:

```kotlin
val context = LocalContext.current
val areNotificationsEnabled by remember {
    NotificationManagerCompat.from(context).areNotificationsEnabled()
}
```

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenter.kt
```diff
@@ -0,0 +1,168 @@
...
+
+class NotificationSettingsPresenter @Inject constructor(
+    private val notificationSettingsService: NotificationSettingsService,
+    private val userPushStoreFactory: UserPushStoreFactory,
+    private val matrixClient: MatrixClient,
+    private val systemNotificationsEnabledProvider: SystemNotificationsEnabledProvider
+) : Presenter<NotificationSettingsState> {
+    @Composable
+    override fun present(): NotificationSettingsState {
+        val userPushStore = userPushStoreFactory.create(matrixClient.sessionId)
```
jmartinesp, line 49: This should probably be inside a `remember` to avoid creating a new one in each recomposition.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenter.kt
```diff
@@ -0,0 +1,168 @@
...
+        val appNotificationsEnabled = userPushStore
+            .getNotificationEnabledForDevice()
+            .collectAsState(initial = false)
+
+        val matrixSettings: MutableState<NotificationSettingsState.MatrixSettings> = remember {
+            mutableStateOf(NotificationSettingsState.MatrixSettings.Uninitialized)
+        }
+
+        LaunchedEffect(Unit) {
+            fetchSettings(matrixSettings)
```
jmartinesp, line 64: Maybe this first `fetchSettings` could be moved inside `observeNotificationSettings` and returned as the first value? I'm not sure how feasible it is with the `debounce` there though.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/edit/EditDefaultNotificationSettingView.kt
```diff
@@ -0,0 +1,70 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.stringResource
+import io.element.android.libraries.designsystem.components.preferences.PreferenceCategory
+import io.element.android.libraries.designsystem.components.preferences.PreferenceView
+import io.element.android.libraries.matrix.api.room.RoomNotificationMode
+import io.element.android.libraries.ui.strings.CommonStrings
+
+@Composable
+fun EditDefaultNotificationSettingView(
```
jmartinesp, line 30: It was a bit difficult to understand the difference between `DefaultNotificationSettings` and `NotificationSettings` screens, could we add some comments or docs?

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/edit/DefaultNotificationSettingOption.kt
```diff
@@ -0,0 +1,98 @@
...
+                .align(Alignment.CenterVertically)
+                .size(48.dp),
+            selected = isSelected,
+            onClick = null // null recommended for accessibility with screenreaders
+        )
+    }
+}
+@DayNightPreviews
+@Composable
+internal fun DefaultNotificationSettingOptionLightPreview() = ElementPreview { ContentToPreview() }
```
bmarty, line 83: When using `@DayNightPreviews`(which should be the default now), the `Light` word should not be in the fun name.

```suggestion
internal fun DefaultNotificationSettingOptionPreview() = ElementPreview { ContentToPreview() }
```

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/SystemNotificationsEnabledProvider.kt
jmartinesp: I see it's called again later. We could still remember the `NotificationManagerCompat` object and call that instead, but feel free to ignore this comment if keeping the provider is easier.

langleyd: Are you suggesting to put it in the view rather than the presenter?

Felt more like something that should be in the presenter and just pass data to the view, but then to keep the presenter testable I didn't want it to have a reference to the context.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/edit/EditDefaultNotificationSettingView.kt
langleyd: No, `ALL_MESSAGES` and `MENTIONS_AND_KEYWORDS_ONLY ` are the only valid values for global defaults on all the clients. Even on wed which has "Off" "On" and "Nosiey" in its legacy settings, "Off" maps to `MENTIONS_AND_KEYWORDS_ONLY` and "On" and "Nosiey" map to `ALL_MESSAGES`.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenter.kt
langleyd: Kind of like Rx's [startWidth](https://reactivex.io/documentation/operators/startwith.html). Not quite sure how to do with flow or if it would make much easier to understand.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/edit/EditDefaultNotificationSettingView.kt
langleyd: Will add a code comment to that effect.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsView.kt
```diff
@@ -0,0 +1,263 @@
...
+
+@Composable
+private fun NotificationSettingsContentView(
+    matrixSettings: NotificationSettingsState.MatrixSettings.Valid,
+    systemSettings: NotificationSettingsState.AppSettings,
+    onNotificationsEnabledChanged: (Boolean) -> Unit,
+    onGroupChatsClicked: () -> Unit,
+    onDirectChatsClicked: () -> Unit,
+    onMentionNotificationsChanged: (Boolean) -> Unit,
+    onCallsNotificationsChanged: (Boolean) -> Unit,
```
ElementBot, line 106: - :warning: Function parameter `onCallsNotificationsChanged` is unused.

jmartinesp: Is this expected?

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenter.kt
jmartinesp: There is also a `startWith` function for Flows, but it's deprecated. Ideally, it should be something like `.onStart { emit(initialValue) }`, but it might not be worth the trouble.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsView.kt
langleyd: Nope :), fixed.

jmartinesp (approved):
LGTM, thanks!

PR #1310, opened 2023-09-13
Optimize `WarmUpRule`

julioromano:
This PR is made of 2 commits.
The first is just a mass rename in order to spare 1 line of code every time `WarmUpRule` is used.
The second tweaks `WarmUpRule` so that the warmup code is only run once per test class.


bmarty (approved):
LGTM, thanks!

jonnyandrew (commented):
Thank you!

PR #1343, opened 2023-09-15
Remove suppressUnsupportedCompileSdk

julioromano:
We're now an AGP 8.1.0 so should be safe to remove as per the comment.


jmartinesp (approved):
LGTM, thanks!

bmarty (approved):
Should be OK, thanks for the cleanup!

PR #1358, opened 2023-09-18
Fix RoomSummary: RoomListEntry.Invalidated 

ganfra:
Currently the Invalidated case was trying to get from memory cache or was defaulting to empty, resulting in slowness when opening the app. 
Now it correctly try to build the RoomSummary and fallback to empty.


bmarty (approved):
Thanks!

PR #1361, opened 2023-09-18
Remove dead code in room summary row

julioromano:
Not sure why it's unused but it's almost 1 year old so probably safe to remove.


julioromano:
Assigning to @ganfra because he was the author of such code. Definitely not a priority though.

ganfra (approved):

PR #1365, opened 2023-09-18
Upgrade rust sdk to .55

julioromano:
Accommodates some breaking API changes.

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
@@ -187,7 +187,13 @@ class RustMatrixRoom(
         _membersStateFlow.value = MatrixRoomMembersState.Pending(prevRoomMembers = currentMembers)
         var rustMembers: List<RoomMember>? = null
         try {
-            rustMembers = innerRoom.members()
+            rustMembers = buildList {
+                while (true) {
+                    // Loading the whole iterator as a stop-gap measure.
+                    // We should probably implement some sort of paging in the future.
+                    addAll(innerRoom.members().nextChunk(1000u) ?: break)
+                }
+            }
             val mappedMembers = rustMembers.parallelMap(RoomMemberMapper::map)
             _membersStateFlow.value = MatrixRoomMembersState.Ready(mappedMembers)
             Result.success(Unit)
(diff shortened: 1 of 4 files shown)


julioromano:
Assigning to @ganfra as he's requested this build together with me.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/roomlist/RoomListExtensions.kt
```diff
@@ -93,7 +93,11 @@ fun RoomListService.syncIndicator(): Flow<RoomListServiceSyncIndicator> =
             }
         }
         tryOrNull {
-            syncIndicator(listener)
+            syncIndicator(
+                200u, // Same value which was previously hardcoded in the rust-sdk
```
ganfra, line 97: Can you use 1000u? we wanted to expand it a bit. Also can we name those constants?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
```diff
@@ -187,7 +187,13 @@ class RustMatrixRoom(
         _membersStateFlow.value = MatrixRoomMembersState.Pending(prevRoomMembers = currentMembers)
         var rustMembers: List<RoomMember>? = null
         try {
-            rustMembers = innerRoom.members()
+            rustMembers = buildList {
+                while (true) {
+                    // Loading the whole iterator as a stop-gap measure.
+                    // We should probably implement some sort of paging in the future.
+                    addAll(innerRoom.members().nextChunk(100u) ?: break)
```
ganfra, line 194: Should we put a higher chunk size for now?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/roomlist/RoomListExtensions.kt
julioromano: OFC!

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
julioromano: 1000 is ok?

ganfra: I guess!

ganfra (approved):

PR #1376, opened 2023-09-19
Fix room member infinite loop

ganfra:
Fix room member infinite loop from https://github.com/vector-im/element-x-android/pull/1365


bmarty (approved):
Good catch, thanks!

PR #1378, opened 2023-09-19
DI: rework how components are created and provided

ganfra:
Rework how dagger components are created and provided.

The way it was done previously, we could end up using the wrong component for a given node. Now, it makes sure the component is provided by a parent node.
Also I took the opportunity to simplify the code so we don't have to pass callback through`plugins`.

It should fix https://github.com/vector-im/element-x-android/issues/1277
It might also fix some crashes we have with `nodeFactories` not found.


bmarty (approved):
Nice work!

PR #1382, opened 2023-09-19
feat: use the raw homeserver URL instead of manually removing the scheme

bnjbvr:
On the one hand, the SDK knows what to do and will clean up the URL if needs be, guessing what the best scheme (http or https) is based on the prefix. On the other hand, trimming the scheme as done here doesn't allow for manual testing on insecure servers, which is handy in e2e testing situations.


bmarty (approved):
LGTM, thanks. Not sure why we try to extract the domain from the url. Maybe influenced by the parameter name of `configureHomeserver`. This is `serverNameOrHomeserverUrl` now, but I think it was something else before.

PR #1386, opened 2023-09-20
Remove usage of async-uniffi

### LINKED ISSUE #1381: async-uniffi is causing memory leaks
While async-uniffi has not a fix for the memory leaks, we should not use any `suspending` method from the rust sdk.
Thanks to https://github.com/matrix-org/matrix-rust-sdk/pull/2577 we now have blocking counterparts.

ganfra:
Closes #1381 

This might have fixed the deadlock from https://github.com/vector-im/element-x-android/issues/1235

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
@@ -55,11 +55,11 @@ import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.cancel
+import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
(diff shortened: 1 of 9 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
```diff
@@ -188,12 +188,12 @@ class RustMatrixRoom(
...
         var rustMembers: List<RoomMember>? = null
         try {
-            rustMembers = innerRoom.members().use { membersIterator ->
+            rustMembers = innerRoom.membersBlocking().use { membersIterator ->
                 buildList {
                     while (true) {
                         // Loading the whole membersIterator as a stop-gap measure.
                         // We should probably implement some sort of paging in the future.
-                        yield()
+                        ensureActive()
```
julioromano, line 196: What difference does this make in practice?
(yield calls ensureActive internally)

julioromano (approved):

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
ganfra: The yield is not working on this scenario, not sure why. Dont have time to investigate more.

PR #1387, opened 2023-09-20
Properly format lastMessage when it belongs to a poll.

julioromano:
Takes care of properly formatting a room's last message when it belongs to a poll.

NB: Polls still aren't exposed as a room's `last_message` from the rust SDK, so this code won't actually run yet. This will happen after integrating rust SDK version 0.1.57 which includes: https://github.com/matrix-org/matrix-rust-sdk/pull/2580

### DIFF
--- a/libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/DefaultRoomLastMessageFormatter.kt
+++ b/libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/DefaultRoomLastMessageFormatter.kt
@@ -96,7 +96,8 @@ class DefaultRoomLastMessageFormatter @Inject constructor(
                 stateContentFormatter.format(content, senderDisplayName, isOutgoing, RenderingMode.RoomList)
             }
             is PollContent -> {
-                prefixIfNeeded(content.question, senderDisplayName, isDmRoom)
+                val message = sp.getString(CommonStrings.common_poll_summary, content.question)
+                prefixIfNeeded(message, senderDisplayName, isDmRoom)
             }
             is FailedToParseMessageLikeContent, is FailedToParseStateContent, is UnknownContent -> {
                 prefixIfNeeded(sp.getString(CommonStrings.common_unsupported_event), senderDisplayName, isDmRoom)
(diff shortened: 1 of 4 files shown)


libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/DefaultRoomLastMessageFormatter.kt
```diff
@@ -95,7 +95,9 @@ class DefaultRoomLastMessageFormatter @Inject constructor(
             is StateContent -> {
                 stateContentFormatter.format(content, senderDisplayName, isOutgoing, RenderingMode.RoomList)
             }
-            is PollContent, // TODO Polls: handle last message
+            is PollContent -> {
+                prefixIfNeeded(content.question, senderDisplayName, isDmRoom)
```
bmarty, line 99: Don't we want to prefix the question with something like "Poll: ", or even better display a picto?

bmarty (approved):
LGTM, thanks.

libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/DefaultRoomLastMessageFormatter.kt
julioromano: Nice idea, forwarded the question to desgin: https://github.com/vector-im/element-meta/issues/2048#issuecomment-1727894105
Not sure though how easy it is to include an icon/image here with the current code that handle the last messages.

julioromano:
Prefixed summary with "Poll: " and added unit tests.

libraries/eventformatter/impl/src/test/kotlin/io/element/android/libraries/eventformatter/impl/DefaultRoomLastMessageFormatterTests.kt
```diff
@@ -58,6 +59,7 @@ import org.robolectric.RobolectricTestRunner
 import org.robolectric.RuntimeEnvironment
 import org.robolectric.annotation.Config
 
+@Suppress("LargeClass")
```
julioromano, line 62: @bmarty I had to add this, is it ok?

PR #1402, opened 2023-09-21
PoC: Create an example TimelineItemPresenter

julioromano:
This tries to create a `TimelineItemPollContentPresenter` and bind it into a map multibinding with anvil. The idea is to retrieve a factory for it in `TimelinePresenter` so that we can properly instantiate presenter for any `TimelineItemEventContent`s that need one in `TimelineItemEventContentView`.

Sadly I couldn't make it work yet.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
@@ -56,7 +56,7 @@ class TimelinePresenter @Inject constructor(
     private val dispatchers: CoroutineDispatchers,
     private val appScope: CoroutineScope,
     private val analyticsService: AnalyticsService,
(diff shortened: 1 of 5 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
```diff
@@ -54,6 +56,7 @@ class TimelinePresenter @Inject constructor(
     private val dispatchers: CoroutineDispatchers,
     private val appScope: CoroutineScope,
     private val analyticsService: AnalyticsService,
+    private val contentPresenterFactories: Map<String, @JvmSuppressWildcards Provider<TimelineItemPresenter>>,
```
ElementBot, line 59: - :warning: Private property `contentPresenterFactories` is unused.

ElementBot, line 59: - :warning: Private property `contentPresenterFactories` is unused.

PR #1403, opened 2023-09-21
Sync Strings

ElementBot:


bmarty (changes requested):
Will update some French strings.

PR #1427, opened 2023-09-25
Navigation fixes

ganfra:
This PR will fix 2 issues : 
- First one is some race condition on the node lifecycles causing a RoomNode to be loaded right after the LoggedInFlowNode is destroyed 
- Second issue his the backstack was not reset to the RoomList when navigating to a Room or Invite

### DIFF
--- a/services/appnavstate/impl/src/main/kotlin/io/element/android/services/appnavstate/impl/DefaultAppNavigationStateService.kt
+++ b/services/appnavstate/impl/src/main/kotlin/io/element/android/services/appnavstate/impl/DefaultAppNavigationStateService.kt
@@ -167,7 +167,7 @@ class DefaultAppNavigationStateService @Inject constructor(
     }
 
     private fun logError(logPrefix: String) {
-        Timber.tag(loggerTag.value).w("$logPrefix must be call first, can't leave.")
+        Timber.tag(loggerTag.value).w("$logPrefix must be call first.")
     }
 
     private fun NavigationState.assertOwner(owner: String): Boolean {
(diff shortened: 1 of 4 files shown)


services/appnavstate/impl/src/main/kotlin/io/element/android/services/appnavstate/impl/DefaultAppNavigationStateService.kt
```diff
@@ -83,7 +82,7 @@ class DefaultAppNavigationStateService @Inject constructor(
         val currentValue = state.value.navigationState
         Timber.tag(loggerTag.value).d("Navigating to space $spaceId. Current state: $currentValue")
         val newValue: NavigationState.Space = when (currentValue) {
-            NavigationState.Root -> error("onNavigateToSession() must be called first")
+            NavigationState.Root -> return logError("onNavigateToSession()")
```
ganfra, line 85: Do not throw but just log and return early.
This could lead to crash because of node lifecycle `onDestroy` callback going from parent to children. 
See here for the appyx related question https://github.com/bumble-tech/appyx/issues/498

services/appnavstate/impl/src/main/kotlin/io/element/android/services/appnavstate/impl/DefaultAppNavigationStateService.kt
```diff
@@ -167,6 +166,10 @@ class DefaultAppNavigationStateService @Inject constructor(
         state.getAndUpdate { it.copy(navigationState = NavigationState.Root) }
     }
 
+    private fun logError(logPrefix: String) {
+        Timber.tag(loggerTag.value).w("$logPrefix must be call first, can't leave.")
```
bmarty, line 170: I think the message is not generic enough.

```suggestion
        Timber.tag(loggerTag.value).w("$logPrefix must be call first.")
```

bmarty (commented):
OK, thanks. I was not comfortable with this fail-fast approach...

The tests have to be updated, let me know if you do not have time to do it.

bmarty (approved):

PR #1428, opened 2023-09-25
Use eventSink method reference in TimelinePresenter.kt

julioromano:
Has been changed in https://github.com/vector-im/element-x-android/pull/1172 but in general method references should always be preferred to lambdas in composable functions (because they have higher stability guarantees).


julioromano:
@jonnyandrew FYI

ganfra (approved):

PR #1457, opened 2023-09-28
Room : makes subscribeToSync/unsubscribeFromSync suspendable 

ganfra:
Rework how we use subscribe/unsubscribe on sync.
Previously the calls were synced and on the main thread, which could lead to break the navigation animation or even causes some ANR.
This is now handled properly with suspendable, and we keep a subscription count to avoid messing with the asynchronous nature of node lifecycle and coroutines.


bmarty (approved):
LGTM, thanks. Maybe add an entry for the changelog? Not sure since it does not change anything for the user, but it may help to investigate regression (if any!)

PR #1467, opened 2023-09-29
Display poll notifications in the notification drawer

julioromano:
Display notifications for polls as if they were message events but prepending the poll question with "Poll: " as we do in the room summary.


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotifiableEventResolver.kt
```diff
@@ -133,9 +133,22 @@ class NotifiableEventResolver @Inject constructor(
...
+                buildNotifiableMessageEvent(
+                    sessionId = userId,
+                    senderId = content.senderId,
+                    roomId = roomId,
+                    eventId = eventId,
+                    noisy = isNoisy,
+                    timestamp = this.timestamp,
+                    senderName = senderDisplayName,
+                    body = stringProvider.getString(CommonStrings.common_poll_summary, content.question),
+                    imageUriString = this.contentUrl,
```
bmarty, line 146: I guess in the case of a poll `this.contentUrl` will always be null.

bmarty (approved):
Thanks!

julioromano:
Hardcoded a null

PR #1477, opened 2023-10-02
Element call url param

bmarty:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Pass parameter `appPrompt` and `confineToRoom` in the fragment part of the URL

FTR: iOS PR: https://github.com/vector-im/element-x-ios/pull/1833

## Motivation and context

Better Element Call URL. Those parameters does not need to reach the server.

## Tests

- Just updated the unit test.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/call/src/main/kotlin/io/element/android/features/call/CallIntentDataParser.kt
+++ b/features/call/src/main/kotlin/io/element/android/features/call/CallIntentDataParser.kt
@@ -54,21 +54,50 @@ class CallIntentDataParser @Inject constructor() {
 }
 
 /**
- * Ensure the uri has the following parameters and value:
+ * Ensure the uri has the following parameters and value in the fragment:
  * - appPrompt=false
  * - confineToRoom=true
  * to ensure that the rendering will bo correct on the embedded Webview.
  */
 private fun Uri.withCustomParameters(): String {
     val builder = buildUpon()
+    // Remove the existing query parameters
     builder.clearQuery()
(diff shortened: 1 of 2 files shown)


julioromano (approved):

dbkr (commented):
Looks okay I think, although if it were me I would probably try to use Android's query string parsing/building to do the work, even though this is really awkward because 1. they only work on whole URLs for some reason, so you'd have to construct an intermediate fake URL builder with the fragment query as the real query string, using that to build a query string and 2. the classes to read and build URLs are separate so you'd need one of each to examine the old fragment query string and one to create the edited one.

I think this is okay though, might just behave oddly if it were given params with no values or something.

bmarty:
> Looks okay I think, although if it were me I would probably try to use Android's query string parsing/building to do the work, even though this is really awkward because 1. they only work on whole URLs for some reason, so you'd have to construct an intermediate fake URL builder with the fragment query as the real query string, using that to build a query string and 2. the classes to read and build URLs are separate so you'd need one of each to examine the old fragment query string and one to create the edited one.
> 
> I think this is okay though, might just behave oddly if it were given params with no values or something.

I agree, there is lots of manual stuff, this is not ideal. Unit tests are helping to reassures us. One other thing, the Uri builder will URL-encode the fragment, and I do not think we want this...

PR #1478, opened 2023-10-02
Fix template

bmarty:
Trying to fix why the template was not working on my AS anymore.

I unzipped the content and add a tiny script to zip it again to be able to install it on AS. Also updated the doc.

Setting the template again locally fixed ny issue, so there is no fixes, but it's better to avoid committing zip files anyway.

Like this, `findAll` action in the Idea may alos make the template files getting visible in the results.


tools/templates/files/IntelliJ IDEA Global Settings
jmartinesp, line 1: Is this expected?

tools/templates/install_templates.sh
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+echo "Zipping the contents of the 'files' directory..."
+
+rm -f ./tmp/file_templates.zip
```
jmartinesp, line 21: If `tmp` dir doesn't exist, the script fails. Maybe it needs an extra check.

tools/templates/files/fileTemplates/Template Presentation Classes.kt.child.2.kt
```diff
@@ -0,0 +1,42 @@
...
+    Box(modifier, contentAlignment = Alignment.Center) {
+        Text(
+            "${NAME} feature view",
+            color = MaterialTheme.colorScheme.primary,
+        )
+    }
+}
+
+@Preview
+@Composable
```
jmartinesp, line 28: It could be a good time to unify these using `@DayNightPreviews` maybe?

docs/_developer_onboarding.md
```diff
@@ -276,11 +276,12 @@ Follow these steps to install and configure the plugin and templates:
 
 1. Install the AS plugin for generating modules :
    [Generate Module from Template](https://plugins.jetbrains.com/plugin/13586-generate-module-from-template)
-2. Import file templates in AS :
+2. Run the script `tools/templates/generate_templates.sh` to generate the template zip file
```
jmartinesp, line 279: This should be `install_templates.sh`, right?

jmartinesp (changes requested):
This doesn't seem to work correctly for me neither on AS Giraffe 2022.3.1 Parch 2 or on the latest Iguana canary version. For some reason I always get just empty directories inside the created module, no gradle files, or presentation classes:

image

tools/templates/files/fileTemplates/Template Presentation Classes.kt.child.2.kt
bmarty: I just realized that: a65eb36

docs/_developer_onboarding.md
bmarty: Yes. I have decided to rename the script actually.

tools/templates/files/IntelliJ IDEA Global Settings
bmarty: I did not change the content of the original zip file. I guess it could be removed 🤷‍♂️

tools/templates/install_templates.sh
bmarty: 89c2130c1d28212d3023b8160398e5e7a832b3b0

jmartinesp:
> This is working here (`Android Studio Giraffe | 2022.3.1 Patch 1`), not sure about the other templates. Did you follow all the steps in the contribution doc?

I did, and the dialog displays the right templates in the right side:

image

And in the preview:

image

But then I just get empty folders when I select 'finish'.

jmartinesp:
Ah! The old templates have a trailing whitespace character that the new ones don't. Could this be the cause somehow?

jmartinesp (approved):
Ok, I think my settings were broken for some reason. I did something that broke them even more to the point of actually having to reset them to the defaults and importing the templates now worked as expected, same for generating a new module, so LGTM. I'll investigate in my other AS instance why I had this weird issue.

PR #1479, opened 2023-10-02
Maestro test: change server (again) to test server which does not support sliding sync

bmarty:
#1066 again, but for gnuradio.org.


bmarty:
All green, incl. Maestro.

PR #1552, opened 2023-10-12
Konsist: add more test

bmarty:
Split existing Konsist test into multiple Test class and add new tests.


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/utils/BooleanProvider.kt
```diff
@@ -20,5 +20,5 @@ import androidx.compose.ui.tooling.preview.PreviewParameterProvider
 
 open class BooleanProvider : PreviewParameterProvider<Boolean> {
     override val values: Sequence<Boolean>
-        get() = sequenceOf(false, true)
+        get() = sequenceOf(true, false)
```
bmarty, line 23: Reorder to avoid breaking screenshot test (this was not used before)

julioromano (approved):

PR #1555, opened 2023-10-12
Feature/bma/fix db migration test

bmarty:
Fix DB migration test.
The default value was missing in the schema.
According to my test, this is not necessary to do a migration, but I will double check.
Reviewer should also test this.

### DIFF
--- a/libraries/session-storage/impl/src/main/sqldelight/io/element/android/libraries/matrix/session/SessionData.sq
+++ b/libraries/session-storage/impl/src/main/sqldelight/io/element/android/libraries/matrix/session/SessionData.sq
@@ -1,6 +1,12 @@
-------------------------------------
---  Current version of the DB: 4  --
-------------------------------------
+--------------------------------------------------------------------
+-- Current version of the DB is the highest value of filename
+-- in the folder `sqldelight/databases`.
+--
+-- When upgrading the schema, you have to create a file .sqm in the
+-- `sqldelight/databases` folder and run the following task to
+-- generate a .db file using the latest schema
+-- > ./gradlew generateDebugSessionDatabaseSchema
+--------------------------------------------------------------------
 
 CREATE TABLE SessionData (
     userId TEXT NOT NULL PRIMARY KEY,


libraries/session-storage/impl/src/main/sqldelight/io/element/android/libraries/matrix/session/SessionData.sq
```diff
@@ -1,13 +1,20 @@
...
     refreshToken TEXT,
     homeserverUrl TEXT NOT NULL,
     slidingSyncProxy TEXT,
+    -- added in version 2
     loginTimestamp INTEGER,
+    -- added in version 3
     oidcData TEXT,
-    isTokenValid INTEGER NOT NULL,
+    -- added in version 4
+    isTokenValid INTEGER NOT NULL DEFAULT 1,
```
bmarty, line 17: `DEFAULT 1` was missing here, and is present in the file `3.sqm`, which migrate the data.

julioromano (approved):

libraries/session-storage/impl/src/main/sqldelight/migrations/0.sqm
```diff
@@ -1,3 +1,6 @@
+-- This file is not striclty necessary, since the first
```
julioromano, line 1: Perhaps this file should be removed.

libraries/session-storage/impl/src/main/sqldelight/io/element/android/libraries/matrix/session/SessionData.sq
```diff
@@ -1,13 +1,20 @@
+------------------------------------
```
julioromano, line 1: I understand the desire of keeping track of things, but this comment will only add additional burden to any future sqldelight migration (you may also forget to update the comment and things will then appear out of sync).

The current version of the DB is always represented by the last *.db file.
Every time you commit a change to the .sq files you must always create a new x.sqm and then generate x+1.db and commit those as well.

libraries/session-storage/impl/src/main/sqldelight/migrations/1.sqm
```diff
@@ -1 +1,3 @@
+-- Migrate DB from version 1
+
```
julioromano, line 2: Same comment as before. This is redundant: Sqldelight explains 1.sqm == migration from version 1.
No problem to keep the comment. Just mentioning that we'll have one more thing to enforce in the future (we'll have to ensure that all *.sqm files have this header)

julioromano (approved):

libraries/session-storage/impl/src/main/sqldelight/io/element/android/libraries/matrix/session/SessionData.sq
ganfra: +1

libraries/session-storage/impl/src/main/sqldelight/migrations/0.sqm
bmarty: I wanted to remove it, but then I got a warning in the IDE in the file 1.sqm, claiming that `SessionData does not exist`.

So I decided to keep it and add a comment. It's fine to have it anyway I think.

libraries/session-storage/impl/src/main/sqldelight/io/element/android/libraries/matrix/session/SessionData.sq
bmarty: You're right. I have updated the comment, also to make the process more visible (it's in the gradle file now)

12996e825353c71f1a5f59d650d500721fcb9849

libraries/session-storage/impl/src/main/sqldelight/migrations/0.sqm
julioromano: > I wanted to remove it, but then I got a warning in the IDE in the file 1.sqm, claiming that `SessionData does not exist`.
> 
> So I decided to keep it and add a comment. It's fine to have it anyway I think.

This might be a "bug" due to the fact that our Sqldelight version is lower than the current one.
So you're ide might be running the sqldelight plugin 2.0.0 but our project is on 1.5.5

We might consider upgrading to 2.0.0  it's been released months ago.

julioromano: Migration is not that difficult, just a few package renames.

bmarty: https://github.com/vector-im/element-x-android/issues/1557

PR #1558, opened 2023-10-12
Move Konsist tests to their own module

bmarty:
- Move Konsist tests to their own module
- Fix issue (merged today on develop, checked by previous version of Konsist tests)

The drawback is that dependencies will have to be added to the `konsist` module on-demand, for instance to check an annotation, a class, etc.


julioromano (approved):
CooL!

Issue #1563, opened 2023-10-13
[Rust sdk] Remove usage of blocking methods


ganfra:
Now that https://github.com/matrix-org/matrix-rust-sdk/pull/2661 has been merged, we should be able to use suspend methods again

PR #1565, opened 2023-10-13
Misc: remove usage of blocking methods #1563

ganfra:
Closes https://github.com/vector-im/element-x-android/issues/1563

If everything works properly then we can merge https://github.com/matrix-org/matrix-rust-sdk/pull/2715

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
@@ -174,7 +174,7 @@ class RustMatrixRoom(
         _membersStateFlow.value = MatrixRoomMembersState.Pending(prevRoomMembers = currentMembers)
         var rustMembers: List<RoomMember>? = null
         try {
-            rustMembers = innerRoom.membersBlocking().use { membersIterator ->
+            rustMembers = innerRoom.members().use { membersIterator ->
                 buildList {
                     while (true) {
                         // Loading the whole membersIterator as a stop-gap measure.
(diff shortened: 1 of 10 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
```diff
@@ -60,6 +60,8 @@ import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.withContext
 import org.matrix.rustcomponents.sdk.EventTimelineItem
+import kotlinx.coroutines.yield
```
ElementBot, line 63: - :warning: The import 'kotlinx.coroutines.yield' is unused.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
```diff
@@ -60,6 +60,8 @@ import kotlinx.coroutines.flow.StateFlow
...
+import org.matrix.rustcomponents.sdk.RequiredState
```
ElementBot, line 64: - :warning: The import 'org.matrix.rustcomponents.sdk.RequiredState' is unused.

julioromano (approved):
LGTM, I've seen you've removed all dispatcher and withContext stuff.
Does this mean uniffi will now take care of running the work on the most appropriate thread?
Or does this mean the dispatcher to use will still come from the calling code (i.e. the app's coroutine context) ?
Or... else? :D

ganfra:
> LGTM, I've seen you've removed all dispatcher and withContext stuff. Does this mean uniffi will now take care of running the work on the most appropriate thread? Or does this mean the dispatcher to use will still come from the calling code (i.e. the app's coroutine context) ? Or... else? :D

There is no need to dispatch anymore as those methods are now truly suspending. But we still have non suspending methods in the codebase, to wrap with `withContext`

PR #1582, opened 2023-10-16
Add script to build the rust sdk locally and from different repo / branch

bmarty:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Add a script to build the Rust SDK.

## Motivation and context

Ease the way to test Rust branch, even from forks, or to test a local version of the rust sdk source

## Screenshots / GIFs

NA

## Tests

- run the script and see if it works as expected

## Tested devices

- MacOS

## Checklist

### DIFF
--- a/libraries/matrix/impl/build.gradle.kts
+++ b/libraries/matrix/impl/build.gradle.kts
@@ -29,11 +29,12 @@ anvil {
 }
 
 dependencies {
+    releaseImplementation(libs.matrix.sdk)
     if (file("${rootDir.path}/libraries/rustsdk/matrix-rust-sdk.aar").exists()) {
         println("\nNote: Using local binary of the Rust SDK.\n")
-        implementation(projects.libraries.rustsdk)
+        debugImplementation(projects.libraries.rustsdk)
     } else {
-        implementation(libs.matrix.sdk)
+        debugImplementation(libs.matrix.sdk)
     }
     implementation(projects.libraries.di)
     implementation(projects.libraries.androidutils)
--- a/libraries/matrix/impl/build.gradle.kts
+++ b/libraries/matrix/impl/build.gradle.kts
@@ -29,8 +29,12 @@ anvil {
 }
 
 dependencies {
-    // implementation(projects.libraries.rustsdk)
-    implementation(libs.matrix.sdk)
+    if (file("${rootDir.path}/libraries/rustsdk/matrix-rust-sdk.aar").exists()) {
+        println("\nNote: Using local binary of the Rust SDK.\n")
+        implementation(projects.libraries.rustsdk)
+    } else {
+        implementation(libs.matrix.sdk)
+    }
     implementation(projects.libraries.di)
     implementation(projects.libraries.androidutils)
     implementation(projects.libraries.network)
--- a/tools/sdk/build_rust_sdk.sh
+++ b/tools/sdk/build_rust_sdk.sh
@@ -46,7 +46,7 @@ fi
 read -p "Do you want to build the app after (yes/no) default to yes? " buildApp
 buildApp=${buildApp:-yes}
 
-# If folder ../matrix-rust-components-kotlin does not exist, close the repo
+# If folder ../matrix-rust-components-kotlin does not exist, clone the repo
 if [ ! -d "../matrix-rust-components-kotlin" ]; then
     printf "\nFolder ../matrix-rust-components-kotlin does not exist. Cloning the repository into ../matrix-rust-components-kotlin.\n\n"
     git clone https://github.com/matrix-org/matrix-rust-components-kotlin.git ../matrix-rust-components-kotlin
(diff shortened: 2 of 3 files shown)


libraries/matrix/impl/build.gradle.kts
```diff
@@ -29,8 +29,12 @@ anvil {
 }
 
 dependencies {
-    // implementation(projects.libraries.rustsdk)
-    implementation(libs.matrix.sdk)
+    if (file("${rootDir.path}/libraries/rustsdk/matrix-rust-sdk.aar").exists()) {
```
julioromano, line 32: Just for safety, could we make it so that this `if` works only for debug builds? (i.e. release builds will always use `   implementation(libs.matrix.sdk)` no matter what?)

julioromano: Yeah but adding a `releaseImplementation()` and haveing the if using only `debugImplementation` could be an additional layer of safety

tools/sdk/build_rust_sdk.sh
```diff
@@ -0,0 +1,80 @@
...
+printf "\nBuilding the SDK...\n\n"
+./scripts/build.sh -p ${rustSdkPath} -m sdk -t aarch64-linux-android -o ../element-x-android/libraries/rustsdk
+
+cd ../element-x-android
+mv ./libraries/rustsdk/sdk-android-debug.aar ./libraries/rustsdk/matrix-rust-sdk.aar
+mkdir -p ./libraries/rustsdk/sdks
+cp ./libraries/rustsdk/matrix-rust-sdk.aar ./libraries/rustsdk/matrix-rust-sdk-${date}.aar
+
+
+if [ ${buildApp} != "yes" ]; then
```
jmartinesp, line 70: I've tried running the script twice, once just hitting enter so it takes the default 'build app?' option, then manually entering 'yes', but I can't see it actually building the app. I guess this should be `==`?

tools/sdk/build_rust_sdk.sh
```diff
@@ -0,0 +1,80 @@
...
+fi
+
+printf "\nResetting matrix-rust-components-kotlin to the latest main branch...\n\n"
+cd ../matrix-rust-components-kotlin
+git reset --hard
+git checkout main
+git pull
+
+printf "\nBuilding the SDK...\n\n"
+./scripts/build.sh -p ${rustSdkPath} -m sdk -t aarch64-linux-android -o ../element-x-android/libraries/rustsdk
```
jmartinesp, line 62: Won't this only work for ARM64 emulators/devices? Maybe the user who builds this wants to use an x86-64 emulator in Linux/Windows.

jmartinesp (changes requested):
Thanks for the script, it's a lot easier than doing all manually! I found a couple of issues though.

Also, maybe we should check if both `ANDROID_NDK_HOME` is set and > v25?

julioromano:
> Thanks for the script, it's a lot easier than doing it all manually! I found a couple of issues, though.
> 
> Also, maybe we should check if both `ANDROID_NDK_HOME` is set and >= v25?

There's v26 since a while and it compiles correctly with it, should we bother keeping the NDK requirement up to date?

libraries/matrix/impl/build.gradle.kts
bmarty: You mean having something like:

```kotlin
    releaseImplementation(libs.matrix.sdk)
    if (file("${rootDir.path}/libraries/rustsdk/matrix-rust-sdk.aar").exists()) {
        println("\nNote: Using local binary of the Rust SDK.\n")
        debugImplementation(projects.libraries.rustsdk)
    } else {
        debugImplementation(libs.matrix.sdk)
    }
```
?

I am wondering now if we may want to test the release build with a local SDK :)

tools/sdk/build_rust_sdk.sh
bmarty: Yes, of course, good catch!

libraries/matrix/impl/build.gradle.kts
julioromano: Yes. That's what I had in mind. In the rare case we'd like to test a release build with a local sdk  we'll have to manually change the gradle file.
This seems to me an added level of safety. But if you feel I'm being paranoid don't be afraid to say it, sometime I am.

bmarty: Works for me. Better safe than sorry!

jmartinesp (approved):
LGTM, thanks for the changes!

tools/sdk/build_rust_sdk.sh
julioromano: You could use `uname -m` to get the architecture and invoke the correct option.
We could reasonably assume, being this for local use, crosscompilation is not a goal.

So if `uname -m` is x86_64 we build the x86_64 version.
If `uname -m` is amd64 we build the aarch64 version.

julioromano: More hints at: https://askubuntu.com/questions/803596/cpu-architecture-scanning-using-bash

PR #1590, opened 2023-10-17
Make sure Konsist tests always run

jmartinesp:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

This PR adds a workaround so Konsist tests are always run.

## Motivation and context

Konsist tests weren't being run in either the CI or locally using gradle because all the tests were considered up to date. By setting the output as never up to date we make sure they're always run.

## Tests

- If the tests fail with the broken test commit, it's working.

### DIFF
--- a/tests/konsist/build.gradle.kts
+++ b/tests/konsist/build.gradle.kts
@@ -32,3 +32,10 @@ dependencies {
     testImplementation(projects.libraries.architecture)
     testImplementation(projects.libraries.designsystem)
 }
+
+// Make sure Konsist tests are always run. This is needed because otherwise we'd have to either:
+// - Add every single module as a dependency of this one.
+// - Move the Konsist tests to a the app module, which was causing issues.
+tasks.withType<Test>().configureEach {
+    outputs.upToDateWhen { false }
+}
--- a/tests/konsist/build.gradle.kts
+++ b/tests/konsist/build.gradle.kts
@@ -35,7 +35,7 @@ dependencies {
 
 // Make sure Konsist tests are always run. This is needed because otherwise we'd have to either:
 // - Add every single module as a dependency of this one.
-// - Move the Konsist tests to a the app module, which was causing issues.
+// - Move the Konsist tests to the `app` module, but the `app` module does not need to know about Konsist.
 tasks.withType<Test>().configureEach {
     outputs.upToDateWhen { false }
 }


build.gradle.kts
```diff
@@ -374,3 +375,7 @@ subprojects {
         }
     }
 }
+
+//tasks.named("check").configure {
+//    dependsOn(":tests:konsist:test")
+//}
```
jmartinesp, line 381: Oops, this is not needed. I'll remove it when I revert the break konsist commit.

app/src/main/kotlin/io/element/android/x/MainNode.kt
```diff
@@ -49,6 +49,8 @@ class MainNode(
 ),
     DaggerComponentOwner {
 
+    private val mBreakKonsistTest = true
```
ElementBot, line 52: - :warning: Private property `mBreakKonsistTest` is unused.

tests/konsist/build.gradle.kts
```diff
@@ -32,3 +32,8 @@ dependencies {
     testImplementation(projects.libraries.architecture)
     testImplementation(projects.libraries.designsystem)
 }
+
+// Make sure Konsist tests are always run
+tasks.withType<Test>().configureEach {
+    outputs.upToDateWhen { false }
+}
```
bmarty, line 39: Konsist test are not always run because the `konsist` module does not depend on any other modules, except `projects.libraries.architecture` and `projects.libraries.designsystem` (see the lines above) because we need access to some external types.
Maybe clarify why this patch is needed?

bmarty (commented):
Thanks, please see my comment. Once the PR has been cleaned up, it can be merged!

tests/konsist/build.gradle.kts
jmartinesp: Done in https://github.com/vector-im/element-x-android/pull/1590/commits/fb0bf385fc2278cf809b6415f225345324bbbb5f.

jmartinesp:
Added some changes to make sure Konsist tests are only run (and fail) in one flow, I'll remove the offending code later.

tests/konsist/build.gradle.kts
```diff
@@ -32,3 +32,10 @@ dependencies {
     testImplementation(projects.libraries.architecture)
     testImplementation(projects.libraries.designsystem)
 }
+
+// Make sure Konsist tests are always run. This is needed because otherwise we'd have to either:
+// - Add every single module as a dependency of this one.
+// - Move the Konsist tests to a the app module, which was causing issues.
```
bmarty, line 38:
```suggestion
// - Move the Konsist tests to the `app` module, but the `app` module does not need to know about Konsist.
```

bmarty (approved):
Just a quick update of the comment, else LGTM

PR #1592, opened 2023-10-18
Feature/fga/setup crypto for pin

ganfra:
This PR introduces a cryptography library to easily encrypt/decrypt data.
It'll be used to store the pin code encrypted to SharedPreferences, and also be able to provide a `CryptoObject` for the `BiometricPrompt`.

Also introduce `PinCodeManager` to have an entry point to store and verify pin. 

Made some code restructuration too (`pin` module is now called `lockscreen`) sorry for the review :/


libraries/cryptography/impl/src/test/kotlin/io/element/android/libraries/cryptography/impl/AESEncryptionDecryptionServiceTest.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.Test
+import java.security.GeneralSecurityException
+import javax.crypto.KeyGenerator
+
+class AESEncryptionDecryptionServiceTest {
+
+    private val encryptionDecryptionService = AESEncryptionDecryptionService()
+
+    @Test
+    fun given_a_valid_key_then_encrypt_decrypt_work() {
```
jmartinesp, line 31: Nit: these can use backticks and whitespaces when running in JVM, in case you want to change them.

features/lockscreen/impl/src/test/kotlin/io/element/android/features/lockscreen/impl/pin/storage/InMemoryPinCodeStore.kt
```diff
@@ -0,0 +1,59 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.features.lockscreen.impl.pin.storage
+
+class InMemoryPinCodeStore : PinCodeStore {
+
+    private var pinCode: String? = null
+    private var remainingAttempts: Int = 3
```
jmartinesp, line 22: Nit: these could be moved to a constant.

features/lockscreen/impl/src/main/kotlin/io/element/android/features/lockscreen/impl/pin/storage/SharedPreferencesPinCodeStore.kt
```diff
@@ -0,0 +1,95 @@
...
+            putInt(REMAINING_PIN_CODE_ATTEMPTS_KEY, remaining)
+        }
+        remaining
+    }
+
+    override suspend fun resetCounter() = withContext(dispatchers.io) {
+        sharedPreferences.edit {
+            remove(REMAINING_PIN_CODE_ATTEMPTS_KEY)
+        }
+    }
```
jmartinesp, line 86: This should never happen in a normal usage, but I'm wondering if using a `Mutex` here could help ensure we don't end up overwriting the remaining attempts.

jmartinesp (approved):
LGTM, a few comments but feel free to ignore them.

features/lockscreen/impl/src/main/kotlin/io/element/android/features/lockscreen/impl/pin/storage/SharedPreferencesPinCodeStore.kt
ganfra: Yeah I can add this for safety :-)

features/lockscreen/impl/src/test/kotlin/io/element/android/features/lockscreen/impl/pin/storage/InMemoryPinCodeStore.kt
ganfra: This is part of the test code, so I guess it's ok like this

PR #1604, opened 2023-10-19
Include desugaring lib also in library modules

julioromano:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Includes the `coreLibraryDesugaring(libs.android.desugar)` dependency in all modules which use one of our gradle plugins.

## Motivation and context

Right now desugaring is enabled also in library modules but the desugar dependency is not included in those.
This causes some unwanted side effects such as being unable to run compose previews in an emu.

This change will also include the desugar dependency in those libraries.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -203,7 +203,6 @@ dependencies {
     implementation(projects.appnav)
     anvil(projects.anvilcodegen)
 
-    coreLibraryDesugaring(libs.android.desugar)
     implementation(libs.appyx.core)
     implementation(libs.androidx.splash)
     implementation(libs.androidx.core)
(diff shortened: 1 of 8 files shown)


ganfra (approved):
Good catch, thanks!

bmarty (approved):
Thanks. `./gradlew tasks` works again with this.

julioromano:
Unfortunately this change triggers lint errors on the library modules which use java8 apis.

I had to reverse course: Will enable desugaring AND include the desugar dep in all modules.
Apparently there are also some news coming in AGP 8.2.0 about this (see https://issuetracker.google.com/issues/259041363 and https://issuetracker.google.com/issues/203113147).
`./gradlew tasks` still works with this approach.

PR #1606, opened 2023-10-19
Add global `context.cacheDir` provider.

julioromano:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Dagger now provides the app's `cacheDir` when requesting a `@CacheDirectory File`  type.

## Motivation and context

To support some upcoming code that needs the `cacheDir` to be changed during tests.

### DIFF
--- a/libraries/di/src/main/kotlin/io/element/android/libraries/di/CacheDir.kt
+++ b/libraries/di/src/main/kotlin/io/element/android/libraries/di/CacheDir.kt
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) 2022 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.libraries.di
+
+import javax.inject.Qualifier
+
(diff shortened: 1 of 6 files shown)


libraries/di/src/main/kotlin/io/element/android/libraries/di/CacheDir.kt
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.libraries.di
+
+import javax.inject.Qualifier
+
+@Qualifier annotation class CacheDir
```
bmarty, line 21: FTR this is what we had in EA: https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/di/FileQualifiers.kt#L31

```kotlin
@Qualifier
@Retention(AnnotationRetention.RUNTIME)
internal annotation class CacheDirectory
```

Maybe use the same name `CacheDirectory`. Also `@Retention` default value is `AnnotationRetention.RUNTIME` , so I think it's fine not to include it, but maybe @ganfra can confirm this?

bmarty (approved):
LGTM, thanks!

libraries/di/src/main/kotlin/io/element/android/libraries/di/CacheDir.kt
julioromano: I just copied whatever we already had from `@ApplicationContext`
As per the naming whatever you prefer I just found it more "mnemonic" to have `@CacheDir private val cacheDir: File` and know that it will point to `context.cacheDir`.

bmarty: OK, it's fine. The PR is approved, you can merge it :)

julioromano:
Renamed it and added docs to all existing qualifiers.

PR #1607, opened 2023-10-19
Extract more content from audio messages.

julioromano:
`TimelineItemAudioContent`:
- Use `java.time.Duration` instead of milliseconds. This will ease up things in the future because currently milliseconds are sent over the wire but in the future seconds will be sent (as per the stable MSC). Using `Duration` will allow our downstream code to be independent of what's passed over the wire.
- Rename `audioSource` property to `mediaSource` to better match its type.

`AudioMessageType`:
- Add and populate new fields `details` and `isVoiceMessage` to be used by voice messages.


jmartinesp (approved):
LGTM!

PR #1610, opened 2023-10-19
Custom room notification settings list

langleyd:
Resolves https://github.com/vector-im/element-x-android/issues/536
Fixes https://github.com/vector-im/element-x-android/issues/1567 (Mostly, will only be fully corrected by the arrival of  [MSC3934](https://github.com/matrix-org/matrix-spec-proposals/pull/3934))

## Screenshots / GIFs

https://github.com/vector-im/element-x-android/assets/6135282/6801f4bc-504b-4064-b2f9-f1181fcd9ee4


features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/notificationsettings/RoomNotificationSettingsNode.kt
```diff
@@ -48,10 +56,18 @@ class RoomNotificationSettingsNode @AssistedInject constructor(
     @Composable
     override fun View(modifier: Modifier) {
         val state = presenter.present()
-        RoomNotificationSettingsView(
-            state = state,
-            modifier = modifier,
-            onBackPressed = this::navigateUp,
-        )
+        if(inputs.showUserDefinedSettingStyle) {
```
julioromano, line 59: I think the pattern we adopted with appyx tends to associate one presenter with one view only.
This is because we can't easily test the Node class so any logic (like this `if``) you put in the Node class will likely not be covered by unit tests.

A better IMHO solution would have been to pass `inputs.showUserDefinedSettingStyle` as input param to `RoomNotificationSettingsPresenter` and then expose it through the state class.
Then create a super-composable function that holds this if logic.
This way we could cover it with screenshot tests at least.

@ganfra what do you think?

Anyhow, not a blocker right now, I just wanted to share this with others to better understand the usage of our patterns.

julioromano (approved):
LGTM, just a few minors. No blockers.

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/notificationsettings/RoomNotificationSettingsNode.kt
langleyd: Makes sense, moved it to down to view.

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/notificationsettings/RoomNotificationSettingsNode.kt
```diff
@@ -48,10 +65,11 @@ class RoomNotificationSettingsNode @AssistedInject constructor(
     @Composable
     override fun View(modifier: Modifier) {
         val state = presenter.present()
```
ElementBot, line 67: - :no_entry_sign: Suspicious indentation: This is indented but is not continuing the previous expression (`val state = presente`...)

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/notificationsettings/RoomNotificationSettingsNode.kt
```diff
@@ -48,10 +65,11 @@ class RoomNotificationSettingsNode @AssistedInject constructor(
...
-        RoomNotificationSettingsView(
-            state = state,
-            modifier = modifier,
-            onBackPressed = this::navigateUp,
-        )
+            RoomNotificationSettingsView(
```
ElementBot, line 68: - :no_entry_sign: Suspicious indentation: This is indented but is not continuing the previous expression (`val state = presente`...)

PR #1593, opened 2023-10-18
Hide keyboard when exiting the room screen

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Adds a `DisposableEffect` to `MessagesView` so the keyboard is hidden when exiting the chat room screen. It seems to take a bit longer than usual, but at least it's hidden automatically.

## Motivation and context

Fixes #1375.

## Screenshots / GIFs

https://github.com/vector-im/element-x-android/assets/480955/39fd21c5-5d07-49cb-a2ea-c55e68362adf

## Tests

- Open a room.
- Tap on the message composer.
- Navigate to other screen (back, to the room settings, to some user's profile, to view a media file, etc.).

The keyboard should be hidden automatically.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 11

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 23
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/vector-im/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request includes a new file under ./changelog.d. See https://github.com/vector-im/element-x-android/blob/develop/CONTRIBUTING.md#changelog
- [x] Pull request includes screenshots or videos if containing UI changes
- [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
- [x] You've made a self review of your PR

PR #1615, opened 2023-10-20
Maestro No need to close the keyboard after fix from #1593

bmarty:


jmartinesp (approved):
LGTM, sorry for the trouble!

PR #1616, opened 2023-10-20
Document the nuances in `UserId` and `SessionId` types.

julioromano:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [x] Other : Docs

### DIFF
--- a/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/core/SessionId.kt
+++ b/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/core/SessionId.kt
@@ -16,4 +16,7 @@
 
 package io.element.android.libraries.matrix.api.core
 
+/**
+ * The [UserId] of the currently logged in user.
+ */
 typealias SessionId = UserId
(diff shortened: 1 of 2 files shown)


bmarty (approved):
LGTM.

To add some context, `SessionId` is used to identify a session, and the value used is the user MatrixId. In the previous SDK we where concatenating the userId and the deviceId in the SDK to let the SDK be able to log in several times with the same user. We do not need to do that anymore, but we like to keep SessionId distinct from UserId.

The SessionId is used to identify the session, for instance when coming from a notification intent. This is preparatory work for multi session support.

julioromano:
> LGTM.
> 
> To add some context, `SessionId` is used to identify a session, and the value used is the user MatrixId. In the previous SDK we where concatenating the userId and the deviceId in the SDK to let the SDK be able to log in several times with the same user. We do not need to do that anymore, but we like to keep SessionId distinct from UserId.
> 
> The SessionId is used to identify the session, for instance when coming from a notification intent. This is preparatory work for multi session support.

Does this mean in the feature the semantic meaning of `SessionId` will change?

Right now is `MXID` of the current user.
In the future it might become "some string that identifies this session but might not be an MXID" ?

bmarty:
i do not think it will change, I am just giving some context here :)

PR #1623, opened 2023-10-23
Sync Strings

ElementBot:


jmartinesp (approved):
LGTM after some changes to the translations and adding some new screenshots.

PR #1625, opened 2023-10-23
Fix emoji shortcut values

langleyd:
Fixes https://github.com/vector-im/element-x-android/issues/1585

## Content

The emoji shortcut values in the action list should match those that come from emojibase/unicode spec. I believe the existing ones are missing the emoji variation selector codepoint.


bmarty:
Recording screenshot: https://github.com/vector-im/element-x-android/actions/runs/6612914842

ganfra (approved):

PR #1630, opened 2023-10-24
Set tracing of `matrix_sdk_ui::timeline` to `trace`

jmartinesp:
Requested by the Rust team do try to debug the gaps in the timeline reported by some users.


jplatte (approved):
Thanks!

bmarty:
This change also impacts the release version. It was already possible to tweak this tracing in the developer option, in the debug builds. This was not enough?

jplatte:
I don't think we should burden external people with going to developer options when reporting bugs. It's there for us, when we are trying specific reproduction steps. However, the timeline gap bugs have been very hard to reproduce.

bmarty:
OK, fine, thanks!

PR #1637, opened 2023-10-25
Voice message bubble design improvements

julioromano:
- Fixes color and size of both text and icons.


julioromano:
Screenshots: https://github.com/vector-im/element-x-android/actions/runs/6642112874

jmartinesp:
I think something broke in the screenshots? Some screenshots were deleted, others added, other apparently renamed... There's no easy way to diff the changes like this. A couple of days ago the same happened in another PR of mine. Maybe the StateProvider already has too many states and we should think about splitting it?

julioromano:
> I think something broke in the screenshots? Some screenshots were deleted, others added, other apparently renamed... There's no easy way to diff the changes like this. A couple of days ago the same happened in another PR of mine. Maybe the StateProvider already has too many states and we should think about splitting it?
> 
> Could this be caused by #1636?

The changes are indeed there and visible in the screenshots.
Though there is also a lot of noise (as in screenshots that don't belong to something that changed in this PR).
Unfortunately you won't see any waveforms in these sshots because the library we're currently using to generate it doesn't paint on first pass so the screenshots are blank.

PR #1642, opened 2023-10-26
PIN : settings and branch flow

ganfra:
This PR adds the `LockScreenSettingsNode` and branches all the screen to the regular flow.
The feature is now only missing the biometric flow, which will come in the next PR.
Figma : https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=13067-149321&mode=dev


features/lockscreen/impl/src/main/kotlin/io/element/android/features/lockscreen/impl/components/PinEntryTextField.kt
```diff
@@ -111,6 +120,7 @@ internal fun PinEntryTextFieldPreview() {
     ElementPreview {
         PinEntryTextField(
             pinEntry = PinEntry.createEmpty(4).fillWith("12"),
+            isSecured = true,
```
bmarty, line 123: Could be nice to have 2 previews, maybe just wrap in a Column?

bmarty (approved):
LGTM, nice work, thanks!

PR #1648, opened 2023-10-26
Secure backup

bmarty:
*The original big PR has been split to smaller one. This PR now include bug fixes and improvement.*

 
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

First implementation of secure backup:
- add a way to create the recovery key
- add a way to change the recovery key
- add a way to disable the keybackup
- add a way to enable the keybackup
- update the logout flow

Figma: https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?node-id=12998%3A32971&mode=dev

## Motivation and context

Fixes #1574
Fixes #1575
Fixes #1577
Fixes #1578

## Screenshots / GIFs

See screenshot in the PR

Demo (including bugs)

SecureBackupDemo

## Tests

- Login to a session and verify it then go to the new `Chat backup` setting entry and test all the new screens.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 23
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/vector-im/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [ ] Pull request is based on the develop branch
- [ ] Pull request includes a new file under ./changelog.d. See https://github.com/vector-im/element-x-android/blob/develop/CONTRIBUTING.md#changelog
- [ ] Pull request includes screenshots or videos if containing UI changes
- [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
- [ ] You've made a self review of your PR

PR #1651, opened 2023-10-26
Preparatory work for SecureBackup: UI update

bmarty:
Part of #1648: iteration on the design kit.


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/Button.kt
```diff
@@ -335,59 +372,70 @@ private fun ButtonCombinationPreview(
...
-                size = size,
-            )
+            ButtonMatrixPreview(style = style, size = size, destructive = false)
+            ButtonMatrixPreview(style = style, size = size, destructive = true)
         }
     }
 }
 
+@Composable
+private fun ButtonMatrixPreview(
```
ElementBot, line 382: - :warning: Composable functions should only be emitting content into the composition from one source at their top level.

See https://mrmans0n.github.io/compose-rules/rules/#do-not-emit-multiple-pieces-of-content for more information.

ganfra (approved):
LGTM!

PR #1653, opened 2023-10-26
SecureBackup: update matrix sdk module.

bmarty:
SDK part of #1648.

### DIFF
--- a/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/encryption/EncryptionService.kt
+++ b/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/encryption/EncryptionService.kt
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2023 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.libraries.matrix.api.encryption
+
+import kotlinx.coroutines.flow.StateFlow
+
+interface EncryptionService {
+    val backupStateStateFlow: StateFlow<BackupState>
+    val recoveryStateStateFlow: StateFlow<RecoveryState>
+    val backupUploadStateStateFlow: StateFlow<BackupUploadState>
+    val enableRecoveryProgressStateFlow: StateFlow<EnableRecoveryProgress>
+
+    suspend fun enableBackups(): Result<Unit>
+
+    suspend fun isLastDevice(): Result<Boolean>
+
+    /**
+     * Enable recovery. Observe enableProgressStateFlow to get progress and recovery key.
+     */
+    suspend fun enableRecovery(waitForBackupsToUpload: Boolean): Result<Unit>
+
+    /**
+     * Change the recovery and return the new recovery key.
+     */
+    suspend fun resetRecoveryKey(): Result<String>
+
+    suspend fun disableRecovery(): Result<Unit>
+
+    /**
+     * Note: accept bot recoveryKey and passphrase.
+     */
+    suspend fun fixRecoveryIssues(recoveryKey: String): Result<Unit>
+
+    /**
+     * Observe [backupUploadStateStateFlow] to get progress.
+     */
+    suspend fun waitForBackupUploadSteadyState(): Result<Unit>
+}
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/encryption/RustEncryptionService.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/encryption/RustEncryptionService.kt
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) 2023 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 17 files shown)


libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/encryption/EncryptionService.kt
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.libraries.matrix.api.encryption
+
+import kotlinx.coroutines.flow.StateFlow
+
+interface EncryptionService {
+    val backupStateStateFlow: StateFlow<BackupState>
```
julioromano, line 22: Just style:
`StateState` :/
unless we have a very strict naming scheme we want to stick to I'd suggest we could drop the `StateFlow` suffix and just use `Flow` even though it doesn't match the actual type name.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/encryption/EncryptionService.kt
```diff
@@ -0,0 +1,52 @@
...
+    val recoveryStateStateFlow: StateFlow<RecoveryState>
+    val backupUploadStateStateFlow: StateFlow<BackupUploadState>
+    val enableRecoveryProgressStateFlow: StateFlow<EnableRecoveryProgress>
+
+    suspend fun enableBackups(): Result<Unit>
```
julioromano, line 27: Is there any way to disable them?

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/encryption/EncryptionService.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    suspend fun isLastDevice(): Result<Boolean>
```
julioromano, line 29: Missing context: last what? Maybe add docstring

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/encryption/RustEncryptionService.kt
```diff
@@ -0,0 +1,133 @@
...
+            service.enableRecovery(
+                waitForBackupsToUpload = waitForBackupsToUpload,
+                progressListener = object : EnableRecoveryProgressListener {
+                    override fun onUpdate(status: RustEnableRecoveryProgress) {
+                        enableRecoveryProgressStateFlow.value = enableRecoveryProgressMapper.map(status)
+                    }
+                }
+            )
+                // enableRecovery returns the encryption key, but we read it from the state flow
+                .let { }
```
julioromano, line 93: Why `.let{}` when `Unit` was enough ?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/encryption/RustEncryptionService.kt
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    fun destroy() {
+        // No way to remove the listeners...
+        service.destroy()
+    }
+
+    override suspend fun enableBackups(): Result<Unit> = withContext(dispatchers.io) {
+        runCatching {
+            service.enableBackups()
```
julioromano, line 76: Just a question, why none of these ffi methods are `suspend`? Didn't we just get rid of all `*blocking` versions of many APIs?

julioromano (approved):
LGTM, just nits.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/encryption/EncryptionService.kt
bmarty: Yes, I agree it's not ideal... I will change it in a coming PR.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/encryption/RustEncryptionService.kt
bmarty: `Unit` is working, but quality check does not like it.

julioromano: Why doesn't it? It's the most correct thing to do in this case kotlin-wise.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/encryption/EncryptionService.kt
bmarty: Yes, there is one, but the name from the SDK is `disableRecovery`, do not ask me why. I will document the API later.
We have to decide if we stick to the SDK name or if we change it to make is clearer.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/encryption/EncryptionService.kt
bmarty: This is the last Session, so last device that logged in. Will document the API.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/encryption/RustEncryptionService.kt
bmarty: Yeah, I agree. Detekt is reporting this:
> ~/workspaces/element-x-android/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/encryption/RustEncryptionService.kt:84:13: A single Unit expression is unnecessary and can safely be removed. [OptionalUnit]

There are several similar reports in the detekt project https://github.com/detekt/detekt, but all the issues seems to be close.

In the mean time, using `.let { }` is the solution, and it's already used in the codebase.

julioromano: Seems like a detekt bug

PR #1663, opened 2023-10-27
Always treat waveform as List<Float>

julioromano:
[MSC3246](https://github.com/matrix-org/matrix-spec-proposals/pull/3246) specifies the waveform as a list of ints because:

> Because floating point numbers are not allowed in Matrix events

Though DSP on audio data is almost always done using their floating point representation.
This PR brings the float<->int rescaling in the `matrix` module so that the application code can always work with float waveform samples.


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/media/AudioDetails.kt
```diff
@@ -21,10 +21,26 @@ import org.matrix.rustcomponents.sdk.UnstableAudioDetailsContent as RustAudioDet
...
+ */
+internal fun List<UShort>.fromMSC3246range(): List<Float> = map { it.toInt() / 1024f }
+
+/**
+ * Resizes the given [0;1] float list as per unstable MSC3246 spec
+ * to a [0;1024] int list to be used for waveform rendering.
+ *
+ * https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/audio-waveform/proposals/3246-audio-waveform.md
+ */
+internal fun List<Float>.toMSC3246range(): List<UShort> = map { (it / 1024).toInt().toUShort() }
```
bmarty, line 46: I think you want to write `it * 1024` instead of `it / 1024`?

bmarty (approved):
Just a small mistake (please fix before merging!), else LGTM

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/media/AudioDetails.kt
julioromano: Indeed :(

PR #1668, opened 2023-10-27
Enable swipe to reply on voice messages

julioromano:
With the new waveform view it does not conflict anymore.


bmarty (approved):
OK, thanks!

PR #1669, opened 2023-10-27
Default enable voice message feature flag

julioromano:
:party


libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
```diff
@@ -47,7 +47,7 @@ enum class FeatureFlags(
         key = "feature.voicemessages",
         title = "Voice messages",
         description = "Send and receive voice messages",
-        defaultValue = false,
+        defaultValue = true,
```
bmarty, line 50: Have you read the comment at line 21? 😛

bmarty (commented):
This does not enable in production app, if this is what you want to achieve (?)

julioromano:
> This does not enable in production app, if this is what you want to achieve (?)

Fixed, Thanks :)

bmarty (approved):

bmarty:
I forgot to mention it, but maybe add an file for a changelog for that? It will be more visible in the release note when we will do the release.

julioromano:
> I forgot to mention it, but maybe add an file for a changelog for that? It will be more visible in the release note when we will do the release.

https://github.com/vector-im/element-x-android/pull/1674

PR #1670, opened 2023-10-27
Top bar in page

bmarty:
Small PR to be able to add optional Back button in `HeaderFooterPage`.

Should not break any screenshot tests.

### DIFF
--- a/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/pages/HeaderFooterPage.kt
+++ b/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/pages/HeaderFooterPage.kt
@@ -18,21 +18,23 @@ package io.element.android.libraries.designsystem.atomic.pages
 
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.consumeWindowInsets
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.systemBarsPadding
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.unit.dp
-import io.element.android.libraries.designsystem.preview.PreviewsDayNight
 import io.element.android.libraries.designsystem.preview.ElementPreview
+import io.element.android.libraries.designsystem.preview.PreviewsDayNight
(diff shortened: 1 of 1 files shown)


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/TopAppBar.kt
```diff
@@ -42,21 +45,26 @@ fun TopAppBar(
...
+                CompositionLocalProvider(LocalContentColor provides ElementTheme.colors.textActionPrimary) {
+                    actions()
+                }
+            },
+            windowInsets = windowInsets,
+            colors = colors,
+            scrollBehavior = scrollBehavior,
+        )
+    } else {
+        Spacer(modifier = modifier.height(64.0.dp /* TopAppBarSmallTokens.ContainerHeight */))
```
bmarty, line 66: `TopAppBarSmallTokens.ContainerHeight` is internal, so cannot use it.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/pages/HeaderFooterPage.kt
```diff
@@ -40,28 +41,34 @@ import io.element.android.libraries.theme.ElementTheme
 @Composable
 fun HeaderFooterPage(
```
ElementBot, line 42: - :warning: Documentation of HeaderFooterPage is outdated

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/TopAppBar.kt
```diff
@@ -42,21 +45,26 @@ fun TopAppBar(
     actions: @Composable RowScope.() -> Unit = {},
     windowInsets: WindowInsets = TopAppBarDefaults.windowInsets,
     colors: TopAppBarColors = TopAppBarDefaults.topAppBarColors(),
-    scrollBehavior: TopAppBarScrollBehavior? = null
+    scrollBehavior: TopAppBarScrollBehavior? = null,
+    isVisible: Boolean = true,
```
julioromano, line 49: This feels odd.

It shouldn't be the responsibility of any composable to provide a way to hide itself.
If you don't want it to be shown you should wrap it with an `if` just outside the point where you call it.

Maybe I didn't understand your intention here.

bmarty: You're right, I will update remove that.

julioromano (approved):

PR #1680, opened 2023-10-30
Sync Strings

ElementBot:


julioromano:
@bmarty Best that you take this as you might have more context of what's going on.

bmarty (approved):
OK!

PR #1682, opened 2023-10-30
Voice message button: Show proper disabled color.

julioromano:
Also changes a bit our own IconButton api to allow to mirror material's and allow color customization.


julioromano:
Screenshots: https://github.com/vector-im/element-x-android/actions/runs/6691203075

julioromano:
Assigning to @jmartinesp for the change to the designsystem module and to @jonnyandrew for voice messages

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/IconButton.kt
```diff
@@ -38,13 +39,13 @@ fun IconButton(
     onClick: () -> Unit,
     modifier: Modifier = Modifier,
     enabled: Boolean = true,
+    colors: IconButtonColors = IconButtonDefaults.iconButtonColors(
```
jmartinesp, line 42: Technically you could have used a `LocalCompositionProvider` and set the `LocalContentColor` to `iconSecondary` too. But I think it makes sense to be able to customize these here too.

jmartinesp (approved):
LGTM design-system wise.

PR #1691, opened 2023-10-30
Fix scaling of composer UI

jonnyandrew:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Stop scaling icons and other non-text UI with the system font scale.

## Motivation and context

The system provides two mechanisms to scale UI: the font size setting and the display size setting. This change ensures that UI components are not affected by both settings and that text can be scaled independently.

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/textcomposer/impl/src/main/kotlin/io/element/android/libraries/textcomposer/TextComposer.kt
+++ b/libraries/textcomposer/impl/src/main/kotlin/io/element/android/libraries/textcomposer/TextComposer.kt
@@ -51,7 +51,6 @@ import androidx.compose.ui.unit.dp
 import io.element.android.libraries.designsystem.components.media.createFakeWaveform
 import io.element.android.libraries.designsystem.preview.ElementPreview
 import io.element.android.libraries.designsystem.preview.PreviewsDayNight
-import io.element.android.libraries.designsystem.text.applyScaleUp
 import io.element.android.libraries.designsystem.theme.components.CircularProgressIndicator
 import io.element.android.libraries.designsystem.theme.components.Icon
 import io.element.android.libraries.designsystem.theme.components.Text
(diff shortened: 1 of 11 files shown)


bmarty:
The idea of using `applyScaleUp` was to keep coherency of UI when changing the system setting for font scale. Is is still OK after these changes?

jonnyandrew:
> The idea of using `applyScaleUp` was to keep coherency of UI when changing the system setting for font scale. Is is still OK after these changes?

The coherency is affected because the text in the editor can now be increased independently from the surrounding UI icons, buttons etc. But, having looked at other apps including system apps and a few other chat apps it seems this is the way it's supposed to work; the font size setting seems only to affect text whereas the display size can be used to scale non-text UI including icons.

bmarty:
> So although the UI isn't as nice looking, users with a large font scale setting would have this same experience across the whole system.

So users will have the same **bad** experience IMO.

jonnyandrew:
It is bad aesthetically but UX wise but the alternative might be worse: consider a user might have increased their display size so that they are able to touch buttons easily and also increased their font size. In this scenario, our composer buttons are scaled twice and only a little width is remaining for the text input. As a result the input text will be wrapped quite aggressively. I'm not sure if this PR is the right solution but it does improve accessibility in this regard.

julioromano (approved):

jonnyandrew:
@bmarty are you happy for me to merge based on [this rationale](https://github.com/vector-im/element-x-android/pull/1691#issuecomment-1785455628)?

bmarty:
> @bmarty are you happy for me to merge based on [this rationale](https://github.com/vector-im/element-x-android/pull/1691#issuecomment-1785455628)?

Yes, OK, thanks for clarifying!

PR #1695, opened 2023-10-30
[Element Call] Keep MatrixClient alive while the call is working

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- When a call starts in `CallType.RoomCall` mode and is in foreground, we get the `MatrixClient` and start the sync.
- When the call screen is put to background or closed, we stop the sync again.

## Motivation and context

Before this fix we were experiencing a bug that consistently made EC work only 50% of the times it was launched. This happened because the EC Widget needs some data from the Rust SDK that can only be retrieved while the sync is active. 

Since in `LoggedInFlowNode` we stop the sync when it's sent to the background, these events we requested were never received and the call only actually worked if it had some cached membership events from the previous attempt to make a call, working only 50% of the time, after a previous failed attempt to establish a call.

## Tests

- Join a call several times in a row.

If they all work, this is working.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 11

## Checklist

### DIFF
--- a/features/call/src/main/kotlin/io/element/android/features/call/ui/CallScreenPresenter.kt
+++ b/features/call/src/main/kotlin/io/element/android/features/call/ui/CallScreenPresenter.kt
@@ -17,6 +17,7 @@
 package io.element.android.features.call.ui
 
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.getValue
@@ -28,6 +29,7 @@ import androidx.compose.runtime.setValue
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedFactory
 import dagger.assisted.AssistedInject
+import io.element.android.appnav.di.MatrixClientsHolder
 import io.element.android.features.call.CallType
 import io.element.android.features.call.data.WidgetMessage
 import io.element.android.features.call.utils.CallWidgetProvider
@@ -37,10 +39,13 @@ import io.element.android.libraries.architecture.Async
 import io.element.android.libraries.architecture.Presenter
 import io.element.android.libraries.architecture.runCatchingUpdatingState
 import io.element.android.libraries.core.coroutine.CoroutineDispatchers
+import io.element.android.libraries.matrix.api.MatrixClient
+import io.element.android.libraries.matrix.api.sync.SyncState
 import io.element.android.libraries.matrix.api.widget.MatrixWidgetDriver
 import io.element.android.libraries.network.useragent.UserAgentProvider
 import io.element.android.services.toolbox.api.systemclock.SystemClock
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
@@ -56,6 +61,8 @@ class CallScreenPresenter @AssistedInject constructor(
     private val userAgentProvider: UserAgentProvider,
     private val clock: SystemClock,
     private val dispatchers: CoroutineDispatchers,
+    private val matrixClientsHolder: MatrixClientsHolder,
+    private val appCoroutineScope: CoroutineScope,
 ) : Presenter<CallScreenState> {
 
     @AssistedFactory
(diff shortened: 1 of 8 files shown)


features/call/src/main/kotlin/io/element/android/features/call/ui/CallScreenView.kt
```diff
@@ -88,9 +90,20 @@ internal fun CallScreenView(
             },
             onWebViewCreated = { webView ->
                 val interceptor = WebViewWidgetMessageInterceptor(webView)
-                state.eventSink(CallScreeEvents.SetupMessageChannels(interceptor))
+                state.eventSink(CallScreenEvents.SetupMessageChannels(interceptor))
             }
         )
+
+        // We need to restart the Matrix client to get the needed call events
+        if (state.isInWidgetMode) {
```
ganfra, line 98: I find it a bit weird to have the view dictate if it should start or stop the sync, could we not manage this entirely in the Presenter?

jmartinesp: We'd need to use  the `OnLifecycleEvent` there instead, which is possible, but then we need to provide a `LocalLifecycleOwner` in a Molecule test, which is both difficult and kind of weird.

jmartinesp: Or maybe we could just replace it with a `DisposableEffect`, but that way we don't stop syncing when the call screen goes to background, right?

jmartinesp: Actually a `DisposableEffect` might be a better solution in the end. I'm replacing it, thanks!

bmarty (approved):
LGTM, thanks!

features/call/src/main/kotlin/io/element/android/features/call/ui/CallScreenPresenter.kt
```diff
@@ -166,6 +175,36 @@ class CallScreenPresenter @AssistedInject constructor(
         }
     }
 
+    @Composable
+    private fun HandleMatrixClientSyncState() {
+        val coroutineScope = rememberCoroutineScope()
+        DisposableEffect(Unit) {
```
ganfra, line 181: Do we really need the `DisposableEffect`? The coroutineScope might be enough

features/call/src/main/kotlin/io/element/android/features/call/ui/CallScreenPresenter.kt
```diff
@@ -166,6 +175,36 @@ class CallScreenPresenter @AssistedInject constructor(
...
+                        }
+                    }
+                    .collect()
+            }
+            onDispose {
+                // We can't use the local coroutine scope here because it will be disposed before this effect
+                appCoroutineScope.launch {
+                    client.syncService().run {
+                        if (syncState.value == SyncState.Running) {
+                            stopSync()
```
ganfra, line 200: Is there no risk of stopping the sync after it's been restarted inside `LoggedInFlowNode`?

features/call/src/main/kotlin/io/element/android/features/call/ui/CallScreenPresenter.kt
jmartinesp: We could manually stop the sync process when either we hang up the call or we receive a hang up message, but it seems safer to do it automatically when this is disposed.

features/call/src/main/kotlin/io/element/android/features/call/ui/CallScreenPresenter.kt
jmartinesp: `LoggedInFlowNode` already tracks the current state and tries to restart the flow automatically in `LoggedInFlowNode.observeSyncStateAndNetworkStatus`, similar to what I do here.

ganfra: By the way, what happens if the network is lost? (airplane)

jmartinesp: Oh, nice catch. Endless loops of syncService starting.

ganfra: I guess this code could be extracted an reusable : 
```
combine(
                    // small debounce to avoid spamming startSync when the state is changing quickly in case of error.
                    syncService.syncState.debounce(100),
                    networkMonitor.connectivity
                ) { syncState, networkStatus ->
                    Pair(syncState, networkStatus)
                }
                    .collect { (syncState, networkStatus) ->
                        Timber.d("Sync state: $syncState, network status: $networkStatus")
                        if (syncState != SyncState.Running && networkStatus == NetworkStatus.Online) {
                            syncService.startSync()
                        }
                    }
```

ganfra (approved):
Let's iterate later.

features/call/src/main/kotlin/io/element/android/features/call/ui/CallScreenPresenter.kt
jmartinesp: For now I just copied it, we'll see how we can extract this to some other component or extension function.

PR #1700, opened 2023-10-31
Add content description to voice message attachment sheet

julioromano:
This will add "Voice message" as content description of the voice message icon in the attachment bottom sheet (after long pressing on a voice message).

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
@@ -589,6 +589,7 @@ private fun ReplyToContent(
     }
 }
 
+@Composable
 private fun attachmentThumbnailInfoForInReplyTo(inReplyTo: InReplyTo.Ready): AttachmentThumbnailInfo? {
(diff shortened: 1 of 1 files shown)


jmartinesp (approved):
The code LGTM and @bmarty already tested it, so approved!

bmarty (changes requested):
Let me check again before merging

bmarty:
The code for message detail bottom sheet is here: https://github.com/vector-im/element-x-android/blob/develop/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/actionlist/ActionListView.kt#L316

You are modifying the code which handle a replied to item in the timeline.

Also in this case, the rendering is this one:

image

So I do not think it is necessary to add again the string "Voice message" to the icon. A contentDecription `null` should be fine in this case.

julioromano:
> The code for message detail bottom sheet is here: https://github.com/vector-im/element-x-android/blob/develop/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/actionlist/ActionListView.kt#L316
> 
> You are modifying the code which handle a replied to item in the timeline.
> 
> Also in this case, the rendering is this one:
> 
> image
> So I do not think it is necessary to add again the string "Voice message" to the icon. A contentDecription `null` should be fine in this case.

You are right. Will close this.

PR #1702, opened 2023-10-31
Feature/fga/biometric unlock

ganfra:
This PR branches the biometric unlock logic.


app/src/main/kotlin/io/element/android/x/MainActivity.kt
```diff
@@ -42,7 +42,7 @@ import timber.log.Timber
 
 private val loggerTag = LoggerTag("MainActivity")
 
-class MainActivity : NodeComponentActivity() {
+class MainActivity : NodeActivity() {
```
ganfra, line 45: Necessary to have FragmentActivity for Biometric api

features/lockscreen/impl/src/main/kotlin/io/element/android/features/lockscreen/impl/biometric/BiometricUnlock.kt
```diff
@@ -0,0 +1,131 @@
...
+) : BiometricUnlock {
+
+    override val isActive: Boolean = true
+
+    private lateinit var cryptoObject: CryptoObject
+
+    override fun setup() {
+        try {
+            val secretKey = ensureKey()
+            val cipher = encryptionDecryptionService.createEncryptionCipher(secretKey)
```
jmartinesp, line 70: While this will work in general, on EA [we had some nasty issues](https://github.com/vector-im/element-android/issues/7064#issuecomment-1241629631) with 'user not authenticated' errors. To avoid this, one possible workaround may be to create the `Cipher` on demand and have 2 internal authentication methods, since it seems like this is a possible workaround:

1. `authenticateWithCryptoObject`, should be called by default, it will do what's inside `authenticate` right now. If the `createEncryptionCipher` inside this authenticate method fails with the 'user not authenticated error', however, we go to the 2nd method.
2. `authenticateWithoutCryptoObject`, will call the same `prompt.authenticate` method without a `CryptoObject`, which will authenticate the user and get rid of the error, then you should be able to initialize the `Cipher` and use it as you'd usually do.

I'm not saying you should implement this now, but we definitely should if we see those kind of issues coming back on EXA.

features/lockscreen/impl/src/main/kotlin/io/element/android/features/lockscreen/impl/biometric/DefaultBiometricUnlockManager.kt
```diff
@@ -0,0 +1,143 @@
...
+     * Returns true if a strong biometric method (i.e.: fingerprint, some face or iris unlock implementations) can be used.
+     */
+    private val canUseStrongBiometricAuth: Boolean
+        get() = lockScreenConfig.isStrongBiometricsEnabled
+            && biometricManager.canAuthenticate(BiometricManager.Authenticators.BIOMETRIC_STRONG) == BiometricManager.BIOMETRIC_SUCCESS
+
+    /**
+     * Returns true if any biometric method (weak or strong) can be used.
+     */
+    private val canUseBiometricAuth: Boolean
```
jmartinesp, line 79: Maybe we should rename this to 'isAuthenticatorAvailable'?

features/lockscreen/impl/src/main/kotlin/io/element/android/features/lockscreen/impl/storage/LockScreenStore.kt
```diff
@@ -32,6 +34,14 @@ interface PinCodeStore : EncryptedPinCodeStorage {
      * Resets the counter of attempts for PIN code and biometric access.
      */
     suspend fun resetCounter()
-}
 
+    /**
+     * Returns whether the biometric unlock is allowed or not.
+     */
+    fun isBiometricUnlockAllowed(): Flow<Boolean>
```
jmartinesp, line 41: Nit: maybe use `enabled` here instead of `allowed`? Allowed sounds like something coming from the OS to me, rather than from a user preference.

jmartinesp (approved):
LGTM thanks! I just have a few nits and one thing we might either want to implement now or as soon as we get some errors we saw back in EA (which we might avoid if we're lucky). 

Also, if biometric auth is enabled, shouldn't it be prompted by default, instead of the PIN code?

Feel free to ignore the nits or implement them in another PR.

features/lockscreen/impl/src/main/kotlin/io/element/android/features/lockscreen/impl/biometric/BiometricUnlock.kt
ganfra: `user not authenticated error` is something different than `KeyPermanentlyInvalidatedException`?

features/lockscreen/impl/src/main/kotlin/io/element/android/features/lockscreen/impl/storage/LockScreenStore.kt
ganfra: It was to reflects what we show on the ui...

ganfra:
> LGTM thanks! I just have a few nits and one thing we might either want to implement now or as soon as we get some errors we saw back in EA (which we might avoid if we're lucky).
> 
> Also, if biometric auth is enabled, shouldn't it be prompted by default, instead of the PIN code?

This is the case, it automatically opens the Biometric bottom sheet, on top of the `PinUnlockView`

> Feel free to ignore the nits or implement them in another PR.

features/lockscreen/impl/src/main/kotlin/io/element/android/features/lockscreen/impl/biometric/BiometricUnlock.kt
ganfra: Ok I see, otherwise I'll remove the flag for now...

ganfra:
> > This is the case, it automatically opens the Biometric bottom sheet, on top of the PinUnlockView
> 
> Ah, true, it does by default. However, for me it doesn't anymore after:
> 
> 1. I dismiss the biometric prompt.
> 2. Unlock using pin code.
> 3. Move the app to background.
> 4. After the grace period, open the app again.
> 
> It won't work automatically, I need to click on 'use biometric'. It does work fine again after restarting the app.

Got it, will fix, thanks!

features/lockscreen/impl/src/main/kotlin/io/element/android/features/lockscreen/impl/biometric/BiometricUnlock.kt
jmartinesp: If you do that we might have to handle key migrations in the future, so be careful.

ganfra: Will handle that later if we encounter the issue, no time for now :/

jmartinesp: No worries, given the deadlines, I think that's the right call too

PR #1705, opened 2023-10-31
Show "Voice message" in voice message push notifications

julioromano:
Don't show the event body anymore as it's not relevant for voice messages.


jonnyandrew (approved):
Thanks! I'm surprised no test broke after this change?

PR #1714, opened 2023-10-31
Pin unlock : implement design for in-app unlock

ganfra:
Same as https://github.com/vector-im/element-x-android/pull/1707 but fixing conflicts was longer than cherry picking the only relevant commit.


jmartinesp (approved):
LGTM, thanks!

PR #1722, opened 2023-11-02
Fix maestro test.

bmarty:
Fix Maestro, wording for screen title is now the same than the wording for the button, so Maestro was clicking on the title. Adding an id should fix the test, and is safer, since the button can have other wording.

### DIFF
--- a/.maestro/tests/account/logout.yaml
+++ b/.maestro/tests/account/logout.yaml
@@ -10,6 +10,7 @@ appId: ${APP_ID}
     id: "sign-out-submit"
 # Ensure cancel cancels
 - tapOn: "Cancel"
+- tapOn:
+    id: "sign-out-submit"
 - tapOn: "Sign out"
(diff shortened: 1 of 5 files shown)


jmartinesp (dismissed):
I'll approve so we can see how Maestro behaves. Thanks!

bmarty:
Thanks. Not sure that it will be enought since chat backup feature is disabled by default... The test will fail a bit later. Let's see.

jmartinesp:
> Thanks. Not sure that it will be enough since chat backup feature is disabled by default... The test will fail a bit later. Let's see.

I think a few more changes are needed, the 'sign out' button needs to be tapped again, and also the title of the confirmation button inside the dialog is no longer 'sign out anyway'.

jmartinesp (approved):

jmartinesp:
Congrats, it's fixed!

PR #1727, opened 2023-11-02
Only build `app` module for Maestro tests.

jmartinesp:
This should save some time since the sample module would no longer be built before running the tests.

Technically, this could still be improved by trying to download the built artefacts from the last `build.yml` run in the same PR and building the apk otherwise, but that seems like an overkill.


julioromano (approved):

bmarty (approved):
The description `Only build app module for Maestro tests.` scared me: I thought the app module will not be built for other jobs :)

PR #1728, opened 2023-11-02
Move `@Preview...` custom annotations to `:libraries:composeutils` module

jmartinesp:
When they were still in `:libraries:designsystem` Showkase was incorrectly creating some empty previews for them, and this apparently prevented having incremental compilation according to some warnings:

```
w: [ksp] No dependencies are reported for ShowkaseMetadata_showkase_io_element_android_libraries_designsystem_preview_previewwithlargeheight which will prevent incremental compilation.

w: [ksp] No dependencies are reported for ShowkaseMetadata_showkase_io_element_android_libraries_designsystem_preview_previewsdaynight which will prevent
incremental compilation.
```

So we should move them to a module where the Showkase processor is not used, at least until this is fixed.

This should hopefully improve build times, if not now, at least in the near future when we can enable the `kotlin.incremental` option again.

### DIFF
--- a/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/bottomsheet/CustomBottomSheetScaffold.kt
+++ b/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/bottomsheet/CustomBottomSheetScaffold.kt
@@ -191,8 +191,8 @@ private fun CustomBottomSheetScaffoldLayout(
 
             val snackbarOffsetX = (layoutWidth - snackbarPlaceable.width) / 2
             val snackbarOffsetY = when (sheetState.currentValue) {
-                SheetValue.PartiallyExpanded -> sheetOffsetY - snackbarPlaceable.height
-                SheetValue.Expanded, SheetValue.Hidden -> layoutHeight - snackbarPlaceable.height
+                PartiallyExpanded -> sheetOffsetY - snackbarPlaceable.height
+                Expanded, SheetValue.Hidden -> layoutHeight - snackbarPlaceable.height
             }
 
             // Placement order is important for elevation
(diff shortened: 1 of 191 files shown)


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/bottomsheet/CustomBottomSheetScaffold.kt
```diff
@@ -253,9 +253,9 @@ private fun CustomStandardBottomSheet(
             .onSizeChanged { layoutSize ->
                 val newAnchors = calculateAnchors(layoutSize)
                 val newTarget = when (state.anchoredDraggableState.targetValue) {
-                    SheetValue.Hidden, SheetValue.PartiallyExpanded -> SheetValue.PartiallyExpanded
-                    SheetValue.Expanded -> {
-                        if (newAnchors.hasAnchorFor(SheetValue.Expanded)) SheetValue.Expanded else SheetValue.PartiallyExpanded
+                    SheetValue.Hidden, PartiallyExpanded -> PartiallyExpanded
```
bmarty, line 256: `SheetValue.Hidden` has not been touch like the others, maybe add the import like for the others?

bmarty (approved):
I wanted to get ride of this warning since a long time, thank!

Can you also update the import in the [template](https://github.com/vector-im/element-x-android/blob/develop/tools/templates/files/fileTemplates/Template%20Presentation%20Classes.kt.child.2.kt#L11)? thanks.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/bottomsheet/CustomBottomSheetScaffold.kt
jmartinesp: Ah sorry, moving the preview annotations added fully qualified imports everywhere and it seems like I inlined some imports that I didn't mean to while trying to fix this in a single operation.

tools/templates/files/fileTemplates/Template Presentation Classes.kt.child.2.kt
```diff
@@ -8,7 +8,7 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.tooling.preview.PreviewParameter
 import io.element.android.libraries.designsystem.preview.ElementPreview
-import io.element.android.libraries.designsystem.preview.PreviewsDayNight
+import io.element.android.libraries.composeutils.annotations.PreviewsDayNight
```
bmarty, line 11: This has to be reordered, sorry it cannot be done automatically on the template. This line should be moved one line above.

jmartinesp:
It seems like if we remove the showkase KSP processor from the module where these annotations are included, they won't be detected by showkase in any other modules, but if we keep them, then we'll get these warnings... I'll try to find a solution tomorrow and close the PR if there's none.

jmartinesp:
The way Showkase works at the moment, I believe there's no way to avoid this other than not using multi preview annotations. While [this can be avoided](https://github.com/airbnb/Showkase/issues/322#issuecomment-1792030614), it would need either some breaking changes on their side or a fork from ours.

I'm closing this for now, maybe we can reopen it in the future.

PR #1734, opened 2023-11-03
Use Parallel GC and bump RAM usage

jmartinesp:
This seems to significantly reduce the CI times in most jobs.

`-XX:+UseParallelGC` at least seems to reduce my build times locally in 15-20s.


bmarty (approved):
LGTM, thanks!

PR #1738, opened 2023-11-03
Introduce AsyncView to avoid repeating ourselves

bmarty:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Introduce AsyncView to avoid repeating ourself.
Maybe naming it AsyncDialogs would be more appropriate, WDYT?

## Motivation and context

Less copy paste, cleaner code, happier developers.

## Screenshots / GIFs

Should not impact (under recording)

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/async/AsyncView.kt
```diff
@@ -0,0 +1,97 @@
...
+ * closed, [onErrorDismiss] will be invoked. If [onRetry] is not null, a retry button will be displayed.
+ * - When loading, display a loading dialog, if [showProgressDialog] is true, with on optional [progressText].
+ */
+@Composable
+fun <T> AsyncView(
+    async: Async<T>,
+    onSuccess: (T) -> Unit,
+    onErrorDismiss: () -> Unit,
+    modifier: Modifier = Modifier,
+    showProgressDialog: Boolean = true,
```
jmartinesp, line 44: Maybe we could use some defaults for this? Instead of directly using `showProgressDialog`, maybe have:

```kotlin
object AsyncViewDefaults {
    @Composable
    fun ProgressDialog(progressText: String? = null) { ... }

    @Composable
    fun SuccessCallback<T>(callback: (data: T) -> Unit) { 
        LaunchedEffect(Unit) { callback(data) } 
    }
}
```

Then this would be:

```kotlin
fun <T> AsyncView(
    ....
    onProgress: @Composable () -> Unit = { AsyncViewDefaults.ProgressDialog() },
    onSuccess: @Composable (Async.Success<T>) -> Unit = {},
)
```

And could be called as:

```kotlin
AsyncView(
    async = Async<Int>,
    ....
    onProgress = { AsyncViewDefaults.ProgressDialog() }, // Not really needed since it's default, just to showcase usage
    onSuccess = AsyncViewDefaults.SuccessCallback { data ->
        someCallback(data)
    },
)
```

And maybe the same for other `on...` parameters?

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/async/AsyncView.kt
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Render an Async object.
+ * - If Success, invoke the callback [onSuccess], only once.
+ * - If Failure, display a dialog with the error, which can be transformed, using [errorMessage]. When
+ * closed, [onErrorDismiss] will be invoked. If [onRetry] is not null, a retry button will be displayed.
+ * - When loading, display a loading dialog, if [showProgressDialog] is true, with on optional [progressText].
+ */
+@Composable
+fun <T> AsyncView(
```
jmartinesp, line 39: The `AsyncView` name might not be what best reflects what this does? It reminds me of `AsyncImage`, and makes me think of this rendering different composables for success, loading, error (which kinds of does, except for success), but maybe not as dialogs.

Maybe we could use something like `AsyncStateRenderer`, or `(Loading/Progress)StateRenderer`? Or something better than `Renderer` for the suffix (naming is hard 😅).

jmartinesp (commented):
This sounds like a good idea to me, but maybe we can make it more flexible? Let me know your opinion on my suggestions, I must be overengineering this component.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/async/AsyncView.kt
bmarty: Yes, naming is hard here (see in the description: `Maybe naming it AsyncDialogs would be more appropriate, WDYT?`) :)

I choose `AsyncView` since it's the `View` for the `Async`... Maybe @ganfra can suggest something better?

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/async/AsyncView.kt
bmarty: That's sound like a good improvement of the API. I was not sure how to properly have default. Your proposal makes it clearer I think, I will see how it look in the code.

bmarty: Doing it for the ProgressDialog, I end up with this:

image

I believe for this particular case (since there will be no particular UI change) it's better to keep a short code at call site, so keep the existing API. We may iterate if the future prove that I was wrong.

Also for `onSuccess`, I did not manage to find a good solution, and I do not want to spend too much time on it. The existing API again is enought I think. Anyway thanks for the proposal!

I will rebase and generate the screenshot again.

bmarty: I will try to have a complete API and keep the existing code as a shortcut.

PR #1748, opened 2023-11-07
Don't leak MediaFileHandle when downloading voice messages

julioromano:
Uses the new `MediaFile.persist()` [API](https://github.com/matrix-org/matrix-rust-sdk/pull/2789) to cache voice messages.
This allows to close the `MediaFile` handle after use and keeping the file.
Also disables rust sdk caching for voice messages as we'll use the app's cache dir for that purpose.

Fixes: https://github.com/vector-im/element-meta/issues/2175

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/voicemessages/timeline/VoiceMessageMediaRepo.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/voicemessages/timeline/VoiceMessageMediaRepo.kt
@@ -24,7 +24,6 @@ import io.element.android.libraries.di.CacheDirectory
 import io.element.android.libraries.di.RoomScope
 import io.element.android.libraries.matrix.api.media.MatrixMediaLoader
 import io.element.android.libraries.matrix.api.media.MediaSource
-import io.element.android.libraries.matrix.api.media.toFile
 import java.io.File
 
 /**
@@ -90,14 +89,15 @@ class DefaultVoiceMessageMediaRepo @AssistedInject constructor(
             source = mediaSource,
             mimeType = mimeType,
             body = body,
+            useCache = false,
         ).mapCatching {
-            val dest = cachedFile.apply { parentFile?.mkdirs() }
-            // TODO By not closing the MediaFile we're leaking the rust file handle here.
-            // Not that big of a deal but better to avoid it someday.
-            if (it.toFile().renameTo(dest)) {
-                dest
-            } else {
-                error("Failed to move file to cache.")
+            it.use { mediaFile ->
+                val dest = cachedFile.apply { parentFile?.mkdirs() }
+                if (mediaFile.persist(dest.path)) {
+                    dest
+                } else {
+                    error("Failed to move file to cache.")
+                }
             }
         }
     }
--- a/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/media/MatrixMediaLoader.kt
+++ b/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/media/MatrixMediaLoader.kt
@@ -35,7 +35,13 @@ interface MatrixMediaLoader {
      * @param source to fetch the data for.
      * @param mimeType: optional mime type.
      * @param body: optional body which will be used to name the file.
+     * @param useCache: if true, the rust sdk will cache the media in its store.
      * @return a [Result] of [MediaFile]
      */
-    suspend fun downloadMediaFile(source: MediaSource, mimeType: String?, body: String?): Result<MediaFile>
+    suspend fun downloadMediaFile(
+        source: MediaSource,
+        mimeType: String?,
+        body: String?,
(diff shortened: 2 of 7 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/voicemessages/timeline/VoiceMessageMediaRepo.kt
```diff
@@ -88,14 +87,15 @@ class DefaultVoiceMessageMediaRepo @AssistedInject constructor(
...
-                dest
-            } else {
-                error("Failed to move file to cache.")
+            it.use { mediaFile ->
+                val dest = cachedFilePath.apply { parentFile?.mkdirs() }
+                if (mediaFile.persist(dest.path)) {
+                    dest
+                } else {
+                    error("Failed to move file to cache.")
+                }
```
bmarty, line 98: Not from this PR, but I am discovering this file, and if `mxcUri2FilePath` throws Exception, I think the app will crash. Do you agree?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/voicemessages/timeline/VoiceMessageMediaRepo.kt
```diff
@@ -88,14 +87,15 @@ class DefaultVoiceMessageMediaRepo @AssistedInject constructor(
...
-            val dest = cachedFilePath.apply { parentFile?.mkdirs() }
-            // TODO By not closing the MediaFile we're leaking the rust file handle here.
-            // Not that big of a deal but better to avoid it someday.
-            if (it.toFile().renameTo(dest)) {
-                dest
-            } else {
-                error("Failed to move file to cache.")
+            it.use { mediaFile ->
+                val dest = cachedFilePath.apply { parentFile?.mkdirs() }
+                if (mediaFile.persist(dest.path)) {
```
bmarty, line 94: I understand that the File will stay in the cache directory forever (i.e. the app will never delete it, but it can be deleted if the cache dir is emptied), and in the clear format (i.e. not encrypted). This can be seen as a security vulnerability.

The Rust SDK could do the same and decrypt again on demand if necessary. The rational:

> This allows to close the MediaFile handle after use and keeping the file.

is fine, but we need to delete the file at some point.

The SDK was deleting the file if the MediaFile was closed?

bmarty (commented):
Please see my remarks.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/voicemessages/timeline/VoiceMessageMediaRepo.kt
julioromano: > The SDK was deleting the file if the MediaFile was closed?

Yes

julioromano: > This can be seen as a security vulnerability.

EA was emptying this dir on every app startup, we could do the same, or rely on the OS to purge the cache dir when it needs disk space.
We forwarded this question to the PMs though as we'd like to understand what are the security requirements around this feature.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/voicemessages/timeline/VoiceMessageMediaRepo.kt
julioromano: Oh I missed it.
Yeah you're right, will fix it in another PR.

julioromano: https://github.com/vector-im/element-x-android/pull/1756

julioromano: This PR has been rebased on top of #1756 so this bug is not present here anymore.

julioromano:
@bmarty I addressed the bug in a separate PR.
As for the security/privacy implications, this PR doesn't to anything different than the app was already doing (apart from closing a rust object which was instead leaked before).
So I think we could merge this and defer this discussion (which I also would like to end up with a decision) somewhere else.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/voicemessages/timeline/VoiceMessageMediaRepo.kt
julioromano: https://matrix.to/#/!kCCQTCfnABLKGGvQjo:matrix.org/$fwaijR8CYDFv8AM0RxvHrUPpwPCIekxEdoFZrdGVFcw?via=matrix.org&via=element.io&via=flipdot.org

PR #1749, opened 2023-11-07
Fix crash on logout

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Replaces a call to `client.setDelegate(null)` with destroying the `TashHandle` returned by `client.setDelegate` now.

## Motivation and context

Fix a crash that happened after a logout, where the delegate was called even after using `client.setDelegate(null)`.

## Tests

- Log in.
- Log out.
- Wait for 30s-1min.

It if doesn't crash (or if Maestro passes), it's fixed.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 11

## Checklist


julioromano (approved):
LGTM but see my remarks in the chat room.
This looks like a behavior change without an API change.
`setDelegate` has been returning a `TaskHandle` for months.
Perhaps best to discuss with the rust folks.

jmartinesp:
We already discussed this, but I'm posting it here so we have some context of why it's needed. Taken from https://github.com/matrix-org/matrix-rust-sdk/commit/ad5761bfb56e15c92ead256be876183d645599e8: 

> The fix here consists in having `Client::set_delegate` return a task handle that needs to be stashed by the FFI
users, and cancelled when the Client gets out of scope. This fixes the leak, by removing the last reference onto
the Client.

PR #1757, opened 2023-11-07
LockScreen polish

ganfra:
This PR improves a bit `LockScreen` feature by : 
- handling secure flag when the pin is setup (preventing to see the app content in recent app list)
- fixing a navigation issue in `LockScreenSettingsFlowNode`
- improving and fixing race condition on `SetupPin` 

About the secure flag,  adding/removing it dynamically is working on some devices, but maybe won't be working on all. 
It has been tested on 2 samsung devices (Android 10 and 9) and Pixel 7 (Android 14) and was working.
I think it's more acceptable than having to read synchronously if we need to add it before `setContentView`...

### DIFF
--- a/features/ftue/impl/src/test/kotlin/io/element/android/features/ftue/impl/DefaultFtueStateTests.kt
+++ b/features/ftue/impl/src/test/kotlin/io/element/android/features/ftue/impl/DefaultFtueStateTests.kt
@@ -57,6 +57,7 @@ class DefaultFtueStateTests {
         val analyticsService = FakeAnalyticsService()
         val migrationScreenStore = InMemoryMigrationScreenStore()
         val permissionStateProvider = FakePermissionStateProvider(permissionGranted = true)
+        val lockScreenService = FakeLockScreenService()
         val coroutineScope = CoroutineScope(coroutineContext + SupervisorJob())
 
         val state = createState(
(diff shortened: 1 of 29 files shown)


features/ftue/impl/src/test/kotlin/io/element/android/features/ftue/impl/DefaultFtueStateTests.kt
```diff
@@ -108,7 +113,11 @@ class DefaultFtueStateTests {
         steps.add(state.getNextStep(steps.lastOrNull()))
         permissionStateProvider.setPermissionGranted()
 
-        // Fourth step, analytics opt in
+        // Fourth step, notifications opt in
```
bmarty, line 116:
```suggestion
        // Fourth step, entering PIN code
```

features/lockscreen/api/src/main/kotlin/io/element/android/features/lockscreen/api/LockScreenService.kt
```diff
@@ -28,5 +35,34 @@ interface LockScreenService {
...
+    isPinSetup()
+        .onEach { isPinSetup ->
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+                activity.setRecentsScreenshotEnabled(!isPinSetup)
+            } else {
+                if (isPinSetup) {
+                    activity.window.setFlags(
+                        WindowManager.LayoutParams.FLAG_SECURE,
+                        WindowManager.LayoutParams.FLAG_SECURE
+                    )
```
julioromano, line 61:
```suggestion
                    activity.window.addFlags(
                        WindowManager.LayoutParams.FLAG_SECURE
                    )
```

julioromano (approved):

PR #1758, opened 2023-11-07
Enable seeking a recorded voice message

jonnyandrew:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Enable seeking the voice message player after recording a voice message.

## Motivation and context

- Part of https://github.com/vector-im/element-meta/issues/2113

## Screenshots / GIFs

[Screen_recording_20231107_162647.webm](https://github.com/vector-im/element-x-android/assets/4940864/2742bd61-d162-4533-8336-ac27e2eb5aa4)

## Tests

- Record a voice message
- Seek the preview player by tapping the waveform
- Play a timeline voice message
- Try seeking the preview player again in the same way

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 13

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/voicemessages/composer/VoiceMessageComposerPlayer.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/voicemessages/composer/VoiceMessageComposerPlayer.kt
@@ -17,96 +17,233 @@
 package io.element.android.features.messages.impl.voicemessages.composer
 
 import io.element.android.libraries.mediaplayer.api.MediaPlayer
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.cancelAndJoin
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.scan
+import kotlinx.coroutines.launch
+import timber.log.Timber
 import javax.inject.Inject
 
 /**
  * A media player for the voice message composer.
  *
  * @param mediaPlayer The [MediaPlayer] to use.
+ * @param coroutineScope
  */
 class VoiceMessageComposerPlayer @Inject constructor(
     private val mediaPlayer: MediaPlayer,
+    private val coroutineScope: CoroutineScope,
 ) {
-    private var lastPlayedMediaPath: String? = null
-    private val curPlayingMediaId
-        get() = mediaPlayer.state.value.mediaId
+    companion object {
+        const val MIME_TYPE = "audio/ogg"
+    }
 
-    val state: Flow<State> = mediaPlayer.state.map { state ->
-        if (lastPlayedMediaPath == null || lastPlayedMediaPath != state.mediaId) {
-            return@map State.NotLoaded
+    private var mediaPath: String? = null
+
+    private var seekJob: Job? = null
+    private val seekingTo = MutableStateFlow<Float?>(null)
+
+    val state: Flow<State> = combine(mediaPlayer.state, seekingTo) { state, seekingTo ->
+        state to seekingTo
+    }.scan(InternalState.NotLoaded) { prevState, (state, seekingTo) ->
+        if (mediaPath == null || mediaPath != state.mediaId) {
+            return@scan InternalState.NotLoaded
         }
 
-        State(
-            isPlaying = state.isPlaying,
+        InternalState(
+            playState = calcPlayState(prevState.playState, seekingTo, state),
             currentPosition = state.currentPosition,
-            duration = state.duration ?: 0L,
+            duration = state.duration,
+            seekingTo = seekingTo,
+        )
+    }.map {
+        State(
+            playState = it.playState,
+            currentPosition = it.currentPosition,
+            progress = calcProgress(it),
         )
     }.distinctUntilChanged()
 
     /**
+     * Set the voice message to be played.
+     */
+    suspend fun setMedia(mediaPath: String) {
+        this.mediaPath = mediaPath
+        mediaPlayer.setMedia(
+            uri = mediaPath,
+            mediaId = mediaPath,
+            mimeType = MIME_TYPE,
+        )
+    }
+
+    /**
      * Start playing from the current position.
      *
-     * @param mediaPath The path to the media to be played.
-     * @param mimeType The mime type of the media file.
+     * Call [setMedia] before calling this method.
      */
-    suspend fun play(mediaPath: String, mimeType: String) {
-        if (mediaPath == curPlayingMediaId) {
-            mediaPlayer.play()
-        } else {
-            lastPlayedMediaPath = mediaPath
-            mediaPlayer.setMedia(
-                uri = mediaPath,
-                mediaId = mediaPath,
-                mimeType = mimeType,
-            )
-            mediaPlayer.play()
+    suspend fun play() {
+        val mediaPath = this.mediaPath
+        if (mediaPath == null) {
+            Timber.e("Set media before playing")
+            return
         }
+
+        mediaPlayer.ensureMediaReady(mediaPath)
+
+        mediaPlayer.play()
     }
 
     /**
      * Pause playback.
      */
     fun pause() {
-        if (lastPlayedMediaPath == curPlayingMediaId) {
+        if (mediaPath == mediaPlayer.state.value.mediaId) {
             mediaPlayer.pause()
         }
     }
 
+    /**
+     * Seek to a given position in the current media.
+     *
+     * Call [setMedia] before calling this method.
+     *
+     * @param position The position to seek to between 0 and 1.
+     */
+    suspend fun seek(position: Float) {
+        val mediaPath = this.mediaPath
+        if (mediaPath == null) {
+            Timber.e("Set media before seeking")
+            return
+        }
+
+        seekJob?.cancelAndJoin()
+        seekingTo.value = position
+        seekJob = coroutineScope.launch {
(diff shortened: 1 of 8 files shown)


features/messages/impl/src/test/kotlin/io/element/android/features/messages/voicemessages/composer/VoiceMessageComposerPresenterTest.kt
```diff
@@ -323,6 +340,7 @@ class VoiceMessageComposerPresenterTest {
             awaitItem().eventSink(VoiceMessageComposerEvents.PlayerEvent(VoiceMessagePlayerEvent.Play))
             awaitItem().eventSink(VoiceMessageComposerEvents.SendVoiceMessage)
             assertThat(awaitItem().voiceMessageState).isEqualTo(aPlayingState().toSendingState())
+            skipItems(1) // Duplicate sending state
```
jonnyandrew, line 343: I needed to add this after [this change](https://github.com/vector-im/element-x-android/pull/1758/files#diff-3f2fbb85a53da92f3e329901ee9c3e3710c53c453b1d816ca18fc02f28e88217R57) but couldn't get to the bottom of why. The `state` flow still emits the exact same values in this test.

libraries/mediaplayer/impl/src/main/kotlin/io/element/android/libraries/mediaplayer/impl/MediaPlayerImpl.kt
```diff
@@ -76,12 +89,16 @@ class MediaPlayerImpl @Inject constructor(
...
+    private val _state = MutableStateFlow(MediaPlayer.State(MediaPlayer.PlayState.Stopped, null, 0L, 0L))
 
     override val state: StateFlow<MediaPlayer.State> = _state.asStateFlow()
 
-    override fun acquireControlAndPlay(uri: String, mediaId: String, mimeType: String) {
+    override fun setMedia(uri: String, mediaId: String, mimeType: String, playWhenReady: Boolean) {
+        player.playWhenReady = false
         player.clearMediaItems()
+        _state.update {
+            it.copy(playState = MediaPlayer.PlayState.Stopped)
```
julioromano, line 100: What is the difference between being Stopped and being Paused?
From the point of view of ExoPlayer there is no difference (we're still calling `prepare()` afterwards so the ExoPlayer will not be stopped, because stop means `STATE_IDLE`), so I feel we're introducing this `Stopped` state because of a need of `VoiceMessageComposerPlayer` rather than exposing an actual state of the ExoPlayer.

libraries/mediaplayer/impl/src/main/kotlin/io/element/android/libraries/mediaplayer/impl/MediaPlayerImpl.kt
```diff
@@ -76,12 +89,16 @@ class MediaPlayerImpl @Inject constructor(
...
     private var job: Job? = null
 
-    private val _state = MutableStateFlow(MediaPlayer.State(false, null, 0L, 0L))
+    private val _state = MutableStateFlow(MediaPlayer.State(MediaPlayer.PlayState.Stopped, null, 0L, 0L))
 
     override val state: StateFlow<MediaPlayer.State> = _state.asStateFlow()
 
-    override fun acquireControlAndPlay(uri: String, mediaId: String, mimeType: String) {
+    override fun setMedia(uri: String, mediaId: String, mimeType: String, playWhenReady: Boolean) {
+        player.playWhenReady = false
```
julioromano, line 97: I don't understand why we would need to set play when ready as false here.

julioromano (commented):
Sorry if I'm being pedantic with changes to `MediaPlayer`.
The intent is to try to keep it as slim and simple as possible because that's the part which interfaces with the ExoPlayer beast which is missing tests.
So the more stuff we add here the more undetected issues we might have in the future.
We should probably think of building some tests so we can more freely enhance MediaPlayer but it is potentially cumbersome (currently in the tech debt list).

libraries/mediaplayer/impl/src/main/kotlin/io/element/android/libraries/mediaplayer/impl/SimplePlayer.kt
```diff
@@ -67,6 +70,7 @@ class SimplePlayerImpl(
         p.addListener(object : Player.Listener {
             override fun onIsPlayingChanged(isPlaying: Boolean) = listener.onIsPlayingChanged(isPlaying)
             override fun onMediaItemTransition(mediaItem: MediaItem?, reason: Int) = listener.onMediaItemTransition(mediaItem)
+            override fun onTimelineChanged(timeline: Timeline, reason: Int) = listener.onTimelineChanged(timeline)
```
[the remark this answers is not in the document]
jonnyandrew: The reason for adding this is to allow seeking the track before it is played which requires the track duration.

At the point `onMediaItemTransition` is called,  the track duration is not available (the value is an 'unset' error code). Previously we only recieved the track duration after playing the track. I added `onTimelineChanged` because it seemed like the most relevant callback and, at the point it's called, the duration is available.

libraries/mediaplayer/impl/src/main/kotlin/io/element/android/libraries/mediaplayer/impl/SimplePlayer.kt
```diff
@@ -77,6 +81,10 @@ class SimplePlayerImpl(
     override val duration: Long
         get() = p.duration
 
+    override var playWhenReady: Boolean
+        get() = p.playWhenReady
+        set(value) { p.playWhenReady = value }
```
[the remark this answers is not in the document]
jonnyandrew: I added this as I felt the API was a bit more explicit (because `play()` only works when ready). But on second thought, I think it goes against the philosophy of the 'simple' player interface! I'll remove.

About the statefulness, FYI [the docs](https://developer.android.com/reference/androidx/media3/common/Player#play()) mention that `play()` and `setPlayWhenReady(true)` are equivalent so I assume `play()` is also stateful. Not sure if that changes any of our assumptions?

libraries/mediaplayer/impl/src/main/kotlin/io/element/android/libraries/mediaplayer/impl/MediaPlayerImpl.kt
jonnyandrew: I added this to make it explicit that calling `setMedia` will pause the playback straight away. I suppose `clearMediaItems` on the following line probably does this implicitly, so I can remove it if it doesn't add any value.

libraries/mediaplayer/impl/src/main/kotlin/io/element/android/libraries/mediaplayer/impl/SimplePlayer.kt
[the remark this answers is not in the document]
jonnyandrew: Removed in https://github.com/vector-im/element-x-android/pull/1758/commits/3febc21abf99daed01e0fa4f494ee2340d85c4ae

libraries/mediaplayer/impl/src/main/kotlin/io/element/android/libraries/mediaplayer/impl/MediaPlayerImpl.kt
jonnyandrew: The `Stopped` state models the state where the player has not yet been played or paused, the distinction being:
- in the `Stopped` state, the player should display the total duration and not display a cursor
- in the `Paused` state, the player should instead display the current player position and a cursor

I added this to the media player itself as I thought other parts of the app will likely want to share this behaviour. But agree that it might be a UI concern and could be a bit confusing for the model to be out of sync with the underlying media player.

I'll see if we can move this out into the presentation layer.

libraries/voicerecorder/api/src/main/kotlin/io/element/android/libraries/voicerecorder/api/VoiceRecorderState.kt
```diff
@@ -16,9 +16,11 @@
 
 package io.element.android.libraries.voicerecorder.api
 
+import androidx.compose.runtime.Immutable
 import java.io.File
 import kotlin.time.Duration
 
+@Immutable
```
julioromano, line 23: Just for my info: What does this do? I mean, what wasn't working well before that prompted you to add this?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/voicemessages/composer/VoiceMessageComposerPlayer.kt
```diff
@@ -17,96 +17,233 @@
...
+        }.apply {
+            invokeOnCompletion {
+                seekingTo.value = null
+            }
+        }
+    }
+
+    private suspend fun MediaPlayer.ensureMediaReady(mediaPath: String): MediaPlayer.State {
+        val state = state.value
+        if (state.mediaId == mediaPath && state.isReady) {
```
julioromano, line 141: Sorry about not checking for this properly in the MediaPlayer. I've opened this: https://github.com/vector-im/element-x-android/pull/1783

julioromano (approved):

libraries/voicerecorder/api/src/main/kotlin/io/element/android/libraries/voicerecorder/api/VoiceRecorderState.kt
jonnyandrew: IIRC it was flagged as a problem when I added `VoiceRecorderState` as a parameter to a composable function [here](https://github.com/vector-im/element-x-android/pull/1758/files#diff-982ecb7d870bf285bd89cdd0681db9f1d98214b4c69a3600591d4e2f8be3bf02R204). It's a class from another module that wasn't built with the compose compilerso the stability couldn't be inferred (see [here](https://developer.android.com/jetpack/compose/performance/stability/fix#multiple-modules) for more info).

PR #1766, opened 2023-11-08
Lower max RAM usage for gradle to avoid CI issues

jmartinesp:
Some CI jobs are failing, presumably due to high RAM usage. Let's see if this fixes it.


bmarty (approved):
LGTM, thanks!

PR #1780, opened 2023-11-09
Allow to seek a voice message before playing it

julioromano:
Previously the only way to start interacting with a voice message was to play it. This change makes it possible to start interacting also with a seek.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/voicemessages/timeline/VoiceMessagePlayer.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/voicemessages/timeline/VoiceMessagePlayer.kt
@@ -17,10 +17,10 @@
 package io.element.android.features.messages.impl.voicemessages.timeline
 
 import com.squareup.anvil.annotations.ContributesBinding
-import io.element.android.libraries.mediaplayer.api.MediaPlayer
 import io.element.android.libraries.di.RoomScope
 import io.element.android.libraries.matrix.api.core.EventId
 import io.element.android.libraries.matrix.api.media.MediaSource
+import io.element.android.libraries.mediaplayer.api.MediaPlayer
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.map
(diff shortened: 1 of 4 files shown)


jonnyandrew (approved):
Code looks good!

Just a few issues I noticed while testing which may or may not be related to the changes:
- The progress bar flickers down on first seeking. This video shows it but you may need to play it frame by frame to see:

[Screencast from 10-11-23 11:32:21.webm](https://github.com/vector-im/element-x-android/assets/4940864/df35ab81-3567-43de-bfc1-d5c82ef1b4b4)

- After clearing the app cache to test, I couldn't play any voice messages:

[Screencast from 10-11-23 11:40:57.webm](https://github.com/vector-im/element-x-android/assets/4940864/e0be6889-630b-4262-a474-a79d19a72e8e)

julioromano:
> * After clearing the app cache to test, I couldn't play any voice messages:

Could you please double check this after the latest rebase? I've just cleared the cache dir twice and this didn't happen.

jonnyandrew:
I just tested and can still reproduce at https://github.com/vector-im/element-x-android/pull/1780/commits/1682b0bbc1b128e73f04a684eacb2ed7b9c82f3a but actually, I can also repro on develop.

The steps I'm using to reproduce are:
1. Open a room and play several voice messages
2. Put the app in the background
3. Clear the cache
4. Resume the app and try to play the same voice messages

So I don't think it should block this PR. I'll raise a separate bug about it.

julioromano:
> I just tested and can still reproduce at [1682b0b](https://github.com/vector-im/element-x-android/commit/1682b0bbc1b128e73f04a684eacb2ed7b9c82f3a) but actually, I can also repro on develop so I don't think it should block this PR.
> 
> I raised a separate bug about it with steps to reproduce: #1788

After seeing exactly what you do in the video I can reproduce.

julioromano:
> * The progress bar flickers down on first seeking. This video shows it but you may need to play it frame by frame to see:

This is because the call to `VoiceMessagePlayer.seekTo()` is `suspend` due to `MediaPlayer.setMedia()` being `suspend` in turn.
We must wait for the media player to be ready before we can call seek on it so that it actually updates the player progress.
In the time until the player is ready the progress will still show 0 hence the flicker.
If you insert an artificial delay in the first line of `MediaPlayer.setMedia()` you can see this glitch even more prominently.

jonnyandrew (approved):

julioromano:
@jonnyandrew waiting to merge this because I'd like https://github.com/vector-im/element-x-android/pull/1795 to land before it.

PR #1782, opened 2023-11-09
Upgrade kover

bmarty:
Migrate to Kover 0.7.5

Report are now located in the `:app` module.

Verify and Html report tested OK locally.
First commits are just removing old empty modules.


app/build.gradle.kts
```diff
@@ -194,6 +194,180 @@ knit {
...
+                ":app",
+                ":samples",
+                ":anvilannotations",
+                ":anvilcodegen",
+                ":samples:minimal",
+                ":tests:testutils",
+                // Exclude `:libraries:matrix:impl` module, it contains only wrappers to access the Rust Matrix
+                // SDK api, so it is not really relevant to unit test it: there is no logic to test.
+                ":libraries:matrix:impl",
+                )
```
ElementBot, line 220: - :no_entry_sign: Unexpected indentation (16) (should be 12)

bmarty:
It seems that we could give a try to 0.7.5.

bmarty:
OK, I have now the same error than when running locally:

> Caused by: kotlinx.kover.gradle.plugin.commons.KoverVerificationException: Rule 'Global minimum code coverage.' violated: instructions covered percentage is 23.521500, but expected minimum is 65

Maybe we could just remove the `excludeTests` block (trying right now to see the impact).

jmartinesp:
> OK, I have now the same error than when running locally:
> 
> > Caused by: kotlinx.kover.gradle.plugin.commons.KoverVerificationException: Rule 'Global minimum code coverage.' violated: instructions covered percentage is 23.521500, but expected minimum is 65
> 
> Maybe we could just remove the `excludeTests` block (trying right now to see the impact).

I think it's not taken into account any screenshot test, it's the same issue we had before. All views have 0% of coverage.

app/build.gradle.kts
```diff
@@ -190,6 +191,221 @@ knit {
...
+            annotatedBy(
+                "io.element.android.libraries.designsystem.preview.PreviewsDayNight",
+                "io.element.android.libraries.designsystem.preview.PreviewWithLargeHeight",
+            )
+        }
+    }
+
+    defaults {
+        // add reports of 'release' Android build variant to default reports
+        mergeWith("release")
```
jmartinesp, line 301: This seems to fix the wrong coverage, it wasn't taking into account any test result from the debug runs or screenshot tests:

```suggestion
        // add reports of both 'debug' and 'release' Android build variant to default reports
        mergeWith("debug")
        mergeWith("release")
```

jmartinesp (approved):
LGTM, let's wait for the new test run to finish and check that codecov is fine before merging.

PR #1786, opened 2023-11-10
[TRY_2] Let TimelineItemPresenter persist state across timeline's LazyColumn item lifecycles

julioromano:
Another attempt, it differs from TRY_1 in that instead of preserving in memory the states in a dedicated class, we cache the presenter objects and move outside of the `present()` function the `MutableState`s we want to keep in between recompositions.

TRY_1 at: https://github.com/vector-im/element-x-android/pull/1743


julioromano:
Superseded by https://github.com/vector-im/element-x-android/pull/1795

PR #1795, opened 2023-11-13
Persist state of VoiceMessagePresenter in memory

julioromano:
Allows [VoiceMessagePresenter] instances to keep their progress and download states while going in and out of the timeline viewport.

This is implemented by caching each instance of a TimelineItem presenter inside the RoomScope. TimelineItem presenters can move some of their state outside of the `present()` function so that such state will survive scrollings of the timeline.


bmarty (approved):
OK, thanks!

PR #1799, opened 2023-11-14
Integrate mentions in the composer

### LINKED ISSUE #1453: Mentions: @room mention pill
From element-x-ios created by [Velin92](https://github.com/Velin92): vector-im/element-x-ios#1829  

This is a special case, but the string @room should be interpreted as a special mention case and display a pill with probably its own different design.
Also it should display the room avatar
It does not need to have any tapping or link behaviour however

jmartinesp:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Integrates a new version of the RTE library with better support for mentions.
- Sets a callback in the mention suggestion selection view to actually insert the selection.
- Creates `MentionSpanProvider` to provide the RTE composer with the custom `ReplacementSpan` for mentions.

## Motivation and context

Closes #1453 

## Screenshots / GIFs

|Light|Dark|
|-|-|
|image|image|

## Tests

- Open a room, type `@` in the composer.
- Select something from the list of suggestions.
- A mention pill should be added to the editor and when sending, it'll be recognised as an actual mention.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 11

## Checklist


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesState.kt
```diff
@@ -28,6 +28,7 @@ import io.element.android.libraries.architecture.Async
 import io.element.android.libraries.designsystem.components.avatar.AvatarData
 import io.element.android.libraries.designsystem.utils.snackbar.SnackbarMessage
 import io.element.android.libraries.matrix.api.core.RoomId
+import io.element.android.libraries.matrix.api.core.UserId
```
ElementBot, line 31: - :warning: The import 'io.element.android.libraries.matrix.api.core.UserId' is unused.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerView.kt
```diff
@@ -30,6 +31,7 @@ import io.element.android.features.messages.impl.voicemessages.composer.VoiceMes
 import io.element.android.features.messages.impl.voicemessages.composer.aVoiceMessageComposerState
 import io.element.android.libraries.designsystem.preview.ElementPreview
 import io.element.android.libraries.designsystem.preview.PreviewsDayNight
+import io.element.android.libraries.matrix.api.core.UserId
```
ElementBot, line 34: - :warning: The import 'io.element.android.libraries.matrix.api.core.UserId' is unused.

libraries/textcomposer/impl/src/main/kotlin/io/element/android/libraries/textcomposer/mentions/MentionSpanProvider.kt
```diff
@@ -0,0 +1,78 @@
...
+    @SuppressLint("ComposableNaming")
+    @Composable
+    fun setup() {
+        currentUserTextColor = ElementTheme.colors.currentUserMentionPillText.toArgb()
+        currentUserBackgroundColor = ElementTheme.colors.currentUserMentionPillBackground.toArgb()
+        otherTextColor = ElementTheme.colors.mentionPillText.toArgb()
+        otherBackgroundColor = ElementTheme.colors.mentionPillBackground.toArgb()
+    }
+
+    fun getMentionSpanFor(text: String, url: String): MentionSpan {
```
ElementBot, line 52: - :warning: Function parameter `text` is unused.

frebib:
This works wonderfully! Nice work
There seems to be a small delay after hitting the @ symbol. Is that intended?

jmartinesp:
> This works wonderfully! Nice work There seems to be a small delay after hitting the @ symbol. Is that intended?

Thanks for your comment, I hadn't noticed it myself, but I just realised the code is not working as expected: it's supposed to return the first 100 matching results for the mention, but it was actually returning *all* possible results, so in large rooms it could be quite slow. I just changed that.

libraries/core/src/main/kotlin/io/element/android/libraries/core/data/FilterUpTo.kt
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Returns a list containing first [count] elements matching the given [predicate].
+ * If the list contains less elements matching the [predicate], then all of them are returned.
+ *
+ * @param count the maximum number of elements to take.
+ * @param predicate the predicate used to match elements.
+ * @return a list containing first [count] elements matching the given [predicate].
+ */
+inline fun <T> Iterable<T>.filterUpTo(count: Int, predicate: (T) -> Boolean): List<T> {
```
ElementBot, line 27: - :warning: Documentation of filterUpTo is outdated

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
```diff
@@ -411,7 +427,7 @@ class MessageComposerPresenter @Inject constructor(
 }
 
 @Immutable
-sealed interface RoomMemberSuggestion {
-    data object Room : RoomMemberSuggestion
-    data class Member(val roomMember: RoomMember) : RoomMemberSuggestion
+sealed interface MentionSuggestion {
```
ganfra, line 430: I'd extract this class outs of this file

libraries/textcomposer/impl/src/main/kotlin/io/element/android/libraries/textcomposer/mentions/MentionSpanProvider.kt
```diff
@@ -0,0 +1,116 @@
...
+    private val currentSessionId: SessionId,
+    private var currentUserTextColor: Int = 0,
+    private var currentUserBackgroundColor: Int = Color.WHITE,
+    private var otherTextColor: Int = 0,
+    private var otherBackgroundColor: Int = Color.WHITE,
+) {
+
+    @Suppress("ComposableNaming")
+    @Composable
+    fun setup() {
```
ganfra, line 50: make private?

ganfra (approved):
Some small remarks otherwise LGTM

libraries/textcomposer/impl/src/main/kotlin/io/element/android/libraries/textcomposer/mentions/MentionSpanProvider.kt
jmartinesp: > make private?

I can make it `internal` at best, since it's used in the `rememberMentionSpanProvider` function.

> Is this method called each time composer  is recomposed?

I think so, yeah. Since those colors don't change I'm not sure if they will be re-calculated and re-assigned to each property, but I don't see any other way to be able to reach the `@Composable` color: we can't use `remember` or `LaunchedEffect` for that.

frebib:
> > This works wonderfully! Nice work There seems to be a small delay after hitting the @ symbol. Is that intended?
> 
> Thanks for your comment, I hadn't noticed it myself, but I just realised the code is not working as expected: it's supposed to return the first 100 matching results for the mention, but it was actually returning _all_ possible results, so in large rooms it could be quite slow. I just changed that.

Interestingly (prior to your changes 8d7f24e) it took about 1 second to open both in a group chat with a few participants (~6 iirc) as well as Matrix HQ

jmartinesp:
> Interestingly (prior to your changes [8d7f24e](https://github.com/vector-im/element-x-android/commit/8d7f24e05b40265f0b802f657689650e9b1f6e06)) for me it consistently takes around 1 second to open both in a group chat with a few participants (~6 iirc) as well as Matrix HQ

It's probably the `.debounce(0.5.seconds)` used to prevent too much computation when the user types really fast. Now that I'm aware of it the delay is quite noticeable, maybe it would be better to just remove this debounce so displaying the results is as immediate as possible.

jmartinesp:
> > Interestingly (prior to your changes [8d7f24e](https://github.com/vector-im/element-x-android/commit/8d7f24e05b40265f0b802f657689650e9b1f6e06)) for me it consistently takes around 1 second to open both in a group chat with a few participants (~6 iirc) as well as Matrix HQ
> 
> It's probably the `.debounce(0.5.seconds)` used to prevent too much computation when the user types really fast. Now that I'm aware of it the delay is quite noticeable, maybe it would be better to just remove this debounce so displaying the results is as immediate as possible.

This should now work better: the initial time to display when using `@` will be as short as possible with no artificial delays, then when we want to filter out some more by continue typing, there will be slightly shorter debounce than before (0.3s between search runs).

PR #1806, opened 2023-11-14
LockScreen : rework LoggedInFlowNode and Back management

ganfra:
What the PR solves : previously, when the LockScreen was shown, it was replacing the View of `LoggedInFlowNode`, so all the Children.Views were removed from the composition. It was causing troubles with flow like attach files or take and send picture.

Why it was done like that : because of how the `BackHandler` is managed by Appyx on the `Backstack`, the children were receiving the onBackPressed from the activity before I could intercept them. So, even if the app was locked it'd pop the backstack instead of putting the app in background.

The fix : Do not remove the Children from the composition, but instead block the back management directly in the `MainActivity`.


app/src/main/kotlin/io/element/android/x/MainActivity.kt
```diff
@@ -61,6 +61,20 @@ class MainActivity : NodeActivity() {
         }
     }
 
+    @Deprecated("")
+    override fun onBackPressed() {
+        // If the app is locked, we need to intercept onBackPressed before it goes to OnBackPressedDispatcher.
+        // Indeed, otherwise we would need to trick Appyx backstack management everywhere.
+        // Without this trick, we would get pop operations on the hidden backstack.
+        if (appBindings.lockScreenService().isLocked) {
+            //Do not kill the app in this case, just go to background.
```
bmarty, line 70:
```suggestion
            // Do not kill the app in this case, just go to background.
```

bmarty (approved):
Excellent, thanks!

PR #1807, opened 2023-11-15
Delete media caches on startup

julioromano:
Clear media caches on app startup

### DIFF
--- a/app/src/main/kotlin/io/element/android/x/ElementXApplication.kt
+++ b/app/src/main/kotlin/io/element/android/x/ElementXApplication.kt
@@ -18,9 +18,8 @@ package io.element.android.x
 
 import android.app.Application
 import androidx.startup.AppInitializer
-import io.element.android.libraries.architecture.bindings
+import io.element.android.features.cachecleaner.api.CacheCleanerInitializer
 import io.element.android.libraries.di.DaggerComponentOwner
-import io.element.android.x.di.AppBindings
 import io.element.android.x.di.AppComponent
 import io.element.android.x.di.DaggerAppComponent
 import io.element.android.x.info.logApplicationInfo
@@ -36,8 +35,8 @@ class ElementXApplication : Application(), DaggerComponentOwner {
         AppInitializer.getInstance(this).apply {
             initializeComponent(CrashInitializer::class.java)
             initializeComponent(TracingInitializer::class.java)
+            initializeComponent(CacheCleanerInitializer::class.java)
         }
         logApplicationInfo()
-        bindings<AppBindings>().cacheCleaner().clearCache()
     }
 }
--- a/app/src/main/kotlin/io/element/android/x/ElementXApplication.kt
+++ b/app/src/main/kotlin/io/element/android/x/ElementXApplication.kt
@@ -18,7 +18,9 @@ package io.element.android.x
 
 import android.app.Application
 import androidx.startup.AppInitializer
+import io.element.android.libraries.architecture.bindings
 import io.element.android.libraries.di.DaggerComponentOwner
+import io.element.android.x.di.AppBindings
 import io.element.android.x.di.AppComponent
 import io.element.android.x.di.DaggerAppComponent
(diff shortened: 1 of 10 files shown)


features/cachecleaner/impl/src/main/kotlin/io/element/android/features/cachecleaner/impl/DefaultCacheCleaner.kt
```diff
@@ -0,0 +1,54 @@
...
+    private val dispatchers: CoroutineDispatchers,
+    @CacheDirectory private val cacheDir: File,
+) : CacheCleaner {
+    companion object {
+        val SUBDIRS_TO_CLEANUP = listOf("temp/media", "temp/voice")
+    }
+
+    override fun clearCache() {
+        scope.launch(dispatchers.io) {
+            runCatching {
```
jmartinesp, line 44: Shouldn't we at least log something if clearing the cache failed?

features/cachecleaner/impl/src/main/kotlin/io/element/android/features/cachecleaner/impl/DefaultCacheCleaner.kt
```diff
@@ -0,0 +1,54 @@
...
+ * Default implementation of [CacheCleaner].
+ */
+@ContributesBinding(AppScope::class)
+class DefaultCacheCleaner @Inject constructor(
+    private val scope: CoroutineScope,
+    private val dispatchers: CoroutineDispatchers,
+    @CacheDirectory private val cacheDir: File,
+) : CacheCleaner {
+    companion object {
+        val SUBDIRS_TO_CLEANUP = listOf("temp/media", "temp/voice")
```
jmartinesp, line 39: 👌

app/src/main/kotlin/io/element/android/x/ElementXApplication.kt
```diff
@@ -36,5 +38,6 @@ class ElementXApplication : Application(), DaggerComponentOwner {
             initializeComponent(TracingInitializer::class.java)
         }
         logApplicationInfo()
+        bindings<AppBindings>().cacheCleaner().clearCache()
```
jmartinesp, line 41: I know I said this didn't have to be an initializer too, but looking at the previous code it seems a bit out of place if it's not 🤔 . Would it take you much time to migrate to using an initializer? If so, we can leave it as is.

jmartinesp (approved):
Approved with a couple of comments.

features/cachecleaner/api/build.gradle.kts
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+plugins {
+    id("io.element.android-library")
+}
+
+android {
+    namespace = "io.element.android.features.cachecleaner.api"
```
ganfra, line 22: I don't think it should be in a feature module, it could even live in an existing one (`androidutils`?).
Also I'd say we don't need an interface for that.

app/src/main/kotlin/io/element/android/x/ElementXApplication.kt
julioromano: Done in: https://github.com/vector-im/element-x-android/pull/1807/commits/47111b9a535007449f848cf09e07f3badcc833b4

features/cachecleaner/api/build.gradle.kts
julioromano: Yeah as you said before the distinction between libraries & features subdirs is kinda moot, we might as well merge them into a single subdir.

This is definitely not a utility class though because it is not general purpose, it is not something that could be reused easily in another context. So I wouldn't put it in `androidutils`.

For the interface yes, we could do without it... but it's there and we've already spent enough time for such a simple thing. I hope it won't bother anyone.

jmartinesp (approved):

PR #1829, opened 2023-11-16
Rust SDK: prefer using interfaces when they exist

bmarty:
Draft, WIP.

The idea is to use the interfaces declared in the FFI layer as much as possible.

Sometimes the object has to be cast to be able to use FFI API. Also the FFI layer does not return systematically interface but sometimes the implementation type.

Adding `useAny` since the interface does not extend `Disposable`. Ideally it could be the case, but not a blocker, since we have the handy `Disposable.destroy(...)` to help (could be renamed to `use` but I prefer to have a different name for clarity).

I think I have use all the possible interface and at all places, but I may have missed some.

This work will help to unit test our implementation, but allowing to inject our own test implementation of the interfaces.

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/MatrixTimelineItemMapper.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/MatrixTimelineItemMapper.kt
@@ -20,9 +20,11 @@ import io.element.android.libraries.matrix.api.core.EventId
 import io.element.android.libraries.matrix.api.timeline.MatrixTimelineItem
 import io.element.android.libraries.matrix.impl.timeline.item.event.EventTimelineItemMapper
 import io.element.android.libraries.matrix.impl.timeline.item.virtual.VirtualTimelineItemMapper
+import io.element.android.libraries.matrix.impl.util.useAny
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 import org.matrix.rustcomponents.sdk.TimelineItem
+import org.matrix.rustcomponents.sdk.TimelineItemInterface
 
 class MatrixTimelineItemMapper(
     private val fetchDetailsForEvent: suspend (EventId) -> Result<Unit>,
@@ -31,7 +33,7 @@ class MatrixTimelineItemMapper(
     private val eventTimelineItemMapper: EventTimelineItemMapper = EventTimelineItemMapper(),
 ) {
 
-    fun map(timelineItem: TimelineItem): MatrixTimelineItem = timelineItem.use {
+    fun map(timelineItem: TimelineItemInterface): MatrixTimelineItem = timelineItem.useAny {
         val uniqueId = timelineItem.uniqueId().toLong()
         val asEvent = it.asEvent()
         if (asEvent != null) {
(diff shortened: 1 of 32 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/MatrixTimelineItemMapper.kt
```diff
@@ -20,9 +20,11 @@ import io.element.android.libraries.matrix.api.core.EventId
 import io.element.android.libraries.matrix.api.timeline.MatrixTimelineItem
 import io.element.android.libraries.matrix.impl.timeline.item.event.EventTimelineItemMapper
 import io.element.android.libraries.matrix.impl.timeline.item.virtual.VirtualTimelineItemMapper
+import io.element.android.libraries.matrix.impl.util.useAny
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 import org.matrix.rustcomponents.sdk.TimelineItem
```
ElementBot, line 26: - :warning: The import 'org.matrix.rustcomponents.sdk.TimelineItem' is unused.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/widget/MatrixWidgetSettings.kt
```diff
@@ -35,13 +36,13 @@ fun MatrixWidgetSettings.Companion.fromRustWidgetSettings(widgetSettings: Widget
...
 suspend fun MatrixWidgetSettings.generateWidgetWebViewUrl(
-    room: Room,
+    room: RoomInterface,
     clientId: String,
     languageTag: String? = null,
     theme: String? = null
 ) = generateWebviewUrl(
     widgetSettings = this.toRustWidgetSettings(),
-    room = room,
+    room = (room as Room) /* TODO SDK Change */,
```
ElementBot, line 45: - :warning: Parentheses in (room as Room) are unnecessary and can be replaced with: room as Room

jmartinesp:
> This work will help to unit test our implementation, but allowing to inject our own test implementation of the interfaces.

Take into account that some of these interfaces return or use concrete FFI class implementations as parameters, not the interfaces, so even if we could create fakes of them we'd still need to create fakes of those objects. The good news is, a couple of weeks ago a PR was merged in uniffi about making the classes `open` so we can fake those too as long as we don't pass them to actual Rust objects. Bad news is, I think we haven't updated to a version with that change in the Rust SDK yet.

bmarty:
> Take into account that some of the Rust binding interfaces return or use concrete FFI class implementations as parameters, not the interfaces

Yes you're right, my plan is to list all those cases (some of them are TODOs in my PR) and share it with the Rust team.

For the parameter, I think the API could work with ids. For instance if the parameter is of type `Room`, it could be changed to `RoomInterface`, but it's probably better to just pass the `roomId: String`.

jmartinesp:
Honestly, before changing the API I'd just wait for the result of the PR I mentioned above to reach the Rust SDK so we don't end up changing any APIs we don't really need to. Fetching stuff from the Rust SDK with ids may not be free of side-effects, since we also need to take into account their lifetime in our code and if the backing Rust object should be kept alive, or if the object auto-starts some logic by instantiating it.

bmarty:
> the PR I mentioned above

Sorry I do not see this PR, which one is it?

jmartinesp:
It's [this one](https://github.com/mozilla/uniffi-rs/pull/1815). I submitted it some time ago, along with another version that used interfaces instead and the uniffi team preferred this version with open classes, as it was easier to implement and maintain.

So with this we could just create fakes by extending the classes, and we wouldn't need to change to using identifiers and fetching new instances.

bmarty:
Will revisit this using the `open` modifier added to all the Uniffi classes.

PR #1835, opened 2023-11-17
Timeline : do not use SubcomposeLayout if not needed

ganfra:
While digging for https://github.com/vector-im/element-x-android/pull/1797 
I tried to remove the usage of `SubComposeLayout` playing with `IntrinsicSize.Max`, but this isn't working for async images. 
So I just simplified so we don't use `SubcomposeLayout` if not needed.


jmartinesp (approved):
The fact that there's no change in the screenshot is proof that it works as expected 😄. Thanks!

PR #1847, opened 2023-11-21
Voice message scrubbing improvements

julioromano:
- Voice messages can be scrubbed (i.e. seeked to) even when they have not been played yet..
- The progress bar is displayed also when paused.
- Multiple voice messages can keep their state when paused.
- Tries to adhere as much as possible at the detailed "green cursor" behavior in the story (but might not be 100% compliant).

Story: https://github.com/vector-im/element-meta/issues/2113

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemVoiceView.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemVoiceView.kt
@@ -96,7 +96,7 @@ fun TimelineItemVoiceView(
         Spacer(Modifier.width(8.dp))
         val context = LocalContext.current
         WaveformPlaybackView(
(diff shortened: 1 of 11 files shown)


jmartinesp (approved):
LGTM, thanks! It's not something that started in this PR, but maybe it would be good to rename `MediaPlayer.State` to something like `MediaPlayer.SharedState` / `GlobalState`? Or add some docs to it, so we know that's shared for all voice players and the internal one is local to this one.

julioromano:
> LGTM, thanks! It's not something that started in this PR, but maybe it would be good to rename `MediaPlayer.State` to something like `MediaPlayer.SharedState` / `GlobalState`? Or add some docs to it, so we know that's shared for all voice players and the internal one is local to this one.

I'd prefer to improve the docs rather than rename stuff otherwise names could get quite lengthy.
I'll be glad to open a doc improvment PR, but so far nothing comes to my mind about what to improve specifically.
Let's talk about it!

PR #1859, opened 2023-11-22
Merge TimelineItemEventRow's textForInReplyTo and attachmentThumbnailInfoForInReplyTo functions

julioromano:
The flow is somewhat misleading so its logic has been merged into `InReplyToDetails.metadata()`.


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
```diff
@@ -637,58 +636,74 @@ private fun ReplyToContent(
...
-    return when (val eventContent = inReplyTo.eventContent) {
-        is MessageContent -> when (eventContent.type) {
-            is LocationMessageType -> stringResource(CommonStrings.common_shared_location)
-            is VoiceMessageType -> stringResource(CommonStrings.common_voice_message)
-            else -> inReplyTo.textContent ?: eventContent.body
-        }
-        is PollContent -> eventContent.question
-        else -> ""
-    }
+private sealed interface InReplyToMetadata {
```
jmartinesp, line 699: I don't really know if it actually makes a difference, but using `@Immutable` with sealed interfaces *should* help in avoiding unnecessary recompositions, since otherwise the compose compiler doesn't know how to treat the interface AFAIK.

jmartinesp (approved):
LGTM, with a comment.

jmartinesp:
Could we also add some mapping tests for `InReplyToMetadata`? I know they usually won't be worth it, but given our coverage objective we should probably add those so coverage doesn't decrease.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
jmartinesp: To the `sealed interface`.

julioromano:
> Could we also add some mapping tests for `InReplyToMetadata`? I know they usually wouldn't be worth it, but given our coverage objective we should probably add those so coverage doesn't decrease.

All done! Might wanna take another look as the tests ended up being quite verbose.

PR #1865, opened 2023-11-22
RoomList : rework a bit the api and make usage of entriesWithDynamicAdapter

ganfra:
This PR change a bit the api  from the matrix module to interacts with RoomList. Internally it now uses entriesWithDynamicAdapter instead of entries.

It's just a first step in the usage of this api as I couldn't get a good config with paging and filtering yet.
So for now it mimics the usage of the entries api, so we don't have other changes in the app.
But it'll allow us to benefit from https://github.com/matrix-org/matrix-rust-sdk/pull/2866


bmarty (approved):
LGTM, thanks for this amazing work!

PR #1869, opened 2023-11-23
Allow polls to be edited

jonnyandrew:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Allow polls to be edited if they have no votes.

## Motivation and context

- Part of https://github.com/vector-im/element-meta/issues/2179

## Tests

- Create a poll
- Edit the poll from the timeline
- Edit the poll from the message actions menu
- Vote on the poll and observe that it's no longer editable

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 13

## Checklist


bmarty:
The conflicts are due to the move of `features/messages/impl/src/test/kotlin/io/element/android/features/messages/` content to `features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl` in #1875

bmarty (commented):
Thanks, nice work.
What happen if a vote is performed by another user while editing the poll? I think this case is not covered. Maybe the server (or the SDK) will reject the edition?

jonnyandrew:
> What happen if a vote is performed by another user while editing the poll? I think this case is not covered. Maybe the server (or the SDK) will reject the edition?

That is indeed the problem with allowing polls to be edited. The spec ([MSC3381](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/polls/proposals/3381-polls.md)) allows it but notes

> Users should be wary of polls changing their question after they have voted. Considering polls can be edited, responses might no longer be relevant. For example, if a poll was opened for "do you like cupcakes?" and you select "yes", the question may very well become "should we outlaw cupcakes?" where your "yes" no longer applies. This MSC considers this problem more of a social issue than a technical one, and reminds the reader that polls should not be used for anything important/serious at the moment.

The app attempts to reduce the chance of it happening (rather than preventing it altogether) by allowing poll to be edited only before the creator is aware of any votes, and we accept that votes that arrive in the meantime may be lost or become invalid.

jonnyandrew:
@bmarty I've addressed the review comments. Am I right based on your edit that there is nothing to be done about https://github.com/vector-im/element-x-android/pull/1869#discussion_r1403711648?

bmarty (approved):
Yes, thanks for the update and the clarification!

PR #1920, opened 2023-11-29
RoomList: introduce incremental loading to improve performances.

ganfra:
This should fix https://github.com/vector-im/element-x-android/issues/1902


frebib:
This is basically instant with this patch. It might even be faster than before! Thanks for fixing :)

bmarty (approved):
LGTM, thanks! Maybe add a changelog file?

PR #1936, opened 2023-11-30
Test android local media actions and other classes

bmarty:
More tests around some classes. There is not a lot to check...

### DIFF
--- a/libraries/mediaviewer/impl/src/test/kotlin/io/element/android/libraries/mediaviewer/impl/local/AndroidLocalMediaFactoryTest.kt
+++ b/libraries/mediaviewer/impl/src/test/kotlin/io/element/android/libraries/mediaviewer/impl/local/AndroidLocalMediaFactoryTest.kt
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2023 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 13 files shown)


libraries/mediaviewer/impl/src/test/kotlin/io/element/android/libraries/mediaviewer/impl/local/AndroidLocalMediaFactoryTest.kt
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+package io.element.android.libraries.mediaviewer.impl.local
+
+import com.google.common.truth.Truth.assertThat
+import io.element.android.libraries.androidutils.filesize.FakeFileSizeFormatter
+import io.element.android.libraries.core.mimetype.MimeTypes
+import io.element.android.libraries.matrix.api.media.MediaFile
+import io.element.android.libraries.matrix.test.media.FakeMediaFile
+import io.element.android.libraries.mediaviewer.api.local.LocalMedia
```
ElementBot, line 24: - :warning: The import 'io.element.android.libraries.mediaviewer.api.local.LocalMedia' is unused.

ganfra (approved):

libraries/mediaviewer/impl/src/test/kotlin/io/element/android/libraries/mediaviewer/impl/local/AndroidLocalMediaFactoryTest.kt
bmarty: fixed on develop: 95b922c2726d9796e9ce2613b729affa7e2c84de

PR #1938, opened 2023-11-30
Feature/fga/user detail direct chat

ganfra:
Closes https://github.com/vector-im/element-x-android/issues/1915

Introduce a new `StartDMAction` so we don't copy/paste code in the different places where we can start a DM.

### DIFF
--- a/features/createroom/test/src/main/kotlin/io/element/android/features/createroom/test/FakeStartDMAction.kt
+++ b/features/createroom/test/src/main/kotlin/io/element/android/features/createroom/test/FakeStartDMAction.kt
@@ -22,7 +22,7 @@ import io.element.android.libraries.architecture.Async
 import io.element.android.libraries.matrix.api.core.RoomId
 import io.element.android.libraries.matrix.api.core.UserId
 import io.element.android.libraries.matrix.test.A_ROOM_ID
-import io.element.android.tests.testutils.simulateLongTask
+import kotlinx.coroutines.delay
 
 class FakeStartDMAction : StartDMAction {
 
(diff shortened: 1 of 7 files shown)


jmartinesp (approved):
LGTM. The process seems a bit convoluted but I guess it's normal given several features are involved. I'm not 100% sure about the custom `StartDMResult` either (couldn't we have used a `Result` with custom success/error values?), but it's not a blocker.

features/createroom/test/src/main/kotlin/io/element/android/features/createroom/test/FakeStartDMAction.kt
```diff
@@ -0,0 +1,39 @@
...
+    private var executeResult: Async<RoomId> = Async.Success(A_ROOM_ID)
+
+    fun givenExecuteResult(result: Async<RoomId>) {
+        executeResult = result
+    }
+
+    override suspend fun execute(userId: UserId, actionState: MutableState<Async<RoomId>>) = simulateLongTask {
+        actionState.value = Async.Loading()
+        actionState.value = executeResult
+    }
```
bmarty, line 38: Since there are 2 state value update, I think this is not necessary to use `simulateLongTask`.
Maybe change to
```kotlin
    override suspend fun execute(userId: UserId, actionState: MutableState<Async<RoomId>>) = {
        actionState.value = Async.Loading()
        delay(1)
        actionState.value = executeResult
    }
```
to be closer to the reality? Or create a new `simulateLongTask` which take a `MutableState` in parameter?

PR #1941, opened 2023-12-01
Fix scaling of timeline images

### LINKED ISSUE #1940: Align images in timeline with iOS: fix scaling strategy
On Android we're applying:

- A Crop content scale for images.
- A min and max aspect ratio values.

On iOS instead they use:

- A Fit content scale.
- Have a min and max height values, but they don't restrict the aspect ratio values themselves.

It seems like this causes some images, specially long vertical ones, to be displayed wrong in Android.

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

The scaling behaviour now matches iOS:

- Fit content scaling, not Crop.
- Min/max height, not min/max aspect ratio. Changes the aspect ratio to take into account width, then height.

## Motivation and context

Fixes #1940.

## Screenshots / GIFs

In the PR.

## Tests

- Send an image with long height.
- Check how it looks in the timeline.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 11

## Checklist


ganfra:
This doesnt work nicely for very large or very high images...

jmartinesp:
Could we test those in the iOS app, to compare what they look like? Maybe I messed up the implementation, or maybe that's just what their algorithm does.

ganfra:
> Could we test those in the iOS app, to compare what they look like? Maybe I messed up the implementation, or maybe that's just what their algorithm does.

Honestly I think we should keep cropping, thats what other messenger apps are doing. We could just have a bigger maxHeight maybe. @VolkerJunginger

jmartinesp:
This is what something similar looks like on iOS:

<details>
<summary>Details</summary>

</details>

seankhl:
Just my two cents as a user: 

I think the WhatsApp and pre-PR solutions are worse, because they are misleading about the contents of the image. It's not clear that I wasn't just sent a black square, whereas at least with the element-x iOS and telegram solutions I can be like okay, there's more detail than I can easily see from the timeline image, let me download it or click and zoom.

The new element-x android solution is okay, but the background color being much larger than the image and the image not being centered in it feels unnecessary and is a bit visually awkward and distracting. In that sense maybe it could be made to conform more to the element-x iOS version.

The final note I'd like to make that the rounded corners of the messages would have a larger negative impact on these very narrow images if the background is not present, as shown in the iOS version. Maybe it would be good to make the background just wide or tall enough to capture the rounded corner, so the image can remain rectilinear while providing the least distracting background? Not sure the exact ideal solution to that, but the rounded corner cutting into the image and obscuring some of it is worth avoiding or at least considering in the design.

jmartinesp:
I was using Telegram desktop today and I saw another possible solution, something like they do here:

image


We could overlay the blurhash image and the actual: this way, the bubble still looks like a bubble and the actual image is displayed. I don't know how it'll look with tiny images though.

seankhl:
The blurhash as background looks good and I like how the rounded corners aren't impacting the image.

I also like how there is no timestamp overlapping with the image in the Telegram desktop version. I think this aspect is worth considering for Element, too. I see no reason why a design couldn't be made that avoids overlapping with the image while being as good as the current one in all other respects. Not trying to criticize, just pointing out that keeping the image completely unmodified and unobscured seems strictly better to me.

jmartinesp:
I was on holidays last week, I'm continuing with this work now. @VolkerJunginger , @amshakal what do you think about the issue found above and the possible solutions? To me, [the blurhash one](https://github.com/vector-im/element-x-android/pull/1941#issuecomment-1837159283) looks like the best option, and as a second option the telegram Android one where images with a width or height smaller than a min value don't actually show a preview (last spoiler tag [here](https://github.com/vector-im/element-x-android/pull/1941#issuecomment-1836069002).

In any case, doing any of these changes would mean doing the equivalent on iOS. I think the blurhash one should be fairly easy to do since all we'd had to do is not replace the loading blurhash background with the loaded image, but overlay them.

ganfra (approved):

tests/uitests/src/test/snapshots/images/ui_S_t[f.messages.impl.timeline.components_TimelineItemEventRowForDirectRoom_null_TimelineItemEventRowForDirectRoom-Day-17_17_null,NEXUS_5,1.0,en].png
Comment (bmarty) [inline tests/uitests/src/test/snapshots/images/ui_S_t[f.messages.impl.timeline.components_TimelineItemEventRowForDirectRoom_null_TimelineItemEventRowForDirectRoom-Day-17_17_null,NEXUS_5,1.0,en].png:1]: I think we need to have a background color, else the bubble does not render nicely...

tests/uitests/src/test/snapshots/images/ui_S_t[f.messages.impl.timeline.components_TimelineItemEventRowForDirectRoom_null_TimelineItemEventRowForDirectRoom-Night-17_18_null,NEXUS_5,1.0,en].png
Comment (bmarty) [inline tests/uitests/src/test/snapshots/images/ui_S_t[f.messages.impl.timeline.components_TimelineItemEventRowForDirectRoom_null_TimelineItemEventRowForDirectRoom-Night-17_18_null,NEXUS_5,1.0,en].png:1]: ... and this will also fix this issue.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/MessageEventBubble.kt
```diff
@@ -99,10 +100,10 @@ fun MessageEventBubble(
     }
 
     // Ignore state.isHighlighted for now, we need a design decision on it.
-    val backgroundBubbleColor = if (state.isMine) {
-        ElementTheme.colors.messageFromMeBackground
-    } else {
-        ElementTheme.colors.messageFromOtherBackground
+    val backgroundBubbleColor = when {
+        !state.displayBackground -> Color.White
```
bmarty, line 104: You should use background color of the bubble, see my remark on the screenshots.

bmarty (changes requested):
Thanks for the update. There is an issue regarding the background color, else I think it's fine.
It's a shame that we do not have a preview with the image and the blurhash behind. Not a P1 to investigate this thought.

tests/uitests/src/test/snapshots/images/ui_S_t[f.messages.impl.timeline.components_TimelineItemEventRowForDirectRoom_null_TimelineItemEventRowForDirectRoom-Day-17_17_null,NEXUS_5,1.0,en].png
Comment (jmartinesp) [reply tests/uitests/src/test/snapshots/images/ui_S_t[f.messages.impl.timeline.components_TimelineItemEventRowForDirectRoom_null_TimelineItemEventRowForDirectRoom-Day-17_17_null,NEXUS_5,1.0,en].png:1]: This is what iOS looks like and what we've decided to use for now, given the deadline, since otherwise it would mean extra work for the iOS team, although this will probably be revisited soon. I took into account the suggestions to have a white background so images with transparency render nicely.

Maybe we could add a subtle border radius for these images so the bubble keeps its shape, something like this?

<details>
<summary>Screenshot</summary>

</details>

Comment (bmarty) [reply tests/uitests/src/test/snapshots/images/ui_S_t[f.messages.impl.timeline.components_TimelineItemEventRowForDirectRoom_null_TimelineItemEventRowForDirectRoom-Day-17_17_null,NEXUS_5,1.0,en].png:1]: Can we have the bubble background color and white background for the image only (and maybe black background on dark theme)?
Like this this is not OK I think, but yes, definitely better with the bubble border.

Comment (bmarty) [reply tests/uitests/src/test/snapshots/images/ui_S_t[f.messages.impl.timeline.components_TimelineItemEventRowForDirectRoom_null_TimelineItemEventRowForDirectRoom-Day-17_17_null,NEXUS_5,1.0,en].png:1]: With the bubble color I've got this:

|Light|Dark|
|-|-|
|image|image|

And I think it's fine.

We may want to improve the rendering of images with transparency in another PR, since applying a white background should be done to all images generated by users: room avatars, user avatars, images in timeline, etc.

@amshakal do you agree?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/MessageEventBubble.kt
```diff
@@ -32,6 +32,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
```
ElementBot, line 35: - :warning: The import 'androidx.compose.ui.graphics.Color' is unused.

ElementBot, line 35: - :warning: The import 'androidx.compose.ui.graphics.Color' is unused.

tests/uitests/src/test/snapshots/images/ui_S_t[f.messages.impl.timeline.components_TimelineItemEventRowForDirectRoom_null_TimelineItemEventRowForDirectRoom-Day-17_17_null,NEXUS_5,1.0,en].png
Comment (amshakal) [reply tests/uitests/src/test/snapshots/images/ui_S_t[f.messages.impl.timeline.components_TimelineItemEventRowForDirectRoom_null_TimelineItemEventRowForDirectRoom-Day-17_17_null,NEXUS_5,1.0,en].png:1]: Yes, this makes sense to me. :)

PR #2024, opened 2023-12-14
Improve back-navigation for invitations

### LINKED ISSUE #2022: Always close the invite list screen when there is no more invite
Motivation :

https://github.com/element-hq/element-meta/issues/2213#issuecomment-1854355265

ganfra:
Always close the invite list screen when there is no more invite : closes #2022

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
@@ -67,6 +67,7 @@ import io.element.android.libraries.di.SessionScope
 import io.element.android.libraries.matrix.api.MatrixClient
 import io.element.android.libraries.matrix.api.core.MAIN_SPACE
 import io.element.android.libraries.matrix.api.core.RoomId
+import io.element.android.libraries.matrix.api.roomlist.RoomList
 import io.element.android.libraries.matrix.api.sync.SyncState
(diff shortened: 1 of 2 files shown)


appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -149,6 +150,23 @@ class LoggedInFlowNode @AssistedInject constructor(
...
+        observeInvitesLoadingState()
+    }
+
+    private fun observeInvitesLoadingState() {
+        lifecycleScope.launch {
+            repeatOnLifecycle(Lifecycle.State.STARTED) {
+                matrixClient.roomListService.invites.loadingState
+                    .collect { inviteState ->
+                        when (inviteState) {
+                            is RoomList.LoadingState.Loaded -> if (inviteState.numberOfRooms == 0) {
```
bmarty, line 162: Question: any idea why it's called `maximumNumberOfRooms` Rust side?

bmarty (approved):
Nice solution, thanks!

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
ganfra: I'd say because it doesn't reflect the actual number of rooms we got from SlidingSync ? but not sure

PR #2040, opened 2023-12-15
fix linkify

### LINKED ISSUE #2038: Messages are not linkified anymore
- send a message containing a link
- observe that the link is not rendered correctly and not clickable

image

bmarty:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Ensure message with body and without formatted body got working links.
Also ensure email addresses are linkified.

## Motivation and context

Closes #2038

## Tests

- Send a message with just a raw link and nothing else
- Observe that with this change the link in the message is rendered as a link and is clickable

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentMessageFactory.kt
```diff
@@ -68,12 +69,15 @@ class TimelineItemContentMessageFactory @Inject constructor(
...
-                htmlDocument = messageType.formatted?.toHtmlDocument(prefix = "* $senderDisplayName"),
-                formattedBody = parseHtml(messageType.formatted, prefix = "* $senderDisplayName"),
-                isEdited = content.isEdited,
-            )
+            is EmoteMessageType -> {
+                val emoteBody = "* $senderDisplayName ${messageType.body}"
+                TimelineItemEmoteContent(
+                    body = emoteBody,
+                    htmlDocument = messageType.formatted?.toHtmlDocument(prefix = "* $senderDisplayName"),
+                    formattedBody = parseHtml(messageType.formatted, prefix = "* $senderDisplayName") ?: emoteBody.withLinks(),
```
ganfra, line 77: If I understand correctly, we always have a `formattedBody` now with those fixes. Could we simplify a bit the `TimelineItemTextBasedContent` interface then? 
It's a bit hard to know what to use currently, and we probably don't need all the fields.
```
sealed interface TimelineItemTextBasedContent : TimelineItemEventContent {
    val body: String
    val htmlDocument: Document?
    val formattedBody: CharSequence?
    val plainText: String
    val isEdited: Boolean
    val htmlBody: String?
        get() = htmlDocument?.body()?.html()
}
```

frebib:
Would we ever consider also detecting and formatting raw user/room mentions in this way too, for clients that only send unformatted bodies?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentMessageFactory.kt
bmarty: > we always have a `formattedBody`

No, `withLinks()` returns `null` if no links are added.

appnav/src/test/kotlin/io/element/android/appnav/loggedin/LoggedInPresenterTest.kt
```diff
@@ -60,8 +60,9 @@ class LoggedInPresenterTest {
         }.test {
             val initialState = awaitItem()
             assertThat(initialState.showSyncSpinner).isFalse()
+            roomListService.postSyncIndicator(RoomListService.SyncIndicator.Show)
             consumeItemsUntilPredicate { it.showSyncSpinner }
-            roomListService.postState(RoomListService.State.Running)
+            roomListService.postSyncIndicator(RoomListService.SyncIndicator.Hide)
```
bmarty, line 65: `consumeItemsUntilPredicate` now expect the predicate to be true at some point, so I had to fix this outdated test.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentMessageFactory.kt
ganfra: Can we still try to simplify the api? 
If we could get a `plainText/formattedText/body/htmlBody` and remove the `htmlDocument` instance?
I'd say, The `formattedText` shouldn't be nullable and should always be displayed in timeline.

bmarty: Yes, but I would prefer to do that in a separate rework PR. OK for you?

ganfra: as u want

ganfra (approved):

PR #2041, opened 2023-12-15
Fix timeline back pagination loop in encrypted room.

ganfra:
Timeline in encrypted room could move to a back pagination loop when reaching the encrypted banner. 

2 commits in this PR : 
- [First commit](f961d81eb068adf13e156335fb314d56c69feae5) let the `matrix` module returns `TimelineItem.uniqueId` as `String` instead of `Long`, as it's already how we transform them after.
- [Second commit](c0ac6e191b89dcea331894e33391b8281553fde4) is the actual fix. In encrypted room when we're reaching the encrypted banner, there was a potential race and was using the wrong timelineItems list to compute the PaginationState. It's now done only in one place.


features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenterTest.kt
```diff
@@ -121,7 +121,7 @@ class TimelinePresenterTest {
     fun `present - on scroll finished send read receipt if an event is before the index`() = runTest {
         val timeline = FakeMatrixTimeline(
             initialTimelineItems = listOf(
-                MatrixTimelineItem.Event(0, anEventTimelineItem())
+                MatrixTimelineItem.Event("0", anEventTimelineItem())
```
bmarty, line 123: I understand that we do not really care about the id in test. So maybe create a `FAKE_ID` const and use it every where, to replace `"0"` and `"0L"`.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/RustMatrixTimeline.kt
```diff
@@ -115,16 +114,54 @@ class RustMatrixTimeline(
...
         }
     }
 
+    private fun paginationStateFlow(): Flow<MatrixTimeline.PaginationState> {
+        return combine(
+            innerTimeline.backPaginationStatusFlow(),
+            timelineItems,
+        ) { paginationStatus, filteredItems ->
+            if (filteredItems.hasEncryptionHistoryBanner()) {
+                return@combine MatrixTimeline.PaginationState(
```
bmarty, line 133: I can observe that sometimes the EncryptionHistory banner was displayed and then some new Event were appearing (so it was back paginating). Is it correct that with this PR the app will not paginate as soon as the banner is present?

bmarty (commented):
LGTM, did not test, some first thoughts.

features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenterTest.kt
ganfra: Yes thats a good idea.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/RustMatrixTimeline.kt
ganfra: Yes this is correct, thats the goal of the PR. It was creating weird behaviour in the poll history too.

bmarty: But the banner was displayed "by mistake", since actually there were some events when back paginating, so we introduce a bug here, no? The root cause of this may be fixed later, but I am afraid that the bug will be more visible.

ganfra: Yeah, we cant know for sure if there is more non encrypted events before. I fix the existing issue. If we want to change this behaviour, we should think about it with iOS and product.

bmarty:
I had the same crash than the one reported by Maestro when attempting to leave the room, but on my side, it was after creating a room:

https://github.com/element-hq/element-x-android-rageshakes/issues/1224

Note: repro rate is not 100%

bmarty (approved):

ganfra:
I still can't reproduce the issue Maestro has, weird state...

jmartinesp:
FWIW, I can't reproduce it either on my side.

ganfra:
I'll give a look again tomorrow.

bmarty:
We will treat Maestro ANR later.

bmarty:
We will treat Maestro ANR later.

PR #2059, opened 2023-12-19
Appyx Overlay

ganfra:
This PR introduce a new `Overlay` nav model (which is just a disguised `Backstack` with zero or one element).
Also rename `BackstackParentNode` to `BaseFlowNode` so it better reflects his usage.


libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/BaseFlowNode.kt
```diff
@@ -0,0 +1,106 @@
...
+
+@Composable
+inline fun <reified NavTarget : Any> BaseFlowNode<NavTarget>.BackstackWithOverlayBox(
+    modifier: Modifier = Modifier,
+    content: @Composable BoxScope.() -> Unit = {},
+) {
+    Box(modifier = modifier) {
+        BackstackView()
+        OverlayView()
+        content()
```
bmarty, line 104: Do we really need a `content` here?

bmarty (approved):
It working fine, thanks!
I can update the failing Konsist test to make it pass.

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/BaseFlowNode.kt
ganfra: I guess in most scenario we don't, but it's not costly, so I don't know if we need it.

PR #2078, opened 2023-12-20
Reaction offset in dm

### LINKED ISSUE #2076: Reactions are offset too far right in DM chats
This is a regression from https://github.com/element-hq/element-x-android/pull/1981

### Steps to reproduce

- Add a reaction to a message in a message *received* in a DM chat (on the left of the timeline)
- Notice that it's a bit further right than it should be

#### What did you expect?
Reactions to be offset from the left of the bubble by the same amount as in non-DM chats

#### What happened instead?
They're too far right

### Your phone model

Pixel 8

### Operating system version

Android 14

bd53b9c5b990e5ecbfe4856dac1ae55840a4bf97

### Homeserver

Synapse develop

### Will you send logs?

Yes

No

bmarty:
Fix #2076 and also apply correct offset to reaction, according to [Figma](https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?node-id=2128%3A106638&mode=dev)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
```diff
@@ -345,7 +345,16 @@ private fun TimelineItemEventRowContent(
                         linkStartOrEnd(event)
                     }
                     .zIndex(1f)
-                    .padding(start = if (event.isMine) 16.dp else 36.dp, end = 16.dp)
+                    .padding(
+                        // Note: due to the applied constraints, start is left for other's message and right for mine
+                        // In design we want a offset of 6.dp compare to the bubble, so start is 22.dp (16 + 6)
```
jmartinesp, line 350: 👌

jmartinesp (approved):

PR #2080, opened 2023-12-20
Room list badges

bmarty:
Follow new rules defined in #2079 to display room badges in the room list.

Closes #2079

PR #2095, opened 2023-12-22
Revert "Room list badges"

jmartinesp:
Reverts element-hq/element-x-android#2080

We found some issues when testing edge cases that didn't appear on our initial tests, we're reverting this for now.


manuroe (approved):
I tested the apk build on a test and my real account. Room list badges work as before.

jmartinesp:
Merging this as is, we need to investigate the failing test at the logout feature. Maybe reverting #2071 is not needed.

PR #2096, opened 2023-12-22
Add warning about updating `com.autonomousapps.dependency-analysis`

jmartinesp:
Versions > 1.27.0 won't allow use local AAR builds of the Rust SDK, until this is fixed a warning should be added so we don't update this dependency accidentally.


jmartinesp:
I don't really get why recording new screenshots was needed. Maybe the ones in `develop` are not up to date?

bmarty (approved):
Thanks. I was disabling this plugin manually to be able to build the app with a local SDK.

PR #2103, opened 2023-12-25
Sync Strings

ElementBot:


bmarty (approved):
OK
Ignoring failing test (in release and in CI only) for now.

PR #2107, opened 2023-12-26
Attempt to fix test on CI (again)

jmartinesp:
PR #2071 fixed issues with the testing flow. I reverted that when we had to downgrade the Rust SDK version before a release, but since we're seeing these same issues in `develop` now maybe that was a mistake.


bmarty (approved):
Thanks!

PR #2117, opened 2023-12-28
Compose upgrade

bmarty:
Replacement for #1867 and do some cleanup on the codebase.

### DIFF
--- a/features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootNode.kt
+++ b/features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootNode.kt
@@ -48,14 +48,12 @@ class CreateRoomRootNode @AssistedInject constructor(
         fun onStartChatSuccess(roomId: RoomId)
     }
 
-    private val callback = object : Callback {
-        override fun onCreateNewRoom() {
-            plugins<Callback>().forEach { it.onCreateNewRoom() }
-        }
+    private fun onCreateNewRoom() {
+        plugins<Callback>().forEach { it.onCreateNewRoom() }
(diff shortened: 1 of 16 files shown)


features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootNode.kt
```diff
@@ -72,8 +70,8 @@ class CreateRoomRootNode @AssistedInject constructor(
             state = state,
             modifier = modifier,
             onClosePressed = this::navigateUp,
-            onNewRoomClicked = callback::onCreateNewRoom,
-            onOpenDM = callback::onStartChatSuccess,
+            onNewRoomClicked = ::onCreateNewRoom,
+            onOpenDM = ::onStartChatSuccess,
```
bmarty, line 74: The compose compiler (or Kotlin compiler?) did not like the previous version

jmartinesp: We could just use `= { callback.onCreateNewRoom() }` there, but this is fine too.

jmartinesp (approved):
LGTM, thanks!

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/root/CreateRoomRootNode.kt
bmarty: Yes, but I prefer to keep the same pattern everywhere.

bmarty:
Failing test is a flaky test that need to be fixed in a dedicated PR.

PR #2130, opened 2023-12-29
Some rework on the codebase

bmarty:
This PR now includes only rework that I want to merge on develop first. So no change in UI and in behavior. Just renaming and clean up.

A following PR will come with the change about notification counts.

<del>
- First commit: #2080 again, which had been reverted in #2095.
- Second commit new version of the SDK, which fix the "RoomInfo.notificationCount always equal to 0" issue
- Thrid commit: map `RoomInfo.notificationCount` to `RoomSummaryDetails.numUnreadMessages`, since `RoomInfo.numUnreadMessages` still have some issue. This thrid commit may be reverted at some point.

The behavior of the notification badges in the room list should be similar to the iOS app with this change.
</del>


jmartinesp (commented):
I'm still seeing an issue I saw before we reverted this: when a room has a previous message with a mention to you, received while the app wasn't running, the unread count and mention won't be gone until a new message in that room (not coming from the current user) is received.

I'm wondering if we should merge it in this state?

jmartinesp:
Could we do the same as iOS, and use a feature flag to decide whether to use the legacy behaviour for this, or this new one?

bmarty:
> Could we do the same as iOS, and use a feature flag to decide whether to use the legacy behaviour for this, or this new one?

Yes, we could. I wanted to avoid this, but if it helps, let's do it.

bmarty:
> I'm still seeing an issue I saw before we reverted this: when a room has a previous message with a mention to you, received while the app wasn't running, the unread count and mention won't be gone until a new message in that room (not coming from the current user) is received.

Can we "accept" this as a limitation, and open an issue somewhere (on the Rut side?), to unblock this PR again?

I do not want to add a feature flag to mitigate this. Let's keep it simple. Do you agree @jmartinesp?

jmartinesp:
> I do not want to add a feature flag to mitigate this. Let's keep it simple. Do you agree @jmartinesp?

I think it would be better to have a FF since the new behaviour is not working that well, and can make the users believe we also have stuck 'notifications' as in EW. Maybe it's better to wait until https://github.com/matrix-org/matrix-rust-sdk/pull/2978 is merged an check if that fixes the issue?

If you think we should merge it as is we can do that, I'm just a bit reluctant to do it because of how users might react if the app starts behaving weirdly here.

bmarty:
https://github.com/matrix-org/matrix-rust-sdk/pull/2978 has just been merged 🎉 . I will do some test with it.

bmarty:
@jmartinesp splitting the PR to 2 parts. First part is some rework, and once merged I will create another PR for the change about notification count.

bmarty:
I will iterate to ensure that existing screenshots do not change.

bmarty:
OK, I will record screenshot because actually the current preview is better (latest message with timestamp). There is less change, so that's fine.

jmartinesp (approved):
LGTM, thanks!

tests/uitests/src/test/snapshots/images/ui_S_t[f.roomlist.impl.components_RoomSummaryRow_null_RoomSummaryRow-Day-9_10_null_1,NEXUS_5,1.0,en].png
Comment (bmarty) [inline tests/uitests/src/test/snapshots/images/ui_S_t[f.roomlist.impl.components_RoomSummaryRow_null_RoomSummaryRow-Day-9_10_null_1,NEXUS_5,1.0,en].png:1]: Having a timestamp and no last message should not happen, but I will fix in a future PR.

PR #2146, opened 2024-01-02
Fix room transition animation happens twice

### LINKED ISSUE #2084: Room transition animation happens twice since #2059
### Steps to reproduce

1. Open a room from the room list
2. Observe that the app now transitions once to the "unloaded room" and again to the "loaded room"

It's mostly visible on slower devices (my phone is new so is pretty quick, but enabling battery saver emulates it a little)

#### What did you expect?
The previous behaviour. Reverting #2059 goes back to the old/expected behaviour.
Specifically, it seems that the issue is that there is a transition from the "unloaded timeline" to the "loaded timeline" whereas before it didn't do that. 

#### What happened instead?
Here's a video I tried to capture of it happening. It's most visible in the last room opened

https://github.com/element-hq/element-x-android/assets/480955/871bf8b6-fab2-4865-ba5c-db422f91c5bc

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

Yes

No

ganfra:
Closes #2084

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/room/RoomFlowNode.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/room/RoomFlowNode.kt
@@ -26,6 +26,7 @@ import androidx.compose.runtime.getValue
 import androidx.compose.ui.Modifier
 import androidx.lifecycle.lifecycleScope
 import com.bumble.appyx.core.modality.BuildContext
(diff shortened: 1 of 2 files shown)


jmartinesp (approved):
LGTM, nice catch!

Issue #2132, opened 2023-12-29
Scrolling in the timeline can be frozen


bmarty:
- Open a room with some messages
- scroll a bit to the history, a few pixels only
- try to scroll back to the bottom: the app behaves like if it was at the bottom of the timeline, i.e. does not scroll to the bottom

To unblock, scroll more to the top, then you can scroll back to the bottom.

`S-Major` since it gives the feeling to the user that the app contains stupid bugs.

Reproduce on API33 emu and on API30 devices.

PR #2150, opened 2024-01-03
Revert compose material3 to 1.2.0-alpha11

ganfra:
Fixes https://github.com/element-hq/element-x-android/issues/2132

I'll dig a bit more in the changes between those versions, and see if an issue already exists on the jetpack repository.


jmartinesp (approved):
It seems to be fixed, thanks!

Issue #2099, opened 2023-12-23
Message bubbles render trailing newlines from formatted HTML body


frebib:
I wouldn't expect such a tall bubble here. The timestamp also looks too far left which is maybe causing this?
![Screenshot_20231223-114648.png](https://github.com/element-hq/element-x-android/assets/775104/cff5136f-9592-4233-a53b-f6f10a7a0490)

```
{"type":"m.room.message","sender":"@echo:nerdhouse.io","content":{"msgtype":"m.notice","body":"a","format":"org.matrix.custom.html","formatted_body":"\u003cp\u003ea\u003c/p\u003e\n"},"origin_server_ts":1703325271499,"unsigned":{"age":93},"event_id":"$Tkq2w8nsVlVnOMcWM9bDX5c_unMN-he2qi6CDg8QP90"}
```

This is the first nightly after 0.4.0

frebib:
I've seen this happen with single emojis and even rich-formatted messages that are short too

bmarty:
This will probably be fixed by #2113

frebib:
Unfortunately #2113 does not fix this

jmartinesp:
> Unfortunately #2113 does not fix this

Are you sure? I can't reproduce this anymore.

bmarty:
I think it's due to the `"formatted_body":"\u003cp\u003ea\u003c/p\u003e\n"`, that I had not seen at first read.

frebib:
Yeah you could argue it's actually correct behaviour and maubot echo is at fault for producing an invalid formatted body?
```
{
  "type": "m.room.message",
  "sender": "@echo:nerdhouse.io",
  "content": {
    "msgtype": "m.notice",
    "body": "✨",
    "format": "org.matrix.custom.html",
    "formatted_body": "<p>✨</p>\n"
  },
  "origin_server_ts": 1704377737269,
  "unsigned": {
    "age": 93
  },
  "event_id": "$n8KLSPGdKjk0Ka4JHGGAdUyO3O8zLsTNzVF9G0CH0tw",
  "room_id": "!ping-v11:maunium.net"
}
```

Most other clients I've used seem to strip the trailing whitespace from these messages.
I don't see any such mentions of how clients are expected to handle this edge-case in the spec with a quick grep

tulir:
[...]

PR #2169, opened 2024-01-05
Trim whitespace at the end of messages

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Trim whitespace *at the end* of both formatted and plain text messages. We don't trim leading whitespace in case it's used for formatting or some nice ASCII art.

```
             (__)    
     `\------(oo)
       ||    (__)
       ||w--||     \|/
   \|/
```

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/2099.

## Tests

- Send a message with one or several trailing whitespaces or new lines.

In the timeline those extra whitespace chars should be gone.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentMessageFactory.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentMessageFactory.kt
@@ -74,7 +74,7 @@ class TimelineItemContentMessageFactory @Inject constructor(
             is EmoteMessageType -> {
                 val emoteBody = "* $senderDisplayName ${messageType.body}"
                 TimelineItemEmoteContent(
-                    body = emoteBody,
+                    body = emoteBody.trimEnd(),
                     htmlDocument = messageType.formatted?.toHtmlDocument(prefix = "* $senderDisplayName"),
                     formattedBody = parseHtml(messageType.formatted, prefix = "* $senderDisplayName") ?: emoteBody.withLinks(),
                     isEdited = content.isEdited,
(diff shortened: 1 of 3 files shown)


bmarty (approved):
LGTM, but this code will trim only the last line, IIUC how `trimEnd()` works. Would it have sense to `trimEnd` all lines of the message? For instance having (with `_` replacing ` ` for clarity):
```
a____
b__
c
```
become
```
a
b
c
```
?
We could replace `trimEnd()` with `lines().joinToString(transform = String::trimEnd, separator = "\n")` to do it.

On the other side I do not really like to apply post treatment on Events that have been sent...

jmartinesp:
> LGTM, but this code will trim only the last line, IIUC how `trimEnd()` works. Would it have sense to `trimEnd` all lines of the message? For instance having (with `_` replacing ` ` for clarity):

I don't think that's the case: `trimEnd()` actually does `trimEnd(Char::isWhitespace)`, which will remove any whitespace character. And `'\n'.isWhitespace()` returns true. The docs say:

> A character is a Java whitespace character if and only if it satisfies one of the following criteria:
It is a Unicode space character (SPACE_SEPARATOR, LINE_SEPARATOR, or PARAGRAPH_SEPARATOR) but is not also a non-breaking space ('\u00A0', '\u2007', '\u202F').
...

frebib:
This doesn't seem to solve the issue in the ping room, because I think you're only trimming the `body` and not the `formattedBody` where the issue actually is?

jmartinesp:
> This doesn't seem to solve the issue in the ping room, because I think you're only trimming the `body` and not the `formattedBody` where the issue actually is?

Sorry, you're right. I applied the `trimEnd()` at the wrong place 🤦 .

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentMessageFactory.kt
```diff
@@ -17,6 +17,7 @@
 package io.element.android.features.messages.impl.timeline.factories.event
 
 import android.text.Spannable
+import android.text.SpannedString
```
ElementBot, line 20: - :warning: The import 'android.text.SpannedString' is unused.

frebib:
Looks great! Thanks

Issue #2172, opened 2024-01-05
No indication that user list is loading when inviting to room


ganfra:
See https://github.com/element-hq/element-meta/issues/1950

PR #2173, opened 2024-01-05
Feature/fga/invite user loader

ganfra:
Closes https://github.com/element-hq/element-x-android/issues/2172


features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/components/SearchUserBar.kt
```diff
@@ -99,6 +101,11 @@ fun SearchUserBar(
                 )
             }
         },
+        contentSuffix = {
```
ganfra, line 104: I prefer using the content suffix here instead of adding a new state as it's not a common pattern in the SearchBar for now.

libraries/usersearch/api/src/main/kotlin/io/element/android/libraries/usersearch/api/UserSearchResult.kt
```diff
@@ -22,3 +22,8 @@ data class UserSearchResult(
     val matrixUser: MatrixUser,
     val isUnresolved: Boolean = false,
 )
+
+data class UserSearchResults(
```
jmartinesp, line 26: Maybe rename it to `UserSearchResultState`? It's quite easy to mistake `UserSearchResult` with `UserSearchResults`.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/SearchBar.kt
```diff
@@ -184,10 +184,10 @@ object ElementSearchBarDefaults {
 @Immutable
 sealed interface SearchBarResultState<in T> {
     /** No search results are available yet (e.g. because the user hasn't entered a search term). */
-    class NotSearching<T> : SearchBarResultState<T>
+    class Empty<T> : SearchBarResultState<T>
```
jmartinesp, line 187: I feel like `NotSearching` was more meaningful than `Empty`, if I saw this class for the first time I'd have some trouble differentiating between `Empty `and `NoResultsFound`, to be honest.

Maybe we could call it `Idle` or `Initial`?

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/userlist/DefaultUserListPresenter.kt
```diff
@@ -57,25 +59,29 @@ class DefaultUserListPresenter @AssistedInject constructor(
...
-            }
+            }.launchIn(this)
         }
 
         return UserListState(
             searchQuery = searchQuery,
             searchResults = searchResults,
             selectedUsers = selectedUsers.toImmutableList(),
             isSearchActive = isSearchActive,
+            isFetchingSearchResults = isFetchingSearchResults,
```
jmartinesp, line 84: I'm wondering if we shouldn't rename `isSearchActive` to something like `displaySearchContent` and `isFetchingSearchResults` to `isSearching` to make it simpler to understand. What do you think?

jmartinesp (approved):
LGTM, thanks. I have a few suggestions about naming, let me know if they make sense.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/SearchBar.kt
ganfra: Yeah but NotSearching wasn't true, it's just an Empty state where you show nothing. But Initial was also on my mind, so I can change if u prefer.

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/userlist/DefaultUserListPresenter.kt
ganfra: Yeah it makes sense

libraries/usersearch/api/src/main/kotlin/io/element/android/libraries/usersearch/api/UserSearchResult.kt
ganfra: It was what I had previously, but it's not really a `State` neither...

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/userlist/DefaultUserListPresenter.kt
ganfra: @jmartinesp  Actually, `active` is the naming used by the `material3.SearchBar`, so maybe we should keep it...

PR #2186, opened 2024-01-08
Sync Strings

ElementBot:


bmarty (approved):
OR for the German translations (fixes) coming from @VolkerJunginger (mentionning you in case you want to double check the incoming changes).

PR #2190, opened 2024-01-08
Attempt to exclude Preview function from code coverage computation.

bmarty:
Preview are using empty lambda that are never covered during the Preview test.
Draft, checking what the CI is reporting.

### DIFF
--- a/libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/coverage/ExcludeFromJacocoGeneratedReport.kt
+++ b/libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/coverage/ExcludeFromJacocoGeneratedReport.kt
@@ -1,0 +1,24 @@
+/*
+ * Copyright (c) 2024 New Vector Ltd
+ *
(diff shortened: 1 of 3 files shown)


libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/coverage/ExcludeFromJacocoGeneratedReport.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package io.element.android.libraries.architecture.coverage
+
+@Retention(AnnotationRetention.RUNTIME)
+@Target(
+    AnnotationTarget.FUNCTION,
+    AnnotationTarget.ANNOTATION_CLASS,
+)
+annotation class ExcludeFromJacocoGeneratedReport
```
bmarty, line 24: According to https://github.com/jacoco/jacoco/wiki/FilteringOptions#annotation-based-filtering, this annotation should meet the requirements to make the targets excluded form Jacoco report and so Kover.

bmarty:
It does not seem to work :/:
image

PR #2191, opened 2024-01-08
Do not use sublist as it crashes if `index > items.count()`.

bmarty:
Fixes https://github.com/element-hq/element-x-android-rageshakes/issues/1301

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
@@ -247,7 +247,8 @@ class TimelinePresenter @AssistedInject constructor(
     }
 
     private fun getLastEventIdBeforeOrAt(index: Int, items: ImmutableList<TimelineItem>): EventId? {
-        for (item in items.subList(index, items.count())) {
+        for (i in index..items.count()) {
+            val item = items[i]
             if (item is TimelineItem.Event) {
                 return item.eventId
             }


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
```diff
@@ -247,7 +247,8 @@ class TimelinePresenter @AssistedInject constructor(
     }
 
     private fun getLastEventIdBeforeOrAt(index: Int, items: ImmutableList<TimelineItem>): EventId? {
-        for (item in items.subList(index, items.count())) {
+        for (i in index..items.count()) {
```
jmartinesp, line 250: Should we also check if the index is within `0..items.count()` to avoid other possible crashes?

bmarty: if `index > items.count()` it will not crash (check on Kotlin playground), but I think it will crash a lot because because `i` can have the value `items.count()`. Good spot!
I do not think index can be < 0 because value is coming from `lazyListState.firstVisibleItemIndex`.

bmarty: https://github.com/element-hq/element-x-android/pull/2191/commits/0756a2d3e44a0b288cbc8f3ec7dc6db235da987c

I also check that `for (i in 4 until 2) {}` is valid (and no op)

jmartinesp (approved):
LGTM, thanks!

PR #2194, opened 2024-01-08
Update timeline items read receipts when the room members are loaded

### LINKED ISSUE #2176: Read receipt avatars & names sometimes fail to show until the room is reopened
### Steps to reproduce

1. Start Element X, open any room
2. Notice how some read receipt avatars are missing and you just see the initial
3. Go back and open the same room again
4. Avatars show up.

(It doesn't happen 100% of the times)

#### What did you expect?
Avatars should always show up (eventually)
#### What happened instead?
They don't show up until the room is reopened

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

Yes

Yes

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

When we fetch an item from the timeline cache, we now always update the read receipts with the latest data from the room members list.

## Motivation and context

Fixes #2176 .

## Tests

- Open a room (ideally a small-ish one so it doesn't take too long to fetch the room members list).
- Check the read receipts.

If after a few seconds they change and now contain the right images and user names, it's working.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


frebib:
This seems to be a lot more reliable than before 👍. Avatars do all eventually load although it can still take a while (several seconds) in larger rooms as it seems to wait for all avatars to be available before showing them. What would be the performance/UX implications of displaying them one at a time as they're available?
Is there any amount of caching for these images in EXA/rust-sdk? It seems to be slow after first loading the app each time but after that's basically instant.

jmartinesp:
> This seems to be a lot more reliable than before 👍. Avatars do all eventually load although it can still take a while (several seconds) in larger rooms as it seems to wait for all avatars to be available before showing them. What would be the performance/UX implications of displaying them one at a time as they're available? Is there any amount of caching for these images in EXA/rust-sdk? It seems to be slow after first loading the app each time but after that's basically instant.

The issue that makes the images take so long to load is that, although we do have a cache for members in the Rust SDK, if I'm not mistaken (so take it with a grain of salt), the current `Room.members()` method from the Rust SDK only returns this cached version *if it was loaded in the current session* / for the current usage of the underlying `Room` object (I can't remember which one was right). 

Technically we could call `Room.getMember(userId)` for each member at the moment, but it will only return something *after this initial member loading has happened*. We could call the profile endpoint if I'm not mistaken, but that wouldn't store the results in the Rust cache of room members.

So to improve this in the future I we have 3 options:

1. Make the `Room.members()` call return the currently cached members first, then the new ones If my understanding of the current behaviour is right.
2. Load members in a paginated way so we only load 100/200/500 at a time instead of thousands in large rooms.
3. Try to lazy load the avatar and display name using the profile endpoint, save them in an alternative cache instead of using the room members one.

Also, loading times on Android are a lot higher on Android than on iOS partly because we have to map those thousand of items to plain Kotlin objects to ensure memory safety given the current UniFFi memory model for Kotlin. This should be fixed in the near future, which should make the app more performant overall, but until then this is what we have.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/TimelineItemsFactory.kt
```diff
@@ -87,7 +87,16 @@ class TimelineItemsFactory @Inject constructor(
                     newTimelineItemStates.add(timelineItemState)
                 }
             } else {
-                newTimelineItemStates.add(cacheItem)
+                val updatedItem = if (cacheItem is TimelineItem.Event && roomMembers.isNotEmpty()) {
+                    eventItemFactory.update(
+                        timelineItem = cacheItem,
+                        receivedMatrixTimelineItem = timelineItems[index] as MatrixTimelineItem.Event,
```
bmarty, line 93: Are you sure that `timelineItems[index]` can always be cast to `MatrixTimelineItem.Event`?

jmartinesp: It uses the unique id to identify cached items, so if the cached item is an event I expect the updated item with that id to also be an event. Maybe that's not always the case?

ganfra: Yes the cached value is obviously the same, otherwise there is a big issue :D

jmartinesp:
It seems like we want to go with option 1 for now:

> Make the Room.members() call return the currently cached members first, then the new ones If my understanding of the current behaviour is right.

So it's probably better to wait until that is merged in the SDK side to also merge this PR.

bmarty (approved):
Sorry for the delay to approve the PR, I think it's fine!

PR #2197, opened 2024-01-09
Make the room settings screen update automatically

### LINKED ISSUE #921: Local echo for new room picture isn't there
### Steps to reproduce

1. Go to a room where you are admin
2. Change the room avatar
3. Save

#### What did you expect?
I see a local echo of the room avatar when I land back in the room detail view after saving the new photo

#### What happened instead?
I still see a blank picture (there was no room avatar before)

### Your phone model

Pixel 6a

### Operating system version

Graphene OS

Nightly

### Homeserver

matrix.org

### Will you send logs?

No

No

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Makes the room name, topic and avatar update with the `RoomInfo` contents.

## Motivation and context

Fixes #921  (the issue name is a bit misleading).

## Screenshots / GIFs

https://github.com/element-hq/element-x-android/assets/480955/a1ccc08a-151a-4b05-aca6-325a0571f258

## Tests

- Go to a room, open its settings.
- Either from another client or this same client, change either the name, topic and avatar or a combination of them.

They should update automatically in this screen.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/RoomDetailsPresenter.kt
```diff
@@ -64,6 +64,12 @@ class RoomDetailsPresenter @Inject constructor(
         val scope = rememberCoroutineScope()
         val leaveRoomState = leaveRoomPresenter.present()
         val canShowNotificationSettings = remember { mutableStateOf(false) }
+        val roomInfo = room.roomInfoFlow.collectAsState(initial = null).value
```
ganfra, line 67: Should we get some composable extension method on room, to get the name, avatar etc etc? Because we have the same logic on `MessagesPresenter`, but we are not using the same fallback...

ganfra: I'd say to use `AsyncData` everywhere

jmartinesp: We need some non-null values for the room name, so I'm not sure using `AsyncData` is the best approach. Maybe we should do something like:

- If `roomInfo` is null, use the fallback values.
- If it's not, use whatever it contains even if it's null, not using the fallback values at all.

And keep the simple `String?` values instead of `AsyncData<String?>`. Or is there some place where we need to check the state of some of the existing `AsyncData` (as if, is it loading, has it failed...) ?

ganfra: Yeah I don't know what is the best strategy here... I'd say if it's loading we should just show a placeholder like we do in MessagesView, otherwise we might have 2 different values, which might be worse?

jmartinesp: I took another look and it seems like for all those `AsyncData` backed values in `MessagesPresenter` we go from `Uninitialized` to `Success` without any intermediate state or failure state. I guess this is because `MatrixRoom.roomInfoFlow` doesn't have any of these states and so we can't really know what's happening behind scenes, and I don't think we can change this.

The best we *could do* is change the `Uninitialized` to `Loading` and have a `Loading` and hopefully a `Success` value later. However, I don't really see how that's really different from just having a nullable value...

jmartinesp: In `RoomDetailsPresenterTests` there is a `present - initial state is created from room if roomInfo is null` test which verifies that we're using fallback values for everything we want to display if there is no `RoomInfo`, so it seems like we *wanted* to have different behaviours. 

The question is whether we want to keep this or not. iOS does use the current values in `RoomProxy` (their `MatrixRoom` counterpart) until they receive a new room info.

jmartinesp: So... what should we do here @ganfra ?

ganfra: Lets merge like this then

ganfra (approved):

PR #2213, opened 2024-01-11
Feature/bma/fdroid

### LINKED ISSUE #1418: FOSS Version
Hi, I wanted to raise some awareness for some objectives that the Android-FOSS community would enjoy seeing. Those are:

1. Support for ntfy-push. This should count as the greatest FOSS-priority.
2. Non-Fdroid FOSS releases, preferrably on GitHub
3. Split APKs

About Nr. 2: While Element-Android has github releases that one can use, they only provide the playstore version of the app. Furthermore, the version available on fdroid often lags behind playstore by a significant timespan. You might make the argument that fdroid only receives stable versions of the app after being tested by playstore users; However for pre-releases, github already provides the necessary facilities. 

I hope this could gain some traction in the early development of Element-X.

bmarty:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Add flavor to be able to build Element X Android for F-Droid

There is still a thing to do about `firebaseAppDistribution`. We cannot use this plugin for F-Droid build, but I am not sure there is some action to do about this. According to https://gitlab.com/fdroid/rfp/-/issues/2556#note_1593343486 it could be stripped out by the F-Droid build recipe. Also note that nothing particular is done about this on the Element Android project (see [here](https://github.com/element-hq/element-android/blob/develop/vector-app/build.gradle#L288))

Creating as a draft PR to see how the CI react to this change.

Also we may have to fix thing when doing the next release.

I have done some test and with the [Ntfy](https://f-droid.org/fr/packages/io.heckel.ntfy/) application installed, UnifiedPush is working as expected, after login to a clean installation of Element X F-Droid 🚀 .

## Motivation and context

Part of #1418

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


app/build.gradle.kts
```diff
@@ -143,6 +147,20 @@ android {
     buildFeatures {
         buildConfig = true
     }
+    flavorDimensions += "store"
+    productFlavors {
+        create("gplay") {
+            dimension = "store"
+            isDefault = true
+            buildConfigField("String", "SHORT_FLAVOR_DESCRIPTION", "\"G\"")
+            buildConfigField("String", "FLAVOR_DESCRIPTION", "\"GooglePlay\"")
```
jmartinesp, line 156: Maybe you can remove the firebase distribution plugin from above and do something like this here:

```kotlin
create("gplay") {
            ...
            apply(plugin = libs.plugins.firebaseAppDistribution.get().pluginId)
            buildTypes.getByName("nightly") {
                project.extensions.configure<AppDistributionExtension>("firebaseAppDistribution") {
                        artifactType = "APK"
                        // We upload the universal APK to fix this error:
                        // "App Distribution found more than 1 output file for this variant.
                        // Please contact firebase-support@google.com for help using APK splits with App Distribution."
                        artifactPath = "$rootDir/app/build/outputs/apk/nightly/app-universal-nightly.apk"
                        // artifactType = "AAB"
                        // artifactPath = "$rootDir/app/build/outputs/bundle/nightly/app-nightly.aab"
                        // This file will be generated by the GitHub action
                        releaseNotesFile = "CHANGES_NIGHTLY.md"
                        groups = "external-testers"
                        // This should not be required, but if I do not add the appId, I get this error:
                        // "App Distribution halted because it had a problem uploading the APK: [404] Requested entity was not found."
                        appId = "1:912726360885:android:e17435e0beb0303000427c"
                    }
            }
        }
```

jmartinesp: Actually, there's no need for the whole `configure...` workaround, it seems like just using `firebaseAppDistribution { ... }` instead should work.

jmartinesp: Actually, my bad, I think this actually still applies it to all the variants... I would probably work if we added the flavors using `register` as that's lazy loaded, but that doesn't seem to be possible.

jmartinesp: In fact, the plugin seems to automatically apply itself to all flavors, so there's probably no way to improve this in any way:

```java
androidExt.getProductFlavors().all((flavor) -> {
    BuildVariantHelper.getExtensions(flavor).add("firebaseAppDistribution", AppDistributionExtension.class);
});
```

licaon-kter:
this recipe
```
  - versionName: 0.4.0
    versionCode: 40004002
    commit: 592bdb0a575a59d31a1402215d99cba40b2bae2f
    subdir: app
    sudo:
      - apt-get update
      - apt-get install -y openjdk-17-jdk-headless
      - update-java-alternatives -a
    gradle:
      - fdroid
    rm:
      - libraries/pushproviders/firebase
    prebuild:
      - sed -i -e '/plugins.firebaseAppDistribution/d' -e '/firebaseAppDistribution
        {/,/^            }/d' -e 's/UniversalApk = .*/UniversalApk = false/' -e 's/include("armeabi-v7a",
        "x86", "arm64-v8a", "x86_64"/include("arm64-v8a"/' -e '/ndk {/,/^        }/d' -e '/gplayImplementation/d'
        build.gradle.kts
      - sed -i -e '/gms/d' ../build.gradle.kts
      - sed -i -e '/firebase/d' ../gradle/libs.versions.toml ../plugins/build.gradle.kts
        ../plugins/src/main/kotlin/extension/DependencyHandleScope.kt
    scanignore:
      - settings.gradle.kts
    scandelete:
      - plugins
  ```

fails: [io.element.android.x_40004002.log.gz](https://github.com/element-hq/element-x-android/files/13919609/io.element.android.x_40004002.log.gz)

what did I miss ?

**Comment by bmarty** [comment]
Thanks for the feedback @licaon-kter.

In the log you provided, I see

```
e: java.rmi.ConnectIOException: error during JRMP connection establishment; nested exception is: 
	java.net.SocketTimeoutException: Read timed out
```

So maybe a network issue during the build? Just try again?

licaon-kter:
Connection to what during build?

jmartinesp:
Maybe some of the plugins (the dependency analysis one, the firebase one if it wasn't properly removed, etc.) need to connect to some external service that's blocked in your network? FWIW, I can't reproduce the issue when building `app:fdroidAssembleRelease`.

enigma9o7:
> Maybe some of the plugins (the dependency analysis one, the firebase one if it wasn't properly removed, etc.) need to connect to some external service that's blocked in your network? FWIW, I can't reproduce the issue when building `app:fdroidAssembleRelease`.

I think the point is a network connection should not be needed to build.

jmartinesp:
`org.owasp.dependencycheck` (repo [here](https://github.com/jeremylong/DependencyCheck)) does mention `java.rmi` in its Dockerfile, but then I wasn't able to find any other references in its source code. It would make sense that it had to access some external service to check for vulnerabilities *while running its task*, but that's not the case here...

And taking a closer look, both these errors happen on the KSP (Kotlin Symbol Processing) phase, which internally uses `java.rmi` for inter-process communication... maybe the KSP process somehow died due to lack of resources and the gradle process just reported it?

licaon-kter:
Builds fine on retry

jmartinesp (approved):

PR #2231, opened 2024-01-15
Replace 'leave room' text with 'leave conversation' for DMs

### LINKED ISSUE #2218: Change 'Leave room' in DMs to 'Leave conversation'
In the room details screen, change the 'Leave room' action to 'Leave conversation'.

The localazy key for this is `screen_room_details_leave_conversation_title` which links to `action.leave_conversation`.

jmartinesp:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [x] Other : add specific strings for DMs when leaving a DM.

## Content

Add alternate texts for:

- Room details screen, 'Leave room' action.
- 'Leave room' confirmation dialog.
- 'Leave room' action in the context menu when long clicking a room in the room list.

So instead of saying 'room' for DMs they say 'conversation'.

## Motivation and context

Closes #2218.

## Screenshots / GIFs

There are new screenshots in the PR.

## Tests

Try to leave a room and a DM either by using the room details screen or the long click context menu in the room list, check whether the right strings are used or not.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist


features/leaveroom/api/src/main/res/values/localazy.xml
```diff
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
+  <string name="leave_conversation_alert_empty_subtitle">"Are you sure that you want to leave this conversation? You\'re the only person here. If you leave, no one will be able to join in the future, including you."</string>
+  <string name="leave_conversation_alert_private_subtitle">"Are you sure that you want to leave this conversation? This conversation is not public and you won\'t be able to rejoin without an invite."</string>
+  <string name="leave_conversation_alert_subtitle">"Are you sure that you want to leave the conversation?"</string>
```
bmarty, line 5: I do not think that a user can see this string, since I guess all the DMs are not public.
So maybe for DM the application should ignore the public/private state of the room?

I just check, and currently, when leaving a DM, the string `leave_room_alert_private_subtitle` is displayed.

features/roomlist/impl/src/main/kotlin/io/element/android/features/roomlist/impl/RoomListPresenter.kt
```diff
@@ -116,10 +119,12 @@ class RoomListPresenter @Inject constructor(
                     }
                     displaySearchResults = !displaySearchResults
                 }
-                is RoomListEvents.ShowContextMenu -> {
+                is RoomListEvents.ShowContextMenu -> coroutineScope.launch {
+                    val isDm = client.getRoom(event.roomListRoomSummary.roomId).use { it?.isDm.orFalse() }
```
bmarty, line 123: Do you think it would be useful to add a field `isDm` to  `RoomListRoomSummary` instead?

bmarty (approved):
LGTM, but maybe the algo would need some change, see my comment.

features/roomlist/impl/src/main/kotlin/io/element/android/features/roomlist/impl/RoomListPresenter.kt
ganfra: Yes it'd be a good idea to expose it in `RoomSummary` first, as we create this with a `RoomInfo` we have already everything we need for this.

jmartinesp: I didn't realise we were building it there, I'll add it there then. I wonder if exposing `isDm` and `isOneToOne` separately actually makes sense. Do we really want to handle something differently if the room is direct but has 3-4 people in it? AFAICT we only distinguish between rooms and DMs.

features/leaveroom/api/src/main/res/values/localazy.xml
jmartinesp: Yeah, you're right. A DM will only be considered a DM when it's direct and has 2 users, so the `private` text here will always be triggered. The other two aren't needed. Thanks!

bmarty: OK, So I think only the string `leave_conversation_alert_subtitle` should be used in this case.

jmartinesp: I'd actually use the contents of `leave_conversation_alert_private_subtitle`:

> Are you sure that you want to leave this conversation? This conversation is not public and you won't be able to rejoin without an invite.

It seems the most accurate one, given the situation.

jmartinesp: Or maybe removing the 'This conversation is not public and" part, just saying "You won't be able to rejoin without an invite", I don't know.

features/leaveroom/impl/src/main/kotlin/io/element/android/features/leaveroom/impl/LeaveRoomPresenterImpl.kt
```diff
@@ -85,6 +86,7 @@ private suspend fun showLeaveRoomAlert(
 ) {
     matrixClient.getRoom(roomId)?.use { room ->
         confirmation.value = when {
+            room.isDm -> Dm(roomId)
```
bmarty, line 89: As discussed with @giomfo IRL today, there is another case that we will need to handle: when the last admin in leaving the room, we should warn the user.

features/leaveroom/api/src/main/res/values/localazy.xml
bmarty: > It seems the most accurate one, given the situation.

I agree, but also the most complex, and we tend to simplify the user life in the context of Element X. That's why a product POV may be helpful. CC @VolkerJunginger

PR #2237, opened 2024-01-16
Be robust with long userId

bmarty:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Fix a crash that was occuring when loging in with a long userId.

Also, in a separate commit, ensure that the client secret used for push is deleted when the session is destroyed.

## Motivation and context

More stable app and no discrimination for long userId. UserId can be up to 255 chars and it was used in a string pattern to build a filename. The app was crashing because of that. The PR rename the existing file to a shorter name, and new created file will have a short name.

## Tests

- Login using an account with a long name, for instance `@a23456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123:matrix.org` and observe that Element X works like a charm. Without this change, the app is crashing.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/pushstore/api/src/main/kotlin/io/element/android/libraries/pushstore/api/clientsecret/PushClientSecret.kt
+++ b/libraries/pushstore/api/src/main/kotlin/io/element/android/libraries/pushstore/api/clientsecret/PushClientSecret.kt
@@ -29,9 +29,4 @@ interface PushClientSecret {
      * Return null if not found.
      */
     suspend fun getUserIdFromSecret(clientSecret: String): SessionId?
-
-    /**
-     * To call when the user signs out.
-     */
-    suspend fun resetSecretForUser(userId: SessionId)
 }
--- a/libraries/pushstore/impl/src/main/kotlin/io/element/android/libraries/pushstore/impl/UserPushStoreDataStore.kt
+++ b/libraries/pushstore/impl/src/main/kotlin/io/element/android/libraries/pushstore/impl/UserPushStoreDataStore.kt
@@ -40,7 +40,7 @@ class UserPushStoreDataStore(
     private val context: Context,
     userId: SessionId,
 ) : UserPushStore {
-    // Hash the sessionId to get ride of exotic chars and take only the first 16 chars.
+    // Hash the sessionId to get rid of exotic chars and take only the first 16 chars.
     // The risk of collision is not high.
     private val preferenceName = "push_store_${userId.value.hash().take(16)}"
 
--- a/libraries/pushstore/impl/src/main/kotlin/io/element/android/libraries/pushstore/impl/UserPushStoreDataStore.kt
+++ b/libraries/pushstore/impl/src/main/kotlin/io/element/android/libraries/pushstore/impl/UserPushStoreDataStore.kt
@@ -100,7 +100,7 @@ class UserPushStoreDataStore(
         context.dataStore.edit {
             it.clear()
         }
-    }
-
+        // Also delete the file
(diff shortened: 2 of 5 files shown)


libraries/pushstore/api/src/main/kotlin/io/element/android/libraries/pushstore/api/clientsecret/PushClientSecret.kt
```diff
@@ -29,9 +29,4 @@ interface PushClientSecret {
      * Return null if not found.
      */
     suspend fun getUserIdFromSecret(clientSecret: String): SessionId?
-
-    /**
-     * To call when the user signs out.
-     */
-    suspend fun resetSecretForUser(userId: SessionId)
```
bmarty, line 36: The method does not need to be exposed in the API, the impl will take care of this.

libraries/pushstore/impl/src/main/kotlin/io/element/android/libraries/pushstore/impl/UserPushStoreDataStore.kt
```diff
@@ -37,7 +40,24 @@ class UserPushStoreDataStore(
...
+        // Migrate legacy data. Previous file can be too long if the userId is too long. The userId can be up to 255 chars.
+        // Example of long file path, with `averylonguserid` replacing a very longer name
+        // /data/user/0/io.element.android.x.debug/files/datastore/push_store_@averylonguserid:example.org.preferences_pb
+        val legacyFile = context.preferencesDataStoreFile("push_store_$userId")
+        if (legacyFile.exists()) {
+            Timber.d("Migrating legacy push data store for $userId")
+            if (!legacyFile.renameTo(context.preferencesDataStoreFile(preferenceName))) {
+                Timber.w("Failed to migrate legacy push data store for $userId")
+            }
+        }
```
bmarty, line 57: Done in the main thread, but maybe not a big deal?

libraries/pushstore/impl/src/main/kotlin/io/element/android/libraries/pushstore/impl/UserPushStoreDataStore.kt
```diff
@@ -80,5 +100,7 @@ class UserPushStoreDataStore(
         context.dataStore.edit {
             it.clear()
         }
+        // Also delete the file
+        context.preferencesDataStoreFile(preferenceName).delete()
```
bmarty, line 104: Now that the preference are emptied, the file can be removed as well to keep the local storage tidied up.

jmartinesp (approved):
Thanks for the fixes! They LGTM, I have a couple of comments, and I'm a bit worried that since we're now observing sessions and modifying files from `PushClientSecretImpl` and it's not a singleton it might cause issues? Could we double check that before merging?

libraries/pushstore/impl/src/main/kotlin/io/element/android/libraries/pushstore/impl/UserPushStoreDataStore.kt
bmarty: Yes, I think it's acceptable. I wanted to include this in the `by preferencesDataStore(name = preferenceName)`, but did not find a good way to do it.

Issue #2236, opened 2024-01-15
OIDC-authed client got spontaneously logged out.


ara4n:
### Steps to reproduce

1. Open app
2. Discover you've been logged out

#### What did you expect?

No unexpected logouts.

#### What happened instead?

@kongo09 got unexpectedly logged out

### Your phone model

Pixel 5

### Application version and app store

0.4.2-nightly 

### Homeserver

element.io

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

kongo09:
It happened after updating the Nightly.

And then happened again later over the day just out of the blue. The second time, I got the following screen with none of the items being true:

![Screenshot_20240115-182446](https://github.com/element-hq/element-x-android/assets/640121/ae02558f-6f1c-4922-b009-7d8fbba04ba8)

spaetz:
It did happen to me on synapse 1.98 yesterday with plain old password auth, no OIDC involved. Not sure if that is the deciding factor.

jmartinesp:
It happened to me several times today while testing on emulators. I've uploaded logs here: https://github.com/element-hq/element-x-android-rageshakes/issues/1355

spaetz:
Just becausr this has been tagged OIDC, the logouts happen in my synapse server with plain old password auth. No OIDC involved.

spaetz:
It happened again this morning (still on EXA 0.4.0), and I noticed that I was logged out of ALL of m sessions by synapse (Element Web claimed I had been logged out for security reasons). And when I logged in via EW, I only had one out of 4 sessions left.
Not sure if this is still caused by this issue or if it is some other issue (in EW or synapse).

spaetz:
Indeed, I keep getting logged out once a day with all my users despite not using EXA, so my issue is clearly not an EXA issue, please ignore my debug logs.

jmartinesp:
FWIW I haven't had any more unexpected logouts after https://github.com/element-hq/element-x-android/pull/2251. @kongo09, have you reproduced this issue again in the latest nightlies (since the end of last week)?

bmarty:
I think we can close the issue now. Feel free to re-open if necessary. Thanks @jmartinesp for the fix!

manuroe:
@VolkerJunginger is it an issue you see on EXA nightly?

VolkerJunginger:
Happens to me on iOS & Android nightly. Most of the time after updating the apps.

PR #2251, opened 2024-01-18
Try mitigating unexpected logouts

jmartinesp:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

- Wraps fetch/save operations for `SessionData` using a `Mutex`, to try to prevent concurrent access.
- Also added some more logs so we can understand exactly where the session refresh is failing.

## Motivation and context

Try to find what's causing #2236 .

## Tests

- Log in, try using the app for some time until a few token refreshes are done. 

If there are no deadlocks or performance issues we can merge it.

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
@@ -20,7 +20,6 @@ import io.element.android.libraries.androidutils.file.getSizeOfFiles
 import io.element.android.libraries.androidutils.file.safeDelete
 import io.element.android.libraries.core.coroutine.CoroutineDispatchers
 import io.element.android.libraries.core.coroutine.childScope
-import io.element.android.libraries.core.hash.md5
 import io.element.android.libraries.matrix.api.MatrixClient
 import io.element.android.libraries.matrix.api.core.ProgressCallback
 import io.element.android.libraries.matrix.api.core.RoomId
@@ -140,8 +139,8 @@ class RustMatrixClient(
                 // TODO handle isSoftLogout parameter.
                 appCoroutineScope.launch {
                     val existingData = sessionStore.getSession(client.userId())
-                    val anonymizedToken = existingData?.accessToken?.md5()
-                    Timber.d("Removing session data with token: $anonymizedToken.")
+                    val anonymizedToken = existingData?.accessToken?.takeLast(4)
+                    Timber.d("Removing session data with token: '...$anonymizedToken'.")
                     if (existingData != null) {
                         // Set isTokenValid to false
                         val newData = client.session().toSessionData(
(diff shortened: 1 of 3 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
```diff
@@ -139,15 +140,24 @@ class RustMatrixClient(
                 // TODO handle isSoftLogout parameter.
                 appCoroutineScope.launch {
                     val existingData = sessionStore.getSession(client.userId())
+                    val anonymizedToken = existingData?.accessToken?.md5()
```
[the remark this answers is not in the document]
bmarty: I think it's probably safer to use `.takeLast(4)` instead of `.md5()`.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
```diff
@@ -158,11 +168,18 @@ class RustMatrixClient(
             Timber.w("didRefreshTokens()")
             appCoroutineScope.launch {
                 val existingData = sessionStore.getSession(client.userId()) ?: return@launch
+                val anonymizedToken = client.session().accessToken.md5()
+                Timber.d("Saving new session data with token: $anonymizedToken. Is new token valid: ${existingData.isTokenValid}")
```
bmarty, line 172: I am confused about the `isTokenValid` set to `existingData.isTokenValid` at line 174 below. Should it be set to `true` in the case of a coming refresh token?

Also, the wording `Is new token valid` should be changed to `Was token valid`.

libraries/session-storage/impl/src/main/kotlin/io/element/android/libraries/sessionstorage/impl/DatabaseSessionStore.kt
```diff
@@ -67,30 +71,38 @@ class DatabaseSessionStore @Inject constructor(
             return
         }
 
-        // Copy new data from SDK, but keep login timestamp
-        database.sessionDataQueries.updateSession(
-            sessionData.copy(
-                loginTimestamp = result.loginTimestamp,
-            ).toDbModel()
-        )
+        sessionDataMutex.withLock {
```
bmarty, line 74: I would wrap the entire function body with the Mutex, to include the db query. But maybe there is a good reason not to do so?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
jmartinesp: Technically, you wrote that code so if you don't know, me neither 😅 . But yes, it's probably better to hardcode a `true` value there and fix the log.

bmarty: Yes, I know that I wrote the code, but it was a long time ago :).

bmarty (approved):
Thanks for the update.

PR #2290, opened 2024-01-25
Add 'send private read receipts' option in advanced settings

### LINKED ISSUE #2204: Advanced Settings: Enable/Disable Sending Read Receipts
From element-x-ios created by [Velin92](https://github.com/Velin92): element-hq/element-x-ios#2319  

We would like to add new advanced setting options that allows users to turn off and on public read receipts.
When they are on we just send the normal `read` type of read receipts, when the setting is off, we instead resort to sending the `privateRead` type of read receipts, so that rooms can still be seen as read but other users will not be able to see our read receipts.

The copies of the setting should be:
title: "Send read receipts"
description: "On disabling, others won't see your read receipt."

jmartinesp:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Adds a 'read receipt' option to advanced settings so the client can save the last read message but won't share it with the other users in the rooms.

## Motivation and context

Closes #2204.

## Tests

- Go to Advanced settings, disable this option.
- Go to any room with unread messages where you have another account in the members. This should automatically mark the room as read.
- Get out of the room, check the room is read.
- From the other account, check the read receipt for your user didn't change in the timeline.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/advanced/AdvancedSettingsView.kt
```diff
@@ -81,6 +81,18 @@ fun AdvancedSettingsView(
                 onChange = { state.eventSink(AdvancedSettingsEvents.SetDeveloperModeEnabled(it)) },
             ),
         )
+        ListItem(
+            headlineContent = {
+                Text(text = "Send read receipts")
+            },
+            supportingContent = {
+                Text(text = "On disabling, others won't see your read receipt.")
```
bmarty, line 89:
```suggestion
                Text(text = "On disabling, others won't see your read receipts.")
```

Also I think those 2 strings should be defined in Localazy.

libraries/preferences/api/src/main/kotlin/io/element/android/features/preferences/api/store/PreferencesStore.kt
```diff
@@ -28,6 +28,9 @@ interface PreferencesStore {
     suspend fun setCustomElementCallBaseUrl(string: String?)
     fun getCustomElementCallBaseUrlFlow(): Flow<String?>
 
+    suspend fun setPrivateReadReceiptsEnabled(enabled: Boolean)
+    fun isPrivateReadReceiptsEnabled(): Flow<Boolean>
```
bmarty, line 32: It's a bit annoying me that this settings is stored in `PreferencesStore`, which should be used for global settings only. The setting `isPrivateReadReceiptsEnabled` should have a Session scope.

That said, IIRC, we do not have (yet) a store for session preferences, *and* we do not support multi sessions yet. Hence, I think logging out and logging in again will keep the setting, and it's probably an unexpected side effect from the user POV.

bmarty (commented):
LGTM, 2 remaks. Also I think the screenshots have to be recorded.

libraries/preferences/api/src/main/kotlin/io/element/android/features/preferences/api/store/PreferencesStore.kt
jmartinesp: I'll see how difficult it is to create a session-scoped preference store.

features/preferences/impl/src/main/res/values/localazy.xml
```diff
@@ -6,6 +6,8 @@
   <string name="screen_advanced_settings_developer_mode">"Developer mode"</string>
   <string name="screen_advanced_settings_developer_mode_description">"Enable to have access to features and functionality for developers."</string>
   <string name="screen_advanced_settings_rich_text_editor_description">"Disable the rich text editor to type Markdown manually."</string>
+  <string name="screen_advanced_settings_send_read_receipts">"Read receipts"</string>
+  <string name="screen_advanced_settings_send_read_receipts_description">"If turned off, your read receipts won\'t be sent to anyone. You will still receive read receipts from other users."</string>
```
bmarty, line 10: > You will still receive read receipts from other users.

Except if they also disabled the public RR :)

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStore.kt
```diff
@@ -0,0 +1,61 @@
...
+import kotlinx.coroutines.flow.map
+import javax.inject.Inject
+
+@ContributesBinding(SessionScope::class)
+@SingleIn(SessionScope::class)
+class DefaultSessionPreferencesStore @Inject constructor(
+    @ApplicationContext context: Context,
+    currentSessionIdHolder: CurrentSessionIdHolder,
+) : SessionPreferencesStore {
+    private val privateReadReceiptsKey = booleanPreferencesKey("privateReadReceipts")
```
bmarty, line 41: Maybe rename to
```suggestion
    private val publicReadReceiptsKey = booleanPreferencesKey("publicReadReceipts")
```

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStore.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    private val store = PreferenceDataStoreFactory.create {
+        context.preferencesDataStoreFile("session_${currentSessionIdHolder.current}_preferences")
```
bmarty, line 44: You have to use a hash of the user id here, because it will crash if the user name is too long. See #2237

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStore.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    override suspend fun setSendPublicReadReceipts(enabled: Boolean) = update(privateReadReceiptsKey, enabled)
+    override fun isSendPublicReadReceiptsEnabled(): Flow<Boolean> = get(privateReadReceiptsKey, true)
+
+    override suspend fun clear() {
+        store.edit { it.clear() }
```
bmarty, line 51: The file may also be removed.

bmarty (commented):
Thanks for the change. There are a few remarks to handle.

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStore.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/2290/commits/0429c3d6c893f0f198e626ad2b24adf5acfa1689

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStore.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/2290/commits/da6c11ab0058741937ebe2324c712da03a0f1824

features/preferences/impl/src/main/res/values/localazy.xml
jmartinesp: I think that's kind of implicit, but please let's not bring back the discussion on the text used here, I already had enough of it 😅 .

bmarty: Yes, sure, I was sort of kidding :).

It reminds me the old days where mobile phone users could pay an option to get incoming call number displayed, and others could pay an option to hide their phone number when they're doing call :)

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStore.kt
```diff
@@ -0,0 +1,63 @@
...
+@SingleIn(SessionScope::class)
+class DefaultSessionPreferencesStore @Inject constructor(
+    @ApplicationContext context: Context,
+    currentSessionIdHolder: CurrentSessionIdHolder,
+) : SessionPreferencesStore {
+    private val sendPublicReadReceiptsKey = booleanPreferencesKey("sendPublicReadReceipts")
+    private val hashedUserId = currentSessionIdHolder.current.value.hash().take(16)
+
+    private val dataStoreFile = context.preferencesDataStoreFile("session_${hashedUserId}_preferences")
+    private val store = PreferenceDataStoreFactory.create { dataStoreFile }
```
bmarty, line 47: Just a small question, with this API, one can create a data store at any location, right?

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStore.kt
```diff
@@ -0,0 +1,63 @@
...
+        dataStoreFile.safeDelete()
+    }
+
+    private suspend fun <T> update(key: Preferences.Key<T>, value: T) {
+        store.edit { prefs -> prefs[key] = value }
+    }
+
+    private fun <T> get(key: Preferences.Key<T>, default: T): Flow<T> {
+        return store.data.map { prefs -> prefs[key] ?: default }
+    }
```
bmarty, line 62: 2 nice functions! I am tempted to extract them and reuse them elsewhere. I'll draft a quick PR.

libraries/preferences/api/src/main/kotlin/io/element/android/features/preferences/api/store/SessionPreferencesStore.kt
```diff
@@ -0,0 +1,26 @@
...
+
+package io.element.android.features.preferences.api.store
+
+import kotlinx.coroutines.flow.Flow
+
+interface SessionPreferencesStore {
+    suspend fun setSendPublicReadReceipts(enabled: Boolean)
+    fun isSendPublicReadReceiptsEnabled(): Flow<Boolean>
+
+    suspend fun clear()
```
bmarty, line 25: Note: this is never called. Not a big deal but should be fixed later.

bmarty (approved):
Thanks for the update! You can merge the PR like this.

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStore.kt
jmartinesp: I'm tempted to think so. This actually creates a `File(dataStoreDir, filename)` with `filename` being the string you pass here, and then calls `mkdirs()` so it *should* work that way.

libraries/preferences/api/src/main/kotlin/io/element/android/features/preferences/api/store/SessionPreferencesStore.kt
jmartinesp: Yeah, I need to figure out the right way to clean up this data when the user is logged out. I'll do it in another PR.

PR #2294, opened 2024-01-25
Ensure login and password exclude `\n`

### LINKED ISSUE #2263: Pasting on the username or password fields can add new lines
### Steps to reproduce

1. In the login screen, paste some text with new lines in either the username or the password fields.

#### What did you expect?

To either not be able to paste it, or can only paste the first line.

#### What happened instead?

The whole text is pasted, but the textfield is a single line one, so you have hidden text and new lines.

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

No

Yes

bmarty:
Closes #2263

### DIFF
--- a/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/loginpassword/LoginPasswordView.kt
+++ b/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/loginpassword/LoginPasswordView.kt
@@ -194,15 +194,17 @@ private fun LoginForm(
                 .onTabOrEnterKeyFocusNext(focusManager)
                 .testTag(TestTags.loginEmailUsername)
                 .autofill(autofillTypes = listOf(AutofillType.Username), onFill = {
-                    loginFieldState = it
-                    eventSink(LoginPasswordEvents.SetLogin(it))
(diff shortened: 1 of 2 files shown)


jmartinesp (approved):
LGTM, thanks!

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/loginpassword/LoginPasswordView.kt
```diff
@@ -272,6 +275,13 @@ private fun LoginForm(
     }
 }
 
+/**
+ * Ensure that the string does not contain any new line characters, which can happen when pasting values.
+ */
+private fun String.sanitize(): String {
+    return replace("\n", "")
```
jmartinesp, line 282: I'm checking with some other apps and most of them translate `\n` into a single whitespace. Maybe we should do the same, but I don't have a strong opinion on this.

bmarty: Yes, I was wondering as well. I think we can keep it like that, it's quite rare to have this user story anyway.

PR #2303, opened 2024-01-26
Remove session preferences on logout.

jmartinesp:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

- Listens to session status in `DefaultSessionPreferencesStoreFactory` and both remove the preferences from its cache and from disk when a user is logged out.

## Motivation and context

We should remove session preferences when the user logs out.

## Tests

- Log in using an account.
- Go to settings -> advanced settings -> disable 'read receipts'.
- Log out.
- Log in again.
- Go to settings -> advanced settings -> 'read receipts' should be enabled again.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 23
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request includes a new file under ./changelog.d. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#changelog
- [ ] Pull request includes screenshots or videos if containing UI changes
- [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
- [x] You've made a self review of your PR

PR #2312, opened 2024-01-29
Fix crash about several DataStores using the same file

### LINKED ISSUE #2308: Lots of crashes with "There are multiple DataStores active for the same file"
### Steps to reproduce

Dunno, but lots of anecdata and rageshakes showing this crash:

`01-26 12:25:00.094 25836 25836 E AndroidRuntime: java.lang.IllegalStateException: There are multiple DataStores active for the same file: /data/user/0/io.element.android.x/files/datastore/session_@user:example.com_preferences.preferences_pb. You should either maintain your DataStore as a singleton or confirm that there is no two DataStore's active on the same file (by confirming that the scope is cancelled).`

#### What did you expect?

no crash

#### What happened instead?

crash

### Your phone model

Pixel 7a

### Operating system version

Android 14

Element X nightly/0.4.2-nightly (Google Pixel 7a; Android 14; UQ1A.240105.004.2024012600; Sdk TODO)

### Homeserver

_No response_

### Will you send logs?

Yes

No

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Fix crash about several DataStores using the same file
	- Create `@SessionCoroutineScope` annotation to pass a session-managed coroutine scope to the DI.
	- Expose this scope from `MatrixClient`.
	- Rework SessionPreferencesStore file creation a bit. Use this coroutine scope to initialize the data store, so it's stopped when the scope is cancelled.
- Make sure the whole item in advanced settings screen triggers the toggle action, to be consistent with the rest of the app.

## Motivation and context

Fixes #2308 (I hope).

## Tests

To be honest, I'm not 100% sure how to trigger this issue. I tried logging out and in again, which was my theory for why this was happening, but sadly I couldn't reproduce the issue.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/element/android/x/di/SessionComponent.kt
+++ b/app/src/main/kotlin/io/element/android/x/di/SessionComponent.kt
@@ -24,9 +24,7 @@ import io.element.android.libraries.architecture.NodeFactoriesBindings
 import io.element.android.libraries.di.AppScope
 import io.element.android.libraries.di.SessionScope
 import io.element.android.libraries.di.SingleIn
-import io.element.android.libraries.di.annotations.SessionCoroutineScope
 import io.element.android.libraries.matrix.api.MatrixClient
-import kotlinx.coroutines.CoroutineScope
 
 @SingleIn(SessionScope::class)
 @MergeSubcomponent(SessionScope::class)
@@ -35,9 +33,6 @@ interface SessionComponent : NodeFactoriesBindings {
     interface Builder {
         @BindsInstance
         fun client(matrixClient: MatrixClient): Builder
-
-        @BindsInstance
-        fun sessionCoroutineScope(@SessionCoroutineScope coroutineScope: CoroutineScope): Builder
 
         fun build(): SessionComponent
     }
--- a/app/src/main/kotlin/io/element/android/x/di/SessionComponent.kt
+++ b/app/src/main/kotlin/io/element/android/x/di/SessionComponent.kt
@@ -24,7 +24,9 @@ import io.element.android.libraries.architecture.NodeFactoriesBindings
 import io.element.android.libraries.di.AppScope
 import io.element.android.libraries.di.SessionScope
 import io.element.android.libraries.di.SingleIn
+import io.element.android.libraries.di.annotations.SessionCoroutineScope
 import io.element.android.libraries.matrix.api.MatrixClient
+import kotlinx.coroutines.CoroutineScope
 
 @SingleIn(SessionScope::class)
 @MergeSubcomponent(SessionScope::class)
@@ -33,6 +35,10 @@ interface SessionComponent : NodeFactoriesBindings {
     interface Builder {
         @BindsInstance
         fun client(matrixClient: MatrixClient): Builder
+
+        @BindsInstance
+        fun sessionCoroutineScope(@SessionCoroutineScope coroutineScope: CoroutineScope): Builder
+
         fun build(): SessionComponent
     }
 
--- a/changelog.d/2308.bugfix
+++ b/changelog.d/2308.bugfix
@@ -1,0 +1,1 @@
+Fix 'There are multiple DataStores active for the same file' crashes
--- a/libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStoreFactory.kt
+++ b/libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStoreFactory.kt
@@ -31,11 +31,7 @@ class DefaultSessionPreferencesStoreFactory @Inject constructor(
 ) {
     private val cache = ConcurrentHashMap<SessionId, DefaultSessionPreferencesStore>()
 
-    fun get(sessionId: SessionId, sessionCoroutineScope: CoroutineScope) = cache.getOrPut(sessionId) {
+    fun get(sessionId: SessionId, sessionCoroutineScope: CoroutineScope): DefaultSessionPreferencesStore = cache.getOrPut(sessionId) {
         DefaultSessionPreferencesStore(context, sessionId, sessionCoroutineScope)
     }
-
-    fun remove(sessionId: SessionId) = cache.remove(sessionId)
-
-    fun clear() = cache.clear()
 }
--- a/libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStoreFactory.kt
+++ b/libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStoreFactory.kt
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2024 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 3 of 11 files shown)


app/src/main/kotlin/io/element/android/x/di/SessionComponent.kt
```diff
@@ -33,6 +35,10 @@ interface SessionComponent : NodeFactoriesBindings {
     interface Builder {
         @BindsInstance
         fun client(matrixClient: MatrixClient): Builder
+
+        @BindsInstance
```
ganfra, line 39: Why using `@BindsInstance` here? I'd just use `@Provides`

ganfra (approved):
One remark otherwise LGTM

app/src/main/kotlin/io/element/android/x/di/SessionComponent.kt
jmartinesp: Won't `@Provides` need the function to be implemented?

ganfra: We already have lots of @Provides in the `SessionMatrixModule`. I'd use the builder only for things we got from the outer world.

jmartinesp: Oh, now I see what you mean. Yeah, a provides is better here.

bmarty:
@jmartinesp  I can repro the crash systematically by:
- opening the app
- open a timeline
- clicking back twice to show the device home
- opening a room timeline from a notification.

But they are other recipes, for instance:
- open a room
- click back twice
- open the app again
- sometimes it crashes the app.

changelog.d/2308.bugfix
```diff
@@ -0,0 +1 @@
+Fix 'There are multiple DataStores active for the same file' crashes
```
bmarty, line 1: I am tempted to say that it's not necessary to add a file for the changelog, since the crash was not on the previous release.

jmartinesp:
@bmarty thanks for the reproducible steps! Sadly, it seems like the fix does not really fix the issue then... I'll keep trying.

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/SessionPreferencesModule.kt
```diff
@@ -0,0 +1,42 @@
...
+import io.element.android.libraries.di.SingleIn
+import io.element.android.libraries.di.annotations.SessionCoroutineScope
+import io.element.android.libraries.matrix.api.user.CurrentSessionIdHolder
+import kotlinx.coroutines.CoroutineScope
+
+@Module
+@ContributesTo(SessionScope::class)
+object SessionPreferencesModule {
+    @Provides
+    @SingleIn(SessionScope::class)
```
bmarty, line 33: I think `@SingleIn(SessionScope::class)` is not necessary here. `defaultSessionPreferencesStoreFactory` is already handling the concurrency access.

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/SessionPreferencesModule.kt
```diff
@@ -0,0 +1,42 @@
...
+    fun providesSessionPreferencesStore(
+        defaultSessionPreferencesStoreFactory: DefaultSessionPreferencesStoreFactory,
+        currentSessionIdHolder: CurrentSessionIdHolder,
```
bmarty, line 36: Since this object has the scope `SessionScope`, can we inject directly the `sessionId: String` instead of the `CurrentSessionIdHolder`?

Oh actually `CurrentSessionIdHolder` is not what I though. It's not really a `Holder`, but more an accessor (or a provider). I think it will be simpler to provide the `sessionId` in `SessionMatrixModule` like some other fields and remove `CurrentSessionIdHolder`. Out of scope of this PR though.

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStoreFactory.kt
```diff
@@ -0,0 +1,41 @@
...
+import java.util.concurrent.ConcurrentHashMap
+import javax.inject.Inject
+
+@SingleIn(AppScope::class)
+class DefaultSessionPreferencesStoreFactory @Inject constructor(
+    @ApplicationContext private val context: Context,
+) {
+    private val cache = ConcurrentHashMap<SessionId, DefaultSessionPreferencesStore>()
+
+    fun get(sessionId: SessionId, sessionCoroutineScope: CoroutineScope) = cache.getOrPut(sessionId) {
```
bmarty, line 34: There is a warning here:
`Declaration has type inferred from a platform call, which can lead to unchecked nullability issues. Specify type explicitly as nullable or non-nullable.`

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStoreFactory.kt
```diff
@@ -0,0 +1,41 @@
...
+        DefaultSessionPreferencesStore(context, sessionId, sessionCoroutineScope)
+    }
+
+    fun remove(sessionId: SessionId) = cache.remove(sessionId)
```
bmarty, line 38: Not used?

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStoreFactory.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    fun clear() = cache.clear()
```
bmarty, line 40: Not used, but I do not see in which case we may want to have such method. So maybe just remove it?

bmarty (commented):
Glad to see that the issue is fixed!
I have some more remarks before we can merge this PR.

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/SessionPreferencesModule.kt
jmartinesp: True, I was just keeping it there in case we can solve the underlying issue and remove `DefaultSessionPreferencesStoreFactory`. But it doesn't do anything at the moment, so we can remove it.

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/SessionPreferencesModule.kt
jmartinesp: > I think it will be simpler to provide the sessionId in SessionMatrixModule like some other fields and remove CurrentSessionIdHolder. Out of scope of this PR though.

The problem is Anvil/Dagger doesn't understand that a `SessionId` is since it's Java based, for the DI library it's just a `String` and it won't allow you to inject it anywhere. That's why we had to create the holder wrapper. But maybe its name can be misleading.

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStoreFactory.kt
jmartinesp: Yeah, I might just remove this and `clear` until I create a way to clear these values in another PR (#2303 ?).

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/SessionPreferencesModule.kt
bmarty: We had [annotations](https://github.com/element-hq/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/di/StringQualifiers.kt#L21) to be able to inject `String`s in the previous project and it was quite handy.

bmarty (approved):
Thanks for the update!

PR #2313, opened 2024-01-29
Cleanup and ensure that canUserJoinCall is live.

bmarty:
Small cleanup, see the commit messages for more details.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
@@ -81,7 +81,6 @@ import io.element.android.libraries.matrix.api.room.MatrixRoom
 import io.element.android.libraries.matrix.api.room.MatrixRoomInfo
 import io.element.android.libraries.matrix.api.room.MatrixRoomMembersState
 import io.element.android.libraries.matrix.api.room.MessageEventType
-import io.element.android.libraries.matrix.api.user.CurrentSessionIdHolder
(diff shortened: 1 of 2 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
```diff
@@ -157,9 +155,9 @@ class MessagesPresenter @AssistedInject constructor(
             mutableStateOf(false)
         }
 
-        LaunchedEffect(currentSessionIdHolder.current) {
+        LaunchedEffect(syncUpdateFlow) {
```
ganfra, line 158: If you want the check to be done at each sync, you should use `syncUpdateFlow.value` instead

ganfra (approved):
One remark otherwise LGTM

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
bmarty: Ah yes, good catch. So there is also a pb here https://github.com/element-hq/element-x-android/blob/develop/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt#L168 ?

bmarty: https://github.com/element-hq/element-x-android/pull/2313/commits/05cc526be98f08f4560c6a1795c091996bbcd2e6

PR #2349, opened 2024-02-05
"Share presence" setting

### LINKED ISSUE #2241: [Task] Setting: Disable sending typing notifications and read receipts

bmarty:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Change "Read receipts" advanced setting used to send private Read Receipt to "Share presence" settings.
When disabled, private Read Receipts will be sent, and no typing notification will be sent. Also Read Receipts and typing notifications will not be rendered in the timeline.

Note that the new boolean to disable rendering on typing notification will be used in a future PR.

## Motivation and context

Closes #2241 

## Screenshots / GIFs

See recorded ones.

## Tests

- Play with the "Share presence" and observe that when it is disabled, no typing notification is sent, no public RR is sent, and RR are not displayed anymore in any the timeline.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


features/preferences/impl/src/main/res/values/localazy.xml
```diff
@@ -8,6 +8,8 @@
   <string name="screen_advanced_settings_rich_text_editor_description">"Disable the rich text editor to type Markdown manually."</string>
   <string name="screen_advanced_settings_send_read_receipts">"Read receipts"</string>
   <string name="screen_advanced_settings_send_read_receipts_description">"If turned off, your read receipts won\'t be sent to anyone. You will still receive read receipts from other users."</string>
+  <string name="screen_advanced_settings_share_presence">"Share presence"</string>
+  <string name="screen_advanced_settings_share_presence_description">"If turned off, you won’t be able to send or receive read receipts or typing notifications"</string>
```
bmarty, line 12: Note that `screen_advanced_settings_send_read_receipts` and `screen_advanced_settings_send_read_receipts_description` may be removed from Localazy when iOS will be ready.

libraries/preferences/api/src/main/kotlin/io/element/android/features/preferences/api/store/SessionPreferencesStore.kt
```diff
@@ -19,8 +19,20 @@ package io.element.android.features.preferences.api.store
 import kotlinx.coroutines.flow.Flow
 
 interface SessionPreferencesStore {
+    suspend fun setSharePresence(enabled: Boolean)
+    fun isSharePresenceEnabled(): Flow<Boolean>
+
     suspend fun setSendPublicReadReceipts(enabled: Boolean)
     fun isSendPublicReadReceiptsEnabled(): Flow<Boolean>
```
jmartinesp, line 26: The internal key should probably be kept for migration, but maybe we should remove the public methods from the interface here? I see it seems to still be used in `TimelinePresenter`, but maybe should we use `sharePresence` there instead?

libraries/preferences/api/src/main/kotlin/io/element/android/features/preferences/api/store/SessionPreferencesStore.kt
```diff
@@ -19,8 +19,20 @@ package io.element.android.features.preferences.api.store
...
 
+    suspend fun setRenderReadReceipts(enabled: Boolean)
+    fun isRenderReadReceiptsEnabled(): Flow<Boolean>
+
+    suspend fun setSendTypingNotifications(enabled: Boolean)
+    fun isSendTypingNotificationsEnabled(): Flow<Boolean>
+
+    suspend fun setRenderTypingNotifications(enabled: Boolean)
+    fun isRenderTypingNotificationsEnabled(): Flow<Boolean>
+
```
jmartinesp, line 36: Is there a reason why we're keeping these separate? When we enabled/disable them we do it to all of them at the same time, so they act as a single one. Or are we thinking about splitting it in the future?

bmarty: It may have sense to invoke separate `set` method, as soon as `setSharePresence` is not invoked anymore.

libraries/preferences/api/src/main/kotlin/io/element/android/features/preferences/api/store/SessionPreferencesStore.kt
jmartinesp: I hadn't thought about forks, but that makes sense, thanks! I still think this is more complicated than it might need to be, but if a decision was made around this then it LGTM.

jmartinesp (approved):

libraries/preferences/api/src/main/kotlin/io/element/android/features/preferences/api/store/SessionPreferencesStore.kt
vaguerant: > I believe forks (and maybe future us) will not want to use `setSharePresence` but may want to add more specific settings.

I mentioned this in #element-x-android:matrix.org but as a user, I would definitely like to see this split into two options:
* share and display read receipts
* share and display typing notifications

The way I described it in the EXA room is that I don't want my presence shared *passively* (read receipts) but I'm happy for my presence to be shared *actively* (typing notifications). For me, the former is uncomfortable over-sharing, while the latter is a sensible and useful result of consciously interacting with the room. My feelings about each are the complete opposite of one another, so combining them into a single setting forces me to compromise one one of them.

Combining sharing and displaying of each into a single option doesn't bother me; I have the same feelings about other people's presence notifications as I do about my own, i.e. I don't want to see *your* read receipts but I do think *your* typing notifications are extremely helpful. It's a very informal poll, but when this was discussed on Matrix, based on messages and emoji reactions, ~3 users agreed with this position (supporting combined sharing/display into a single setting) and 1 disagreed (preferring separate settings for sharing and display).

Personally, I feel like *four* different settings in the UI relating to presence is a bit bloated for what is shaping up to be the "default" Matrix client on Android, but combining them all into *one* is overly broad.

PR #2353, opened 2024-02-06
Do not use a key for the LazyColumn.

bmarty:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Do not use a key for the LazyColumn or the scroll will not behave as expected if a room is moved to the top of the list.

## Motivation and context

Fix a crash since the SDK may provide several times the same room, so the LazyList is complaining that the key is not unique. This duplication will be fixed later in the SDK.

## Tests

- Scroll to the bottom of a rather large room list (> 50 rooms)
- Without this PR, there is a crash
- With this PR the last room may be displayed twice (and this has to be fixed SDK side)

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


jmartinesp (approved):
I added this to uncover any issues with duplicate rooms that might exist, so I guess it fulfilled its objective 😅 .

If it's something frequent we should definitely remove this code for now.

PR #2355, opened 2024-02-06
Test timeline view

bmarty:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [x] Other : more UI tests

## Content

Add test on some Composable.
Also avoid passing the Timeline eventSink to all the Timeline items, it's quite not OK. Introduce interfaces `EventFromTimelineItem` and `TimelineItemPollEvents` to have better visibility scopes.

## Motivation and context

Improve test coverage and improve code quality.

image

## Tests

- No behavior change

## Tested devices

- NA

## Checklist


.github/workflows/tests.yml
```diff
@@ -82,7 +82,9 @@ jobs:
       - name: ☂️ Upload coverage reports to codecov
         if: always()
         uses: codecov/codecov-action@v4
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+        with:
+          fail_ci_if_error: true
+          verbose: true
```
jmartinesp, line 87: Should we remove the `verbose` flag here?

tests/testutils/src/main/kotlin/io/element/android/tests/testutils/EnsureNeverCalled.kt
```diff
@@ -27,3 +27,9 @@ class EnsureNeverCalledWithParam<T> : (T) -> Unit {
         throw AssertionError("Should not be called and is called with $p1")
     }
 }
+
+class EnsureNeverCalledWithTwoParams<T, U> : (T, U) -> Unit {
```
jmartinesp, line 31: About this and the other classes above: maybe we should just use a function?

Something like:

```kotlin
fun ensureNeverCalledWithTwoParams<T, U>(p1: T, p2: U) {
    throw AssertionError("Should not be called and is called with $p1 and $p2")
}
// then use it like
...
onReactionClicked = ::ensureNeverCalledWithTwoParams,
...
```

The syntax looks a bit weirder at the usage point. Feel free to ignore if you think the class approach is cleaner.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineEvents.kt
```diff
@@ -22,16 +22,23 @@ sealed interface TimelineEvents {
     data object LoadMore : TimelineEvents
     data class SetHighlightedEvent(val eventId: EventId?) : TimelineEvents
     data class OnScrollFinished(val firstIndex: Int) : TimelineEvents
+
+    // Events coming from a timeline item
+    sealed interface EventFromTimelineItem : TimelineEvents
+
+    // Events coming from a poll item
+    sealed interface TimelineItemPollEvents : EventFromTimelineItem
```
jmartinesp, line 30: Nit: maybe this should wrap all the existing `TimelineItemPollEvents` below?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineEvents.kt
```diff
@@ -22,16 +22,23 @@ sealed interface TimelineEvents {
     data object LoadMore : TimelineEvents
     data class SetHighlightedEvent(val eventId: EventId?) : TimelineEvents
     data class OnScrollFinished(val firstIndex: Int) : TimelineEvents
+
+    // Events coming from a timeline item
+    sealed interface EventFromTimelineItem : TimelineEvents
+
+    // Events coming from a poll item
```
jmartinesp, line 29: Maybe these comments could be documentation ones?

jmartinesp (approved):
LGTM, thanks! I have a few suggestions, but feel free to ignore them if you don't think they're worth the time.

The new `EventFromTimelineEvent` and `TimelineItemPollEvents` hierarchy was a bit difficult to understand at first, since I thought they both were at the same level, but once it 'clicked' in my head it made sense.

tests/testutils/src/main/kotlin/io/element/android/tests/testutils/EnsureNeverCalled.kt
bmarty: That's another possible approach, yes. At first I wanted to have parity with `EnsureCalledOnce` and consort.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineEvents.kt
bmarty: Yes, it could, and I did it first, but the problem is that is make the access more complex, since all the hierarchy has to be imported, or explicitly written. I wanted to keep it flat, even if at first sight, it's a bit harder to read the parents chain.

bmarty: I also thought about splitting `TimelinePresenter.handleEvents()` into smaller function, by casing the interfaces type, but we may consider this when we will have more Events, because for now it's just adding cognitive complexity.

PR #2357, opened 2024-02-07
Enable trace logging for the Olm Account

poljar:
This should help us to track down UTD sources, it ensures that we log the state of a Olm session when we try to decrypt an `m.olm.*` to-device message.

This should be applied after we update to https://github.com/matrix-org/matrix-rust-sdk/pull/3100.

Matching EX-iOS PR: https://github.com/element-hq/element-x-ios/pull/2427


jmartinesp:
I'll create a new SDK version then, if those changes are needed.

poljar:
> I'll create a new SDK version then, if those changes are needed.

Oh, they aren't strictly needed, but they would be helpful since we otherwise still miss some nice information.

jmartinesp:
The Rust SDK version that contains the changes for this PR is in https://github.com/element-hq/element-x-android/pull/2359

poljar:
> The Rust SDK version that contains the changes for this PR is in #2359

Thanks for taking care of the SDK bump.

PR #2366, opened 2024-02-08
Remove modifier parameter in private methods

bmarty:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [X] Other : improve test coverage

## Content

This PR removes the `Modifier` parameter in private methods when it never take other values than the default one.
There is a right balance to find between following Compose UI recommendation to provide a Modifier in the parameter list and improving the coverage result. When the Modifier is never passed into parameter, the line appears as uncovered for instance here:

image

With this change, the coverage is improved because it removes what may be considered as false positive. If necessary, it's still possible to add the parameter back.

As a bonus, it fixes some mistakes in the existing code, that may appears if a Modifier was actually set.

image

## Motivation and context

Improve test coverage and so reduce risk of regression

## Screenshots / GIFs

No UI change

## Tests

- No behavior change.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/MessagesReactionButton.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/MessagesReactionButton.kt
@@ -126,9 +126,8 @@ private val ADD_EMOJI_SIZE = 16.dp
 @Composable
 private fun TextContent(
     text: String,
-    modifier: Modifier = Modifier,
 ) = Text(
-    modifier = modifier
+    modifier = Modifier
         .height(REACTION_EMOJI_LINE_HEIGHT.toDp()),
     text = text,
     style = ElementTheme.typography.fontBodyMdRegular,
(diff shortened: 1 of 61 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/MessagesReactionButton.kt
```diff
@@ -138,27 +137,24 @@ private fun TextContent(
...
-    modifier: Modifier = Modifier,
 ) = Row(
     verticalAlignment = Alignment.CenterVertically,
-    modifier = modifier,
 ) {
     // Check if this is a custom reaction (MSC4027)
     if (reaction.key.startsWith("mxc://")) {
         AsyncImage(
-            modifier = modifier
+            modifier = Modifier
```
bmarty, line 157: Note that `modifier` was used twice here, which is not correct.

features/networkmonitor/api/src/main/kotlin/io/element/android/features/networkmonitor/api/ui/ConnectivityIndicatorView.kt
```diff
@@ -57,7 +56,7 @@ fun ConnectivityIndicatorView(
         enter = fadeIn() + expandVertically(),
         exit = fadeOut() + shrinkVertically(),
     ) {
-        StatusBarPaddingSpacer(modifier)
+        StatusBarPaddingSpacer()
```
bmarty, line 59: Note that `modifier` was used twice here, which is not correct.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/MessagesReactionButton.kt
jmartinesp: It's weird the linters never found the issue.

jmartinesp (approved):

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/MessagesReactionButton.kt
bmarty: Yes, maybe because of the form `= Row(`. I may check later...

PR #2392, opened 2024-02-14
Add the user certificates as additional certificates to the ClientBuilder

poljar:
Now, this is a story all about how
Certificates work in Android town
And I'd like to take a minute
Enter, close the door
I'll tell you how I've figured out the inner workings of the Keystore

Well it all boils down to the fact that Google got scared
It said, "Your certs are movin' to a place you'll never find 'em".

So the directory, user certificates are stored, is hard to find, and possibly not readable by your application[[1]]. Instead, we need to use the Keystore[[2]] API, specifically we'll need to open the `AndroidCAStore` Keystore type.

The various Keystore types are supposedly documented[[3]], but I'm failing to find a logical path that would lead you to conclude that:
1. System certificates can or should be accessed using the Keystore, specifically the `AndroidCAStore` type
2. User certificates can be found in the same Keystore type as the system certificates

So this was mostly found using random googling, swearing, and a couple of educated guesses.

[1]: https://android-developers.googleblog.com/2016/07/changes-to-trusted-certificate.html
[2]: https://developer.android.com/reference/java/security/KeyStore
[3]: https://docs.oracle.com/en/java/javase/17/docs/specs/security/standard-names.html#keystore-types
 
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

The Rust SDK, if configured to use the [`rustls`](https://github.com/rustls/rustls) crate for SSL reads the system certificates using the [`openssl-probe`](https://github.com/alexcrichton/openssl-probe) crate. The `openssl-probe` crate reads the certificates from some predefined directory. This directory doesn't contain any additional certificates the user installed on the device.

Like previously mentioned, user-installed certificates should™ be gathered via the `Keystore` API and that one might not be available from Rust land. We can't easily modify the `openssl-probe` to use the `Keystore` API so here we are doing it the dumb way. We load the certs from the Kotlin side and push them to the Rust side.

It would probably be nice to write a test for the method that gathers the certificates, but I have no idea how to do this.

This PR requires: https://github.com/matrix-org/matrix-rust-sdk/pull/3126

## Motivation and context

Sorry, no motivation was found. Somebody told me to do this.

## Screenshots / GIFs

[mitmproxy.webm](https://github.com/element-hq/element-x-android/assets/552026/8c8d72d5-7d85-44a8-af0b-1e70df88b42c)

## Tests

- Generate a fresh set of self-signed certificates using `mitmproxy`, `mitmproxy` does this automatically when it starts up the first time.
- Push the `mitmproxy` certificate to the emulated device using `adb push mitmproxy-ca-cert.cer /data/local`
- Install the certificate as a user certificate under **Security & privacy** => **More security settings** => **Encryption & credentials**
- Run `mitmproxy` in reverse proxy mode `mitmproxy --mode reverse:http://localhost:8008 -k --listen-port 8010`
- Forward the 8010 port through adb `adb reverse tcp:8010 tcp:8010`
- Try to connect to the `mitmproxy` reverse proxy using `https://localhost:8010` as the homeserver address

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

Android 14.0 ("UpsideDownCake") | x86_64 | API 34

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
@@ -27,7 +27,9 @@ import kotlinx.coroutines.withContext
 import org.matrix.rustcomponents.sdk.ClientBuilder
 import org.matrix.rustcomponents.sdk.Session
 import org.matrix.rustcomponents.sdk.use
+import timber.log.Timber
 import java.io.File
+import java.security.KeyStore
 import javax.inject.Inject
 
 class RustMatrixClientFactory @Inject constructor(
@@ -46,6 +48,7 @@ class RustMatrixClientFactory @Inject constructor(
             .username(sessionData.userId)
             .passphrase(sessionData.passphrase)
             .userAgent(userAgentProvider.provide())
+            .addRootCertificates(getAdditionalCertificates())
             // FIXME Quick and dirty fix for stopping version requests on startup https://github.com/matrix-org/matrix-rust-sdk/pull/1376
             .serverVersions(listOf("v1.0", "v1.1", "v1.2", "v1.3", "v1.4", "v1.5"))
             .use { it.build() }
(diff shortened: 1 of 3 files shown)


poljar:
What do I need to do to get this merged/reviewed? The CI failures seem to be caused by the fact that I created the PR from a fork and a random 500 error codecov threw.

In other words, the CI failures are unrelated to the changes I made.

jmartinesp:
I tried to test this, but I can't get anything in mitmproxy, probably because the OkHttp request we do to `https://localhost:8081` fails before the Rust SDK can do anything. What did you do to have the results seen in the video?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -68,6 +71,57 @@ class RustMatrixClientFactory @Inject constructor(
...
+* The Rust HTTP client doesn't include user-installed certificates in its internal certificate
+* store. This means that whatever the user installs will be ignored.
+*
+* While most users don't need user-installed certificates some special deployments or debugging
+* setups using a proxy might want to use them.
+*
+* @return A list of byte arrays where each byte array is a single user-installed certificate
+*         in encoded form.
+*/
+fun getAdditionalCertificates(): List<ByteArray> {
```
ganfra, line 86: We should extract this method behind an interface, so we can easily have an empty implementation for release mode, or stuff like that.

bmarty: I can handle this.

poljar: The same method should be used for release builds. While I did test this with `mitmproxy`, people/customers want to use EX with a server that is using a custom CA.

bmarty:
Thanks @poljar, I will merge the PR in a branch I own to be able to add more commit. This will then be reviewed as a regular PR.

poljar:
> I tried to test this, but I can't get anything in mitmproxy, probably because the OkHttp request we do to `https://localhost:8081` fails before the Rust SDK can do anything. What did you do to have the results seen in the video?

I think that this is due to `Synapse` not being reachable, you can use `matrix.org` to test this if you don't have a local `Synapse` running:

```bash
mitmproxy --mode reverse:https://matrix.org --listen-port 8010
```

Just note that once well-known has been handled, after logging in, the reverse proxy won't be used by the Client anymore.

jmartinesp:
> I think that this is due to `Synapse` not being reachable, you can use `matrix.org` to test this if you don't have a local `Synapse` running:
> 
> ```shell
> mitmproxy --mode reverse:https://matrix.org --listen-port 8010
> ```
> 
> Just note that once well-known has been handled, after logging in, the reverse proxy won't be used by the Client anymore.

I can confirm it works as expected when using `matrix.org`, it's a problem with my local setup. Thanks again for the help!

PR #2415, opened 2024-02-19
Display banned users in room member list

### LINKED ISSUE #2256: [Task] Show sections for roles in room-member list

jmartinesp:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Adds a segmented control button row on top of the member list when the user can ban/unban users, there are banned users and the room is not a DM.
- Display the banned users in a list in a separate 'tab'.
- When this other tab is selected, the search works on the banned users instead of the joined members.

## Motivation and context

Closes #2256.

## Screenshots / GIFs

Inside the PR.

## Tests

- Go to a room where you're not either an admin or a moderator, open the room member list. It should appear just as before.
- Now go into a room where you're either an admin or a moderator and there are banned users (EX Android one?). The segmentation control should be present.
- Change to the 'banned' tab, the banned users should be there, with no name or avatar (as they could be offensive). The click action doesn't work at the moment, it's part of a different task.
- Search for some of these users.
- Go back to the member list tab, check it works as before.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListPresenter.kt
```diff
@@ -57,6 +62,13 @@ class RoomMemberListPresenter @Inject constructor(
             value = room.canInvite().getOrElse { false }
         }
 
+        val canDisplayBannedUsers by produceState(initialValue = false) {
+            room.membersStateFlow
+                .map { members -> members.roomMembers()?.any { it.membership == RoomMembershipState.BAN }.orFalse() }
+                .onEach { hasBannedUsers -> value = hasBannedUsers && room.canBan().getOrElse { false } && !room.isDm }
```
bmarty, line 68: The test `!room.isDm` is a bit late, maybe check it before checking the roomMembers?
Note that on a DM there should not be a lot of user.

(I am wondering why the fact that a DM should be taken into account to display or not the banned users, it seems that it may lead to problem if the room actually has banned users. Similarly, I do not see the point to hide the banned users to user who cannot unban. Sorry to rise the point during the PR review, they have maybe been discussed with product/design. It it is the case, just ignore me :) ).

tests/uitests/src/test/snapshots/images/ui_S_t[f.roomdetails.impl.members_RoomMemberList_null_RoomMemberList-Day-2_3_null_8,NEXUS_5,1.0,en].png
Comment (bmarty) [inline tests/uitests/src/test/snapshots/images/ui_S_t[f.roomdetails.impl.members_RoomMemberList_null_RoomMemberList-Day-2_3_null_8,NEXUS_5,1.0,en].png:1]: Could be nice to add a preview of a list of banned users.

bmarty (approved):
Thanks, a few remarks, but I guess you can ignore me.

jmartinesp:
> > The ui looks a bit weird with the search and the segmented control, no?
> 
> I agree, I would just add a "Banned" section, at the bottom of the list, so that banned users may also appears when the list is filtered (I did not check how the search works now)

If I were an admin/moderator in a large room (>1000 users?) I certainly wouldn't want to scroll all those users to be able to reach the banned section at the bottom. This is done so users with permissions (admin, moderators) can have a quick and easy way to reach this section.

I don't think the design is bad (I would have flagged it sooner if I had), and I only see 2 alternatives myself:

1. Having a separate screen for banned users, accessed from a different point in some menu.
2. Making the sections collapsible, although if the UI for those isn't really obvious the users might miss it.

But again, I don't think this UI is bad, in the context of only users with moderation permissions actually seeing them.

bmarty:
I agree that sections should be collapsible, so it will fix the problem for large rooms.

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListPresenter.kt
jmartinesp: > The test !room.isDm is a bit late, maybe check it before checking the roomMembers?

True, it would be a lot faster that way.

> (I am wondering why the fact that a DM should be taken into account to display or not the banned users

A DM in this context is a direct room with 2 users, you and the other user. So if you are banned you won't be able to access this view and if the other user is banned you won't be able to talk to anyone anyway, so it would be easier to just leave the room.

PR #2421, opened 2024-02-20
Last session recovery

bmarty:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Add a way to let the user enter a recovery key when there is no other sessions available.

Also update the UI flow, when clicking on the verification banner : either directly open the verification screens (instead of opening the settings), or directly open the "enter recovery key" screen (instead of opening the verification screens, which is non-sense since there are no other sessions).

- This PR also fixes an issue were the various states regarding crypto was not updated (b09294c)
- and the `isLastDevice` is now a `StateFlow`, to help having more dynamic UI.
- Finally the flag `FeatureFlags.SecureStorage` is removed.

## Motivation and context

Closes #2400 

## Screenshots / GIFs

RecoveryKey_OneSession

## Tests

- Log an account in, with no other session available.
- The banner display the special case to enter a recovery key.
- If there are other sessions, the banner will be the classical verify session banner

When verifying a session, it's also possible to choose to enter a recovery key instead.

See the recorded screenshot for more details.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 23
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [ ] Pull request is based on the develop branch
- [ ] Pull request includes a new file under ./changelog.d. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#changelog
- [ ] Pull request includes screenshots or videos if containing UI changes
- [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
- [ ] You've made a self review of your PR

PR #2429, opened 2024-02-22
Test recovery key

bmarty:
Complement for #2421: handle @ganfra 's remarks, and add Maestro test on entering Recovery Key to verify the session.
The RecoveryKey has been added to the GH secret, but will be visible on Maestro recording, this is OK I think, this is just a test account. The account password is still not visible.

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/encryption/RustEncryptionService.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/encryption/RustEncryptionService.kt
@@ -25,7 +25,7 @@ import io.element.android.libraries.matrix.api.encryption.EncryptionService
 import io.element.android.libraries.matrix.api.encryption.RecoveryState
 import io.element.android.libraries.matrix.api.sync.SyncState
 import io.element.android.libraries.matrix.impl.sync.RustSyncService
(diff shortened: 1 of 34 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/encryption/RustEncryptionService.kt
```diff
@@ -66,10 +65,15 @@ internal class RustEncryptionService(
...
-    private var recoveryStateListenerTaskHandle: TaskHandle? = null
 
-    private val backupStateFlow = MutableStateFlow(service.backupState().let(backupStateMapper::map))
+    private val backupStateFlow = mxCallbackFlow {
+        val listener = object : BackupStateListener {
+            override fun onUpdate(status: RustBackupState) {
+                trySend(backupStateMapper.map(status))
+            }
+        }
+        service.backupStateListener(listener)
```
ganfra, line 75: I'd have offer this as extension methods as we already do that in several places (RoomList, Timeline...)

bmarty: Yes, sorry, I forgot this: 7d4cf6d574c7ed51a72dff11d1fa875bc34d0549

ganfra (approved):

PR #2430, opened 2024-02-22
Pass user certificates to the Rust SDK

bmarty:
See #2392, which is the first commit of this PR.
The second commit is just cleanup. A user certificate added to the device is still detected and used after this change.

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/certificates/DefaultUserCertificatesProvider.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/certificates/DefaultUserCertificatesProvider.kt
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) 2024 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


jmartinesp (approved):

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/certificates/DefaultUserCertificatesProvider.kt
```diff
@@ -37,41 +38,54 @@ class DefaultUserCertificatesProvider @Inject constructor() : UserCertificatesPr
...
+        try {
+            keyStore.load(null)
+        } catch (e: Exception) {
+            Timber.w(e, "Failed to load AndroidCAStore keystore")
+            return emptyList()
+        }
+        val aliases = try {
+            keyStore.aliases()
+        } catch (e: Exception) {
+            Timber.w(e, "Failed to get aliases from AndroidCAStore keystore")
```
jmartinesp, line 58: Couldn't you merge the first 2 blocks? I think the third could be merged too if we use different catch blocks.

bmarty: I wanted to have specific log to eventually investigate what was failing. But probably the Exception will contain enough details. WDYT?

jmartinesp: Yeah, I think so. Also, the stacktrace should point at which function threw it.

bmarty: 9fed8689682dacd9382030ce766f210ff479e608

Can't really merge the first block, since `keyStore` is used later to get the entry.

PR #2433, opened 2024-02-22
Test direct logout dialog and RoomDetailsView

bmarty:
Not a lot more to say, this PR adds a bunch of UI tests and screenshot tests

image


features/logout/impl/src/main/kotlin/io/element/android/features/logout/impl/direct/DefaultDirectLogoutView.kt
```diff
@@ -50,3 +54,14 @@ class DefaultDirectLogoutView @Inject constructor() : DirectLogoutView {
         )
     }
 }
+
+@PreviewsDayNight
+@Composable
+fun DefaultDirectLogoutViewPreview(
```
ElementBot, line 60: - :warning: Composables annotated with @Preview that are used only for previewing the UI should not be public.

See https://mrmans0n.github.io/compose-rules/rules/#preview-composables-should-not-be-public for more information.

jmartinesp (approved):

PR #2449, opened 2024-02-27
Add French and German screenshots

### LINKED ISSUE #2454: Screenshot in foreign languages
The tools we are using to perform screenshot test allow the usage of other languages than English. For now, only English is used, to both record and test screenshots.

But we do not test the translated versions of the application. Several type of problems can occur:
- layout issue, for instance if the translation is longer than the English version, and this is quite frequent, especially in languages like German. If this occurs, 3 solutions: update the translation to a shorter version, fix the layout (Composable), assume the current behavior (like in the dialog below, button are stacked instead of being on the same line):
image

Other example of layout issue: text alignment problem:
image

- translation issue like typo, grammar mistake, etc. They are sometimes easier to see in the context of their usage.
- translation inconsistency in a specific screen. Wording is translated key by key, and the translators do not always have a clear idea of where the string can be used. This  can lead to inconsistency in the global screen. For instance, see below an French exmaple: the tab `Banni` should have the plural form (`Bannis`), but, this is the translation of `Banned` so technically the translation is correct, but this appear as a wording issue in this screen:
image

## What can we do

We want to enforce the quality of the application in the most used languages, like German, French, Spanish, Russian, Ukrainian.

We have the advantage of having native people for some of those languages working in the company, and the community can help for other languages.

## Constraint

- We want the CI to stay fast, especially when running the test. So the tests will only check the English screenshots. Same for recording the screenshot with the current GitHub action. It will only record the English version.

- Screenshot recording for other languages can be performed during the Localazy sync Github action. So PR from this GA will contain both the string changes and the screenshot of impacted screens. The check of those screenshots will still have to be done by human people. New screenshot will appear in case of new translated screen (so the screenshot is now different than the English one), or when a translation is updated, so reviewers can check that the change is OK.

- We want to limit the usage of LFS quota, so only screenshots which are different than the English version will be stored. Lots of screenshots do not contain string resource, so this will limit the number of screenshots recorded per language. It's not possible to record only those screenshots, but a solution could be to record them all, then delete those which are similar to the English version.

- Screenshot file for other languages will have to be stored in a different folder than the English screenshot, and in one subfolder per language, so that reviewer can have a global view (from the `Finder` app for instance) of all the screnshots for a specific languages

## Definition of done

- There is a folder per language with all the screenshot files that are different than the English version
- The CI is not slower than before to check a regular PR
- The Localazy GA records screenshots for other languages

bmarty:
Closes #2454 (see the details here).

Create a parent class for screenshot test. The class now have two child classes:
- The class S is testing only the English version.
- The class T is testing the other languages (`T` stands for translations), and is used on demand (by adding parameter to the project: `./gradlew recordPaparazziDebug -PallLanguages`)

The added script `generateAllScreenshots.py` is taking care of:
- Record the screenshots in all desired languages ("fr" and "de" for now)
- Remove screenshot similar to the english version
- Move the screenshot in the folder `screenshots/<lang>`

The script is working fine locally. We will have to check if it's working fine on the CI has well, and eventually fix it.

The action which sync Localazy has been updated to invoke this script after a Localazy sync, so a PR with Localazy sync may include changes on those screenshots. Additionnally it may also update the English screenshot, so a human action to record the screenshot in cas of English change on the Localazy string is not required anymore.

I have already detected some layout issues, but I do not want to fix them now, I want to fix them once the screenshots are getting recorded once (so after this PR has been merged and after a Localazy sync has been triggerd) so we can see the change in the screenshots, and confirm that the fix is done.

Current `test` task is not impacted with those new screenshots. The only impact is the number of files under LFS, which is about 400+ files per languages.

Also the current action to record the screenshots should not be impacted by these changes, and it will still record only the English version of the screenshots.

### DIFF
--- a/tests/uitests/src/test/kotlin/ui/T.kt
+++ b/tests/uitests/src/test/kotlin/ui/T.kt
@@ -1,12 +1,11 @@
 /*
- * Copyright 2022 The Android Open Source Project
- * Copyright (c) 2022 New Vector Ltd
+ * Copyright (c) 2024 New Vector Ltd
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     https://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


tests/uitests/src/test/kotlin/ui/T.kt
```diff
@@ -0,0 +1,51 @@
...
+import com.google.testing.junit.testparameterinjector.TestParameterInjector
+import org.junit.Test
+import org.junit.runner.RunWith
+
+/**
+ * Screenshot test for the Locale other then English.
+ */
+@RunWith(TestParameterInjector::class)
+class T : ScreenshotTest() {
+    /**
```
ElementBot, line 29: - :warning: A member is named after the class. This might result in confusion. Either rename the member or change it to a constructor.

jmartinesp (approved):
LGTM, it seems like a good strategy.

Let's see how it goes, but thanks for your work on this!

bmarty:
Thanks. Running a Localazy sync to see! https://github.com/element-hq/element-x-android/actions/runs/8092594439

PR #2455, opened 2024-02-28
Sync Strings

ElementBot:


bmarty:
I have triggered a sync but it was supposed to be run on the branch `feature/bma/frenchScreenshot`... So this is not the expected result.

bmarty (approved):

PR #2465, opened 2024-02-29
Sync Strings

ElementBot:


bmarty:
I take care of this. The empty tag error should not occur anymore since we have a post-treatment after Localazy import. Something is not working. Also the screenshots has not been generated :/. It should be after #2449 has been merged.

PR #2466, opened 2024-02-29
Fix localazy sync

bmarty:
Those changes should fix all the problems identified here: https://github.com/element-hq/element-x-android/pull/2465#issuecomment-1970638968

### DIFF
--- a/.github/workflows/sync-localazy.yml
+++ b/.github/workflows/sync-localazy.yml
@@ -12,6 +12,15 @@ jobs:
     if: github.repository == 'element-hq/element-x-android'
     steps:
       - uses: actions/checkout@v4
+      - name: Use JDK 17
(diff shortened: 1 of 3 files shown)


jmartinesp:
Can we launch the flow on this branch so we check if these changes do fix the issues?

bmarty:
> Can we launch the flow on this branch so we check if these changes do fix the issues?

I do not think it will work (I already tried on the other PR), since the localazy workflow is creating a PR, it's not committing on the branch.

bmarty:
Note: I have tested the script locally and the issue with the xml formatter is fixed.

jmartinesp (approved):
Ok, let's check if it works then.

bmarty:
https://github.com/element-hq/element-x-android/actions/runs/8094108848

PR #2467, opened 2024-02-29
Sync Strings

ElementBot:


bmarty (changes requested):
There is a pb with the file `tests/uitests/src/test/snapshots/images/ui_S_t[l.designsystem.atomic.atoms_ElementLogoAtomLarge_null_ElementLogoAtomLarge-Day_0_null,NEXUS_5,1.0,en].png` , it should not have been updated, and it appears to be empty now.

Also the workflow has triggered this error, which is quite weird:

```
 > Task :tests:uitests:testDebugUnitTest

ui.T > t[l.designsystem.atomic.atoms_ElementLogoAtomLarge_null_ElementLogoAtomLarge-Day_0_null,NEXUS_5,1.0,fr] FAILED
    kotlin.io.NoSuchFileException at T.kt:45

4755 tests completed, 1 failed
```

Weird because the test should not be run by this workflow 🤔 .

The two problems are probably related.

bmarty:
Second attempt was successful: https://github.com/element-hq/element-x-android/pull/2468

PR #2511, opened 2024-03-08
Disable special log level for read receipts

bnjbvr:
Read receipts have been considered quite stable over the last few weeks, so there's no need to keep the trace level for the associated logs for the time being.


libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/tracing/TracingFilterConfiguration.kt
```diff
@@ -67,7 +66,6 @@ enum class Target(open val filter: String) {
     MATRIX_SDK_SLIDING_SYNC("matrix_sdk::sliding_sync"),
     MATRIX_SDK_BASE_SLIDING_SYNC("matrix_sdk_base::sliding_sync"),
     MATRIX_SDK_UI_TIMELINE("matrix_sdk_ui::timeline"),
-    MATRIX_SDK_BASE_READ_RECEIPTS("matrix_sdk_base::read_receipts"),
```
bmarty, line 70: We could eventually keep this line to be able to manually tweak the tracing configuration from the app.

bmarty (approved):
OK, thanks!

PR #2532, opened 2024-03-12
Fix layout poll

bmarty:
Fix layout for empty poll history, only visible in some other language or on small screens, for instance in French:

image

Now the text will be centered.


bmarty:
Too high coverage error will be fixed in another PR and can be ignored here.

ganfra (approved):

PR #2533, opened 2024-03-12
Sync Strings

ElementBot:
- Update Strings from Localazy

PR #2534, opened 2024-03-12
Fix string import

bmarty:
I have noticed some duplication of string when importing from Localazy (See #2533).

This PR update the `config.json` file to only import desired string in the `:features:messages:impl` module.

As a side effect, some unused strings are now imported to the main string module. These are duplicated strings which are not used on Android. We will have to do some cleanup later.

### DIFF
--- a/tools/localazy/downloadStrings.sh
+++ b/tools/localazy/downloadStrings.sh
@@ -51,7 +51,7 @@ echo "Moving files from values-id to values-in..."
 find . -type d -name 'values-id' -execdir mv {}/translations.xml {}/../values-in/translations.xml \;
 
 echo "Deleting all the folders values-id..."
-find . -type d -name 'values-id' -exec rm -rf {} \;
+find . -type d -name 'values-id' -exec rm -rf {} 2> /dev/null \;
 set -e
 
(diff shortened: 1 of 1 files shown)


tools/localazy/config.json
```diff
@@ -128,17 +128,20 @@
...
+        "screen_room_attachment.*",
+        "screen_room_encrypted.*",
+        "screen_room_invite.*",
+        "screen\\.room\\.mention.*",
+        "screen_room_message.*",
+        "screen_room_no_permission.*",
+        "screen_room_reactions.*",
+        "screen_room_retry.*",
+        "screen_room_timeline.*",
+        "screen_room_typing.*"
```
bmarty, line 144: Could be nice to remove some strings key on Localazy to have a common prefix like `screen_room_timeline_`, but this require to do some change in the codebase both on Android and iOS codebase.

jmartinesp (approved):
Thanks!

tools/localazy/config.json
bmarty: (will do in https://github.com/element-hq/element-x-android/issues/2535)

PR #2545, opened 2024-03-13
Use `userId` on display name change state events

### LINKED ISSUE #2125: Show new name just once, not twice in display name change events
See issue title.

It looks quite confusing at first glance, especially with long names that include sentences or `away` text:
Screenshot_20230927-125549_Element X

I think it would be sufficient to change the string like so: `$old_name changed their display name to $new_name`.

I don't see anything better than simply showing `old -> new` instead of what's right now.

### Additional context

Maybe it'd be nicer for readability if the display names in that event were somehow highlighted, or just looked different than the `changed their display name` text, especially with long names like in screenshot above.

Yes

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Replaces the first usage of the user's display name with its user id in display name change state events.

## Motivation and context

Partially fixes #2125 . It may not be the ideal solution, but until we decide how to handle this in both platforms, we should just follow what iOS currently does.

## Tests

- Change the display name of a user.
- Open a room that user has joined.

You should see a `{userId} changed their display name from {old} to {new}` state event.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/DefaultRoomLastMessageFormatter.kt
+++ b/libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/DefaultRoomLastMessageFormatter.kt
@@ -97,7 +97,7 @@ class DefaultRoomLastMessageFormatter @Inject constructor(
                 roomMembershipContentFormatter.format(content, senderDisplayName, isOutgoing)
             }
             is ProfileChangeContent -> {
-                profileChangeContentFormatter.format(content, senderDisplayName, isOutgoing)
+                profileChangeContentFormatter.format(event.sender, content, senderDisplayName, isOutgoing)
             }
             is StateContent -> {
                 stateContentFormatter.format(content, senderDisplayName, isOutgoing, RenderingMode.RoomList)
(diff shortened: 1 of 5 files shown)


libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/DefaultRoomLastMessageFormatter.kt
```diff
@@ -97,7 +97,7 @@ class DefaultRoomLastMessageFormatter @Inject constructor(
                 roomMembershipContentFormatter.format(content, senderDisplayName, isOutgoing)
             }
             is ProfileChangeContent -> {
-                profileChangeContentFormatter.format(content, senderDisplayName, isOutgoing)
+                profileChangeContentFormatter.format(event.sender, content, senderDisplayName, isOutgoing)
```
bmarty, line 100: Not a big deal, but the code would be more beautiful if `event.sender` was the second parameter and not the first one.

bmarty (approved):
Thanks!

libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/ProfileChangeContentFormatter.kt
```diff
@@ -24,6 +25,7 @@ class ProfileChangeContentFormatter @Inject constructor(
     private val sp: StringProvider,
 ) {
     fun format(
+        senderId: UserId,
```
bmarty, line 28: Also, now that we pass the `senderId`, the parameter `senderIsYou` may be removed? The MatrixClient will have to be injected in the constructor. I understand that this is maybe temporary code, so this remark can be ignored I guess.

PR #2556, opened 2024-03-15
Test and verify coverage only on debug variant

jmartinesp:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Only test and verify coverage on debug variant.

## Motivation and context

- Make tests faster: it's a few minutes faster, but I honestly expected the diff to be higher.
- Avoid the bug in Kover found in https://github.com/element-hq/element-x-android/actions/runs/8292938071/job/22695100068?pr=2553.

## Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,7 +48,7 @@ jobs:
         with:
           cache-read-only: ${{ github.ref != 'refs/heads/develop' }}
 
-      - name: ⚙️ Run unit tests, debug and release
+      - name: ⚙️ Run unit tests for debug variant
         run: ./gradlew testDebugUnitTest $CI_GRADLE_ARG_PROPERTIES
 
       - name: 📸 Run screenshot tests
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -49,13 +49,13 @@ jobs:
           cache-read-only: ${{ github.ref != 'refs/heads/develop' }}
 
       - name: ⚙️ Run unit tests, debug and release
-        run: ./gradlew test $CI_GRADLE_ARG_PROPERTIES
+        run: ./gradlew testDebugUnitTest $CI_GRADLE_ARG_PROPERTIES
 
       - name: 📸 Run screenshot tests
         run: ./gradlew verifyPaparazziDebug $CI_GRADLE_ARG_PROPERTIES
 
       - name: 📈Generate kover report and verify coverage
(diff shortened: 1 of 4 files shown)


jmartinesp:
I don't see a good way to keep the coverage, since the missing tests can only be run on release mode given our `inline value class` implementations.

.github/workflows/tests.yml
```diff
@@ -48,22 +48,22 @@ jobs:
         with:
           cache-read-only: ${{ github.ref != 'refs/heads/develop' }}
 
-      - name: ⚙️ Run unit tests, debug and release
-        run: ./gradlew test $CI_GRADLE_ARG_PROPERTIES
+      - name: ⚙️ Run unit tests for debug variant
+        run: ./gradlew testDebugUnitTest $CI_GRADLE_ARG_PROPERTIES
```
bmarty, line 52: I am wondering if `./gradlew testGplayDebugUnitTest` would be enough.

app/build.gradle.kts
```diff
@@ -189,7 +188,7 @@ androidComponents {
             val abiCode = abiVersionCodes[name] ?: 0
             // Assigns the new version code to output.versionCode, which changes the version code
             // for only the output APK, not for the variant itself.
-            output.versionCode.set((output.versionCode.get() ?: 0) * 10 + abiCode)
+            output.versionCode.set((output.versionCode.orNull ?: 0) * 10 + abiCode)
```
bmarty, line 191: 👍

app/build.gradle.kts
```diff
@@ -215,26 +214,6 @@ knit {
     }
 }
 
-val ciBuildProperty = "ci-build"
```
bmarty, line 218: The parameter `ci-build` is also used in `nightlyReports.yml`, I think it can be removed from this file too.

bmarty (approved):
LGTM, some small remarks.

.github/workflows/tests.yml
jmartinesp: Probably. I kept it like that in case we add some `fdroid` only tests in the future. I'll change it and see if anything changes.

jmartinesp: Actually, that will only run `:app:testGplayDebugUnitTest` according to `--dry-run`, since the flavor only exists in that module.

bmarty: OK 👍

PR #2561, opened 2024-03-18
Sync Strings

ElementBot:


bmarty (approved):
OK, not a lot of changes, since a recent sync occurred in #2557

PR #2564, opened 2024-03-18
Remove unnecessary `Room.updateMembers()` calls.

jmartinesp:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Remove unnecessary `Room.updateMembers()` calls.

Some of them can be directly removed since we have a way to automatically get member info updates based on membership changes.

Others can be replaced by a simpler `getUpdatedMember(userId)` method. This might still do a full member sync, but it's quite unlikely, and even if it does it we won't have to map every fetched user before we can use the result.

## Motivation and context

After testing the improved room member list and `updateMembers()` sometimes causing 45MB of data to be loaded for the Matrix HQ room, I want to reduce these calls as much as possible. It should also improve performance since we avoid unnecessary data mapping.

## Tests

There should be no changes. I tested the following flows:

- In a DM, have the other user leave the room. Then tap on the message composer to re-invite them.
- Open the room details screen of a user, block and unblock them.
- Invite a new user to a room.
- Open the room member list.

As far as I could tell they all still worked as expected.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


bmarty (approved):
LGTM at first sight. Thanks!

PR #2569, opened 2024-03-19
Correctly format notification text for reply in a thread #2568

### LINKED ISSUE #2568: Text in notification for reply in thread is not well formatted
### Steps to reproduce

1. Alice and Bob are in a room A
2. Alice opens room A
3. Alice sends a message "Message"
4. Bob sees a notification with this message and dismiss it, just to clear the notification panel
5. Alice starts a thread from this message
6. Alice reply to the first message with the text "Reply", in the thread

Observe the notification rendering on Bobs device

_Note_: the issue is not observed when replying to a message but not in a thread. In this case, the notification will only contains the reply. This is maybe another issue which is out of scope of this one.

#### What did you expect?

- `In reply to` is localized
- the replied message should not be rendered, for long message, the reply content will not be visible
- The expected content should be something like (for instance in French):

```
En réponse à @alice:matrix.org: Reply
```

#### What happened instead?

```
In reply to@alice:domain.org
MessageReply
```

4 problems:
- `In reply to` is not localized
- A space is missing before the userId
- The first message is rendered
- The first message and the reply are not separated by a space.

### Your phone model

Any

### Operating system version

Any

0.4.6

### Homeserver

Any

### Will you send logs?

No

Yes

bmarty:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Ensure `PlainTextNodeVisitor` correctly format `mx-reply` content.

## Motivation and context

Fixes #2568

## Tests

- See #2568

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/ToPlainText.kt
+++ b/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/ToPlainText.kt
@@ -16,9 +16,12 @@
 
 package io.element.android.libraries.matrix.ui.messages
 
+import io.element.android.libraries.matrix.api.core.MatrixPatterns
 import io.element.android.libraries.matrix.api.timeline.item.event.FormattedBody
 import io.element.android.libraries.matrix.api.timeline.item.event.MessageFormat
 import io.element.android.libraries.matrix.api.timeline.item.event.TextMessageType
+import io.element.android.libraries.ui.strings.CommonStrings
+import io.element.android.services.toolbox.api.strings.StringProvider
 import org.jsoup.nodes.Document
 import org.jsoup.nodes.Element
 import org.jsoup.nodes.Node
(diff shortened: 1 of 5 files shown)


libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/ToPlainText.kt
```diff
@@ -29,50 +32,87 @@ import org.jsoup.select.NodeVisitor
  * Converts the HTML string in [TextMessageType.formatted] to a plain text representation by parsing it and removing all formatting.
  * If the message is not formatted or the format is not [MessageFormat.HTML], the [TextMessageType.body] is returned instead.
  */
-fun TextMessageType.toPlainText() = formatted?.toPlainText() ?: body
+fun TextMessageType.toPlainText(stringProvider: StringProvider? = null) =
```
jmartinesp, line 35: Wouldn't it be better to always pass a non-null `StringProvider`?

jmartinesp:
I'm not too sure about the current implementation: wouldn't it be better to add the `in reply to` part to the notification builder? If we change it here, it will also be changed for the text in the 'replied to' timeline events, in the room list for the last event, etc.

libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/ToPlainText.kt
bmarty: I tested it, but it's quite painful since this method is used in data class constructor, and other place where we do not have a stringProvider.

bmarty:
> I'm not too sure about the current implementation: wouldn't it be better to add the `in reply to` part to the notification builder? If we change it here, it will also be changed for the text in the 'replied to' timeline events, in the room list for the last event, etc.

I believe the reply part is handle differently on the timeline Event. The change in the code should only impact how event from notification are built (with type `NotificationContent.MessageLike.RoomMessage`). See some screenshots below, there is not regression:

image

image

libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/ToPlainText.kt
bmarty: Actually my bad, it's only he case for `Document.toPlainText`. I will give it a try.

jmartinesp:
> I believe the reply part is handle differently on the timeline Event. The change in the code should only impact how event from notification are built (with type `NotificationContent.MessageLike.RoomMessage`). See some screenshots below, there is not regression:

Oh, it's not happening because the HTML somehow doesn't contain the `mx-reply` there, weird!

i.e. this is what I see in the debugger:

```
<html>
 <head></head>
 <body>
  Some text
 </body>
</html>
```
And this is the formatted body from the view source code screen:

```
<mx-reply><blockquote><a href=\"https://matrix.to/#/!some-link">In reply to</a> <a href=\"https://matrix.to/#/@some-user\">@some-user:matrix.org</a><br><a href=\"https://matrix.to/#/@some-other-user:matrix.org\">some-user-uesr</a>: original text.</blockquote></mx-reply>Some text
```

bmarty:
Yes, I guess the SDK is doing some cleanup for us.

libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/ToPlainText.kt
bmarty: So I have tried it and it's adding lots of noise (Let me know if you want to see the code). Since there is no added value, I think we can keep the parameter optional. Do you agree @jmartinesp ?

jmartinesp:
So how about we do it in some different way? We could create something to detect if a message contains this `mx-reply` block and then extract both the message we replied to and the new one, and format them in some class related to the notifications, where we can access the `StringProvider`. It's a bit more work, but at least we wouldn't have to make the `ToPlainText` functions aware of these kind of replies.

Alternatively, if we do want to have this as part of the `toPlainText()` result everywhere, I think the second best thing we could do is create some `PlainTextFormatter` that can be injected instead of these functions, and use it in `TimelineItemContentMessageFactory` and other parts of the app.

jmartinesp:
Also, the current rendering seems a bit broken:

share_994317422670100723

bmarty:
Yes the current rendering is broken, see https://github.com/element-hq/element-x-android/issues/2568

I did not test with the mention though.

jmartinesp:
> Yes the current rendering is broken, see #2568
> 
> I did not test with the mention though.

To be clear, by 'current rendering' I meant with this branch, not with the code in `develop`.

bmarty:
I cannot repro the issue anymore, something may have changed SDK side. I will also close the issue when I will know what has changed.

PR #2592, opened 2024-03-22
Replace leftover hardcoded string with the right text

jmartinesp:


bmarty (approved):
Thanks. I wish lint will check issues like this at some point...

PR #2618, opened 2024-03-28
Fix analytics

### LINKED ISSUE #2612: Fix analytics issue.
This code:

```kotlin
        isSpace = MatrixPatterns.isSpaceId(this.roomId.value),
```

https://github.com/element-hq/element-x-android/blob/main/services/analytics/api/src/main/kotlin/io/element/android/services/analytics/api/extensions/ViewRoomExt.kt#L29

is not correct. `isSpaceId` returns `true` for all valid `roomId`, so the analytics are wrong.

bmarty:
spaceId have the same syntax as roomId. Use MatrixRoom.isSpace instead of MatrixPatterns.isSpaceId (#2612)

Also remove useless orFalse() calls.

Closes #2612

### DIFF
--- a/changelog.d/2612.bugfix
+++ b/changelog.d/2612.bugfix
@@ -1,0 +1,1 @@
+Fix analytics issue around room considered as space by mistake.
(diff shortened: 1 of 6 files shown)


ganfra (approved):
LGTM

PR #2672, opened 2024-04-08
Add `SessionData.needsVerification` field

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Add `SessionData.needsVerification` field, which will be used to decide whether to display the verification step in FTUE flow or not. It's enabled by default so previous users of the app will have to verify their session if they haven't done it yet.
- Add a 'skip verification' button for debug builds.
- Checks verification actually worked before finishing the flow, otherwise displays the 'failed' screen so the user can retry.

## Motivation and context

- Allows us to add a skip button in session verification flow for debug builds.
- We can have the verification state almost instantly for `LoggedInFlowNode`.
- It doesn't depend on network availability to know the verification state and display the UI.

## Tests

With a debug build, log into a new account and:

- Check the skip button is present and works as expected.
- Verify the recovery key flow now displays the 'completed' screen when successful.
- Verify the emoji verification still works as expected.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist


features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/FtueSessionVerificationFlowNode.kt
```diff
@@ -83,7 +88,12 @@ class FtueSessionVerificationFlowNode @AssistedInject constructor(
                     .params(SecureBackupEntryPoint.Params(SecureBackupEntryPoint.InitialTarget.EnterRecoveryKey))
                     .callback(object : SecureBackupEntryPoint.Callback {
                         override fun onDone() {
-                            callback.onDone()
+                            lifecycleScope.launch {
+                                // Mark the session as verified
+                                sessionVerificationService.skipVerification()
```
bmarty, line 93: It's weird to me to see a call to `sessionVerificationService.skipVerification()` here. The service should know when the session is verified, there should be no need to call `skipVerification` here (and it's not skipping, since it's done IIUC).

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/verification/RustSessionVerificationService.kt
```diff
@@ -165,8 +181,27 @@ class RustSessionVerificationService(
...
+            runCatching {
+                withTimeout(5.seconds) {
+                    while (!verificationController.isVerified()) {
+                        delay(100)
+                    }
+                }
+            }
+                .onSuccess {
+                    val existingSession = sessionStore.getSession(sessionId) ?: error("No session to restore with id $sessionId")
+                    sessionStore.updateData(existingSession.copy(needsVerification = false))
```
bmarty, line 197: I see that setting `needsVerification` to `false` is done here, so I guess my first comment should just be "just remove this code"?

bmarty: Or also set the boolean in case of failure?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/verification/RustSessionVerificationService.kt
```diff
@@ -188,6 +223,14 @@ class RustSessionVerificationService(
...
     }
 
+    override suspend fun skipVerification(): Boolean = runCatching {
+        val existingSession = sessionStore.getSession(client.userId()) ?: error("No session with id ${client.userId()}")
+        sessionStore.updateData(existingSession.copy(needsVerification = false))
+        true
+    }
+        .onFailure { Timber.e(it, "Failed to skip verification") }
+        .getOrDefault(false)
+
```
bmarty, line 233: The returned boolean does not seem to be used, maybe this API should return `Result<Unit>` instead of `Boolean`?

libraries/matrix/test/src/main/kotlin/io/element/android/libraries/matrix/test/verification/FakeSessionVerificationService.kt
```diff
@@ -20,17 +20,22 @@ import io.element.android.libraries.matrix.api.verification.SessionVerificationD
...
 import io.element.android.libraries.matrix.api.verification.VerificationFlowState
+import io.element.android.tests.testutils.lambda.LambdaNoParamRecorder
+import io.element.android.tests.testutils.lambda.lambdaRecorder
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 
-class FakeSessionVerificationService : SessionVerificationService {
+class FakeSessionVerificationService(
+    var skipVerificationResult: LambdaNoParamRecorder<Boolean> = lambdaRecorder<Boolean>(ensureNeverCalled = false, block = { false })
```
bmarty, line 30: `ensureNeverCalled` is set to false by default, so `ensureNeverCalled = false` could be removed, but maybe to set the default value of `skipVerificationResult` it should be set to `true`, to actually correctly cover other unrelated tests?
```suggestion
    var skipVerificationResult: LambdaNoParamRecorder<Boolean> = lambdaRecorder<Boolean>(ensureNeverCalled = true, block = { false })
```

No sure about it though, but this is what I would do.

libraries/session-storage/api/src/main/kotlin/io/element/android/libraries/sessionstorage/api/SessionData.kt
```diff
@@ -18,16 +18,32 @@ package io.element.android.libraries.sessionstorage.api
...
     val oidcData: String?,
+    /** The Sliding Sync Proxy URL for this session, if any. */
     val slidingSyncProxy: String?,
+    /** The timestamp of the last login. May be `null` in very old sessions. */
     val loginTimestamp: Date?,
+    /** Whether the [accessToken] is valid or not. */
     val isTokenValid: Boolean,
+    /** The login type used to authenticate the session. */
     val loginType: LoginType,
+    /** The optional passphrase used to encrypt data in the SDK local store. */
```
bmarty, line 45: Thanks for the doc, I though after reading it above that it was the passphrase of the user, and it would be an error to store it in DB :)

libraries/session-storage/impl/src/main/sqldelight/migrations/5.sqm
```diff
@@ -0,0 +1,3 @@
+-- Migrate DB from version 5
```
bmarty, line 1: Maybe document here that for migrating users, we do not force the verification of the session?

features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/VerifySelfSessionStateMachine.kt
```diff
@@ -216,5 +219,8 @@ class VerifySelfSessionStateMachine @Inject constructor(
 
         /** Reset the verification flow to the initial state. */
         data object Reset : Event
+
+        /** Skip the verification flow. */
+        data object Skip : Event
```
bmarty, line 224: I would not touch the Verification state machine, skipping the verification should only affect the FTUE state. Except if we want to add a specific "Skipped verification" screen?

bmarty (commented):
Thanks for the PR.

I know this is a feature only for developers, but seeing this screen after skipping is a bit confusing:

image

I would directly jump into the next FTUE screen (notification permission / analytics)

Also, after skipping, I should still be able to verify the session, and the option is not displayed anymore in the settings. Or does skipping mean that it's not possible to verify the session anymore? Again, since this is only an option for developers?

I think the feedback above ^ are also duplicates of inline comments, sorry for the repetition :)

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/FtueSessionVerificationFlowNode.kt
jmartinesp: That only works for session verification through emoji, since we have the `didFinish` callback, but it won't work for verification through recovery key.

We need something else to mark the session as verified. Since the logic is exactly the same as `skipVerification` I could just rename it to something more generic, or I could add the same code in `SecureBackupEnterRecoveryKeyPresenter`. WDYT?

jmartinesp:
> I would directly jump into the next FTUE screen (notification permission / analytics)

Done, good call.

> Also, after skipping, I should still be able to verify the session, and the option is not displayed anymore in the settings. Or does skipping mean that it's not possible to verify the session anymore? Again, since this is only an option for developers?

One of the requirements of a previous story was to remove any way to verify the session after the FTUE flow, so sadly that's not possible anymore. We could bring it back just for debug builds, but I'm not sure if it's worth the effort.

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/FtueSessionVerificationFlowNode.kt
jmartinesp: I've done it in the presenter instead to avoid cyclic dependencies between `SessionVerificationService` and `EncryptionService`.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/verification/RustSessionVerificationService.kt
```diff
@@ -188,12 +195,21 @@ class RustSessionVerificationService(
         _verificationFlowState.value = VerificationFlowState.Initial
     }
 
+    override suspend fun saveVerifiedState(verified: Boolean) = tryOrFail {
+        val existingSession = sessionStore.getSession(client.userId())
+            ?: error("Failed to save verification state. No session with id ${client.userId()}")
+        sessionStore.updateData(existingSession.copy(needsVerification = !verified))
+        // Wait until the new state is saved
```
bmarty, line 202: Why we need to wait here?

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/FtueSessionVerificationFlowNode.kt
bmarty: I do not understand, isn't it possible to inject `private val sessionStore: SessionStore` to `RustEncryptionService`?

And write:

```kotlin
    override suspend fun recover(recoveryKey: String): Result<Unit> = withContext(dispatchers.io) {
        runCatching {
            service.recover(recoveryKey)
            val existingSession = sessionStore.getSession(client.userId())
                ?: error("Failed to save verification state. No session with id ${client.userId()}")
            sessionStore.updateData(existingSession.copy(needsVerification = false))
        }.mapFailure {
            it.mapRecoveryException()
        }
    }
```

bmarty (commented):
Sorry, I have more remarks

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/verification/RustSessionVerificationService.kt
jmartinesp: Because otherwise if you go back immediately to any of the screens that use `needsVerificationFlow`, the state is still `true`. I found it odd too since I expected the new value to be emitted by the time `sessionStore.updateData` is finished, but that didn't seem to be the case.

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/FtueSessionVerificationFlowNode.kt
jmartinesp: I wanted to reuse the method in `SessionVerificationService` to centralise this kind of change, but I can do it like this too.

libraries/session-storage/impl/src/main/sqldelight/migrations/5.sqm
bmarty: After internal discussion, I am not sure about the choice to let existing unverified user access the application. Maybe just force users to verify their session, even after an upgrade of the application?
But the value ("is the session verified") would have to be read from the SDK.
CC @pmaier1

features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/VerifySelfSessionView.kt
```diff
@@ -114,20 +140,21 @@ private fun HeaderContent(verificationFlowStep: FlowStep) {
...
             is SessionVerificationData.Emojis -> R.string.screen_session_verification_compare_emojis_title
         }
+        is FlowStep.Skipped -> return
     }
     val subtitleTextId = when (verificationFlowStep) {
         is FlowStep.Initial, FlowStep.AwaitingOtherDeviceResponse -> R.string.screen_identity_confirmation_subtitle
         FlowStep.Canceled -> R.string.screen_session_verification_cancelled_subtitle
         FlowStep.Ready -> R.string.screen_session_verification_ready_subtitle
-        FlowStep.Completed -> R.string.screen_identity_confirmation_subtitle
+        FlowStep.Completed -> R.string.screen_identity_confirmed_subtitle
```
bmarty, line 149: Good spot!

bmarty (approved):
LGTM thanks!

PR #2686, opened 2024-04-10
Prevent the word "Element" from being included in string resources.

### LINKED ISSUE #2685: [Task] Prevent "Element" term to be used in Localazy
We want to ease forks of the application to use their name in some strings which reference the application, and so it's not possible to use "Element" in the string value added to Localazy.
We already use placeholder `%1$s` for instance, that are replaced at runtime by the application name (which can be "Element", or the forked application name), but this issue is about to ensure that `Element` is not used again by mistake.

We also want to create a special value to name the Element desktop application. It can just be `Element` for now (or `Element Web`?), but forks may want to use something else. This value will be added as a new field of `BuildMeta`: `desktopApplicationName`.

The detected strings containing `Element` so far are:

- [x] `screen_qr_code_login_initial_state_item_1`: "Open Element on a desktop device"
**Replace with placeholder and use value for desktop app**
- [x] `screen_qr_code_login_initial_state_title`: "Open Element on another device to get the QR code"
**Replace with placeholder**
- [x] `screen_create_room_action_invite_people`: "Invite people to Element"
**This string is not used anymore and can be deleted on Localazy**
- [x] `screen_advanced_settings_element_call_base_url`: "Custom Element Call base URL"
**Only used in developer option, so not in prod, can be ignored**
- [x] `screen_advanced_settings_element_call_base_url_description`: "Set a custom base URL for Element Call."
**Only used in developer option, so not in prod, can be ignored**
- [x] `screen_create_new_recovery_key_list_item_1`: "Open Element in a desktop device"
**Replace with placeholder and use element desktop string and change to `Open Element on a desktop device`**
- [x] `screen_server_confirmation_message_login_element_dot_io`: "A private server for Element employees."
**only used on iOS and for element.io homeserver, so this is fine to ignore it**
- [x] `screen_onboarding_welcome_message`: "Welcome to the fastest Element ever. Supercharged for speed and simplicity." **Replace with placeholder**
- [x] `screen_onboarding_welcome_title`: "Be in your element"
**will probably be changed on the forks, so we can ignore**

Strings where placeholder are added need some change on the code (Android and iOS) to replace the placeholder by the application name.

A tool will be added to the project to detect errors when importing string from Localazy. It will be up to the developer to fix the issue during string importation if there are reported by the script.

Note: there are also some iOS string to update:

image

bmarty:
Add verification on String content when importing string from Localazy. In particular, the term `Element` should not be used.

The developer will still have to check the output of the script, this is not a blocking failure, and the CI will not check for forbidden strings.

Also fixing (or ignoring) existing errors, see the list in #2685.

Closes #2685


tools/localazy/checkForbiddenTerms.py
```diff
@@ -0,0 +1,64 @@
...
+        "Element",
+        [
+            # Those 2 strings are only used in debug version
+            "screen_advanced_settings_element_call_base_url",
+            "screen_advanced_settings_element_call_base_url_description",
+            # only used for element.io homeserver, so it's fine
+            "screen_server_confirmation_message_login_element_dot_io",
+            # "Be in your element", will probably be changed on the forks, so we can ignore.
+            "screen_onboarding_welcome_title",
+        ]
```
jmartinesp, line 20: This format seems a bit weird, couldn't we get something similar with a dictionary?

jmartinesp (approved):
LGTM, just a minor comment about the `term - exceptions` mapping. Thanks for writing the script!

tools/localazy/checkForbiddenTerms.py
bmarty: Yes, good idea: https://github.com/element-hq/element-x-android/pull/2686/commits/b0ff24347f76b96892a2bf221749ca685af51b00

bmarty:
@jmartinesp I have added a few commits, let me know if it's still OK for you. Thanks!

jmartinesp (approved):
The new changes look good to me, thanks. There seem to be some other CI issues though.

PR #2741, opened 2024-04-22
Iterate on `loadingNode` to render the same way JoinRoomView will render

bmarty:
Ready to be reviewed.

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/room/RoomFlowNode.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/room/RoomFlowNode.kt
@@ -17,10 +17,9 @@
 package io.element.android.appnav.room
 
 import android.os.Parcelable
-import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.runtime.Composable
-import androidx.compose.ui.Alignment
+import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.getValue
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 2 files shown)


frebib:
Not sure this is correct. Looks to have the wrong placeholder for me 
https://github.com/element-hq/element-x-android/assets/775104/b85c6f44-9296-42b3-852f-30dacb87d4d0

bmarty:
@frebib you're right, good catch, thanks! Let's iterate to look more like the existing room loading view, which will be more frequent than clicking on an unknown room permalink.

appnav/src/main/kotlin/io/element/android/appnav/room/RoomFlowNode.kt
```diff
@@ -172,10 +175,14 @@ class RoomFlowNode @AssistedInject constructor(
...
 
-    private fun loadingNode(buildContext: BuildContext) = node(buildContext) {
-        Box(modifier = it.fillMaxSize(), contentAlignment = Alignment.Center) {
-            CircularProgressIndicator()
-        }
+    private fun loadingNode(buildContext: BuildContext) = node(buildContext) { modifier ->
+        val networkStatus by networkMonitor.connectivity.collectAsState()
+        LoadingRoomNodeView(
+            state = LoadingRoomState.Loading,
+            hasNetworkConnection = networkStatus == NetworkStatus.Online,
```
ganfra, line 182: This is a bit weird to show the hasNetwork connection here and not in room preview :/

ganfra (approved):

PR #2743, opened 2024-04-23
Handle properly the migration of DefaultSessionPreferencesStore #2742

### LINKED ISSUE #2742: Application freeze when going to Advanced Settings
### Steps to reproduce

1. Go to the setting and "Advanced settings"

#### What did you expect?

Advanced setting screen is displayed

#### What happened instead?

The app freeze

bmarty:
Closes #2742 

Can be tested with the debugger, to ensure that the migration occurs only once.


libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStore.kt
```diff
@@ -53,7 +51,15 @@ class DefaultSessionPreferencesStore(
     private val renderTypingNotificationsKey = booleanPreferencesKey("renderTypingNotifications")
 
     private val dataStoreFile = storeFile(context, sessionId)
-    private val store = PreferenceDataStoreFactory.create(scope = sessionCoroutineScope) { dataStoreFile }
+    private val store = PreferenceDataStoreFactory.create(
+        scope = sessionCoroutineScope,
+        migrations = listOf(
```
jmartinesp, line 56: 👀  TIL

jmartinesp (approved):
LGTM, thanks!

PR #2757, opened 2024-04-26
Fix potential infinite loading at first startup.

bmarty:
There is an issue with datastore 1.1.0, that I cannot repro on unit test.

To repro with the application:
- Clear the cache of the application and run the app. Nearly each time, there is an infinite loading. If not do it again.
- Due to the `DefaultMigrationStore` not behaving as expected

Stick to 1.0.0 for now.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -10,7 +10,12 @@ firebaseAppDistribution = "4.2.0"
 
 # AndroidX
 core = "1.13.0"
-datastore = "1.1.0"
+# Warning: there is an issue with 1.1.0, that I cannot repro on unit test.
+# To repro with the application:
+# Clear the cache of the application and run the app. Nearly each time, there is an infinite loading
+# Due to the DefaultMigrationStore not bahaving as expected.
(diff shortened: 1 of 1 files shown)


gradle/libs.versions.toml
```diff
@@ -10,7 +10,12 @@ firebaseAppDistribution = "4.2.0"
 
 # AndroidX
 core = "1.13.0"
-datastore = "1.1.0"
+# Warning: there is an issue with 1.1.0, that I cannot repro on unit test.
+# To repro with the application:
+# Clear the cache of the application and run the app. Nearly each time, there is an infinite loading
+# Due to the DefaultMigrationStore not bahaving as expected.
+# Stick to 1.0.0 for now, and ensure that this scenario cannot be reproduced when upgrading the version.
```
ElementBot, line 17: - :warning: A newer version of androidx.activity:activity-compose than 1.8.2 is available: 1.9.0

jmartinesp (approved):
LGTM. I could reproduce the issue with datastore `1.1.0`, I suspect it could be related to trying to set the pref's value several times in a short amount of time?

It only happened when I tried to set `applicationMigrationVersion` to 0 to reset the migrations and then it's changed to 1 a few ms later.

PR #2759, opened 2024-04-26
Permalink timeline

bmarty:
Draft PR.


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -152,46 +154,66 @@ fun TimelineView(
...
 @Composable
 private fun BoxScope.TimelineScrollHelper(
-    isTimelineEmpty: Boolean,
+    hasAnyEvent: Boolean,
     lazyListState: LazyListState,
     newEventState: NewEventState,
+    isLive: Boolean,
     forceJumpToBottomVisibility: Boolean,
+    focusRequestState: FocusRequestState,
+    onClearFocusRequestState: () -> Unit,
```
ElementBot, line 208: - :warning: Lambda parameters in a @Composable that are referenced directly inside of restarting effects can cause issues or unpredictable behavior.

If restarting the effect is ok, you can add the reference to this parameter as a key in that effect, so when the parameter changes, a new effect is created.
However, if the effect is not to be restarted, you will need to use `rememberUpdatedState` on the parameter and use its result in the effect.

See https://mrmans0n.github.io/compose-rules/rules/#be-mindful-of-the-arguments-you-use-inside-of-a-restarting-effect for more information.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemRow.kt
```diff
@@ -47,69 +58,110 @@ internal fun TimelineItemRow(
...
-                onReactionLongClick = onReactionLongClick,
-                onMoreReactionsClick = onMoreReactionsClick,
-                onReadReceiptClick = onReadReceiptClick,
-                eventSink = eventSink,
-                modifier = modifier,
+    }
+}
+
+@Composable
+private fun Modifier.focusedEvent(
```
ElementBot, line 139: - :warning: Using @Composable builder functions for modifiers is not recommended, as they cause unnecessary recompositions.
You should consider migrating this modifier to be based on Modifier.Node instead.

See https://mrmans0n.github.io/compose-rules/rules/#avoid-modifier-extension-factory-functions for more information.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemVirtualRow.kt
```diff
@@ -16,24 +16,48 @@
...
+import io.element.android.features.messages.impl.timeline.model.virtual.TimelineItemLastForwardIndicatorModel
+import io.element.android.features.messages.impl.timeline.model.virtual.TimelineItemLoadingIndicatorModel
 import io.element.android.features.messages.impl.timeline.model.virtual.TimelineItemReadMarkerModel
+import io.element.android.features.messages.impl.timeline.model.virtual.TimelineItemRoomBeginningModel
 
 @Composable
 fun TimelineItemVirtualRow(
     virtual: TimelineItem.Virtual,
+    timelineRoomInfo: TimelineRoomInfo,
+    eventSink: (TimelineEvents.EventFromTimelineItem) -> Unit,
```
ElementBot, line 43: - :warning: Lambda parameters in a @Composable that are referenced directly inside of restarting effects can cause issues or unpredictable behavior.

If restarting the effect is ok, you can add the reference to this parameter as a key in that effect, so when the parameter changes, a new effect is created.
However, if the effect is not to be restarted, you will need to use `rememberUpdatedState` on the parameter and use its result in the effect.

See https://mrmans0n.github.io/compose-rules/rules/#be-mindful-of-the-arguments-you-use-inside-of-a-restarting-effect for more information.

features/location/impl/src/test/kotlin/io/element/android/features/location/impl/common/actions/AndroidLocationActionsTest.kt
```diff
@@ -18,9 +18,11 @@ package io.element.android.features.location.impl.common.actions
 
 import com.google.common.truth.Truth.assertThat
 import io.element.android.features.location.api.Location
+import org.junit.Ignore
 import org.junit.Test
 import java.net.URLEncoder
 
+@Ignore
```
ElementBot, line 25: - :warning: Test is ignored without giving any explanation

features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl/timeline/TimelineItemIndexerTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import io.element.android.libraries.matrix.api.core.EventId
+import io.element.android.libraries.matrix.test.AN_EVENT_ID
+import org.junit.Test
+
+class TimelineItemIndexerTest {
+    @Test
+    fun `test TimelineItemIndexer`() {
+        val eventIds = mutableListOf<EventId>()
+        val data = listOf(
+            aTimelineItemEvent().also { eventIds.add(it.eventId!!) },
```
jmartinesp, line 31: Maybe make sure all ids used here don't match? They're random, so there's not much of a chance of it happening, so up to you.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/MatrixRoom.kt
```diff
@@ -98,7 +98,9 @@ interface MatrixRoom : Closeable {
 
     val syncUpdateFlow: StateFlow<Long>
 
-    val timeline: MatrixTimeline
+    val liveTimeline: Timeline
```
jmartinesp, line 101: Maybe add docs for this and `timelineFocusedOnEvent`?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/postprocessor/LastForwardIndicatorsPostProcessor.kt
```diff
@@ -0,0 +1,75 @@
...
+
+private fun createLastForwardIndicator(identifier: String): MatrixTimelineItem {
+    return MatrixTimelineItem.Virtual(
+        uniqueId = "last_forward_indicator_$identifier",
+        virtual = VirtualTimelineItem.LastForwardIndicator
+    )
+}
+
+private fun List<MatrixTimelineItem>.latestEventIdentifier(): String {
+    return findLast {
```
jmartinesp, line 67: It could by `findLast { it is MatrixTimelineItem.Event` } without then `when`.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/postprocessor/RoomBeginningPostProcessor.kt
```diff
@@ -16,23 +16,38 @@
...
 import io.element.android.libraries.matrix.api.timeline.item.event.MembershipChange
 import io.element.android.libraries.matrix.api.timeline.item.event.OtherState
 import io.element.android.libraries.matrix.api.timeline.item.event.RoomMembershipContent
 import io.element.android.libraries.matrix.api.timeline.item.event.StateContent
+import io.element.android.libraries.matrix.api.timeline.item.virtual.VirtualTimelineItem
 
 /**
- * This timeline post-processor removes the room creation event and the self-join event from the timeline for DMs.
+ * This timeline post-processor removes the room creation event and the self-join event from the timeline for DMs
+ * or add the RoomBeginning item for non DM room.
```
jmartinesp, line 29: I think I'd have split this in 2 processors, but it's no big deal.

jmartinesp (commented):
LGTM in general, I have a few comments but nothing blocking. I'll wait until it's ready for the final review.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/virtual/TimelineLoadingMoreIndicator.kt
```diff
@@ -16,35 +16,58 @@
...
-            .fillMaxWidth()
-            .wrapContentHeight()
-            .padding(8.dp),
+        modifier = modifier.fillMaxWidth(),
         contentAlignment = Alignment.Center,
     ) {
-        CircularProgressIndicator(
-            strokeWidth = 2.dp,
-        )
+        when (direction) {
```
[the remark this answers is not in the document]
bmarty: The idea would be to use the `CircularProgressIndicator` in both cases, which looks more appropriate, but to limit the glitch when hiding the `FORWARDS` indicator, we use a indicator with a small height.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/TimelineItemsFactory.kt
```diff
@@ -96,10 +98,12 @@ class TimelineItemsFactory @Inject constructor(
                 } else {
                     cacheItem
                 }
+                newTimelineById[updatedItem.identifier()] = updatedItem
```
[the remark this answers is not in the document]
bmarty: It seems that you're right, I'll remove it.

features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl/timeline/TimelineItemIndexerTest.kt
bmarty: Yes, I guess it's fine. I preferred to keep the randomness and store ids in `eventIds`, another solution would have been to create `AN_EVENT_ID_3` and `AN_EVENT_ID_4`. Let's revisit later if it's worth it.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/MatrixRoom.kt
bmarty: Yes, added. Also I believe we will sort of rework this API to distinguish the live timeline and the detached timeline.
We should not use the detached timeline to send an event for instance.

bmarty:
I think there is an issue with `enterSpecialMode`, which is called on the current timeline here:
https://github.com/element-hq/element-x-android/blob/bf87b975fc987e966659013f413638edd2e53831/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt#L270

But then message edition or reply is invoked on the live timeline, in RustMatrixRoom.

`RustMatrixRoom.enterSpecialMode` is not used anymore, which is part of the problem.

I think special modes, and so reply and edition have to occurs on the current timeline, because the Event have to be known. But we will not have local echo, which are only handled by the live timeline.

jmartinesp (approved):

PR #2832, opened 2024-05-13
Fix localazy import

bmarty:
[Latest Localazy sync](https://github.com/element-hq/element-x-android/actions/runs/9055194003/job/24875946486) was failing because of Python version not supporting `match` structure. The first commit fixes this. I have updated the version every where Python is used, hopefully there will be no bad side effect on the other GA.

The second commit ensure that the data for simplified Chinese are correct. I'll let the next Localazy sync run the script and update the generated file, to validate it's working as expected (it's OK locally).

### DIFF
--- a/tools/localazy/importSupportedLocalesFromLocalazy.py
+++ b/tools/localazy/importSupportedLocalesFromLocalazy.py
@@ -27,6 +27,8 @@ def normalizeForResourceConfigurations(locale):
             return "in"
         case "zh_TW#Hant":
             return "zh-rTW"
+        case "zh#Hans":
+            return "zh-rCN"
         case _:
             return locale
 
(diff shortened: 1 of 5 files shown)


tools/localazy/importSupportedLocalesFromLocalazy.py
```diff
@@ -27,6 +27,8 @@ def normalizeForResourceConfigurations(locale):
             return "in"
         case "zh_TW#Hant":
             return "zh-rTW"
+        case "zh#Hans":
+            return "zh-rCN"
```
bmarty, line 31: Localazy will import string for simplified Chinese in the folders `values-zh`, and not `values-zh-rCN. I am not sure this need to match the value here.

PR #2839, opened 2024-05-14
It's not possible to join room from the RoomDirectory screen anymore.

bmarty:
Cleanup after #2827


jmartinesp (approved):
Thanks for the cleanup! I left if there in case we wanted to re-enable it in the future with some other UI, but it's probably not the case so it's better to just get rid of it.

bmarty:
(Yes, I needed to do some cleanup to simplify another coming PR)

Issue #2912, opened 2024-05-24
Move Push provider settings to the notification settings


bmarty:
We added a way to change the push provider in #2873 . The setting has been added to the `Advanced settings` screen since it's quite an advanced setting indeed.

It's also having sense to have it to the `Notification` settings screen, but we did not want to annoy regular users with this power user setting.

Maybe another solution which will check the 2 requirements would be to move the setting to the `Notifications` setting screen, in an new `Advanced` section, but only displayed if several push providers are detected. With this solution:

- Gplay user with only Firebase will not see this setting
- Gplay users with a UnifiedPush provider installed (such as `ntfy`) can be considered as power users, since they have installed an alternative PushProvider and so will see the setting in the `Notifications` setting screen. 
- FDroid users will not have Firebase support, but will need to install a UnifiedPush provider. The `Notifications` setting screen may be a place to alert the user that no PushProvider is installed if this is the case. We are already alerting if Notification permission is not granted (see the screenshot below).

|Only one Push provider detected|Several push providers detected|
|-|-|
|<img width="323" alt="image" src="https://github.com/element-hq/element-x-android/assets/3940906/76b451fe-d068-4e5a-b8cc-062232988f80">|<img width="325" alt="image" src="https://github.com/element-hq/element-x-android/assets/3940906/8a9796e9-cccf-46c7-80e8-55d3c8bd92f2">|

bmarty:
(Approved internally by @amshakal)

damnms:
I am running 0.7.5 - i dont have this advanced feature, even if i turn on dev options. Where can i configure the push provider?

PR #2922, opened 2024-05-27
Iterate on preference design

bmarty:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Let Preference Composable use ListItem under the hood.

## Motivation and context

Let the application globally follow the design.

(And preparatory work for #2912)

## Screenshots / GIFs

See recorded once.

## Tests

- Navigate to the various preference screen and check that there is no issues.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/preferences/PreferenceCategory.kt
```diff
@@ -19,53 +19,41 @@ package io.element.android.libraries.designsystem.components.preferences
...
-import io.element.android.libraries.designsystem.theme.components.Text
+import io.element.android.libraries.designsystem.theme.components.ListSectionHeader
 
 @Composable
 fun PreferenceCategory(
     modifier: Modifier = Modifier,
     title: String? = null,
-    showDivider: Boolean = true,
+    showTopDivider: Boolean = true,
+    showBottomDivider: Boolean = false,
```
jmartinesp, line 35: `showBottomDivider` is never true, should we remove it? The `ListSectionHeader` component was built to always use only the top one ( which I admit is a bit awkward) so it shouldn't be needed.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/preferences/PreferenceSwitch.kt
```diff
@@ -53,62 +42,57 @@ fun PreferenceSwitch(
...
+            showIconAreaIfNoIcon = showIconAreaIfNoIcon,
+        ),
+        headlineContent = {
             Text(
                 style = ElementTheme.typography.fontBodyLgRegular,
                 text = title,
-                color = enabled.toEnabledColor(),
             )
+        },
+        supportingContent = {
```
jmartinesp, line 62: This should probably be `supportingContent = subtitle?.let { ... }`, I think otherwise the preference will add some spacing that's not really needed.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/preferences/PreferenceText.kt
```diff
@@ -67,34 +63,26 @@ fun PreferenceText(
...
+            tintColor = tintColor,
+        ),
+        headlineContent = {
             Text(
                 style = ElementTheme.typography.fontBodyLgRegular,
                 text = title,
                 color = tintColor ?: enabled.toEnabledColor(),
             )
+        },
+        supportingContent = {
```
jmartinesp, line 85: Same here about the spacing, maybe inline the `if` here and add composable blocks to each branch, and `null` for the `else`.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/preferences/PreferenceText.kt
```diff
@@ -108,35 +96,35 @@ fun PreferenceText(
...
-            val endBadgeStartPadding = if (currentValue != null || loadingCurrentValue) 8.dp else 16.dp
-            RedIndicatorAtom(
-                modifier = Modifier
-                    .align(Alignment.CenterVertically)
-                    .padding(start = endBadgeStartPadding)
-            )
-        }
-    }
+        },
+        trailingContent = ListItemContent.Custom {
```
jmartinesp, line 100: Same about nullability here. If there's nothing to display it should be `null`, not an empty row.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/preferences/components/PreferenceIcon.kt
```diff
@@ -31,13 +30,39 @@ import androidx.compose.ui.tooling.preview.Preview
...
+fun preferenceIcon(
+    icon: ImageVector? = null,
+    @DrawableRes iconResourceId: Int? = null,
+    showIconBadge: Boolean = false,
+    tintColor: Color? = null,
+    enabled: Boolean = true,
+    showIconAreaIfNoIcon: Boolean = false,
+): ListItemContent.Custom? {
+    return if (icon != null || iconResourceId != null || showIconAreaIfNoIcon) {
+        ListItemContent.Custom {
```
jmartinesp, line 49: This isn't wrong, but maybe it'd be better to use the exiting `ListItemContent.Icon` when possible, maybe adding the badge option if needed, or a nullable version for the empty space. I'm not sure though, since that may complicate its usage.

jmartinesp (commented):
Thanks for the work on this. To be honest I was going in the opposite direction, slowly replacing `Preference*` components with plain `ListItem` ones since that's what Compound gives us, so maybe we should first discuss which approach to take. 

I'm not opposed to having EXA only wrappers but maybe they're not really needed in some cases, I'm not sure if this is one of them.

Besides that, I found a few things that could be improved since I think they'll affect paddings and internal margins.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/preferences/PreferenceSwitch.kt
bmarty: Yes, having an empty Composable does not seems to have an impact, but let's do what you're suggesting to be sure.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/preferences/components/PreferenceIcon.kt
bmarty: Yes, there is room for improvement. "adding the badge option if needed" to `ListItemContent.Icon` should be my choice but I prefer to work on this later.

bmarty:
> I'm also quite confused about the coverage drop, the amount of covered lines should be pretty much the same 🫤 .

Yes, let's ignore this for now...

bmarty:
> Thanks for the work on this. To be honest I was going in the opposite direction, slowly replacing `Preference*` components with plain `ListItem` ones since that's what Compound gives us, so maybe we should first discuss which approach to take.
> 
> I'm not opposed to having EXA only wrappers but maybe they're not really needed in some cases, I'm not sure if this is one of them.

Having `Preference*` Composable helps to have compact code, but yes, maybe one day we could get ride of them, and directly use `ListItem`. I consider those changes as a step to this direction.

jmartinesp (approved):
Thanks for the changes, LGTM!

PR #2928, opened 2024-05-28
Move push provider setting

### LINKED ISSUE #2912: Move Push provider settings to the notification settings

bmarty:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [x] Other : iterate on setting screens

## Content

Move push provider setting to the "Notifications" screen and display it only when several push provider are available.

## Motivation and context

Closes #2912

## Screenshots / GIFs

See recorded ones

## Tests

- Install ntfy and check that changing the push provider from the new location (Settings/Notifications) still works

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/preferences/impl/src/test/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenterTests.kt
+++ b/features/preferences/impl/src/test/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenterTests.kt
@@ -20,7 +20,7 @@ import app.cash.molecule.RecompositionMode
 import app.cash.molecule.moleculeFlow
 import app.cash.turbine.test
 import com.google.common.truth.Truth.assertThat
-import io.element.android.libraries.architecture.AsyncAction
+import io.element.android.libraries.architecture.AsyncData
 import io.element.android.libraries.matrix.api.MatrixClient
 import io.element.android.libraries.matrix.api.room.RoomNotificationMode
 import io.element.android.libraries.matrix.test.A_THROWABLE
@@ -247,7 +247,7 @@ class NotificationSettingsPresenterTests {
             presenter.present()
         }.test {
             val initialState = awaitLastSequentialItem()
-            assertThat(initialState.currentPushDistributor).isEqualTo(AsyncAction.Success("aDistributorName0"))
+            assertThat(initialState.currentPushDistributor).isEqualTo(AsyncData.Success("aDistributorName0"))
             assertThat(initialState.availablePushDistributors).containsExactly("aDistributorName0", "aDistributorName1")
             initialState.eventSink.invoke(NotificationSettingsEvents.ChangePushProvider)
             val withDialog = awaitItem()
@@ -261,10 +261,10 @@ class NotificationSettingsPresenterTests {
             withDialog.eventSink(NotificationSettingsEvents.SetPushProvider(1))
             val withNewProvider = awaitItem()
             assertThat(withNewProvider.showChangePushProviderDialog).isFalse()
-            assertThat(withNewProvider.currentPushDistributor).isEqualTo(AsyncAction.Loading)
+            assertThat(withNewProvider.currentPushDistributor).isInstanceOf(AsyncData.Loading::class.java)
             skipItems(1)
             val lastItem = awaitItem()
-            assertThat(lastItem.currentPushDistributor).isEqualTo(AsyncAction.Success("aDistributorName1"))
+            assertThat(lastItem.currentPushDistributor).isEqualTo(AsyncData.Success("aDistributorName1"))
             cancelAndIgnoreRemainingEvents()
         }
     }
@@ -288,9 +288,9 @@ class NotificationSettingsPresenterTests {
             withDialog.eventSink(NotificationSettingsEvents.SetPushProvider(1))
             val withNewProvider = awaitItem()
             assertThat(withNewProvider.showChangePushProviderDialog).isFalse()
-            assertThat(withNewProvider.currentPushDistributor).isEqualTo(AsyncAction.Loading)
+            assertThat(withNewProvider.currentPushDistributor).isInstanceOf(AsyncData.Loading::class.java)
             val lastItem = awaitItem()
-            assertThat(lastItem.currentPushDistributor).isInstanceOf(AsyncAction.Failure::class.java)
+            assertThat(lastItem.currentPushDistributor).isInstanceOf(AsyncData.Failure::class.java)
         }
     }
 


features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenter.kt
```diff
@@ -68,6 +76,61 @@ class NotificationSettingsPresenter @Inject constructor(
...
+                        pushProvider to distributor
+                    }
+                }
+        }
+        // List of Distributor names
+        val distributorNames = remember {
+            distributors.map { it.second.name }
+        }
+
+        var currentDistributorName by remember { mutableStateOf<AsyncAction<String>>(AsyncAction.Uninitialized) }
```
jmartinesp, line 93: Should this be `AsyncData` instead?

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenter.kt
```diff
@@ -68,6 +76,61 @@ class NotificationSettingsPresenter @Inject constructor(
...
+        var refreshPushProvider by remember { mutableIntStateOf(0) }
+
+        LaunchedEffect(refreshPushProvider) {
```
jmartinesp, line 96: Is `refreshPushProvider` really needed? It only changes when the push provider is set, so it seems like it's doing the same job twice.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenter.kt
```diff
@@ -68,6 +76,61 @@ class NotificationSettingsPresenter @Inject constructor(
...
+            pushService.getAvailablePushProviders()
+                .flatMap { pushProvider ->
+                    pushProvider.getDistributors().map { distributor ->
+                        pushProvider to distributor
+                    }
+                }
+        }
+        // List of Distributor names
+        val distributorNames = remember {
+            distributors.map { it.second.name }
```
jmartinesp, line 90: Maybe use `toImmutableList()` here instead of in the return so it's only done once?

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsState.kt
```diff
@@ -46,4 +50,10 @@ data class NotificationSettingsState(
         val systemNotificationsEnabled: Boolean,
         val appNotificationsEnabled: Boolean,
     )
+
+    /**
+     * Whether the advanced settings should be shown.
+     * This is true if the current push distributor is in a failure state or if there are multiple push distributors available.
+     */
+    val showAdvancedSettings: Boolean = currentPushDistributor.isFailure() || availablePushDistributors.size > 1
```
jmartinesp, line 58: This should probably use a `get()` option since otherwise I think doing a `state.copy(availableDistributors = ...)` would change the distributors but not this value, and the current state of the `currentPushDistributor` won't be taken into account.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsView.kt
```diff
@@ -69,7 +80,7 @@ fun NotificationSettingsView(
             NotificationSettingsState.MatrixSettings.Uninitialized -> return@PreferencePage
             is NotificationSettingsState.MatrixSettings.Valid -> NotificationSettingsContentView(
                 matrixSettings = state.matrixSettings,
```
jmartinesp, line 82: Should we avoid passing both `state.matrixSettings` and `state` separately here?

jmartinesp (approved):
It LGTM in general, but I think there are some point which could be improved. Thanks for the changes!

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenter.kt
bmarty: Could be but `currentDistributorName` is also updated when the user change the distributor, so it's also an AsyncAction... The boundaries between the 2 sealed interface is thin.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenter.kt
bmarty: Yes, you're right, probably a remaining of an old version of the code. Will remove.

bmarty: Actually I prefer to remove the line
```
currentDistributorName = AsyncAction.Success(distributor.name)
```
So that we read again the value from the module.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenter.kt
bmarty: Sorry I do not understand your remark.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsState.kt
bmarty: `state.copy()` will create a new object and so `showAdvancedSettings` will be computed again. Or am I missing something?

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsView.kt
bmarty: I do that to avoid having to cast again `state.matrixSettings`. The type is `MatrixSettings` and here there is a smart cast to `MatrixSettings.Valid`.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenter.kt
jmartinesp: In [line 168](https://github.com/element-hq/element-x-android/pull/2928/files#diff-992aff3ec5cbd1d0a22fae581a1bb1929dfc078d1ffc22711782f4878c4e8822R168) we get this and apply `.toImmutableList()` in every recomposition. Since it won't change ever, it's better to just do it inside the remember, right?

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenter.kt
jmartinesp: It seems a bit more complex, but it should ensure everything is up to date 👍 .

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsView.kt
jmartinesp: Ah, gotcha. I didn't notice the cast.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsState.kt
jmartinesp: Oh, never mind 🤦 . I somehow forgot this is equivalent to:

```kotlin
val showAdvancedSettings: Boolean

init {
	showAdvancedSettings = ...
}
```

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenter.kt
jmartinesp: I mean, having an 'action' called `currentDistributorName` is kind of weird, that's why I thought `AsyncData` fits it better. More so if we're going to use the `LaunchedEffect` to fetch it.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenter.kt
bmarty: 3349df5a24084679af5dfd1a668675196011c1c2

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSettingsPresenter.kt
bmarty: OK: 9a24e7d4afb0d1aba94808f4327949baccc68883

PR #2939, opened 2024-05-29
Fix small issue introduced in #2924.

bmarty:
This code is not used yet, and also the value will always be the same for the 2 API, but let's use the correct API anyway.


jmartinesp (approved):
Oops, nice catch!

PR #2941, opened 2024-05-29
Sync Strings

ElementBot:

Note: doing an extra sync as `pt-BR` have been moved by to `pt` on Localazy interface.


bmarty (approved):
OK

PR #2948, opened 2024-05-29
Konsist upgrade

bmarty:
Replacement for #2529.

Fix API break and fix existing issues.


jmartinesp (approved):
LGTM, thanks! 

Just a note about the Stable/Immutable sealed interfaces: I believe this was true a few versions of compose ago, but maybe we should check this again using the compose compiler reports and see if sealed interfaces can now autodetect its own stability.

EDIT: it seems like it still stands if the sealed interface is used outside the module, so it's not a bad idea to always check for this.

PR #2984, opened 2024-06-05
Incoming share

### LINKED ISSUE #1980: Sharing Support
Share Media like a Picture to Element X -> Room / DM

Easier to send files 

_No response_

### Additional context

_No response_

Yes

bmarty:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Add a quick support to handle sharing from other application (text and files).
There is no preview, the data is sent once the room is selected.

There is a new feature flag `IncomingShare` that can be set to false to disable the feature.

## Motivation and context

Closes #1980 

## Screenshots / GIFs

|Demo|Feature flag demo|
|-|-|
|AndroidShare|AndroidShareFeatureFlagDemo|

## Tests

- Install the app and login to an account
- From another application share text or media (image, etc.)

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
@@ -16,6 +16,7 @@
 
 package io.element.android.appnav
 
+import android.content.Intent
 import android.os.Parcelable
 import androidx.compose.foundation.layout.Box
 import androidx.compose.runtime.Composable
@@ -54,6 +55,7 @@ import io.element.android.features.roomdirectory.api.RoomDescription
 import io.element.android.features.roomdirectory.api.RoomDirectoryEntryPoint
 import io.element.android.features.roomlist.api.RoomListEntryPoint
 import io.element.android.features.securebackup.api.SecureBackupEntryPoint
+import io.element.android.features.share.api.ShareEntryPoint
 import io.element.android.features.userprofile.api.UserProfileEntryPoint
 import io.element.android.libraries.architecture.BackstackView
 import io.element.android.libraries.architecture.BaseFlowNode
@@ -98,6 +100,7 @@ class LoggedInFlowNode @AssistedInject constructor(
     private val networkMonitor: NetworkMonitor,
     private val ftueService: FtueService,
     private val roomDirectoryEntryPoint: RoomDirectoryEntryPoint,
+    private val shareEntryPoint: ShareEntryPoint,
     private val matrixClient: MatrixClient,
     snackbarDispatcher: SnackbarDispatcher,
 ) : BaseFlowNode<LoggedInFlowNode.NavTarget>(
@@ -219,6 +222,9 @@ class LoggedInFlowNode @AssistedInject constructor(
 
         @Parcelize
         data object RoomDirectorySearch : NavTarget
+
+        @Parcelize
+        data class IncomingShare(val intent: Intent) : NavTarget
     }
 
     override fun resolve(navTarget: NavTarget, buildContext: BuildContext): Node {
@@ -375,6 +381,20 @@ class LoggedInFlowNode @AssistedInject constructor(
                     })
                     .build()
             }
+            is NavTarget.IncomingShare -> {
+                shareEntryPoint.nodeBuilder(this, buildContext)
+                    .callback(object : ShareEntryPoint.Callback {
+                        override fun onDone(roomIds: List<RoomId>) {
+                            navigateUp()
+                            if (roomIds.size == 1) {
+                                val targetRoomId = roomIds.first()
+                                backstack.push(NavTarget.Room(targetRoomId.toRoomIdOrAlias()))
+                            }
+                        }
+                    })
+                    .params(ShareEntryPoint.Params(intent = navTarget.intent))
+                    .build()
+            }
         }
     }
 
@@ -410,6 +430,17 @@ class LoggedInFlowNode @AssistedInject constructor(
                 NavTarget.UserProfile(
                     userId = userId,
                 )
+            )
+        }
+    }
+
+    internal suspend fun attachIncomingShare(intent: Intent) {
+        waitForNavTargetAttached { navTarget ->
+            navTarget is NavTarget.RoomList
+        }
+        attachChild<Node> {
+            backstack.push(
+                NavTarget.IncomingShare(intent)
             )
         }
     }
--- a/features/share/api/src/main/kotlin/io/element/android/features/share/api/ShareService.kt
+++ b/features/share/api/src/main/kotlin/io/element/android/features/share/api/ShareService.kt
@@ -1,0 +1,23 @@
+/*
+ * Copyright (c) 2024 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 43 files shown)


features/share/impl/src/main/kotlin/io/element/android/features/share/impl/ShareIntentHandler.kt
```diff
@@ -0,0 +1,125 @@
...
+        val uri: Uri,
+        val mimeType: String,
+    )
+
+    /**
+     * This methods aims to handle incoming share intents.
+     *
+     * @return true if it can handle the intent data, false otherwise
+     */
+    override suspend fun handleIncomingShareIntent(
```
jmartinesp, line 63: Nit: this should probably be in the interface, right?

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -375,6 +381,20 @@ class LoggedInFlowNode @AssistedInject constructor(
                     })
                     .build()
             }
+            is NavTarget.IncomingShare -> {
+                shareEntryPoint.nodeBuilder(this, buildContext)
+                    .callback(object : ShareEntryPoint.Callback {
+                        override fun onDone(roomIds: List<RoomId>) {
+                            navigateUp()
+                            if (roomIds.size == 1) {
```
jmartinesp, line 389: At the moment this is always the case, right? Do we want to limit sharing to just 1 room at a time?

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/compat/Compat.kt
```diff
@@ -16,9 +16,32 @@
...
 
+import android.content.Intent
 import android.content.pm.ApplicationInfo
 import android.content.pm.PackageManager
+import android.content.pm.ResolveInfo
 import android.os.Build
+import android.os.Parcelable
+
+inline fun <reified T> Intent.getParcelableExtraCompat(key: String): T? = when {
+    Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU -> getParcelableExtra(key, T::class.java)
```
jmartinesp, line 27: Maybe use `IntentCompat.getParcelableExtra`?

features/share/impl/src/main/kotlin/io/element/android/features/share/impl/SharePresenter.kt
```diff
@@ -0,0 +1,112 @@
...
+    private val shareIntentHandler: ShareIntentHandler,
+    private val matrixClient: MatrixClient,
+    private val mediaPreProcessor: MediaPreProcessor,
+) : Presenter<ShareState> {
+    @AssistedFactory
+    interface Factory {
+        fun create(intent: Intent): SharePresenter
+    }
+
+    private val shareActionState: MutableState<AsyncAction<List<RoomId>>> = mutableStateOf(AsyncAction.Uninitialized)
```
jmartinesp, line 49: Since this is already a property, wouldn't it be better to use `by` so we can avoid using `.value`? `CoroutineScope.share` wouldn't need it as a parameter either in that case.

jmartinesp (approved):
LGTM, some comments but feel free to ignore them.

features/share/impl/src/main/kotlin/io/element/android/features/share/impl/SharePresenter.kt
bmarty: Sorry, I do not get your point about `by`. Can you share some code snippet please?

jmartinesp: So something like:

```kotlin
private var shareActionState by mutableStateOf<AsyncAction<List<RoomId>>>(AsyncAction.Uninitialized)

fun onRoomSelected(roomIds: List<RoomId>) {
    appCoroutineScope.share(intent, roomIds)
}

@Composable
override fun present(): ShareState {
    fun handleEvents(event: ShareEvents) {
        when (event) {
            ShareEvents.ClearError -> shareActionState = AsyncAction.Uninitialized
        }
    }

    return ShareState(
        shareAction = shareActionState,
        eventSink = { handleEvents(it) }
    )
}

...
```

That wouldn't be possible if the variable was part of the `present` function, but we can do it if it's a property in the presenter class.

bmarty: Strangely it does not compile here:

image

bmarty: because the type is not correct with `by`:

image

jmartinesp: Ah, true, I completely forgot about `runCatchingUpdatingState`, sorry. No need to change it then.

amshakal:
I am happy with this solution :)

features/share/impl/src/main/kotlin/io/element/android/features/share/impl/DefaultShareService.kt
```diff
@@ -0,0 +1,76 @@
...
+            .firstOrNull { it.name.endsWith(".ShareActivity") }
+            ?.let { shareActivityInfo ->
+                ComponentName(
+                    shareActivityInfo.packageName,
+                    shareActivityInfo.name,
+                )
+            }
+            ?: return Unit.also {
+                Timber.w("ShareActivity not found")
+            }
```
jmartinesp, line 56: Maybe move this to a private method so this one just gets the component name if it exists and registers / unregisters it?

features/share/api/src/main/kotlin/io/element/android/features/share/api/ShareService.kt
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.features.share.api
+
+import kotlinx.coroutines.CoroutineScope
+
+interface ShareService {
```
jmartinesp, line 21: The `*Service` suffix makes me think it should be implemented by extending an Android Service. Maybe `Helper`? Or leave it as is, it's not a big issue.

jmartinesp (commented):
Thanks for the changes to dynamically enable/disable the feature, I didn't know this was even possible! I added a couple of comments about naming/style, but again feel free to ignore them.

features/share/api/src/main/kotlin/io/element/android/features/share/api/ShareService.kt
bmarty: We already have a lot of `Service` classes, the rule we had on EA for classes extending `android.app.Service` is to name them with `AndroidService` suffix, you can see all the children of [VectorAndroidService](https://github.com/element-hq/element-android/blob/develop/vector/src/main/java/im/vector/app/core/services/VectorAndroidService.kt#L28) for instance.

features/share/impl/src/main/kotlin/io/element/android/features/share/impl/DefaultShareService.kt
bmarty: 2a980a925df7a07963024aec0fad03de7b6592a3

PR #2991, opened 2024-06-06
Sliding sync : update RoomSubscription values as per EXI.

ganfra:
Should fix https://github.com/element-hq/element-x-android-rageshakes/issues/2144


bmarty (approved):
LGTM, thanks!

PR #2993, opened 2024-06-07
Timeline : use the new LiveBackPaginationStatus api

ganfra:
This will handle the case where the timeline will back paginate on his own.


jmartinesp (approved):
LGTM, although I'm not sure if we should have both the `paginateBackwards` result and the listener updating the pagination status at the same time. It's probably fine as they should return the same result.

ganfra:
> LGTM, although I'm not sure if we should have both the `paginateBackwards` result and the listener updating the pagination status at the same time. It's probably fine as they should return the same result.

Yes it shouldn't change anything and as it's a StateFlow if the value are the same it won't emit anything downstream.

PR #3029, opened 2024-06-13
Improve how active calls work

jmartinesp:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

- Sending the `m.call.notify` event is now done in `CallScreenPresenter` once we know the sync is running.
- You can mark a call of both external url or room type as joined.
- Hanging up checks the current active call type and will only remove it if it matches.

## Motivation and context

@toger5 mentioned some ways in which the EXA implementation of ringing calls could be improved.

## Tests

- Just tests calling between clients in several scenarios.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14.

## Checklist


bmarty:
I am currently testing the feature.

I think this commit https://github.com/element-hq/element-x-android/pull/3029/commits/e18c8d064fa48f291bdc957039a9055bfdb79124 can be dropped, I have fixed the failing test in #3036

bmarty (approved):
The changes make sense, and my tests went well, thanks!

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/ActiveCallManager.kt
```diff
@@ -62,23 +60,22 @@ interface ActiveCallManager {
 
     /**
      * Hangs up the active call and removes any associated UI.
+     * @param callType The type of call that the user hung up, either an external url one or a room one.
      */
-    fun hungUpCall()
+    fun hungUpCall(callType: CallType)
```
toger5, line 65: there seems to be a conflict between the docstring and the method name.
`hangUpCall` or
```
 * Called when the active call has been hung up and all associated UI was removed.
```

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/ActiveCallManager.kt
```diff
@@ -62,23 +60,22 @@ interface ActiveCallManager {
...
 
     /**
      * Called when the user joins a call. It will remove any existing UI and set the call state as [CallState.InCall].
```
toger5, line 68:
```suggestion
     * Called after the user joined a call. It will remove any existing UI and set the call state as [CallState.InCall].
```

jmartinesp: https://github.com/element-hq/element-x-android/pull/3029/commits/2ff3ee933b157510cba8799d26bb4915ae3092d5

jmartinesp:
@toger5 let me know if any other changes are needed.

jmartinesp:
> Is this related: [element-hq/element-call#2421](https://github.com/element-hq/element-call/issues/2421)

I don't think so. We don't really have some event we receive as a push that says 'this user hung up, the call is now over' we can use to cancel the ringing, right?

jmartinesp:
Maybe we could try using the room info and observe the call status from there as long as we have an active call now that I think about it. But I'd add that on a separate PR.

toger5:
@jmartinesp yea that is how its intended to be done. When the ring starts, we want to run the sync loop and once we receive an empty call member event (room info updates to call has stopped) we update the state again.

Initially it was planned to do all this in the rust sdk directly but the app team decided this is quicker on the platform side.

jmartinesp:
Ok, then I'll merge this as is and then continue the work on another PR once I'm free from my current tasks.

toger5:
This sounds really good. Can you evaluate with @stefanceriu if there is a chance to move some of that logic into the rust sdk?

Maybe its also worth to confirm with the MSC: https://github.com/matrix-org/matrix-spec-proposals/blob/toger5/matrixrtc-call-ringing/proposals/4075-call-notify-event.md
if the mobile implementation covers all the cases. (That should mention all the cases in which we should start and stop to ring.)

jmartinesp:
PoC of a platform implementation for this: https://github.com/element-hq/element-x-android/pull/3047

Issue #3014, opened 2024-06-11
Ringing does not stop in the room even when caller hangs up


rajil:
Bug filed [here](https://github.com/element-hq/element-call/issues/2421).

bmarty:
I think this is fixed by https://github.com/element-hq/element-x-android/pull/3047

rajil:
No, it is not fixed as tested against 0.4.16-nightly (develop 3d27f26d). The call is not torn down when caller A hangs up. The ring continues on devices B and C. The improvement is that if B and C join the call they see each other (i.e. video conference gets established).

PR #3047, opened 2024-06-18
Cancel ringing call notification on call cancellation

jmartinesp:
## Content

Listen to active call state for DMs with ringing calls so if the other user hangs up before you've replied the ringing stops and the call is cancelled.

## Motivation and context

Improve the call ringing implementation. Fixes https://github.com/element-hq/element-x-android/issues/3014.

## Tests

- With a client using a session A, call a user in another client using session B from a DM.
- The user in session B should see a ringing call notification.
- From session A, hang up.
- The ringing should stop soon after in session B.

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): 14, 13.

## Checklist

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/ActiveCallManager.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/ActiveCallManager.kt
@@ -25,14 +25,25 @@ import io.element.android.features.call.impl.notifications.CallNotificationData
 import io.element.android.features.call.impl.notifications.RingingCallNotificationCreator
 import io.element.android.libraries.di.AppScope
 import io.element.android.libraries.di.SingleIn
+import io.element.android.libraries.matrix.api.MatrixClientProvider
 import io.element.android.libraries.push.api.notifications.ForegroundServiceType
 import io.element.android.libraries.push.api.notifications.NotificationIdProvider
 import io.element.android.libraries.push.api.notifications.OnMissedCallNotificationHandler
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.drop
+import kotlinx.coroutines.flow.filter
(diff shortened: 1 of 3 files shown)


jmartinesp:
So I created this PR as a draft so everyone could test it (especially Timo, who suggested the implementation), it's functional and it could be merged (maybe after some cleanup, I can't remember) but we don't have a counterpart on iOS and I don't think we have a plan to do this work there any time soon.

It depends on whether we want this even if iOS doesn't have it or not. @manuroe ?

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/ActiveCallManager.kt
```diff
@@ -173,6 +189,34 @@ class DefaultActiveCallManager @Inject constructor(
...
+                matrixClientProvider.getOrRestore(callType.sessionId).getOrNull()
+                    ?.getRoom(callType.roomId)
+                    ?.roomInfoFlow
+                    ?.map { it.hasRoomCall }
+                    ?: flowOf()
+            }
+            // We only want to check if the room active call status changes
+            .distinctUntilChanged()
+            // Skip the first one, we're not interested in it (if the check below passes, it had to be active anyway)
+            .drop(1)
```
bmarty, line 211: I am a bit scared about this `drop`. If it is not mandatory, can we just remove it?
Or maybe replace by
```kotlin
            // We only want to check if the room active call status changes
            .distinctUntilChanged()
            .filter { roomHasActiveCall -> !roomHasActiveCall }
            .onEach {
                // The call was cancelled
                incomingCallTimedOut()
            }
            .launchIn(coroutineScope)
```

(Unite test are still passing with this)

bmarty (approved):
Just one remark, else LGTM!

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/ActiveCallManager.kt
jmartinesp: IIRC it's not, because in a real life situation the ringing this flow will start before we get the room info, and by default the `roomHasActiveCall` will be `false` by default, instantly cancelling any incoming call as a result. We could replace it by some operator that takes pairs of events instead so we can compare `hadCall == true && newValue == false`, but that's technically doing exactly the same as the drop as it wouldn't trigger until there are 2 elements in the flow.

bmarty: ok

PR #3051, opened 2024-06-19
Resolve display names in mentions in real time

jmartinesp:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

- Add a `UserProfileCache` that will be passed as a local composition provider so composables inside a room will be able to check user profile info given a user id. The alternative would be to pass some `getUserProfile: (UserId) -> MatrixUser` lambda in the state or the whole member list.
- Do the same for `MentionSpanProvider` so we don't have to initialise them in presenters anymore and do some weird tricks so the tests don't crash 🎉 .
- Using the 2 values above, given a `MentionSpan` in some text message, before we display it we update the `text` contents of this `MentionSpan` with the latest display name of the user, from the profile cache. Note this is only done for user mentions, but a similar approach could be used in the future for room pills.
- When sending mentions, use the user id for the fallback text instead of the current display name of the user, since this can contain outdated or incorrect data.

## Motivation and context

Currently, the plain and rich text editors both send mentions using the display name of the user at that time as the fallback link text, which might contain incorrect data.

On the other side, the editors and timeline rendering were just using this fallback text to display the pill contents.

## Tests

- Using both the plain and rich text editors, send messages with mentions.
- Check the messages display the updated display name for the mentioned users, and the source contains only the user id as the text for the link.

Hopefully this won't affect timeline performance 🤞 .

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


jmartinesp:
> When using the user for the fallback, it impacts the rendering in the room list:

Ouch, I forgot about that case. Sadly I don't think we can use the same strategy here since we'd have to load the room, timeline, members, etc... I can't see any way to load the display name at that point 🫤 .

> Also the display name for mention is not live

It's not really live, no: it will update the display names when room members are reloaded (going to settings or the room members screen, i.e). The problem is we don't have a real time observer for members from the SDK, so this will only happen when we manually send a request to `/members`. So it's ready to be live, but we're missing the Rust SDK changes that would make this work as expected... and there's no ETA for when that will be done.

libraries/matrixui/build.gradle.kts
```diff
@@ -40,6 +40,7 @@ dependencies {
     implementation(projects.libraries.core)
     implementation(projects.libraries.uiStrings)
     implementation(projects.libraries.testtags)
+    implementation(projects.services.toolbox.api)
```
bmarty, line 43: I think this dependency can be removed too (was for SystemClock)

bmarty (approved):
LGTM, thanks!

PR #3064, opened 2024-06-21
Improve click behavior on room timeline title

### LINKED ISSUE #3063: In the room timeline topbar, clicking on the room avatar has no effect
- open a room
- try to click on the room avatar. It should open the room settings
- click on the room name: it open the room settings

Clicking on the room avatar should also open the room settings to increase accessibility.

Also the click effect could be improved, currently this is a rectangle, could have a better shape with round corner.

Toolbar

bmarty:
|Before|After|
|-|-|
|Toolbar|ToolbarFixed

I have to confess that I am not sure why with this change the avatar is now clickable, I thought there where more thing to fix.

Also not sure about the screenshot recording change, since it's just adding a 8 dp invisible padding at the end of the placeholder.

🤷‍♂️ 

Closes #3063


ganfra (commented):
I'm not sure why it's needed as it already works properly on my side?
Also, why have you changed the shape of the ripple? it's better with a rectangle and aligned to what we have everywhere else

bmarty:
We already have rounded corner ripple effect on the codebase, and I think they are prettier.

@amshakal which version of the click effect do you prefer? Left or right?

bmarty:
Let's merge and revisit later (or never)

PR #3100, opened 2024-06-26
Unify the way we decide whether a room is a DM or a group room

jmartinesp:
## Content

- Add `fun isDm(isDirect: Boolean, activeMembersCount: Int): Boolean`.
- Add extension functions that call this `isDm` using existing data for `MatrixRoom` and `MatrixRoomInfo`.
- Apply these changes through the app.

## Motivation and context

Up until now, different parts of the app used different definitions of what a DM is or when should we distinguish between a group room and a direct one. The idea is to always use this function as the source of truth in all parts of the app.

## Tests

Most things should work as before, but unencrypted direct rooms with 2 people should also be considered DMs and have the alternative UI/logic, matching Element Web's behaviour (except when there are more than 2 people in the room).

Notifications now also use the more strict definition of what a DM is in order to display custom avatar images or texts.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/DmChecker.kt
jmartinesp, line 1: This is where the magic happens. I'm bad at naming classes 🥲 .

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/DmChecker.kt
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.libraries.matrix.api.room
+
+/**
+ * Verifies if a room is considered a direct message one for Element.
+ */
+object DmChecker {
```
bmarty, line 22: I would not introduce a class for this, but rather create some extensions, like `RoomInfo.isElementDm()`, `MatrixRoomInfo.isElementDm()`, etc.

If all those extensions have the same name, and all in the same file, I think it can work.

jmartinesp: I'm afraid if we have 3, 4 of these extensions with exactly the same code inside them we might want to update them and forget about some of them. That was why I created this object.

We could make it so the extensions all called the same private function underneath though, so we could ensure the check is done with the same piece of code for every one of them.

bmarty: . o O (Why do we have so many types to describe the same thing?)

Still it will be more clean and less error prone to use an extension at the call site.

jmartinesp: There are a couple of places where it wouldn't be possible to just use extensions:

- `RoomSummaryDetailsFactory.create`: it uses a Rust `RoomInfo` which is not available in the `matrix.api` module.
- `NotificationMapper.map`, we only have the `NotificationRoomInfo`, which also comes from Rust and is not available in `matrix.api`.

Either we keep the `DmChecker` object for those or we create a public `fun isRoomDm(isDirect: Boolean, activeMemeberCount: Int): Boolean` that can be called for other cases.

WDYT?

libraries/matrix/test/src/main/kotlin/io/element/android/libraries/matrix/test/room/FakeMatrixRoom.kt
```diff
@@ -91,6 +90,8 @@ class FakeMatrixRoom(
     canRedactOwn: Boolean = false,
     canRedactOther: Boolean = false,
 ) : MatrixRoom {
+    override val isOneToOne: Boolean = activeMemberCount <= 2L
```
bmarty, line 93: Do we really need to override the default value (which is `activeMemberCount == 2L`) here? It looks like this is against what the spec says.

bmarty (commented):
Thanks, only a few remarks left, let me know what you think.

jmartinesp:
@frebib FWIW this PR removes the 'isEncrypted' check for DMs, so I think it should be quite similar to what your previous PR did (only here we try to centralise the check for the whole app).

frebib:
Nice, thanks. I'll give it a try later

features/invite/impl/src/test/kotlin/io/element/android/features/invite/impl/response/AcceptDeclineInvitePresenterTest.kt
```diff
@@ -265,7 +265,7 @@ class AcceptDeclineInvitePresenterTest {
         return InviteData(
             roomId = roomId,
             roomName = name,
-            isDirect = isDirect
+            isDm = isDirect
```
bmarty, line 268: Maybe rename also the parameter.

PR #3110, opened 2024-06-28
Fix incorrect 'device verified' screen when app was opened with no network connection

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Fix checking for session verification status after session restoration, a different check is needed.
- Fix deadlock when reading the `skipVerification` preference. I'd rather just remove it, but it's needed as a workaround on `AppMigration02`.

## Motivation and context

It turns out `encryptionService.verificationState()` runs a network request that will cause a deadlock when it fails and that will trigger a timeout that will assume your session needs to be verified.

Also fixed another deadlock that caused the screen to remain blank sometimes after logging in, because DataStore got stuck when checking the `skipVerification` state for some reason I don't fully understand. We'll have to keep an eye on similar behaviours through the app.

## Tests

- Open your app with no network connection, it should load just fine.
- Restore network connection, then log out.
- Log in again.
- You should see the session verification screen again.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/verification/RustSessionVerificationService.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/verification/RustSessionVerificationService.kt
@@ -61,7 +61,7 @@ class RustSessionVerificationService(
     private val verificationStateListenerTaskHandle = encryptionService.verificationStateListener(object : VerificationStateListener {
         override fun onUpdate(status: VerificationState) {
             Timber.d("New verification state: $status")
-            updateVerificationStatus(status)
+            sessionCoroutineScope.launch { updateVerificationStatus() }
         }
     })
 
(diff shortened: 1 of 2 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/verification/RustSessionVerificationService.kt
```diff
@@ -200,11 +202,36 @@ class RustSessionVerificationService(
...
-            VerificationState.VERIFIED -> SessionVerifiedStatus.Verified
-            VerificationState.UNVERIFIED -> SessionVerifiedStatus.NotVerified
+    private suspend fun updateVerificationStatus() {
+        if (verificationFlowState.value == VerificationFlowState.Finished) {
+            // Calling `encryptionService.verificationState()` performs a network call and it will deadlock if there is no network
+            // So we need to check that *only* if we know there is network connection, which is the case when the verification flow just finished
+            Timber.d("Updating verification status: flow just finished")
+            runCatching {
+                encryptionService.waitForE2eeInitializationTasks()
+            }.onSuccess {
```
bmarty, line 212: Maybe do something on failure?

bmarty (approved):
LGTM, thanks for fixing those annoying bugs!

PR #3144, opened 2024-07-05
Add icon for "Mark as read" and "Mark as unread" actions.

bmarty:
Icons are coming from Compound 0.0.7 which has just been released.

See recorded screenshots.


ganfra (approved):
Amazing.

PR #3164, opened 2024-07-08
Add a custom sliding sync proxy setting in developer options

jmartinesp:
## Content

Added a new `customSlidingSyncProxy` preference in `AppPreferencesStore` that will be loaded when restoring the session so it can be used to test a different proxy.

## Motivation and context

Implements the Android part of https://github.com/element-hq/element-meta/issues/2468

## Tests

- Go to developer options, Rust SDK section.
- Select `custom sliding sync proxy`.
- Set a custom value.
- Kill and restart the app, the new proxy should be used.
- Undo what you did by deleting the custom proxy text and restarting the app.
- The default proxy should be used.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStore.kt
+++ b/libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStore.kt
@@ -49,7 +49,7 @@ class DefaultSessionPreferencesStore(
     private val renderReadReceiptsKey = booleanPreferencesKey("renderReadReceipts")
     private val sendTypingNotificationsKey = booleanPreferencesKey("sendTypingNotifications")
     private val renderTypingNotificationsKey = booleanPreferencesKey("renderTypingNotifications")
-    private val skipSessionVerification = booleanPreferencesKey("skipSessionVerification")
+    private val skipSessionVerificationKey = booleanPreferencesKey("skipSessionVerification")
 
     private val dataStoreFile = storeFile(context, sessionId)
     private val store = PreferenceDataStoreFactory.create(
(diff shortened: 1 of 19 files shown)


libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStore.kt
```diff
@@ -49,7 +49,7 @@ class DefaultSessionPreferencesStore(
     private val renderReadReceiptsKey = booleanPreferencesKey("renderReadReceipts")
     private val sendTypingNotificationsKey = booleanPreferencesKey("sendTypingNotifications")
     private val renderTypingNotificationsKey = booleanPreferencesKey("renderTypingNotifications")
-    private val skipSessionVerification = booleanPreferencesKey("skipSessionVerification")
+    private val skipSessionVerificationKey = booleanPreferencesKey("skipSessionVerification")
```
bmarty, line 52: 👍

bmarty (approved):
LGTM, thanks a lot!

jmartinesp:
Apparently this PR doesn't match what we want to do since we want to change a part of the sliding sync request URL that can't be overridden with this new option. I'm wondering if we should still merge it in case we'll find some use for it in the future 🤔 .

libraries/preferences/impl/src/main/kotlin/io/element/android/libraries/preferences/impl/store/DefaultSessionPreferencesStore.kt
bmarty: (but ideally this renaming should be done in a separate commit: for next time!)

bmarty:
> we want to change a part of the sliding sync request URL

I am wondering how a user could customize this at runtime.

jmartinesp:
> I am wondering how a user could customize this at runtime.

That's the thing, we can't. That's hardcoded in Ruma, and that's why this PR didn't fit what we needed. It's probably better if I just close it.

PR #3118, opened 2024-07-01
Give locale and theme to Element Call

bmarty:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Use Element Call with correct theme and language.

## Motivation and context

Improve user experience and closes #3079 

## Screenshots / GIFs

image

## Tests

- Change system theme to dark
- Change system language to not English (for instance French)
- Start a call and observe that Element Call is rendered in dark mode and with the selected language (is supported)

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 23
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [ ] Pull request is based on the develop branch
- [ ] Pull request title will be used in the release note, it clearly define what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
- [ ] You've made a self review of your PR

PR #3165, opened 2024-07-08
Always follow the desired theme for Pin, Incoming Call and Element Call screens

### LINKED ISSUE #3153: Element Call: Still not following the app theme.
### Steps to reproduce

1. Update to 0.4.16, then manually switch the color scheme to dark or depending on the system (if the system color scheme is already set to dark);
2. Enter a room;
3. Initiate a call driven by Element Call.

#### What did you expect?
n/a

#### What happened instead?
According to the 0.4.16 changelog, if the Element X theme is dark at this time, then Element Call should also be dark during display. However, it does not appear to do so.

As shown in the picture:
Dark mode is not enabled on the test device at this time, but Element X is manually configured for dark mode.
Screenshot_20240706-125639

--------
**Element Call is my self-hosted instance, version 0.5.17**
#3118 shows the screenshots I still see after updating to 0.4.16.

### Your phone model

OnePlus MT2110

### Operating system version

PixelOS (Android 14)

Element X for Android 0.4.16

### Homeserver

Synapse 1.110.0

### Will you send logs?

No

No

bmarty:
## Content

Fix several issue around theming: Pin and Incoming call was not using the application setting for theme.
ElementCall was reading the current theme but before it was set.

## Motivation and context

Dark theme everywhere when selected manually by the user or by following the system setting.

Closes #3153

## Screenshots / GIFs

image

## Tests

- Set dark theme
- Start a call
- Observe that we now have a dark Element Call rendering.
- Set up PIN code
- Put the app to bg then to fg
- Now: dark screen to enter pin code. Previously: light version
- Same for incoming call screen.

I was confused by the icon change in the screenshot at #3118 , but now the icons are always the same, which is also a bit confusing. But this is on the web part.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/ElementThemeApp.kt
+++ b/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/ElementThemeApp.kt
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2024 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.libraries.designsystem.theme
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
+import io.element.android.compound.theme.ElementTheme
+import io.element.android.compound.theme.Theme
+import io.element.android.compound.theme.isDark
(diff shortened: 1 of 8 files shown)


jmartinesp (approved):
Thanks, LGTM!

bmarty:
@jmartinesp I have added another commit, can you double check please?

Also I have found an issue about assets stored in `-nightly` resource folder, I will create a separate issue.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/ElementThemeApp.kt
```diff
@@ -0,0 +1,46 @@
...
+import io.element.android.compound.theme.mapToTheme
+import io.element.android.libraries.preferences.api.store.AppPreferencesStore
+
+/**
+ * Theme to use for all the regular screens of the application.
+ * Will manage the light / dark theme based on the user preference.
+ */
+@Composable
+fun ElementThemeApp(
+    appPreferencesStore: AppPreferencesStore,
```
jmartinesp, line 35: Having the design system library depending on the preferences one is a bit weird to be honest. Maybe this should be in a separate module that depends on both? Although that also sounds like an overkill.

bmarty: Yes, it's not a big deal to me, and it's better than duplicated code. Also in the future (EXE) it's better to centralize where the application theme is configured.

Issue #3179, opened 2024-07-11
Opening rooms is excruciatingly slow


frebib:
This has been happening for a while now, but I've put up with it. It's getting to the point where the app is borderline unusable with how slow it is. Re-opening the app speeds up room opens to a reasonable rate (not even close to how fast WhatsApp is though, mind you) but after 2-3 times, it slows down again.

This video is real time and very representative of what it's like to use EXA for me

https://github.com/user-attachments/assets/f716ce47-00c5-447d-a1ee-71ff88624fb0

Yes, it really took ~7 seconds for that last room to open.

It seems to do this even when offline loading the "in memory cache" of the last few messages in the timeline, so it's probably not even a slow HS (not that my HS is often slow). I can even reproduce this to a slightly lesser extent with Element-X Nightly and a very fast conduwuit server with an account that only has about 10 rooms.

I haven't tried logging out and back in, but clearing the cache seems to do nothing. I can submit logs or even run custom patches to help debug, if required.

DoM1niC:
same Problem here, have to swipe away to increase the perf

PR #3186, opened 2024-07-12
Performance : improve time to open a room.

ganfra:
## Content

Currently when getting a room through the `client.getRoom(roomId)` method, it also initialises a `MatrixTimeline` and automatically add the timeline listener.
This is not ok for performance as we use the method in different contexts where we don't need to listen to timeline events. 

The timeline will now be subscribed only when needed. It's still hidden by the sdk, so it remains transparent for the client .

Also, rust sdk has removed his timeline cache, getting a `fullRoom` takes some time as we need to initialise the inner timeline. I've now added a small `lruCache` to keep those in memory.

Last point, the app was triggering some back pagination because of some race condition when building UI items.

## Motivation and context
Closes https://github.com/element-hq/element-x-android/issues/3179

## Tests

- Open a room
- Close the room
- Re-open the same room
- Repeat and everything should stay quick and memory low.

## Tested devices

- [X] Physical
- [X] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustRoomFactory.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustRoomFactory.kt
@@ -62,12 +62,12 @@ class RustRoomFactory(
     private val mutex = Mutex()
     private var isDestroyed: Boolean = false
 
-    private data class RustRoomObjects(
+    private data class RustRoomReferences(
         val roomListItem: RoomListItem,
         val fullRoom: Room,
     )
 
-    private val cache = lruCache<RoomId, RustRoomObjects>(
+    private val cache = lruCache<RoomId, RustRoomReferences>(
         maxSize = CACHE_SIZE,
         onEntryRemoved = { evicted, roomId, oldRoom, _ ->
             Timber.d("On room removed from cache: $roomId, evicted: $evicted")
@@ -104,22 +104,22 @@ class RustRoomFactory(
                 Timber.d("Room factory is destroyed, returning null for $roomId")
                 return@withContext null
             }
-            var roomObjects: RustRoomObjects? = getRoomObjects(roomId)
-            if (roomObjects == null) {
+            var roomReferences: RustRoomReferences? = getRoomReferences(roomId)
+            if (roomReferences == null) {
                 // ... otherwise, lets wait for the SS to load all rooms and check again.
                 roomListService.allRooms.awaitLoaded()
-                roomObjects = getRoomObjects(roomId)
+                roomReferences = getRoomReferences(roomId)
             }
-            if (roomObjects == null) {
+            if (roomReferences == null) {
                 Timber.d("No room found for $roomId, returning null")
                 return@withContext null
             }
-            val liveTimeline = roomObjects.fullRoom.timeline()
+            val liveTimeline = roomReferences.fullRoom.timeline()
             RustMatrixRoom(
                 sessionId = sessionId,
                 isKeyBackupEnabled = isKeyBackupEnabled(),
-                roomListItem = roomObjects.roomListItem,
-                innerRoom = roomObjects.fullRoom,
+                roomListItem = roomReferences.roomListItem,
+                innerRoom = roomReferences.fullRoom,
                 innerTimeline = liveTimeline,
                 sessionCoroutineScope = sessionCoroutineScope,
                 notificationSettingsService = notificationSettingsService,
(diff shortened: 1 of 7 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/TimelineItemsSubscriber.kt
```diff
@@ -0,0 +1,108 @@
...
+                        onNewSyncedEvent()
+                    }
+                    postDiffs(diffs)
+                }
+                .launchIn(coroutineScope)
+        }
+        subscriptionCount++
+    }
+
+    suspend fun unsubscribeIfNeeded() = mutex.withLock {
```
bmarty, line 73: This function is not used, can you double check please?

bmarty:
@ganfra do you think that merging this PR can close #3179?

frebib:
> @ganfra do you think that merging this PR can close #3179?

With this PR, performance seems better, but IMHO still not close to what it _should_ be. I don't know if EX or rust-sdk is to blame. I still see ~1s+ room opening times. For contrast, WhatsApp opens rooms instantly

ganfra:
Currently we are in a state where the rust sdk has no cached timeline, so it's creating one each time. It'll be back later, so should be faster.
We also have some slowness coming from using JNA (though uniffi). 
And in the end we might have some overhead from kotlin codebase, like this PR is improving.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/TimelineItemsSubscriber.kt
ganfra: Ok so I kept it, because it's used now!

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustRoomFactory.kt
```diff
@@ -55,8 +58,23 @@ class RustRoomFactory(
     private val getSessionData: suspend () -> SessionData,
 ) {
     @OptIn(ExperimentalCoroutinesApi::class)
-    private val createRoomDispatcher = dispatchers.io.limitedParallelism(1)
+    private val dispatcher = dispatchers.io.limitedParallelism(1)
     private val mutex = Mutex()
+    private var isDestroyed: Boolean = false
+
+    private data class RustRoomObjects(
```
jmartinesp, line 65: `RustRoomReferences`? Naming is hard, up to you.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustRoomFactory.kt
```diff
@@ -70,30 +88,41 @@ class RustRoomFactory(
             )
         }
 
-    suspend fun create(roomId: RoomId): MatrixRoom? = withContext(createRoomDispatcher) {
-        var cachedPairOfRoom: Pair<RoomListItem, Room>?
+    suspend fun destroy() {
```
jmartinesp, line 91: Is this ever called? I couldn't find anything calling it.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustRoomFactory.kt
```diff
@@ -55,8 +58,23 @@ class RustRoomFactory(
...
+        val roomListItem: RoomListItem,
+        val fullRoom: Room,
+    )
+
+    private val cache = lruCache<RoomId, RustRoomObjects>(
+        maxSize = CACHE_SIZE,
+        onEntryRemoved = { evicted, roomId, oldRoom, _ ->
+            Timber.d("On room removed from cache: $roomId, evicted: $evicted")
+            oldRoom.roomListItem.close()
+            oldRoom.fullRoom.close()
```
jmartinesp, line 75: Clever!

jmartinesp (commented):
I'm a bit wary about the never called `RustRoomFactory.destroy()`

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustRoomFactory.kt
ganfra: Oooops, it was, but I reverted some part and forgot about this one, thanks!

bmarty (approved):
LGTM and tested OK, thanks!

Issue #2966, opened 2024-06-03
@room pills don't render correctly when sent using Markdown input


frebib:
### Steps to reproduce

1. Use the non-RTE input field and send an @room mention
2. Note that the event doesn't send a formatted body, nor does the message bubble render the @room as a pill

#### What did you expect?
`@room` to render as a pill

#### What happened instead?
It rendered as plain text

### Will you send logs?

No

### Are you willing to provide a PR?

No

bmarty:
Without formating, this event is sent:

```
  "content": {
    "body": "@room Test",
    "m.mentions": {
      "room": true
    },
    "msgtype": "m.text"
  },
```

and with formatting:

```
  "content": {
    "body": "@room test 2",
    "format": "org.matrix.custom.html",
    "formatted_body": "@room test 2",
    "m.mentions": {},
    "msgtype": "m.text"
  },
```
 and this is how this render on Web:

<img width="132" alt="image" src="https://github.com/element-hq/element-x-android/assets/3940906/8f628ea4-028d-4210-a354-bcd649b4e705">

and on EXA this is the opposite:

<img width="160" alt="image" src="https://github.com/element-hq/element-x-android/assets/3940906/d0213d71-a6c4-43a9-bac1-ee4ec4ecd949">

Also this [line](https://github.com/element-hq/element-x-android/blob/develop/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemTextView.kt#L91) is crashing in debug build, since `mentionSpan.rawValue` is equal to `@room` and this is not a valid userId.

We have several problems to fix, not only a rendering problem.

frebib:
Intentional mentions are working now, but the markdown input is still not sending a formatted body, so the room pill still doesn't render as a pill. EW does render a pill even without a formatted_body though. Not sure what the intention is here
```
  "content": {
    "msgtype": "m.text",
    "body": "@room ",
    "m.mentions": {
      "room": true
    }
  },
```

jmartinesp:
> Intentional mentions are working now, but the markdown input is still not sending a formatted body, so the room pill still doesn't render as a pill.

Oh, true. I did all my tests with messages with both an `@room` mention and a explicit one so they were always formatted and I never found this issue. Let me check if I can fix this remaining case.

jmartinesp:
This should now be fixed with #3201.

PR #3193, opened 2024-07-15
Restore intentional mentions in the markdown/plain text editor

jmartinesp:
## Content

- Add `TextPillificationHelper` to add pills to existing MD text by using regexes searching for raw ids and links pointing to permalinks with ids.
- Use the helper when loading a draft, setting the text from the edit mode, or loading the MD version of the rich text content when toggling format.
- Split `MentionSpanProvider` into `MentionSpanProvider` and `MentionSpanTheme`, which contains the info about colors, typography and spacing to use in `MentionSpan`.
- Modify `MentionSpan` so those values can be update after the span creation.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/2966

## Screenshots / GIFs

No UI changes, I just renamed a preview.

## Tests

- In a room, without formatting enabled, write `@room and @someotheruser`.
- Enable formatting, check the pills still work.
- Disable formatting, the pills should still be there.
- Go back to the room list, then back to the room. The draft should contain the pills too.
- Send the message, check the pills are rendered fine.
- Edit the message, check the pills are added correctly to the editor and once edited the message renders fine.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
```diff
@@ -397,7 +402,22 @@ class MessageComposerPresenter @Inject constructor(
             canCreatePoll = canCreatePoll.value,
             attachmentsState = attachmentsState.value,
             memberSuggestions = memberSuggestions.toPersistentList(),
-            eventSink = { handleEvents(it) }
+            resolveMentionDisplay = remember(mentionSpanTheme) {
```
ganfra, line 405: Can you just put that in a val before the actual creation of the MessageComposerState?

jmartinesp: As a remembered lambda, you mean?

features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl/utils/TextPillificationHelperTest.kt
```diff
@@ -0,0 +1,124 @@
...
+                PermalinkData.UserLink(UserId("@user:server.com"))
+            }),
+            permalinkBuilder = FakePermalinkBuilder(permalinkForUserLambda = {
+                Result.success("https://matrix.to/#/@user:server.com")
+            }),
+        )
+        val pillified = helper.pillify(text)
+        val mentionSpans = pillified.getMentionSpans()
+        assertThat(mentionSpans).hasSize(1)
+        assertThat(mentionSpans.firstOrNull()?.type).isEqualTo(MentionSpan.Type.USER)
```
ganfra, line 53: Maybe change to this? mentionsSpans.firstOrNull()?.let { assertThat(...)....}

ganfra (approved):
2 small remarks, otherwise LGTM

features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl/utils/TextPillificationHelperTest.kt
bmarty: this is not equivalent test.

jmartinesp: The assertion error wouldn't be the same, but we have the `hasSize()` check before. I could also either extract the value to a variable and keep it nullable or use `first()` instead and have the tests crash in that case.

bmarty: I'd vote for using `first()`, it will not crash since line 52 is checking for the collection size.

ganfra: Yeah sorry that was what I had in mind, not `firstOrNull`

PR #3201, opened 2024-07-16
Fix pillification not working for non formatted message bodies

jmartinesp:
## Content

Pure Markdown bodies weren't being 'pillified' so their mentions were turned into UI elements in the timeline.

A new `pillifiedBody` property was added to `TimelineItemTextBasedContent` to fix this.

## Motivation and context

#2966

## Screenshots / GIFs

There shouldn't be any changes 🤞 .

## Tests

- Send a message with a single `@room` mention.
- Check the message is properly rendered in the timeline with the `@room` pill.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemTextView.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemTextView.kt
@@ -77,20 +77,11 @@ internal fun getTextWithResolvedMentions(content: TimelineItemTextBasedContent):
     val userProfileCache = LocalRoomMemberProfilesCache.current
     val lastCacheUpdate by userProfileCache.lastCacheUpdate.collectAsState()
     val mentionSpanTheme = LocalMentionSpanTheme.current
-    val textWithMentions = if (content.formattedBody != null) {
-        remember(content.formattedBody, mentionSpanTheme, lastCacheUpdate) {
-            content.formattedBody?.let { formattedBody ->
-                updateMentionSpans(formattedBody, userProfileCache)
-                mentionSpanTheme.updateMentionStyles(formattedBody)
-                formattedBody
-            }
-        }
-    } else {
-        remember(content.pillifiedBody, mentionSpanTheme, lastCacheUpdate) {
-            updateMentionSpans(content.pillifiedBody, userProfileCache)
-            mentionSpanTheme.updateMentionStyles(content.pillifiedBody)
-            content.pillifiedBody
-        }
+    val formattedBody = content.formattedBody ?: content.pillifiedBody
+    val textWithMentions = remember(formattedBody, mentionSpanTheme, lastCacheUpdate) {
+        updateMentionSpans(formattedBody, userProfileCache)
+        mentionSpanTheme.updateMentionStyles(formattedBody)
+        formattedBody
     }
     return SpannableString(textWithMentions)
 }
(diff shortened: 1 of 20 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemTextView.kt
```diff
@@ -77,14 +77,22 @@ internal fun getTextWithResolvedMentions(content: TimelineItemTextBasedContent):
...
+                updateMentionSpans(formattedBody, userProfileCache)
+                mentionSpanTheme.updateMentionStyles(formattedBody)
+                formattedBody
+            }
+        }
+    } else {
+        remember(content.pillifiedBody, mentionSpanTheme, lastCacheUpdate) {
+            updateMentionSpans(content.pillifiedBody, userProfileCache)
+            mentionSpanTheme.updateMentionStyles(content.pillifiedBody)
+            content.pillifiedBody
```
bmarty, line 92: Is all this block equivalent to this simpler form?

```kotlin
    val formattedBody = content.formattedBody ?: content.pillifiedBody
    val textWithMentions = remember(formattedBody, mentionSpanTheme, lastCacheUpdate) {
        updateMentionSpans(formattedBody, userProfileCache)
        mentionSpanTheme.updateMentionStyles(formattedBody)
        formattedBody
    }
```

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/event/TimelineItemTextBasedContent.kt
```diff
@@ -19,13 +19,30 @@ package io.element.android.features.messages.impl.timeline.model.event
...
+    /** The formatted body of the event, already parsed and with the DOM translated to Android spans. */
     val formattedBody: CharSequence?
+
+    /** The plain text version of the event body. This is the Markdown version without actual Markdown formatting. */
     val plainText: String
+
+    /** Whether the event has been edited. */
     val isEdited: Boolean
+
+    /** The raw HTML body of the event. */
```
bmarty, line 45: Kudos for adding some documentation 👍

features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl/utils/FakeTextPillificationHelper.kt
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.features.messages.impl.utils
+
+class FakeTextPillificationHelper(
+    private val pillifyLambda: (CharSequence) -> CharSequence = { it }
```
bmarty, line 20: I prefer to have a default lambda content with `lambdaError()` so that we ensure that it's not called when it's not expected, which is actually doing more test. Note that in this particular case, this is maybe OK to keep the current default implementation. I let you choose what's best.

bmarty (approved):
Thanks, the pillification is working as expected. But there are still difference when sending `@room` from Element Web and from EXA (tested in a DM):

image

Also on EXA I do not have completion for `@room`, I have it on Web. This may be fixed later, not related to this PR.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemTextView.kt
jmartinesp: Yeah, nice catch!

features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl/utils/FakeTextPillificationHelper.kt
jmartinesp: This component is sometimes used as a dependency deep in the hierarchy so I think it's better to not have to customise it for testing other components that may depend on it for functionality but don't check its output.

PR #3208, opened 2024-07-17
Add logging to help debug forced logout issues

jmartinesp:
## Content

Add extra logging to try to debug spontaneous logouts in OIDC sessions:
- An extra log level for OIDC token refreshes in the SDK.
- More logs to make sure we're always using the same `RustMatrixClient` instance when saving the new tokens to persistent storage.
- Use the same hashing for tokens in the SDK and the app so we can compare them 1:1.

## Motivation and context

We want to get rid of those logouts.

## Tests

Not much to test other than checking your logs for the new content.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/util/Token.kt
```diff
@@ -0,0 +1,26 @@
...
+package io.element.android.libraries.matrix.impl.util
+
+import java.security.MessageDigest
+
+private val sha256 by lazy { MessageDigest.getInstance("SHA-256") }
+
+@OptIn(ExperimentalStdlibApi::class)
+internal fun anonymizeToken(token: String): String {
+    return sha256.digest(token.toByteArray()).toHexString()
+}
```
bmarty, line 26: Writting an externsion like
```kotlin
@OptIn(ExperimentalStdlibApi::class)
internal fun String.toAnonymizedToken(): String {
    return sha256.digest(toByteArray()).toHexString()
}
```

would simplify the all the call sites.

Also maybe we could just log the last digits, it will make the log less big I guess.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
```diff
@@ -158,16 +159,22 @@ class RustMatrixClient(
 
     private val isLoggingOut = AtomicBoolean(false)
 
+    @OptIn(ExperimentalStdlibApi::class)
     private val clientDelegate = object : ClientDelegate {
+        private val clientTag = "RustMatrixClient#${this@RustMatrixClient.hashCode().toHexString()}"
         override fun didReceiveAuthError(isSoftLogout: Boolean) {
-            Timber.w("didReceiveAuthError(isSoftLogout=$isSoftLogout)")
+            Timber.w("[$clientTag]: didReceiveAuthError(isSoftLogout=$isSoftLogout)")
```
bmarty, line 166: Why not using Timber tag facility?

```kotlin
Timber.tag(clientTag).w("didReceiveAuthError(isSoftLogout=$isSoftLogout)")
```

bmarty (commented):
Thanks for improving logs here.

jmartinesp:
I think both issues should be handled by the latest commits. I also added a new commit to ensure we only run a single token persistence coroutine at a time.

bmarty (approved):

PR #3219, opened 2024-07-19
Use the right filename for log files so they're sorted in rageshakes

jmartinesp:
## Content

Use the `log` suffix when creating log files so the rageshake server recognises its format and doesn't rename them to just `logs-0001.log.gz`, `logs-0002.log.gz`, etc.

## Motivation and context

This way we'll have instead `logs-2024-07-19-11.log.gz`, `logs-2024-07-19-10.log.gz`, etc., stating the hour of logs included inside in their filename.

## Tests

You can upload a test bug report after letting the app run for a while and collect new logs, then check the rageshakes.

https://github.com/element-hq/element-x-android-rageshakes/issues/2410

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/element/android/x/initializer/TracingInitializer.kt
+++ b/app/src/main/kotlin/io/element/android/x/initializer/TracingInitializer.kt
@@ -71,8 +71,6 @@ class TracingInitializer : Initializer<Unit> {
         return WriteToFilesConfiguration.Enabled(
             directory = bugReporter.logDirectory().absolutePath,
             filenamePrefix = "logs",
-            // DO NOT CHANGE: suffix *MUST* be "log" for the rageshake server to not rename the file to something generic
-            filenameSuffix = "log",
             // Keep a maximum of 1 week of log files.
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/io/element/android/x/initializer/TracingInitializer.kt
```diff
@@ -71,7 +71,7 @@ class TracingInitializer : Initializer<Unit> {
         return WriteToFilesConfiguration.Enabled(
             directory = bugReporter.logDirectory().absolutePath,
             filenamePrefix = "logs",
-            filenameSuffix = null,
+            filenameSuffix = "log",
```
ganfra, line 74: Add a small comment so we don't change this?

ganfra (approved):

app/src/main/kotlin/io/element/android/x/initializer/TracingInitializer.kt
jmartinesp: Yeah, good call.

PR #3229, opened 2024-07-19
Rework FakeMatrixRoom so that it contains only lambdas.

bmarty:
This is actually adding more tests, to ensure that only expected methods are used by unit tests.


libraries/matrix/test/src/main/kotlin/io/element/android/libraries/matrix/test/room/FakeMatrixRoom.kt
```diff
@@ -84,126 +84,88 @@ class FakeMatrixRoom(
     override val activeMemberCount: Long = 234L,
     val notificationSettingsService: NotificationSettingsService = FakeNotificationSettingsService(),
     override val liveTimeline: Timeline = FakeTimeline(),
-    private var roomPermalinkResult: () -> Result<String> = { Result.success("room link") },
-    private var eventPermalinkResult: (EventId) -> Result<String> = { Result.success("event link") },
-    var sendCallNotificationIfNeededResult: () -> Result<Unit> = { Result.success(Unit) },
-    canRedactOwn: Boolean = false,
-    canRedactOther: Boolean = false,
+    private var roomPermalinkResult: () -> Result<String> = { lambdaError() },
```
jmartinesp, line 87: I tend to make these public so we don't have to add extra `given___` methods to modify them, but it's no big deal.

jmartinesp (approved):

PR #3233, opened 2024-07-22
Fix linkification not working for `Spanned` strings in text messages

jmartinesp:
## Content

Make `CharSequence.withFixedURLSpans()` handle any type of CharSequence, not only `Spannable`. If the text wasn't `Spannable`,  a new one will be created from it.

## Motivation and context

Previously, the RTE would always return a `SpannableString` as a result, and as the code checked for `this is Spannable` it worked fine.

Now a `SpannedString` is returned instead (the result of a `buildSpannedString { ... }` block), so the code was skipped.

This issue was found as a regression after upgrading the RTE version to `2.37.7` in https://github.com/element-hq/element-x-android/pull/3218.

## Tests

- Open a room where a plain link was sent inside a message with `formatted_body`.
- The link should be auto linkified again.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


ganfra (approved):
Should we add some more tests ?

jmartinesp:
> Should we add some more tests ?

Yep, on it.

PR #3240, opened 2024-07-24
Timeline UI | MessageShield Support

BillCarsonFr:
Fixes https://github.com/element-hq/crypto-internal/issues/345
Depends on https://github.com/matrix-org/matrix-rust-sdk/pull/3679
Design: https://www.figma.com/design/2aqj9A5WRvGY3tcdTrBek4/Unencrypted-message-warning-in-Timeline?node-id=910-77038&t=pF5GGduo3EsrFG2Q-0
 
## Content

Show E2E message shields. 

image
image

image

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


frebib:
I tried this out for nothing more than sheer curiosity, and there were a couple things I noticed that were a bit weird (aside from the fact that literally every encrypted message was marked as unauthentic)
- The bubble on messages received from other users (on the left hand side) are the same colour as the background (at least in dark mode) so it looks pretty odd and like there is a lot of whitespace
- The bubble sits above the timestamp, which balloons the timeline a LOT vertically, because there's an empty line at the bottom of every message where there wasn't one before

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
```diff
@@ -590,3 +643,83 @@ internal fun TimelineItemEventRowPreview() = ElementPreview {
...
+        )
+    }
+}
+
+@Composable
+@ReadOnlyComposable
+private fun TimelineItem.Event.shieldPosition(): MessageShieldPosition {
+    val shield = this.messageShield ?: return MessageShieldPosition.None
+
+    // sdk returns raw human readable strings, add i18n support
```
bmarty, line 705: @Jorge Could the SDK return an enum instead?

jmartinesp: I think you should discuss this with @pixlwave (and not poor `@jorge` 😅 ).

pixlwave: Ah yeah I fixed this… You can now switch on the `ShieldState::code` which is an enum 😁

On EXI we map `ShieldState` to this [`EncryptionAuthenticity`](https://github.com/element-hq/element-x-ios/blob/develop/ElementX/Sources/Services/Timeline/TimelineItemContent/EncryptionAuthenticity.swift) representation for our sanity.

pixlwave: @bmarty For completeness here is the view we use for the timestamp with send/encryption warnings: [`TimelineItemSendInfoLabel`](https://github.com/element-hq/element-x-ios/blob/develop/ElementX/Sources/Screens/Timeline/View/Style/TimelineItemSendInfoLabel.swift)

And the PRs https://github.com/element-hq/element-x-ios/pull/3051, https://github.com/element-hq/element-x-ios/pull/3100 and https://github.com/element-hq/element-x-ios/pull/3116

pixlwave: > Same for the color, it should be a enum (if the first enum is not enough) with a notion of criticity. "color" is a concept for the application (UI) level.

FWIW, both myself and @stefanceriu raised this [here](https://github.com/element-hq/element-x-ios/pull/3051#discussion_r1680882169) and [here](https://github.com/element-hq/element-x-ios/pull/3116#discussion_r1705128756) but the crypto team don't want to change it 🤷‍♂️

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/MessageShieldView.kt
```diff
@@ -0,0 +1,131 @@
...
+}
+
+@Composable
+internal fun MessageShield.toIcon(): ImageVector {
+    return when (this) {
+        is MessageShield.AuthenticityNotGuaranteed,
+        is MessageShield.UnverifiedIdentity -> CompoundIcons.Admin()
+        is MessageShield.UnknownDevice,
+        is MessageShield.UnsignedDevice -> CompoundIcons.HelpSolid()
+        is MessageShield.SentInClear -> CompoundIcons.KeyOff()
```
bmarty, line 102: TODO: need to check this mapping.

bmarty:
Updated, maybe need to add a feature flag as per iOS app (?)

pixlwave:
> Maybe need to add a feature flag as per iOS app

Tbh the only reason I've kept the iOS flag around is so we could enable this on both apps at the same time. So entirely up do you, if you don't want to add one I'll remove it from iOS when this PR is merged.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/MessageShieldView.kt
pixlwave: This should be the final mapping: https://github.com/element-hq/element-x-ios/blob/cf35056c787fbae4a219e2a442738b6a84164a95/ElementX/Sources/Services/Timeline/TimelineItemContent/EncryptionAuthenticity.swift#L59-L65 (I realise now that I linked you to a bunch of PRs that iteratively changed it over time).

bmarty: Thanks. I have updated the code.
But there is no shield at all, which is against the SDK recommendation [here](https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk-common/src/deserialized_responses.rs#L196) and [here](https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk-common/src/deserialized_responses.rs#L204).

Now I have this (from the Preview):

image

@BillCarsonFr can you confirm please?

bmarty (approved):
Implementation should be equivalent to iOS, let's merge and iterate if necessary.

PR #3257, opened 2024-07-30
Feature/fga/push subscribe to room

ganfra:
## Content
Add a feature flag so we can try subscribing to room whenever we receive a notifiable push.

## Motivation and context
To improve time to see timeline when opening the app from push notification.

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 23
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [ ] Pull request is based on the develop branch
- [ ] Pull request title will be used in the release note, it clearly define what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
- [ ] You've made a self review of your PR

PR #3260, opened 2024-07-31
Feature/fga/start sync on push

ganfra:
## Content
https://github.com/element-hq/element-x-android/pull/3257 wasn't enough, so try to sync and waits for the event in the timeline if the app is in background.

## Motivation and context
Improve on top of https://github.com/element-hq/element-x-android/pull/3257

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/SyncOnNotifiableEvent.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/SyncOnNotifiableEvent.kt
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) 2024 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.libraries.push.impl.push
+
+import io.element.android.libraries.core.coroutine.CoroutineDispatchers
+import io.element.android.libraries.featureflag.api.FeatureFlagService
+import io.element.android.libraries.featureflag.api.FeatureFlags
+import io.element.android.libraries.matrix.api.MatrixClientProvider
+import io.element.android.libraries.matrix.api.core.EventId
+import io.element.android.libraries.matrix.api.room.MatrixRoom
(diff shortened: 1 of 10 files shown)


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/SyncOnNotifiableEvent.kt
```diff
@@ -0,0 +1,87 @@
...
+        }
+        val client = matrixClientProvider.getOrRestore(notifiableEvent.sessionId).getOrNull() ?: return@withContext
+        client.getRoom(notifiableEvent.roomId)?.use { room ->
+            room.subscribeToSync()
+
+            // If the app is in foreground, sync is already running, so just add the subscription.
+            if (!appForegroundStateService.isInForeground.value) {
+                val syncService = client.syncService()
+                syncService.startSyncIfNeeded()
+                room.waitsUntilEventIsKnown(eventId = notifiableEvent.eventId, timeout = 10.seconds)
```
jmartinesp, line 57: When I was doing similar tests what I did here was just wait until the `syncState` was `Running`, which I believe is enough to mean the first sync has been received. But we can keep this to make it more explicit, it will probably even work better.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/SyncOnNotifiableEvent.kt
```diff
@@ -0,0 +1,87 @@
...
+                        is MatrixTimelineItem.Event -> timelineItem.eventId == eventId
+                        else -> false
+                    }
+                }
+            }
+        }
+    }
+
+    private suspend fun SyncService.startSyncIfNeeded() {
+        if (syncCounter.getAndIncrement() == 0) {
```
jmartinesp, line 77: Wouldn't it be better to just check the `syncService` state? What if some other part of the app started a sync on its own?

jmartinesp (approved):
LGTM, although I have my doubts about `startSyncIfNeeded()` and its `stop` counterpart.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/SyncOnNotifiableEvent.kt
ganfra: No waiting for running is not enough, the data usually comes after several sync iterations

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/SyncOnNotifiableEvent.kt
ganfra: The counter is just for avoid cancelling if there are other push. 
If the sync is already started from somewhere else, this won't have any effect. 
If the app is backgrounded again it will cancel the sync anyway.
So, in the end, this is just best effort.

PR #3291, opened 2024-08-09
Fix `build_rust_sdk.sh` script to work on linux

erikjohnston:
A few linux distributions don't have `gdate`, but I think they should all have `date`.


jmartinesp (approved):
Thanks for the changes!

PR #3295, opened 2024-08-11
Big emoji-only messages

### LINKED ISSUE #1438: Implement large emoji
### Your use case

image

Left: Element X iOS. Right: Element X Android.

We should match the implementation here (& Element Web) where we enlarge emoji when sent on their own.

_No response_

### Additional context

_No response_

No

frebib:
## Content

Render emoji-only messages in a larger text

Adapted from SpiritCroc's SchildNext implementation from https://github.com/SchildiChat/schildichat-android-next/commit/7eba87f02e884b9b921aa4e74dd0a737e4da31ad

Authored-by: SpiritCroc <dev@spiritcroc.de>
Submitted-by: Joe Groocock <me@frebib.net>

## Motivation and context

Parity with Element-X iOS, Element Web, and pretty much any other chat client ever.

Fixes #1438

## Screenshots / GIFs

|Before|After|
|-|-|
|Screenshot_20240811-175500|Screenshot_20240811-175137|

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 15

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/event/TimelineItemEventContentProvider.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/event/TimelineItemEventContentProvider.kt
@@ -42,6 +42,7 @@ class TimelineItemEventContentProvider : PreviewParameterProvider<TimelineItemEv
         aTimelineItemTextContent(),
         aTimelineItemUnknownContent(),
         aTimelineItemTextContent().copy(isEdited = true),
+        aTimelineItemTextContent(body = "😁")
     )
 }
 
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/Emoji.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/Emoji.kt
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2024 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.features.messages.impl.utils
+
+import com.sigpwned.emoji4j.core.Grapheme.Type.EMOJI
+import com.sigpwned.emoji4j.core.Grapheme.Type.PICTOGRAPHIC
+import com.sigpwned.emoji4j.core.GraphemeMatchResult
+import com.sigpwned.emoji4j.core.GraphemeMatcher
+
+/**
+ * Returns true if the string consists exclusively of "emoji or pictographic graphemes".
+ */
+fun String.containsOnlyEmojis(): Boolean {
+    if (isEmpty()) return false
+
+    val matcher = GraphemeMatcher(this)
+    var m: GraphemeMatchResult? = null
+    var contiguous = true
+    var previous = 0
+    while (contiguous && matcher.find()) {
+        m = matcher.toMatchResult()
+        // Many non-"emoji" characters are pictographics. We only want to identify this specific range
+        // https://en.wikipedia.org/wiki/Miscellaneous_Symbols_and_Pictographs
+        val isEmoji = m!!.grapheme().type == EMOJI || (m.grapheme().type == PICTOGRAPHIC && m.group() in "🌍".."🗺")
+        contiguous = isEmoji and (m.start() == previous)
+        previous = m.end()
+    }
+
+    return contiguous and (m?.end() == length)
+}
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/Emoji.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/Emoji.kt
@@ -35,7 +35,7 @@ fun String.containsOnlyEmojis(): Boolean {
         m = matcher.toMatchResult()
         // Many non-"emoji" characters are pictographics. We only want to identify this specific range
         // https://en.wikipedia.org/wiki/Miscellaneous_Symbols_and_Pictographs
-        val isEmoji = m!!.grapheme().type == EMOJI || (m.grapheme().type == PICTOGRAPHIC && m.group() in "🌍".."🗺")
+        val isEmoji = m!!.grapheme().type == EMOJI || m.grapheme().type == PICTOGRAPHIC && m.group() in "🌍".."🗺"
         contiguous = isEmoji and (m.start() == previous)
         previous = m.end()
     }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -193,6 +193,7 @@ matrix_analytics_events = "com.github.matrix-org:matrix-analytics-events:0.23.1"
 
 # Emojibase
 matrix_emojibase_bindings = "io.element.android:emojibase-bindings:1.1.3"
+sigpwned_emoji4j = "com.sigpwned:emoji4j-core:15.1.0"
 
 # Di
 inject = "javax.inject:javax.inject:1"
--- a/features/messages/impl/build.gradle.kts
+++ b/features/messages/impl/build.gradle.kts
@@ -72,6 +72,7 @@ dependencies {
     implementation(libs.androidx.constraintlayout.compose)
     implementation(libs.androidx.media3.exoplayer)
     implementation(libs.androidx.media3.ui)
+    implementation(libs.sigpwned.emoji4j)
     implementation(libs.vanniktech.blurhash)
     implementation(libs.telephoto.zoomableimage)
     implementation(libs.matrix.emojibase.bindings)
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemTextView.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemTextView.kt
@@ -35,6 +35,7 @@ import io.element.android.features.messages.impl.timeline.components.layout.Cont
 import io.element.android.features.messages.impl.timeline.components.layout.ContentAvoidingLayoutData
 import io.element.android.features.messages.impl.timeline.model.event.TimelineItemTextBasedContent
 import io.element.android.features.messages.impl.timeline.model.event.TimelineItemTextBasedContentProvider
+import io.element.android.features.messages.impl.utils.containsOnlyEmojis
 import io.element.android.libraries.designsystem.preview.ElementPreview
 import io.element.android.libraries.designsystem.preview.PreviewsDayNight
 import io.element.android.libraries.matrix.api.core.UserId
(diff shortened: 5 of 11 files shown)


frebib:
cc @SpiritCroc

gradle/libs.versions.toml
```diff
@@ -22,6 +22,7 @@ lifecycle = "2.7.0"
 activity = "1.9.1"
 media3 = "1.4.0"
 camera = "1.3.4"
+emoji2 = "1.5.0-beta01"
```
frebib, line 25: Note for the reviewer- I'm not sure if non-release versions are acceptable, but given that this library hasn't had a release in over a year, and this newer build adds support for the newer Unicode 15.1 emojis I figured it might be okay

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/event/TimelineItemEventContentProvider.kt
```diff
@@ -61,6 +61,8 @@ class TimelineItemTextBasedContentProvider : PreviewParameterProvider<TimelineIt
         aTimelineItemNoticeContent().copy(formattedBody = buildSpanned("Notice")),
         aTimelineItemTextContent(),
         aTimelineItemTextContent().copy(formattedBody = buildSpanned("Text")),
+        // Emojis should render larger than normal text
```
frebib, line 64: For some reason this doesn't seem to work in the preview. I'd guess it's because the preview doesn't have the emoji library initialised (https://github.com/element-hq/element-x-android/pull/3295/files#diff-78bca535fd89d019ca42a19f0ee787cd9e0ad3d8072433a4a7f6c18c8b6b19b5R42)? Not sure how to make that work

SpiritCroc:
Thanks!

## Sign-off

Signed-off-by: Tobias Büttner <dev@spiritcroc.de>

frebib:
```
> Task :app:compileGplayDebugKotlin FAILED
e: warnings found and -Werror specified
w: file:///home/runner/work/element-x-android/element-x-android/app/src/main/kotlin/io/element/android/x/ElementXApplication.kt:42:26 'constructor BundledEmojiCompatConfig(Context)' is deprecated. Deprecated in Java
```

(big) 😑

Yet, not a single other match for this on github passes the second argument: https://github.com/search?q=BundledEmojiCompatConfig&type=code

This library is becoming more effort than it's worth. We don't even need the font compat rubbish 💩

frebib:
I contemplated the options for a while, and opted to stick with androidx-emoji2 for a couple of reasons
- Implementing emoji detection is hard.. EX iOS does it, but it's not perfect (and they even opted for a simpler implementation for the sake of maintainability https://github.com/element-hq/element-x-ios/commit/2019984918a50b45c03ac804aee58ce48d956f7b)
- I tried to integrate https://github.com/vanniktech/Emoji but that has the same problems/questionable implementation, plus it would require either androidx-emoji2 or yet another vanniktech library for emoji data. I also note that Benoit calls out the vanniktech library as being "3rd party" here https://github.com/matrix-org/emojibase-bindings/issues/16 which seems to me like a negative.
- Using a regex, which seems like the most straightforward (and honestly, probably performant too) option is actually very complex due to the questionable nature of the design of the unicode emoji system. It'd be a burden to maintain in the long run, even if it would work now.
- The "is this string an emoji-only string" implementation can be swapped out in future if a better option comes along, like Java 21's `String.isEmoji()`

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/event/TimelineItemEventContentProvider.kt
jmartinesp: I don't think there's an easy way to preview these, we could add some check along with `LocalInspectionMode.current` maybe, but that seems like quite the hack.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/Emoji.kt
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.features.messages.impl.utils
+
+import androidx.emoji2.text.EmojiCompat
+import timber.log.Timber
+
+fun String.containsOnlyEmojis(maxEmojis: Int = Integer.MAX_VALUE): Boolean {
```
jmartinesp, line 22: Can we add some docs?

jmartinesp (commented):
Thanks for the PR!

I'm a bit torn about this: we already have the [emojibase](https://github.com/matrix-org/emojibase-bindings) library for using emojis, getting the categories, etc., so I don't really like the idea of adding another one with mostly the same functionality - even though it's the official one. It also probably adds some small overhead on the app init time.

On the other side, the implementation is quite simple and works well, and duplicating the logic of `EmojiCompat` seems like an overkill.

Also, the logic isn't exactly the same as iOS: they only use a large text style if the message contains a single emoji and nothing else, so maybe we should do the same.

@bmarty any thoughts about this?

frebib:
> we already have the emojibase library for using emojis, getting the categories, etc.

Fully agree here. In fact, I'm quite frustrated that we need a library for this at all. It should just be pattern matching really. If the spec wasn't so awful, I'd do it with a regex.
I'm happy to drop this lib and find another way to do it- I'm not totally satisfied with it either tbqh

> they only use a large text style if the message contains a single emoji and nothing else, so maybe we should do the same.

But then how will I get my large chefs kiss?
😗🤌

frebib:
Based on that excerpt from the docs, it makes using this library a non-starter. It's dead weight. I'll go back to the drawing board yet again. Maybe I'll try to understand the unicode spec a bit more and write some lightweight logic to detect emojis.

bmarty:
We probably need a product decision about rendering messages with only Emojis with a bigger font. This very old feature has been introduced before the Matrix protocol supports reaction. So now that we have reaction, maybe we do not need this feature.

Also, ideally, the "only emoji" detection could be done by the Rust SDK, so that the algorithm is shared with iOS.

(note: tracked by #1438)

CC @mxandreas

frebib:
I took some more time trying to implement this with a regex, but after reading into the unicode spec, I realised that it would be a pretty monumental feat to implement something that correctly catches all various of the cases and variants that can apply to specific emojis (of which there are a LOT). I guess it sorta explains why many of these libraries ship huge tables- it's so they can account for all of the variants. For example here is just _one_ set of variants for one of the many ranges of emoji graphemes: https://en.wikipedia.org/wiki/Miscellaneous_Symbols_and_Pictographs#Table_of_emojis_with_modifiers

I settled on implementing with https://github.com/sigpwned/emoji4j as it seems relatively well maintained and simple/lightweight. The implementation has emoji-specific grapheme selection which is what we want. There are a few edge cases that it doesn't account for (but I think they're all from unicode 15.1.. it only supports up to 15).

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/Emoji.kt
frebib: Had to remind myself how to write ~JavaDoc~ KDoc(?), but done

mxandreas:
> We probably need a product decision about rendering messages with only Emojis with a bigger font. This very old feature has been introduced before the Matrix protocol supports reaction. So now that we have reaction, maybe we do not need this feature.

I can see @amshakal 's comment on the original ticket that this is nice to have (but I understand it is still something to have), she is probably better to comment what were the arguments for it.

frebib:
CI OOMed (again). It's pretty flaky.
Can someone re-run it please?

jmartinesp:
Ok, the screenshot test failure looks like this:

image

I think it's because in the mocked data to display we forgot to enforce a locale. If you open go [here](https://github.com/element-hq/element-x-android/blob/2f0e9ebd93638f3632c8f4b6bf91d6f8efe82d33/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/AggregatedReactionProvider.kt#L43) and add `Locale.US` to the time formatter it'll probably fix the issue we're seeing after recording the screenshots again.

frebib:
Perfect, thanks for the help! I'd have never found that 😆 
I pushed an extra commit to fix those screenshots 🤞🏻

amshakal:
> > We probably need a product decision about rendering messages with only Emojis with a bigger font. This very old feature has been introduced before the Matrix protocol supports reaction. So now that we have reaction, maybe we do not need this feature.
> 
> I can see @amshakal 's comment on the original ticket that this is nice to have (but I understand it is still something to have), she is probably better to comment what were the arguments for it.

I recommend we implement large-sized emojis in our messaging app because this aligns with user expectations based on the behavior of standard messaging apps. Most apps render a single or a few emojis larger because they have the space, and this small detail can enhance the user experience by adding a sense of delight and improving visual accessibility.

While this isn’t a core functional feature, it’s a subtle enhancement that users have come to appreciate. As a baseline, we could start by making a single emoji larger, which is the minimum implementation seen across popular messaging platforms. For example, WhatsApp uses this for up to three emojis, Signal for up to four, Slack doesn’t limit emoji size, and Telegram reduces emoji size incrementally after six.

Ultimately, I support adding this feature, even if it's just for single emojis. However, I’m also open to allowing multiple large emojis, similar to other platforms, if we want to provide a more flexible user experience.

jmartinesp (approved):
LGTM then!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/Emoji.kt
```diff
@@ -0,0 +1,61 @@
...
+fun String.containsOnlyEmojis(): Boolean {
+    val matcher = GraphemeMatcher(this)
+
+    var m: GraphemeMatchResult? = null
+    var contiguous = true
+    var prevend = 0
+    while (matcher.find()) {
+        m = matcher.toMatchResult()
+        contiguous = m!!.isEmoji() and contiguous and (m.start() == prevend)
+        prevend = m.end()
```
bmarty, line 57: I am wondering about the performance impact of this algorithm.

Maybe if `contiguous` becomes false, we can directly let the function return `false`, by breaking the loop, or change the `while` parameter to `while (contiguous && matcher.find())`?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/Emoji.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    return isNotEmpty() and contiguous and (m?.end() == length)
```
bmarty, line 60: If the String is empty, it can return false at the beginning of this method, no need to instanciate a `GraphemeMatcher`.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/Emoji.kt
```diff
@@ -0,0 +1,61 @@
...
+    // Many non-"emoji" characters are pictographics. We only want to identify this specific range
+    // https://en.wikipedia.org/wiki/Miscellaneous_Symbols_and_Pictographs
+    return grapheme().type == PICTOGRAPHIC && group() in "🌍".."🗺"
+}
+
+/**
+ * Returns true if the string contains any "emoji or pictographic graphemes".
+ * @param max Upper limit of the number of emojis that should be in the string.
+ */
+fun String.containsEmojis(max: Int = Integer.MAX_VALUE): Boolean {
```
bmarty, line 38: This method is only used in test, so maybe just remove it? Private methods above could also be removed.

bmarty (commented):
Some remarks, then I guess the PR can be merged, thanks!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/Emoji.kt
frebib: All reasonable concerns. I don't know why I didn't do this 🤔  It'll probably speed up the average case a bit

frebib:
One final remark- it seems that maybe this isn't a unanimously desired feature. Perhaps additionally to this, we could add a toggle for it, like Element Web has?

bmarty:
> One final remark- it seems that maybe this isn't a unanimously desired feature. Perhaps additionally to this, we could add a toggle for it, like Element Web has?

We try to limit the number of settings EX has, so I guess it's fine if this behavior cannot be disabled. Thanks!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/Emoji.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    val matcher = GraphemeMatcher(this)
+    var m: GraphemeMatchResult? = null
+    var contiguous = true
+    var previous = 0
+    while (contiguous && matcher.find()) {
+        m = matcher.toMatchResult()
+        // Many non-"emoji" characters are pictographics. We only want to identify this specific range
+        // https://en.wikipedia.org/wiki/Miscellaneous_Symbols_and_Pictographs
+        val isEmoji = m!!.grapheme().type == EMOJI || (m.grapheme().type == PICTOGRAPHIC && m.group() in "🌍".."🗺")
```
bmarty, line 38:
```suggestion
        val isEmoji = m!!.grapheme().type == EMOJI || m.grapheme().type == PICTOGRAPHIC && m.group() in "🌍".."🗺"
```

Remove unnecessary parenthesis

PR #3345, opened 2024-08-27
Remove feature flag `PictureInPicture`

bmarty:


libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
```diff
@@ -120,13 +120,6 @@ enum class FeatureFlags(
         defaultValue = { true },
         isFinished = false,
     ),
-    PictureInPicture(
-        key = "feature.pictureInPicture",
-        title = "Picture in Picture for Calls",
-        description = "Allow the Call to be rendered in PiP mode",
-        defaultValue = { it.buildType != BuildType.RELEASE },
-        isFinished = false,
```
ganfra, line 128: Should we not use `isFinished = true` first?

ganfra (approved):

ganfra:
Ok for me, just one question.
Also we should probably clean up a bit our feature flags at some point.

Issue #3274, opened 2024-08-06
Logged out without apparent reason


benparsons:
### Steps to reproduce

1. When I open the app, I find I am logged out

Logs sent to https://github.com/element-hq/element-x-android-rageshakes/issues/2491

#### What did you expect?

App to load

#### What happened instead?

I am at the login screen

### Your phone model

Samsung S20FE

### Operating system version

Android 13

### Application version and app store

0.5.1 Nightly

### Homeserver

element.io

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

benparsons:
Prior info:

- https://github.com/element-hq/element-x-android/issues/2999
- https://github.com/element-hq/element-x-android/issues/3028
- https://github.com/element-hq/element-x-android/issues/3176

benparsons:
Happened again today, logs sent

PR #3378, opened 2024-09-02
Make sure to save the tokens the Client might return when its session is restored

jmartinesp:
## Content

When we're restoring a Rust `Client` object with some session data, the client might automatically refresh the tokens it uses. Usually, this would get reported by the `ClientDelegate` methods, but since that's only added *after* the session is restored, the update doesn't trigger the delegate callback and the new tokens aren't stored locally. This means next time we restore the session we'd be using discarded credentials and the server will log us out.

I'm creating a new `RustClientSessionDelegate` that implements both:
- `ClientSessionDelegate`: used by iOS mainly up until now, it's passed to the `ClientBuilder` so it will be able to handle token refreshes while the Client is restoring the session, solving the problem above.
- `ClientDelegate`: what we were using for handling token refreshes and refresh errors that should end in a logout. Since the previous delegate is handling token refreshes now, we only need the logout part.

## Motivation and context

Found some promising explanation in [a rageshake](https://github.com/element-hq/element-x-android-rageshakes/issues/2621#issuecomment-2324525182).

This might fix https://github.com/element-hq/element-x-android/issues/3274.

## Tests

This is quite difficult to test since it depends on the access token expiration time, which is different in each homeserver and also you need some luck (or bad luck) to make it fail, so even if it doesn't fail in some hours or days the fix might not do anything.

I think the best we can do is include it, then check if there haven't been any recent reports of spontaneous logouts.

When I tried to test it, I did the next:

- Set up an OIDC account, log in, verify, etc., and have a room in common that will trigger notifications.
- Wait until a token refresh happens, then kill the app by swiping them from the recents menu (this will kill the app but not prevent notifications from being received).
- Wait until the access token expiry date happens and don't touch the app in the meantime.
- Send a message that will trigger a notification in the OIDC session.
- Check the logs, the restoration of the client and the refresh of the tokens should happen in that order and there should be no issues.

However, I'm not 100% certain this is the right way to reproduce the previous error.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/MatrixClient.kt
+++ b/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/MatrixClient.kt
@@ -88,10 +88,10 @@ interface MatrixClient : Closeable {
      * Logout the user.
      * Returns an optional URL. When the URL is there, it should be presented to the user after logout for
      * Relying Party (RP) initiated logout on their account page.
+     * @param userInitiated if false, the logout came from the HS, no request will be made and the session entry will be kept in the store.
      * @param ignoreSdkError if true, the SDK will ignore any error and delete the session data anyway.
-     * @param forced if true, the logout came from the HS, no request will be made and the data will be kept.
      */
-    suspend fun logout(ignoreSdkError: Boolean, forced: Boolean): String?
+    suspend fun logout(userInitiated: Boolean, ignoreSdkError: Boolean): String?
 
     /**
      * Retrieve the user profile, will also eventually emit a new value to [userProfile].
--- a/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/MatrixClient.kt
+++ b/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/MatrixClient.kt
@@ -89,8 +89,9 @@ interface MatrixClient : Closeable {
      * Returns an optional URL. When the URL is there, it should be presented to the user after logout for
      * Relying Party (RP) initiated logout on their account page.
      * @param ignoreSdkError if true, the SDK will ignore any error and delete the session data anyway.
+     * @param forced if true, the logout came from the HS, no request will be made and the data will be kept.
      */
-    suspend fun logout(ignoreSdkError: Boolean): String?
+    suspend fun logout(ignoreSdkError: Boolean, forced: Boolean): String?
 
     /**
      * Retrieve the user profile, will also eventually emit a new value to [userProfile].
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
@@ -51,8 +51,16 @@ class RustClientSessionDelegate(
     // To make sure only one coroutine affecting the token persistence can run at a time
     private val updateTokensDispatcher = coroutineDispatchers.io.limitedParallelism(1)
 
-    /** This [RustMatrixClient] needs to be set up as soon as possible so `didReceiveAuthError` can work properly. */
-    var client: RustMatrixClient? = null
+    // This Client needs to be set up as soon as possible so `didReceiveAuthError` can work properly.
+    private var client: RustMatrixClient? = null
+
+    /**
+     * Sets the [ClientDelegate] for the [RustMatrixClient], and keeps a reference to the client so it can be used later.
+     */
+    fun bindClient(client: RustMatrixClient) {
+        this.client = client
+        client.setDelegate(this)
+    }
 
     override fun saveSessionInKeychain(session: Session) {
         appCoroutineScope.launch(updateTokensDispatcher) {
@@ -83,9 +91,10 @@ class RustClientSessionDelegate(
             clientLog.v("didReceiveAuthError -> do the cleanup")
             // TODO handle isSoftLogout parameter.
             appCoroutineScope.launch(updateTokensDispatcher) {
-                val currentClient = this@RustClientSessionDelegate.client
+                val currentClient = client
                 if (currentClient == null) {
                     clientLog.w("didReceiveAuthError -> no client, exiting")
+                    isLoggingOut.set(false)
                     return@launch
                 }
                 val existingData = sessionStore.getSession(currentClient.sessionId.value)
@@ -102,7 +111,7 @@ class RustClientSessionDelegate(
                 } else {
                     clientLog.d("No session data found.")
                 }
-                client?.logout(ignoreSdkError = true, forced = true)
+                client?.logout(userInitiated = false, ignoreSdkError = true)
             }.invokeOnCompletion {
                 if (it != null) {
                     clientLog.e(it, "Failed to remove session data.")
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
@@ -1,0 +1,124 @@
+/*
+ * Copyright (c) 2024 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.element.android.libraries.matrix.impl
+
+import io.element.android.libraries.core.coroutine.CoroutineDispatchers
+import io.element.android.libraries.matrix.impl.mapper.toSessionData
+import io.element.android.libraries.matrix.impl.paths.getSessionPaths
+import io.element.android.libraries.matrix.impl.util.anonymizedTokens
+import io.element.android.libraries.sessionstorage.api.SessionStore
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.launch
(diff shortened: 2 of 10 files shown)


bmarty (commented):
LGTM, but maybe a better fix is to invoke `client.restoreSession(sessionData.toSession())` after creating the `RustMatrixClient` instance? So move the line 67 into this `also` block? So in this case, the callback `didRefreshTokens()` will be invoked, since the `clientDelegate` will be set?

jmartinesp:
I don't think we can because `RustMatrixClient` needs a `SyncService` parameter that we can only get once the session is restored. We'd have to maybe lazily initialize it, and make sure that doesn't break any existing behaviours. 

In any case, I *think* we'll implement a fix in the Rust SDK instead. I'm leaving the PR as a draft until either that solution is ready or we decide a client side one might be better.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
```diff
@@ -0,0 +1,123 @@
...
+    private val clientLog = Timber.tag("$this")
+
+    // Used to ensure a single coroutine will be modifying the token data at a time.
+    private val isLoggingOut = AtomicBoolean(false)
+
+    // To make sure only one coroutine affecting the token persistence can run at a time
+    private val updateTokensDispatcher = coroutineDispatchers.io.limitedParallelism(1)
+
+    /** This [RustMatrixClient] needs to be set up as soon as possible so `didReceiveAuthError` can work properly. */
+    var client: RustMatrixClient? = null
```
jmartinesp, line 54: I hate that this is needed, but I don't see any better way to centralise handling the token refreshes in a single point without being able to access the client for a logout.

jmartinesp:
> In any case, I think we'll implement a fix in the Rust SDK instead. I'm leaving the PR as a draft until either that solution is ready or we decide a client side one might be better.

Well, this is no longer true. We saw we could reuse the same flow as iOS with `ClientSessionDelegate`. This part probably needs a refactor in the SDK, but it's probably not the time to tackle it.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
```diff
@@ -536,9 +467,9 @@ class RustMatrixClient(
         deleteSessionDirectory(deleteCryptoDb = false)
     }
 
-    override suspend fun logout(ignoreSdkError: Boolean): String? = doLogout(
-        doRequest = true,
-        removeSession = true,
+    override suspend fun logout(ignoreSdkError: Boolean, forced: Boolean): String? = doLogout(
```
bmarty, line 470: now that `doLogout` is called only once, I guess there is no need to have a sub private method anymore. Also the `forced` parameter is maybe a bit hard to understand, maybe rename to `from401` or something better?

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/MatrixClient.kt
```diff
@@ -89,8 +89,9 @@ interface MatrixClient : Closeable {
      * Returns an optional URL. When the URL is there, it should be presented to the user after logout for
      * Relying Party (RP) initiated logout on their account page.
      * @param ignoreSdkError if true, the SDK will ignore any error and delete the session data anyway.
+     * @param forced if true, the logout came from the HS, no request will be made and the data will be kept.
```
bmarty, line 92: Actually most of the data will be deleted, but we will keep the entry in the Session database, so that the application is able to display a "You've been signed out" screen ([here](https://github.com/element-hq/element-x-android/blob/develop/appnav/src/main/kotlin/io/element/android/appnav/RootFlowNode.kt#L114)). Maybe make the comment a bit clearer?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
```diff
@@ -0,0 +1,124 @@
...
+    private val clientLog = Timber.tag("$this")
+
+    // Used to ensure several calls to `didReceiveAuthError` don't trigger multiple logouts
+    private val isLoggingOut = AtomicBoolean(false)
+
+    // To make sure only one coroutine affecting the token persistence can run at a time
+    private val updateTokensDispatcher = coroutineDispatchers.io.limitedParallelism(1)
+
+    /** This [RustMatrixClient] needs to be set up as soon as possible so `didReceiveAuthError` can work properly. */
+    var client: RustMatrixClient? = null
```
bmarty, line 55: Sorry, I do not see where the var `client` is set. Changing the visibility to `private` is OK, so I guess something is missing.

Indeed, deleting the session from another one does not work, I got 401, but nothing happen.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
```diff
@@ -0,0 +1,124 @@
...
+    override fun didReceiveAuthError(isSoftLogout: Boolean) {
+        clientLog.w("didReceiveAuthError(isSoftLogout=$isSoftLogout)")
+        if (isLoggingOut.getAndSet(true).not()) {
+            clientLog.v("didReceiveAuthError -> do the cleanup")
+            // TODO handle isSoftLogout parameter.
+            appCoroutineScope.launch(updateTokensDispatcher) {
+                val currentClient = this@RustClientSessionDelegate.client
+                if (currentClient == null) {
+                    clientLog.w("didReceiveAuthError -> no client, exiting")
+                    return@launch
```
bmarty, line 89: Maybe when this happen set `isLoggingOut` back to false, to get another chance to do the cleanup later?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    override fun didRefreshTokens() {
+        // This is done in `saveSessionInKeychain(Session)` instead.
+    }
+
+    override fun retrieveSessionFromKeychain(userId: String): Session {
+        // This should never be called, as it's only used for multi-process setups
+        error("retrieveSessionFromKeychain should never be called for Android")
```
bmarty, line 122: Maybe do not crash the app but just log an error?

bmarty (changes requested):
There is something missing, see the comment (link is coming)

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
jmartinesp: True, I completely forgot to add this in `RoomMatrixClient.init`. Good catch!

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
jmartinesp: Then I need to create a fake `Session` object, or retrieve it from disk so it conforms to the method signature.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
jmartinesp: I have inlined the doLogout call and `forced` is now inverted and called `userInitiated`.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    override fun didReceiveAuthError(isSoftLogout: Boolean) {
+        clientLog.w("didReceiveAuthError(isSoftLogout=$isSoftLogout)")
+        if (isLoggingOut.getAndSet(true).not()) {
+            clientLog.v("didReceiveAuthError -> do the cleanup")
+            // TODO handle isSoftLogout parameter.
+            appCoroutineScope.launch(updateTokensDispatcher) {
+                val currentClient = this@RustClientSessionDelegate.client
```
bmarty, line 86:
```suggestion
                val currentClient = client
```

works too :)

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
bmarty: Nicer, thanks!

libraries/matrix/test/src/main/kotlin/io/element/android/libraries/matrix/test/FakeMatrixClient.kt
```diff
@@ -170,8 +170,8 @@ class FakeMatrixClient(
         clearCacheLambda()
     }
 
-    override suspend fun logout(ignoreSdkError: Boolean, forced: Boolean): String? = simulateLongTask {
-        return logoutLambda(ignoreSdkError, forced)
+    override suspend fun logout(userInitiated: Boolean, ignoreSdkError: Boolean): String? = simulateLongTask {
+        return logoutLambda(ignoreSdkError, userInitiated)
```
bmarty, line 174: There is a swap in the parameters, this is not a big deal, but it can be confusing. Also when `userInitiated` is `false`, `ignoreSdkError` does not have sense.

Maybe we should create a small sealed interface to handle this:

```kotlin
    sealed interface LogoutParameter {
        data class UserInitiated(val ignoreSdkError: Boolean) : LogoutParameter
        data object ServerInitiated : LogoutParameter
    }
```

Not a blocker for this PR though, I can do it later, just sharing the idea here :)

bmarty (approved):
Thanks for the update, all my tests are passing now!

Hopefully no more wild sign out!

libraries/matrix/test/src/main/kotlin/io/element/android/libraries/matrix/test/FakeMatrixClient.kt
jmartinesp: Can we add this in a separate PR then? So I can merge this and add it to the current release.

Issue #3185, opened 2024-07-12
Use matrix.org's push gateway by default with UnifiedPush


p1gp1g:
#### What would you like to do?

Sygnal, the push gateway behind `https://matrix.org/_matrix/push/v1/notify` supports WebPush. This can be used to push to UnifiedPush endpoints, which is compatible with WebPush

#### Why would you like to do it?

This avoid relying on an external service (the current push gateway hosted by unifiedpush.org)

#### How would you like to achieve it?

- Add a new appId to matrix.org's [sygnal config](https://github.com/matrix-org/sygnal/blob/main/sygnal.yaml.sample):
    - (example) name: im.vector.app.unifiedpush
    - type: webpush
- Matrix.org's sygnal may have to be configured to accept all endpoints (which is the goal of webpush). It is not a security issue (SSRF) if the server can't access any internal networks or local services that aren't exposed on the Internet. Else, local IP must be blocked, [like synapse does](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#ip_range_whitelist).
- Change the [default gateway](https://github.com/element-hq/element-x-android/blob/61c021668619fa400784c3ec5c8098baf1089c90/libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushConfig.kt#L24) to use `https://matrix.org/_matrix/push/v1/notify`

And it should work.

#### Note

The discovery request to catch self-hosted gateway is still used and available

### Have you considered any alternatives?

A new push kind would be added to matrix specifications to allow webpush, that would bring encrypted push messages in the same time (https://github.com/matrix-org/matrix-spec-proposals/pull/3013)

### Additional context

-

### Are you willing to provide a PR?

No

devurandom:
See-also: https://github.com/element-hq/element-x-android/issues/3031
See-also: https://github.com/element-hq/element-x-android/issues/3571

Issue #3031, opened 2024-06-13
Not receiving notifications on the F-Droid version with unified push


K4LCIFER:
### Steps to reproduce

1. Install unified push (e.g. ntfy.sh)
2. Install Element X via F-Droid
3. Receive a message
4. Observe that no notification is received.

#### What did you expect?

I would expect to receive a notification.

#### What happened instead?

I didn't receive a notification

### Your phone model

Google Pixel 6

### Operating system version

Android 14

### Application version and app store

0.4.14

### Homeserver

matrix.org

### Will you send logs?

No

### Are you willing to provide a PR?

No

K4LCIFER:
> To unblock you, can you navigate to Setting/Notifications then scroll to the bottom. If the current push provider is displayed and in Error, you can click on it and select again "Ntfy". It should help.

I don't appear to have any push provider information on that page.
![picture](https://github.com/element-hq/element-x-android/assets/7829231/41c846ad-be23-4bc2-87e6-855ec35c8851)

> Also note that if your a matrix.org user, push may be rate limited by the ntfy server.

I don't think that's the case here. I've received notifications from other services at higher rates through Ntfy than through Element.

> It's also possible to run the notification troubleshoot tests to see what could go wrong.

It appears that there is no issue and it is able to find the UnifiedPush provider.
![picture1](https://github.com/element-hq/element-x-android/assets/7829231/eb8eb1d3-3614-4b3b-bbf2-de37aabb28a0)

alnmy:
Same issue, GrapheneOS with self-hosted ntfy.sh instance that has other notifications working but not from Element X. Also same results from the troubleshoot tests.

alnmy:
> Same issue, GrapheneOS with self-hosted ntfy.sh instance that has other notifications working but not from Element X. Also same results from the troubleshoot tests.

My issue has been resolved. Essentially, my push server was behind a VPN which my server could not access. I would consider it a bug that the notification troubleshooter reports the "test push loop back" as being completed, when the server could not do this.

bmarty:
The `"Test Push loop back"` test does not involve the Matrix homeserver. But yes, I agree that the communication homeserver -> push gateway is not covered by the troubleshoot tests, and this is annoying. This is the only part that is not covered.

This is discussed at https://github.com/matrix-org/matrix-spec-proposals/pull/4078

anytimesoon:
[...]

PR #3388, opened 2024-09-04
Unified push endpoint: do not fallback to default endpoint in case of failure and add troubleshoot test.

bmarty:
## Content

When the UnifiedPush endpoint is not detected as a valid Matrix gateway, the application was using a default endpoint.

Now, the endpoint is still checked, but is used even if it is not valid. This allow to use the provided endpoint even if an unexpected error occurs (network error, SSL certificat error, etc.)

A test is added to the troubleshoot test suite to check again the endpoint and alert the user in case of issue. For instance here, the test is failing:

image

The second commit is adding more unit test on the other troubleshoot tests.

## Motivation and context

Make UnifiedPush solution work in more cases and environments.

## Tests

- Configure notification to use UnifiedPush
- Run troubleshoot test suite and see the new test.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/permissions/impl/src/main/kotlin/io/element/android/libraries/permissions/impl/troubleshoot/NotificationTroubleshootCheckPermissionTest.kt
+++ b/libraries/permissions/impl/src/main/kotlin/io/element/android/libraries/permissions/impl/troubleshoot/NotificationTroubleshootCheckPermissionTest.kt
@@ -37,7 +37,7 @@ class NotificationTroubleshootCheckPermissionTest @Inject constructor(
     private val permissionStateProvider: PermissionStateProvider,
     private val sdkVersionProvider: BuildVersionSdkIntProvider,
     private val permissionActions: PermissionActions,
-    private val stringProvider: StringProvider,
+    stringProvider: StringProvider,
 ) : NotificationTroubleshootTest {
     override val order: Int = 0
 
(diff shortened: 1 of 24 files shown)


jmartinesp (approved):
LGTM, thanks!

p1gp1g:
Hi,
I was reading the code to understand why and I found the gateway resolver always return the custom gateway.

I was a bit surprised because only some provider have one. The matrix gateway and the discovery request is not part of the UnifiedPush specs, but we did it to help self-hosting the matrix gateway. Hopefully, the matrix gateway will not be required some time, when matrix will support web push pushkind ([MSC4174])

For instance, Conversation.im is a known distributor that doesn't have any matrix gateway, and it seems someone opened an issue for that: https://github.com/element-hq/element-x-android/issues/3571

In a second step, it may makes sense to move to matrix.org's gateway as a fallback (see #3185), and use sygnal's web push pusher. The new UnifiedPush's library generate web push keys and decrypt the push messages. That way it will make the transition to [MSC4174] moving from `kind=http` to `kind=webpush`.

BTW, do you know how to advocate for MSCs ?  [MSC4173] would be very useful too.


PS: The new lib will also fix the error raised when EAX tries to start a foreground service from the background when it doesn't have the right to (https://github.com/element-hq/element-x-android/issues/3031#issuecomment-2430189130) once supported by distributors

[MSC4173]: https://github.com/matrix-org/matrix-spec-proposals/pull/4173
[MSC4174]: https://github.com/matrix-org/matrix-spec-proposals/pull/4174

PR #3402, opened 2024-09-05
Fix Element Call closing automatically on API 34

jmartinesp:
## Content

The leave hint listener for PiP in the Call screen is now registered once the UI is ready.

## Motivation and context

It seems like registering a user leave hint listener way too early was causing the activity to try to enter PiP erroneously and that led to the activity closing instead.

## Tests

- Try to start a call in Android 14.
- Most of the time (always, in my case) the screen would previously close automatically. It should now work as expected.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
@@ -35,9 +35,12 @@ import androidx.annotation.RequiresApi
 import androidx.appcompat.app.AppCompatActivity
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberUpdatedState
+import androidx.compose.runtime.setValue
 import androidx.core.app.PictureInPictureModeChangedInfo
 import androidx.core.content.IntentCompat
 import androidx.core.util.Consumer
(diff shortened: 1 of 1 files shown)


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
```diff
@@ -126,14 +129,23 @@ class ElementCallActivity :
...
     @Composable
     private fun ListenToAndroidEvents(pipState: PictureInPictureState) {
+        val lifecycleState by lifecycle.currentStateFlow.collectAsState()
+
         val pipEventSink by rememberUpdatedState(pipState.eventSink)
-        DisposableEffect(Unit) {
-            val onUserLeaveHintListener = Runnable {
-                pipEventSink(PictureInPictureEvents.EnterPictureInPicture)
+        var onUserLeaveHintListener by remember { mutableStateOf<Runnable?>(null) }
+        DisposableEffect(lifecycleState.isAtLeast(Lifecycle.State.STARTED)) {
```
ganfra, line 136: I don't understand the `DisposableEffect(lifecycleState.isAtLeast(Lifecycle.State.STARTED))`, this will still launch the effect with value `false`, and then with value `true`

jmartinesp: I just realised I had added the wrong check to the `if` above (copy&paste issue), so now it should:

- Launch the side effect with `false`, but do nothing.
- Launch the side effect again with `true`, this time register the listener.
- If we go out of the call and then back again, the listener won't be registered again.
- When the activity is finished the listener will also be removed.

I tried with simpler versions using `LaunchedEffect` and a lifecycle observer, but neither of them removed the listener when the activity was closed. Although maybe as the activity is being closed, it's not needed at all.

jmartinesp: Actually, I was able to simplify it in https://github.com/element-hq/element-x-android/pull/3402/commits/b791c54a9e2d4285d6489bf03caeb1fd86a5d0c4. It seems like using `lifecycleScope.launch` is enough to make it wait until the UI is ready.

bmarty (approved):
My test is now passing.
FTR, starting a call on API 34 was failing, since the Activity was finished just after being started.

ganfra (approved):

PR #3419, opened 2024-09-06
Ensure Call is not hang up when user is asked to grant system permissions

bmarty:
## Content

Ensure Call is not hang up when user is asked to grant system permissions

When a system permission pop up appear, the callback onUserLeaveHint is invoked and then the application try to enter pip, but Element Call does not allow it and then the call is terminated.

The fix prevent this effect when permission request is in progress.

## Motivation and context

Replace fix #3402 

## Tests

- Revoke Camera permission
- Start an Element Call
- observe that the call is not terminated
- grant permission
- the call start
- press the home button
- the call enters pip mode

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
@@ -42,7 +42,6 @@ import androidx.core.app.PictureInPictureModeChangedInfo
 import androidx.core.content.IntentCompat
 import androidx.core.util.Consumer
 import androidx.lifecycle.Lifecycle
-import androidx.lifecycle.lifecycleScope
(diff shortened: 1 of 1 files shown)


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
```diff
@@ -234,6 +234,7 @@ class ElementCallActivity :
                 }
             }
             callback(permissionsToGrant.toTypedArray())
+            requestPermissionCallback = null
```
bmarty, line 237: Independently of the PR fix, it's a good thing to reset this var.

jmartinesp (approved):
It's simple and it fixes the issue, thanks!

PR #3429, opened 2024-09-09
Add banner for optional migration to simplified sliding sync

### LINKED ISSUE #3427: Offer user to transition to SSS (Android)
1. Follow the logic defined in the parent issue to prompt the user to log out to complete the migration to SSS.
2. For prompt choose platform-native implementation (whatever is the easiest to implement).

jmartinesp:
## Content

- Add `MatrixClient.isNativeSlidingSyncSupported()` and `MatrixClient.isUsingNativeSlidingSync` to check whether the home server supports native sliding sync and we're already using it.
- Add `NativeSlidingSyncMigrationBanner` composable to the `RoomList` screen when the home server supports native sliding sync but the current session is not using it.
- Add an extra logout successful action to the logout flow, create `EnableNativeSlidingSyncUseCase` so it can be used there.

## Motivation and context

Closes #3427.

## Screenshots / GIFs

In the PR.

## Tests

- With simplified sliding sync disabled in the client, log into a homeserver which has it enabled (i.e. element.io).
- You should see the migration banner.
- When tapping in log out & upgrade, a confirmation will be displayed.
- Once logged out if you log in again you'll be using SSS.
- The banner won't be visible then.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/roomlist/impl/src/main/kotlin/io/element/android/features/roomlist/impl/RoomListPresenter.kt
+++ b/features/roomlist/impl/src/main/kotlin/io/element/android/features/roomlist/impl/RoomListPresenter.kt
@@ -29,6 +29,7 @@ import io.element.android.features.invite.api.response.AcceptDeclineInviteState
 import io.element.android.features.invite.api.response.InviteData
 import io.element.android.features.leaveroom.api.LeaveRoomEvent
 import io.element.android.features.leaveroom.api.LeaveRoomPresenter
+import io.element.android.features.logout.api.direct.DirectLogoutPresenter
 import io.element.android.features.networkmonitor.api.NetworkMonitor
 import io.element.android.features.networkmonitor.api.NetworkStatus
 import io.element.android.features.roomlist.impl.datasource.RoomListDataSource
@@ -88,6 +89,7 @@ class RoomListPresenter @Inject constructor(
     private val acceptDeclineInvitePresenter: Presenter<AcceptDeclineInviteState>,
     private val fullScreenIntentPermissionsPresenter: FullScreenIntentPermissionsPresenter,
     private val notificationCleaner: NotificationCleaner,
+    private val logoutPresenter: DirectLogoutPresenter,
 ) : Presenter<RoomListState> {
     private val encryptionService: EncryptionService = client.encryptionService()
     private val syncService: SyncService = client.syncService()
@@ -115,13 +117,15 @@ class RoomListPresenter @Inject constructor(
 
         val contextMenu = remember { mutableStateOf<RoomListState.ContextMenu>(RoomListState.ContextMenu.Hidden) }
 
+        val directLogoutState = logoutPresenter.present()
+
         fun handleEvents(event: RoomListEvents) {
             when (event) {
                 is RoomListEvents.UpdateVisibleRange -> coroutineScope.launch {
                     updateVisibleRange(event.range)
                 }
                 RoomListEvents.DismissRequestVerificationPrompt -> securityBannerDismissed = true
-                RoomListEvents.DismissRecoveryKeyPrompt -> securityBannerDismissed = true
+                RoomListEvents.DismissBanner -> securityBannerDismissed = true
                 RoomListEvents.ToggleSearchResults -> searchState.eventSink(RoomListSearchEvents.ToggleSearchVisibility)
                 is RoomListEvents.ShowContextMenu -> {
                     coroutineScope.showContextMenu(event, contextMenu)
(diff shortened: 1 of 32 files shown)


features/roomlist/impl/src/main/kotlin/io/element/android/features/roomlist/impl/RoomListPresenter.kt
```diff
@@ -176,6 +182,7 @@ class RoomListPresenter @Inject constructor(
             derivedStateOf {
                 when {
                     currentSecurityBannerDismissed -> SecurityBannerState.None
+                    needsSlidingSyncMigration -> SecurityBannerState.NeedsNativeSlidingSyncMigration
```
bmarty, line 185: Are you sure that the `NeedsNativeSlidingSyncMigration` banner has higher priority than the `SetUpRecovery` and `RecoveryKeyConfirmation` banner?

It will fully hide the two latter, since if `NeedsNativeSlidingSyncMigration` is dismissed, `currentSecurityBannerDismissed` will be true and the other banner will never be displayed again (until the user logout and login again).

libraries/matrix/test/src/main/kotlin/io/element/android/libraries/matrix/test/FakeMatrixClient.kt
```diff
@@ -77,6 +77,8 @@ class FakeMatrixClient(
     private val clearCacheLambda: () -> Unit = { lambdaError() },
     private val userIdServerNameLambda: () -> String = { lambdaError() },
     private val getUrlLambda: (String) -> Result<String> = { lambdaError() },
+    var isNativeSlidingSyncSupportedLambda: suspend () -> Boolean = { true },
+    var isUsingNativeSlidingSyncLambda: () -> Boolean = { true },
```
bmarty, line 81: I prefer to use `private val` for lambda like this. Also, they seem to be unused, maybe add a quick unit test to check the value of `needsSlidingSyncMigration`?

libraries/preferences/api/src/main/kotlin/io/element/android/libraries/preferences/api/store/EnableNativeSlidingSyncUseCase.kt
```diff
@@ -0,0 +1,23 @@
...
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
+class EnableNativeSlidingSyncUseCase @Inject constructor(
+    private val appPreferencesStore: AppPreferencesStore,
+    private val appCoroutineScope: CoroutineScope,
+) {
+    operator fun invoke() {
+        appCoroutineScope.launch {
+            appPreferencesStore.setSimplifiedSlidingSyncEnabled(true)
```
bmarty, line 20: There is maybe an issue if the user decides to login using another homeserver... I agree this is an edge case and this is temporary code, but with the current codebase, but we should maybe have a setting per server...

bmarty (approved):
LGTM, some small non-blocking remarks.

features/roomlist/impl/src/main/kotlin/io/element/android/features/roomlist/impl/RoomListPresenter.kt
jmartinesp: Yes, you're right, this one should have the lowest priority.

libraries/matrix/test/src/main/kotlin/io/element/android/libraries/matrix/test/FakeMatrixClient.kt
jmartinesp: I usually prefer using the `public var` because that way you can set up both the value on the constructor or later in the tests if you need to change it without having to create a new method just for setting it.

libraries/preferences/api/src/main/kotlin/io/element/android/libraries/preferences/api/store/EnableNativeSlidingSyncUseCase.kt
jmartinesp: Since this forces you to log out, I don't think most users will try it and then change their mind and log into another home server. Besides, changing this would required quite a few logic changes and maybe UI ones for the developer options toggle, and it seems like an overkill for something like this.

What we could do, maybe, is that when this is enabled, the 1st login attempt will be tried using a native sliding sync config, then if it fails with a not supported error the 2nd one will use the previous proxy one as a fallback. WDYT? Maybe I can add this in a separate PR.

bmarty: OK, let's not do anything then.
Also there is another (limited) risk of user seeing the banner and decide to logout from the settings. In this case, there will be no usage of the SSS and the banner will reappear.

PR #3430, opened 2024-09-09
Pinned messages list : hide reactions

ganfra:
## Content
Allow disabling computation of reaction and read receipts from the timelineItemFactory.

## Motivation and context
Reactions should be hidden in the pinned messages list.

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/TimelineItemsFactory.kt
```diff
@@ -26,15 +29,20 @@ import kotlinx.coroutines.flow.distinctUntilChanged
...
+    @Assisted context: TimelineItemsFactoryContext,
     private val dispatchers: CoroutineDispatchers,
     private val eventItemFactory: TimelineItemEventFactory,
     private val virtualItemFactory: TimelineItemVirtualFactory,
     private val timelineItemGrouper: TimelineItemGrouper,
     private val timelineItemIndexer: TimelineItemIndexer,
 ) {
+    @AssistedFactory
+    interface Creator {
+        fun create(context: TimelineItemsFactoryContext): TimelineItemsFactory
```
bmarty, line 43: We now have creator of factories :)

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/TimelineItemsFactoryContext.kt
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+package io.element.android.features.messages.impl.timeline.factories
+
+/**
+ * Some data used to configure the creation of timeline items.
+ * @param computeReadReceipts when false, read receipts will be empty.
+ * @param computeReactions when false, reactions will be empty.
+ */
+data class TimelineItemsFactoryContext(
```
bmarty, line 15: Maybe rename to `TimelineItemsFactoryParams`? Context is sort of a reserved word in Android dev...

ganfra: I find Params a bit weird here, do you have other ideas? I wasn't sure how to name it..

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/pinned/list/PinnedMessagesListPresenter.kt
```diff
@@ -55,7 +56,7 @@ import kotlin.time.Duration.Companion.milliseconds
 class PinnedMessagesListPresenter @AssistedInject constructor(
     @Assisted private val navigator: PinnedMessagesListNavigator,
     private val room: MatrixRoom,
-    private val timelineItemsFactory: TimelineItemsFactory,
+    timelineItemsFactoryCreator: TimelineItemsFactory.Creator,
```
jmartinesp, line 59: `TimelineItemsFactory.Factory` was a bit redundant, I guess 😅 .

jmartinesp (approved):
LGTM, but I agree with @bmarty in some renaming for the `Context` class.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/TimelineItemsFactoryContext.kt
jmartinesp: I can only think of either `Params` or `Config`.

ganfra: This is a bit stupid to pass params which never changes, so I'd use @AssistedInject instead if you don't like the `lateinit`

jmartinesp: It's done on `init`, so it's not that bad, although it's true a refactor could easily break it. The alternative might be creating a `TimelineItemsFactoryContextBuilder` and I think none of us want to reach that level of design pattern overload 😅 .

ganfra: I don't like creating several bools for config, it should be in one wrapper IMO.
So either I pass the wrapper to the create function, but it's a bit stupid because it won't change overtime, either I also use an `AssistedFactory` 😅

bmarty: Yes, to be clean we will need an assisted factory.

ganfra: Done.

bmarty (approved):
Thanks for the update!

PR #3449, opened 2024-09-12
Fix events blinking at the beginning of DM

bmarty:
## Content

Do not wait for the `m.room.create` Event to be known, or for the start of the timeline to be fully loaded to filter out Events from the beginning of DM. Use the new added field `RoomInfo.creator`, if available, to filter out the creator join Event.

This way the application is able to filter out the join Event of the creator of the DM even if the timeline is not fully loaded.

## Motivation and context

Stable UI.

## Screenshots / GIFs

|Before|After|
|-|-|
|DM_Event_Before|DM_Event_After|

*Notes*: 
- to be able to see the effect I have set up the network type of the emulator to `Edge`, that's why loading the timeline is so slow.

image

- I have recorded the "After" case before the "Before" case, by checkout a commit before the fix, so please ignore the time displayed on the emulator :)

## Tests

- Create a DM, or have a DM with no messages sent
- Clear the cache from the app
- Navigate to the room list, let it getting loaded
- Reduce the speed of the network, in order to better see the effect
- Open the DM
- Observe the Events displayed
- Observe that the Events `m.room.create` and the join event of the creator are never displayed, or that the number of collapsed Event are 2 (and not 3 or 4)

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/postprocessor/LoadingIndicatorsPostProcessor.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/postprocessor/LoadingIndicatorsPostProcessor.kt
@@ -16,9 +16,11 @@ import io.element.android.services.toolbox.api.systemclock.SystemClock
 class LoadingIndicatorsPostProcessor(private val systemClock: SystemClock) {
     fun process(
         items: List<MatrixTimelineItem>,
+        isInit: Boolean,
         hasMoreToLoadBackward: Boolean,
         hasMoreToLoadForward: Boolean,
     ): List<MatrixTimelineItem> {
+        if (!isInit) return items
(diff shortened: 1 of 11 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/postprocessor/LoadingIndicatorsPostProcessor.kt
```diff
@@ -16,9 +16,11 @@ import io.element.android.services.toolbox.api.systemclock.SystemClock
 class LoadingIndicatorsPostProcessor(private val systemClock: SystemClock) {
     fun process(
         items: List<MatrixTimelineItem>,
+        isInit: Boolean,
```
ganfra, line 19: Maybe it should be name `isTimelineInit`? same for the other one.

ganfra (approved):
Thanks, just one small remark.

bmarty:
The fix will be in release 0.6.1.

PR #3450, opened 2024-09-12
Improve code coverage metrics

bmarty:
This PR is adding more tests on the :matrix:impl module.

Let's merge this PR, it's big enough and add more tests later.

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/AuthenticationException.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/AuthenticationException.kt
@@ -8,14 +8,21 @@
 package io.element.android.libraries.matrix.impl.auth
 
 import io.element.android.libraries.matrix.api.auth.AuthenticationException
-import org.matrix.rustcomponents.sdk.ClientBuildException as RustAuthenticationException
+import org.matrix.rustcomponents.sdk.ClientBuildException
 
 fun Throwable.mapAuthenticationException(): AuthenticationException {
     val message = this.message ?: "Unknown error"
     return when (this) {
-        is RustAuthenticationException.Generic -> AuthenticationException.Generic(message)
-        is RustAuthenticationException.InvalidServerName -> AuthenticationException.InvalidServerName(message)
-        is RustAuthenticationException.SlidingSyncVersion -> AuthenticationException.SlidingSyncVersion(message)
+        is ClientBuildException -> when (this) {
+            is ClientBuildException.Generic -> AuthenticationException.Generic(message)
+            is ClientBuildException.InvalidServerName -> AuthenticationException.InvalidServerName(message)
+            is ClientBuildException.SlidingSyncVersion -> AuthenticationException.SlidingSyncVersion(message)
+            is ClientBuildException.Sdk -> AuthenticationException.Generic(message)
+            is ClientBuildException.ServerUnreachable -> AuthenticationException.Generic(message)
+            is ClientBuildException.SlidingSync -> AuthenticationException.Generic(message)
+            is ClientBuildException.WellKnownDeserializationException -> AuthenticationException.Generic(message)
+            is ClientBuildException.WellKnownLookupFailed -> AuthenticationException.Generic(message)
+        }
         else -> AuthenticationException.Generic(message)
     }
 }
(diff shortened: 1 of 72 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/exception/ClientException.kt
```diff
@@ -12,7 +12,11 @@ import org.matrix.rustcomponents.sdk.ClientException as RustClientException
 
 fun Throwable.mapClientException(): ClientException {
     return when (this) {
-        is RustClientException.Generic -> ClientException.Generic(msg)
+        is RustClientException -> {
+            when (this) {
+                is RustClientException.Generic -> ClientException.Generic(msg)
```
jmartinesp, line 17: Maybe use `is RustClientException.Generic` directly in the first branch?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/AuthenticationException.kt
```diff
@@ -8,14 +8,21 @@
...
 fun Throwable.mapAuthenticationException(): AuthenticationException {
     val message = this.message ?: "Unknown error"
     return when (this) {
-        is RustAuthenticationException.Generic -> AuthenticationException.Generic(message)
-        is RustAuthenticationException.InvalidServerName -> AuthenticationException.InvalidServerName(message)
-        is RustAuthenticationException.SlidingSyncVersion -> AuthenticationException.SlidingSyncVersion(message)
+        is ClientBuildException -> when (this) {
+            is ClientBuildException.Generic -> AuthenticationException.Generic(message)
+            is ClientBuildException.InvalidServerName -> AuthenticationException.InvalidServerName(message)
+            is ClientBuildException.SlidingSyncVersion -> AuthenticationException.SlidingSyncVersion(message)
```
jmartinesp, line 19: I hope the Client builder retry wasn't affected by this! I think it uses the raw `ClientBuildException.SlidingSyncVersion` since it comes directly from the SDK.

libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/room/preview/RoomPreviewMapperTest.kt
```diff
@@ -0,0 +1,96 @@
...
+import io.element.android.libraries.matrix.api.room.RoomType
+import io.element.android.libraries.matrix.api.room.preview.RoomPreview
+import io.element.android.libraries.matrix.test.A_ROOM_ALIAS
+import io.element.android.libraries.matrix.test.A_ROOM_ID
+import org.junit.Test
+import org.matrix.rustcomponents.sdk.RoomPreview as RustRoomPreview
+
+class RoomPreviewMapperTest {
+    @Test
+    fun `map should map values 1`() {
```
jmartinesp, line 20: Kind of a generic test name 😅 .

jmartinesp (approved):
LGTM, thanks a lot for this work!

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/exception/ClientException.kt
bmarty: This is not equivalent, using a dedicated `when` ensures that all the `RustClientException` sub-classes are mapped.

libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/room/preview/RoomPreviewMapperTest.kt
bmarty: Yeah, I tried not to spend hours looking for perfect names for tests like this :)

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/AuthenticationException.kt
bmarty: Yes, it should be OK.

But just to understand, there is no code behavior change here, this is just about ensuring that all the `ClientBuildException` sub classes are mapped, or am I missing something?

jmartinesp: No, it's fine. I double checked and we're using the exception from the SDK instead of the mapped one, so even if anything changed here it wouldn't matter.

PR #3458, opened 2024-09-13
Add forced logout flow when the proxy is no longer available

jmartinesp:
## Content

- Added `MatrixClient.isSlidingSyncProxyAvailable()` function.
- Modifies `ErrorDialog` component so it can be made non-dismissable if needed, added a separate submit action from the dismiss one.
- Check if a forced migration to native sliding sync is needed (the user is using the proxy, which is no longer available), and display the error dialog on top of all screens.
- Rework the login process so it tries to log in using native sliding sync if available and enabled, then using the proxy. Also enable native sliding sync support by default.

## Motivation and context

Second part of https://github.com/element-hq/element-x-android/issues/3427.

## Screenshots / GIFs

-

## Tests

This is quite difficult to test unless you either change the check in `LoggedInFlowNode` or make the return value of `Client::available_sliding_sync_versions()` in Rust to just the native one. If you manage to do one of them:

- Open the app, the dialog should be displayed.
- It must not be dismissable and should prevent using the app.
- Restarting the app should make no change.
- Tapping on logout and upgrade should do exactly that.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
```diff
@@ -534,6 +534,10 @@ class RustMatrixClient(
         return client.availableSlidingSyncVersions().contains(SlidingSyncVersion.Native)
     }
 
+    override suspend fun isSlidingSyncProxySupported(): Boolean {
+        return client.availableSlidingSyncVersions().any { it is SlidingSyncVersion.Proxy }
```
ganfra, line 538: Can't this fail?

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
ganfra, line 1: Can we try to move logic to the LoggedInPresenter and LoggedInView?

features/login/impl/src/main/res/values-be/translations.xml
ganfra, line 1: PR should avoid updating all translations, just the ones you need for your work would be better :)

ganfra (approved):
Some remarks, otherwise LGTM

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
jmartinesp: Apparently not, if the SDK can't load the data (network error, one of the config files not existing...) it just returns an empty list.

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
jmartinesp: I can try 🤞 .

features/login/impl/src/main/res/values-be/translations.xml
jmartinesp: That's actually what I did, but I had to remove the translations in other languages for a couple of strings because otherwise the code quality checks wouldn't pass.

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/3458/commits/b6b286afac4d621c8ab8295c65c051ea2d019a4c

PR #3461, opened 2024-09-13
Require acknowledgement to send to a verified user if their identity changed or if a device is unverified.

ganfra:
## Content
This PR makes the following changes:

Build clients with the requirements to fail when sending to once verified users who aren't (fully) verified any more
Show a description of the send failures for verified users in the ActionListView
Add a bottom sheet to resolve failures for verified users by tapping on the description above, or by tapping on the failure in the timeline.
If there is more than one user ID in the failure, the sheet iterates through them 1 by 1, resolving each one and then dismissing itself at the end.

## Motivation and context
Closes https://github.com/element-hq/element-x-android/issues/3459
Closes https://github.com/element-hq/element-x-android/issues/3460

## Tests

- Makes sure to verify a user with another client
- Reset the security or login with a new device for this verified user
- Send a message with EX in an encrypted room with this user
- See the message failing to send 
- Click on the error icon or long click on the item
- See the bottomsheet and play with the options

## Tested devices

- [X] Physical
- [ ] Emulator
- OS version(s):

## Checklist


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/actionlist/ActionListPresenter.kt
```diff
@@ -115,19 +116,47 @@ class DefaultActionListPresenter @AssistedInject constructor(
...
     }
 
+    private suspend fun buildVerifiedUserSendFailure(
+        timelineItem: TimelineItem.Event,
+    ): ActionListState.VerifiedUserSendFailure {
+        return when (val sendState = timelineItem.localSendState) {
+            is LocalEventSendState.Failed.VerifiedUserHasUnsignedDevice -> {
+                val userId = sendState.devices.keys.firstOrNull()
+                if (userId == null) {
+                    ActionListState.VerifiedUserSendFailure.None
```
jmartinesp: Is this OK? I guess it shouldn't ever happen except if there is some bug in the SDK, but maybe we should still display the UI with some placeholder value and log the issue?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/actionlist/ActionListPresenter.kt
```diff
@@ -115,19 +116,47 @@ class DefaultActionListPresenter @AssistedInject constructor(
...
+                } else {
+                    val displayName = room.userDisplayName(userId).getOrNull() ?: userId.value
+                    ActionListState.VerifiedUserSendFailure.UnsignedDevice(displayName)
+                }
+            }
+            is LocalEventSendState.Failed.VerifiedUserChangedIdentity -> {
+                val userId = sendState.users.firstOrNull()
+                if (userId == null) {
+                    ActionListState.VerifiedUserSendFailure.None
```
jmartinesp: Same here, whether we decide to treat this as 'the SDK failed to provide some user but the status is right' or 'the status is wrong since there is no user' I think we should at least log these issues.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/MatrixRoom.kt
```diff
@@ -349,5 +350,9 @@ interface MatrixRoom : Closeable {
      */
     suspend fun clearComposerDraft(): Result<Unit>
 
+    suspend fun ignoreDeviceTrustAndResend(devices: Map<UserId, List<DeviceId>>, transactionId: TransactionId): Result<Unit>
+    suspend fun withdrawVerificationAndResend(userIds: List<UserId>, transactionId: TransactionId): Result<Unit>
```
jmartinesp, line 354: Could we add some docs here? I'm not really sure what 'withdraw verification' means in this context.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/sendfailure/resolve/ResolveVerifiedUserSendFailureView.kt
```diff
@@ -0,0 +1,156 @@
...
+
+@Composable
+private fun VerifiedUserSendFailure.title(): String {
+    return when (this) {
+        is VerifiedUserSendFailure.UnsignedDevice -> stringResource(id = CommonStrings.screen_resolve_send_failure_unsigned_device_title, userDisplayName)
+        is VerifiedUserSendFailure.ChangedIdentity -> stringResource(
+            id = CommonStrings.screen_resolve_send_failure_changed_identity_title,
+            userDisplayName
+        )
+        VerifiedUserSendFailure.None -> ""
```
jmartinesp, line 121: Maybe these should be `error("This method should never be called for this state")`? Up to you, it would allow us to catch any regressions faster (and maybe make some users angry 😅 ).

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/sendfailure/resolve/VerifiedUserSendFailureResolver.kt
```diff
@@ -0,0 +1,64 @@
...
+
+package io.element.android.features.messages.impl.crypto.sendfailure.resolve
+
+import androidx.compose.runtime.mutableStateOf
+import io.element.android.libraries.matrix.api.core.TransactionId
+import io.element.android.libraries.matrix.api.room.MatrixRoom
+import io.element.android.libraries.matrix.api.timeline.item.event.LocalEventSendState
+import timber.log.Timber
+
+class VerifiedUserSendFailureResolver(
```
jmartinesp, line 16: Some docs, maybe?

jmartinesp (approved):
Thanks for the code, it works pretty well! I have a few comments, but none are either really important or urgent.

PR #3479, opened 2024-09-17
Account deactivation.

bmarty:
## Content

Add ability for the user to delete the account in-app, if this is a password based account.

## Motivation and context

Now the the application is able to create account, we need to provide a way to delete it.

## Screenshots / GIFs

AccountDeactivation

Notes:
- I do not know why the first icon in the list is not tinted, probably a cache issue. Let's see how the screanshots are getting recorded.
- the "Deactivate" button in the confirmation dialog is now red (destructive)

## Tests

- Create an account (can be done in-app on matrix.org)
- Navigate to the setting a deactivate the account

Observe that the account has been deleted (this is not possible to login again) and the application is back to fresh install state.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 23
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [ ] Pull request is based on the develop branch
- [ ] Pull request title will be used in the release note, it clearly define what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
- [ ] You've made a self review of your PR

Issue #3478, opened 2024-09-17
Sync state


bmarty:
We stop the sync before login out here: https://github.com/element-hq/element-x-android/blob/main/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt#L466, but there are 2 problems:

- we do not start it again in case of logout failure (which is mitigated by the second problem)
- if there is network coverage, the sync is automatically started again here: https://github.com/element-hq/element-x-android/blob/main/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt#L179

We should probably let the MatrixClient expose a `canStartSync` flow to avoid the sync to be automatically started again by `observeSyncStateAndNetworkStatus()`

bmarty:
We are not stopping the sync anymore, but instead we remove the client listener (done in #3480). So I guess we can close this issue.

PR #3480, opened 2024-09-17
Make sure the logout action doesn't cause a crash

jmartinesp:
## Content

Making sure to unregister the delegate should fix the first part of the issue.

For the other one, adding `RustSyncService.isServiceReady` to check if we should start/stop the service, which is enabled by default and set to false on destroy should help.

We should be extra careful to also include this same behaviour to account deactivation after https://github.com/element-hq/element-x-android/pull/3479 if this PR is merged.

## Motivation and context

Some reasons why this could happen:
1. The `ClientDelegate` could receive a `didReceiveAuthError` callback call on a logout, which could trigger another logout when every Rust object had already been destroyed.
2. Even though we stop the sync before logging out, `LoggedInFlowNode` will try to start it again automatically when it detects we still have internet connection.

## Tests

- Log in, log out, several times.

If it doesn't create a crash anymore, it's fixed. Also, failures while logging out should restore the client delegate and allow the user to keep refreshing their token and receive auth errors.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
@@ -50,7 +50,6 @@ class RustClientSessionDelegate(
      */
     fun bindClient(client: RustMatrixClient) {
         this.client = client
-        client.setDelegate(this)
     }
 
     override fun saveSessionInKeychain(session: Session) {
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt
@@ -16,15 +16,22 @@ import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.stateIn
-import org.matrix.rustcomponents.sdk.SyncServiceInterface
 import org.matrix.rustcomponents.sdk.SyncServiceState
 import timber.log.Timber
+import java.util.concurrent.atomic.AtomicBoolean
+import org.matrix.rustcomponents.sdk.SyncService as InnerSyncService
 
 class RustSyncService(
-    private val innerSyncService: SyncServiceInterface,
+    private val innerSyncService: InnerSyncService,
     sessionCoroutineScope: CoroutineScope
 ) : SyncService {
+    private val isServiceReady = AtomicBoolean(true)
+
     override suspend fun startSync() = runCatching {
+        if (!isServiceReady.get()) {
+            Timber.d("Can't start sync: service is not ready")
+            return@runCatching
+        }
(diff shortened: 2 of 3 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
```diff
@@ -50,7 +50,6 @@ class RustClientSessionDelegate(
      */
     fun bindClient(client: RustMatrixClient) {
         this.client = client
-        client.setDelegate(this)
```
jmartinesp, line 53: This was never really needed, as `RustMatrixClient` automatically sets its inner `Client` delegate.

bmarty:
> We should be extra careful to also include this same behaviour to account deactivation after https://github.com/element-hq/element-x-android/pull/3479 if this PR is merged.

Done in 1473afc10c3bb0c5fd61ecdaec219628f44e1f5a if you want to have a quick look @jmartinesp

bmarty (approved):
LGTM, thanks
Related to #3478

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt
```diff
@@ -16,28 +16,47 @@ import kotlinx.coroutines.flow.distinctUntilChanged
...
 
 class RustSyncService(
-    private val innerSyncService: SyncServiceInterface,
+    private val innerSyncService: InnerSyncService,
     sessionCoroutineScope: CoroutineScope
 ) : SyncService {
+    private val isServiceReady = AtomicBoolean(true)
+
     override suspend fun startSync() = runCatching {
+        if (!isServiceReady.get()) {
```
ganfra, line 31: Do we really need those checks, the failure will be catched by `runCatching` anyway

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt
```diff
@@ -16,28 +16,47 @@ import kotlinx.coroutines.flow.distinctUntilChanged
...
+            return@runCatching
+        }
         Timber.i("Start sync")
         innerSyncService.start()
     }.onFailure {
         Timber.d("Start sync failed: $it")
     }
 
     override suspend fun stopSync() = runCatching {
+        if (!isServiceReady.get()) {
```
ganfra, line 42: Same

ganfra (approved):

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt
jmartinesp: This way we can have more detailed logging in case we run into a similar issue in the future, but you're right, it's not really needed.

ganfra: With the runcatching we are already logging the exception if I'm correct

PR #3482, opened 2024-09-17
Use FFI interfaces rather than implementation

bmarty:
## Content

Use FFI interfaces rather than implementation

## Motivation and context

This is good practice this way we will be able to add more unit tests.

## Tests

- Smoke test.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
```diff
@@ -604,7 +605,7 @@ class RustMatrixRoom(
     override fun getWidgetDriver(widgetSettings: MatrixWidgetSettings): Result<MatrixWidgetDriver> = runCatching {
         RustWidgetDriver(
             widgetSettings = widgetSettings,
-            room = innerRoom,
+            room = innerRoom as Room,
```
bmarty, line 608: This would need to be avoided too.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/widget/MatrixWidgetSettings.kt
```diff
@@ -26,13 +27,13 @@ fun MatrixWidgetSettings.Companion.fromRustWidgetSettings(widgetSettings: Widget
...
 suspend fun MatrixWidgetSettings.generateWidgetWebViewUrl(
-    room: Room,
+    room: RoomInterface,
     clientId: String,
     languageTag: String? = null,
     theme: String? = null
 ) = generateWebviewUrl(
     widgetSettings = this.toRustWidgetSettings(),
-    room = room,
+    room = room as Room,
```
bmarty, line 36: This would need to be avoided too.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/widget/RustWidgetDriver.kt
```diff
@@ -45,7 +46,7 @@ class RustWidgetDriver(
         val coroutineScope = CoroutineScope(coroutineContext)
         coroutineScope.launch {
             // This call will suspend the coroutine while the driver is running, so it needs to be launched separately
-            driverAndHandle.driver.run(room, widgetCapabilitiesProvider)
+            driverAndHandle.driver.run(room as Room, widgetCapabilitiesProvider)
```
bmarty, line 49: This would need to be avoided too.

libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/sdk/FakeRoomInterface.kt
```diff
@@ -0,0 +1,355 @@
...
+import org.matrix.rustcomponents.sdk.RtcApplicationType
+import org.matrix.rustcomponents.sdk.StateEventType
+import org.matrix.rustcomponents.sdk.TaskHandle
+import org.matrix.rustcomponents.sdk.Timeline
+import org.matrix.rustcomponents.sdk.TypingNotificationsListener
+import org.matrix.rustcomponents.sdk.UserPowerLevelUpdate
+import uniffi.matrix_sdk.RoomMemberRole
+import uniffi.matrix_sdk.RoomPowerLevelChanges
+
+class FakeRoomInterface(
```
bmarty, line 36: Ideally this would be a generated class.

jmartinesp:
The only downsides I see to this are having to use `(FooInterface as? Foo)?.destroy()` and `(FooInterface as? Foo)?.use { ... }` since neither the `Disposable` nor the `Closeable` interfaces are part of those `*Interface` components.

If we don't care about destroying the Rust objects ourselves and we leave that to the `Cleaner` using these interfaces should have no real downsides AFAICT.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
```diff
@@ -128,15 +131,17 @@ class RustMatrixClient(
...
 
     private val rustSyncService = RustSyncService(syncService, sessionCoroutineScope)
     private val pushersService = RustPushersService(
         client = client,
         dispatchers = dispatchers,
     )
-    private val notificationProcessSetup = NotificationProcessSetup.SingleProcess(syncService)
+
+    // Note: need to force casting to ClientSyncService for now
+    private val notificationProcessSetup = NotificationProcessSetup.SingleProcess((syncService as ClientSyncService))
```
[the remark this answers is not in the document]
jmartinesp: I don't think this is possible. The receiving Rust code expects a concrete Rust implementation.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
```diff
@@ -279,20 +280,20 @@ class RustMatrixAuthenticationService @Inject constructor(
...
-                   .run { config() }
-                   .build()
+                return rustMatrixClientFactory
+                    .getBaseClientBuilder(
+                        sessionPaths = sessionPaths,
+                        passphrase = pendingPassphrase,
+                        slidingSyncType = slidingSyncType,
+                    )
+                    .config()
+                    .build()
```
jmartinesp, line 296: There's a weird extra whitespace here 🤔 .

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
```diff
@@ -304,7 +305,7 @@ class RustMatrixAuthenticationService @Inject constructor(
                 passphrase = pendingPassphrase,
                 slidingSyncType = getSlidingSyncProxy(),
             )
-            .run { config() }
+            .config()
```
jmartinesp, line 308: 👌

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
```diff
@@ -251,18 +251,19 @@ class RustMatrixAuthenticationService @Inject constructor(
...
                         .toSessionData(
                             isTokenValid = true,
                             loginType = LoginType.QR,
                             passphrase = pendingPassphrase,
                             sessionPaths = emptySessionPaths,
                         )
-                    sessionStore.storeData(sessionData)
-                    SessionId(sessionData.userId)
+                } finally {
+                    Disposable.destroy(client)
```
jmartinesp, line 263: It may be an overkill, but maybe we could create a general `.use { ... }` extension function? Something like:

```kotlin
inline fun <T : Any?, R> T.use(block: (T) -> R) =
    try {
        block(this)
    } finally {
        try {
            // N.B. our implementation is on the nullable type `Disposable?`.
            Disposable.destroy(this)
        } catch (e: Throwable) {
            // swallow
        }
    }
```

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustMatrixRoom.kt
jmartinesp: I don't think this is possible either, same reason: Rust code expects to receive Rust concrete types, unless we exported this as a trait instead, maybe.

libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/room/member/RoomMemberListFetcherTest.kt
```diff
@@ -204,32 +203,6 @@ class RoomMemberListFetcherTest {
...
-        membersNoSyncCallCount++
-        return getMembersNoSync()
-    }
-
-    override fun close() {
-        // No-op
-    }
-}
-
 class FakeRoomMembersIterator(
```
jmartinesp, line 206: Maybe this should be moved too?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
bmarty: I double checked, a space was missing on all those lines.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
bmarty: Yes, I though about it :)

libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/room/member/RoomMemberListFetcherTest.kt
bmarty: Ah yes, probably

libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/sdk/FakeRoomInterface.kt
```diff
@@ -0,0 +1,355 @@
...
+    override suspend fun memberDisplayName(userId: String): String? {
+        TODO("Not yet implemented")
+    }
+
+    override suspend fun members(): RoomMembersIterator {
+        membersCallCount++
+        return getMembers()
+    }
+
+    override suspend fun membersNoSync(): RoomMembersIterator {
```
bmarty, line 235: @jmartinesp `members()` and `membersNoSync()` should return `RoomMembersIteratorInterface` instead, so that we can provide a Fake implementation.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
```diff
@@ -353,7 +354,7 @@ class RustMatrixAuthenticationService @Inject constructor(
     }
 
     private fun clear() {
-        currentClient?.close()
+        Disposable.destroy(currentClient)
```
bmarty, line 357: There is maybe one thing, the `close()` method is annotated with `@Synchronized` and `destroy()` is not, but has an `AtomicBoolean` to protect from multiple execution, so I think this is fine. WDYT @jmartinesp ?

libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/room/member/RoomMemberListFetcherTest.kt
bmarty: Done in https://github.com/element-hq/element-x-android/pull/3482/commits/3d883916e6158f22a53eb35c6586b6dc42491b49

libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/sdk/FakeRoomInterface.kt
jmartinesp: We *could* create a `RoomMembersInterface` trait in Rust, export that and make `RoomMembersIterator` implement it, but then on the FFI layer we'd have an interface created for the trait and then a class as the implementation:

```kotlin
public interface RoomMembersIteratorInterfaceInterface { // MOAR INTERFACES!
    
    fun `len`(): kotlin.UInt
    
    fun `nextChunk`(`chunkSize`: kotlin.UInt): List<RoomMember>?
    
    companion object
}

open class RoomMembersIteratorInterface: Disposable, AutoCloseable, RoomMembersIteratorInterfaceInterface {
...
```

And then:

```kotlin
// Same name as the open class above, but as an interface now?
public interface RoomMembersIteratorInterface {
    
    fun `len`(): kotlin.UInt
    
    fun `nextChunk`(`chunkSize`: kotlin.UInt): List<RoomMember>?
    
    companion object
}

// So is this using the interface above or the open class? I guess it's the interface
open class RoomMembersIterator: Disposable, AutoCloseable, RoomMembersIteratorInterface {
...
```

As the concrete Rust implementation ¯\\\_(ツ)_/¯. Does it make sense? Probably note. Is it useful for us in any way? Nope.

Some time ago I tried pushing a PR to the UniFFI project to ensure Rust functions took/returned the `*Interface` versions of the different exported structs and traits, but they said it might be too complex to maintain and cause some unexpected issues so I had to settle with just being able to create `open` versions of the concrete Rust classes, which is what we're currently using for `RoomMembersIterator` here.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
jmartinesp: I think it should be fine.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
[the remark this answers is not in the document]
bnjbvr: Yeah we don't have a notion of a Sync service interface in the Rust code, I think it's an abstraction on your side?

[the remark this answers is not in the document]
bmarty: On our side we should only manipulate `*Interface` interfaces, but we are forced to use Rust implementation of the interfaces in some locations.

libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/sdk/FakeRoomInterface.kt
bmarty: I do not understand, `RoomMembersIteratorInterface` already exists. But OK about about the limitation of the UniFFI.

bmarty: OK, let's extend open class then, since it seems to be the long term solution and it's not that bad.
If a method is not extended and used in the test, the test fails, to this is enough to have a not-so-bad devx.

https://github.com/element-hq/element-x-android/pull/3482/commits/eaaea7eb2c9cc9edb46777403c4b847de4cc4542

jmartinesp: `RoomMembersIteratorInterface` is an interface generated in the Kotlin bindings by UniFFI, not something that exists in Rust. It's supposed to be used for fakes, but every SDK call needs a concrete Rust type instead of the generated interface, so I don't really see how useful it can be. Apparently it does work well enough for the Mozilla folks, because they didn't want to change it 🫤 .

bmarty:
I think the PR is now mergeable.
The changes it contains will allow to add more unit test (maybe in #3450)

ganfra:
I'm not sure to see the interest to replace the classes by interface inside the Rust implementation classes... if they are called `RustBlablaService`, we are expecting to use Rust classes, not fakes. Most of our code inside this module is already just wrapper to avoid exposing the rust.

bmarty:
OK, will open another PR with just some cleanup then.

Issue #3496, opened 2024-09-19
In the user profile screen, touching the mxid should copy it to clipboard 


benparsons:
#### Why would you like to do it?

Useful to extract mxids.

### Are you willing to provide a PR?

No

jmartinesp:
This was done some time ago but we forgot to close the issue 😅 .

PR #3502, opened 2024-09-19
Tapping user mxid or room alias copies it to the clipboard

frebib:
## Content

Make user id and room alias text in room/user view pages clickable and copy the text to the clipboard on click.

One small issue with this implementation is the snackbars don't appear until the view is closed.. not sure why, because I copy-pasted some other code that does the same thing and it works fine. Coroutines in Kotlin are ~shit~ weird. Suggestions welcome 🙏🏻 

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/3496

(Do note that I have no idea what I'm doing.. I merely mimic what I see until it works)

## Screenshots / GIFs

https://github.com/user-attachments/assets/50915a67-b92f-467d-8c1f-ec93fa00a83a

## Tests

Finger touchy screeny

- Step 1
- Step 2
- Step ...

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 14

## Checklist


bmarty:
> One small issue with this implementation is the snackbars don't appear until the view is closed.. not sure why, because I copy-pasted some other code that does the same thing and it works fine. Coroutines in Kotlin are ~shit~ weird. Suggestions welcome 🙏🏻

You need the View where SnackBar may be rendered to have a SnackBarHost.

You can look for `val snackbarHostState = rememberSnackbarHostState(snackbarMessage = state.snackbarMessage)` in the code base and add the code to all the View which may have to render a Snackar. I know this is painful :).

frebib:
Perfect, thanks Benoit. I'll give that a go in a few days and see how I get on

frebib:
Snackbars are fixed, thanks!
As per the contributing guidelines, I'll still need to add translations to localazy for the snackbar message. I appreciate that this doesn't have a design, but perhaps we could agree on working for that message? I've currently just copied what Element Android used, which is simply `Copied to clipboard`. Once we're agreed, I'll go ahead and work out how to submit that.
I should probably take a stab at trying to add some presenter tests for this new functionality at some point too

libraries/ui-strings/src/main/res/values/localazy.xml
```diff
@@ -130,6 +130,7 @@
   <string name="common_call_invite">"Call in progress (unsupported)"</string>
   <string name="common_call_started">"Call started"</string>
   <string name="common_chat_backup">"Chat backup"</string>
+  <string name="common_copied_to_clipboard">"Copied to clipboard"</string>
```
bmarty, line 133: I have added the string to Localazy at https://localazy.com/p/element/source-language/_a7211042194211892347. (there is no action required from you since the key is the same).

bmarty:
Thanks for the update. I am wondering if the Snackbar interferes too much with the System clipboard alert (but I did not test the code yet):

image

frebib:
> Thanks for the update. I am wondering if the Snackbar interferes too much with the System clipboard alert

Yeah I thought about this too. I'm not sure which subset of Android this little pop-up applies to. My guess would be either Pixel, or some modern Android version. There are likely some devices that don't have this and on those the snackbar would be useful?

There are other places in the app that also use this snackbar popup when copying to the clipboard. I've added it here for consistency, but I'd be happy to remove it (It'll greatly simplify the patch) if the decision is taken to remove them from other places in the app too. I'm not sure what iOS does

jmartinesp:
> My guess would be either Pixel, or some modern Android version

I'm pretty sure this [started on Android 13](https://9to5google.com/2022/03/18/android-13-dp2-clipboard-overlay-pixel-bottom-search-more/).

Issue #3420, opened 2024-09-06
F-Droid can't build


licaon-kter:
### Steps to reproduce

ref: https://gitlab.com/fdroid/fdroiddata/-/jobs/7763982630#L6856

diff log: [elx40005021.log](https://github.com/user-attachments/files/16909302/elx40005021.log) see line 33?

we are using _this one weird trick_ in the recipe: https://gitlab.com/fdroid/fdroiddata/-/blob/master/metadata/io.element.android.x.yml#L40

so you don't have `main`  there now? remember https://github.com/element-hq/element-x-android/issues/1418#issuecomment-2129067795

### Outcome

.

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

/LE: fyi https://gitlab.com/fdroid/fdroiddata/-/commit/56269f5965ec1a9e1feb20192bfd74d333cb7300

bmarty:
The `gitBranchName` has been move (in #3381) and is now set here: https://github.com/element-hq/element-x-android/blob/main/app/build.gradle.kts#L301

```kotlin
    val gitBranchName = providers.of(GitBranchNameValueSource::class.java) {}.get()
```

I guess the trick at https://gitlab.com/fdroid/fdroiddata/-/blob/master/metadata/io.element.android.x.yml#L40 need to be updated to alter this line instead.

linsui:
[diff.zip](https://github.com/user-attachments/files/16927129/diff.zip)

The branch name patch has been update. The current problem is that two classes have different order. I have tried to build it many times with different cpu core numbers and the result never changed. I thought there is difference between the github ubuntu runner and our debian VM.

bmarty:
There is no way to "ignore" that difference?

linsui:
No, we can't ignore it. We have to fix it.

licaon-kter:
wow, so close
```
   --- ElementXAndroidv0.6.0_armv7/smali/io/element/android/x/di/AppComponent.smali
── +++ io.element.android.x_40006001/smali/io/element/android/x/di/AppComponent.smali
┄ Ordering differences only
 @@ -1,15 +1,15 @@
  .class public interface abstract Lio/element/android/x/di/AppComponent;
  .super Ljava/lang/Object;
  .source "SourceFile"
  
  # interfaces
  .implements Lio/element/android/libraries/architecture/NodeFactoriesBindings;
 -.implements Lio/element/android/x/di/AppBindings;
[...]

PR #3518, opened 2024-09-23
Rename AppBindings to XAppBindings

bmarty:
Rename AppBindings to XAppBindings to attempt to fix reproducible build issue

Attempt to fix difference reported here: https://github.com/element-hq/element-x-android/issues/3420#issuecomment-2349031957

If this does not work, we may want to move `XAppBindings` to a sub package...


jmartinesp (approved):
It's kind of sad we have to workaround this, but let's try it out 🥲 .

jmartinesp (changes requested):
It seems like this doesn't fix the issue for reproducible builds, so there's no point in the rename... Thanks for trying to fix it, though.

PR #3527, opened 2024-09-24
Handle no network error when starting Element Call.

### LINKED ISSUE #3506: Element Call without network
The Element Call experience when there is no network available is quite not ideal. An error is displayed in the WebView:

image

Also, it's possible to enter PiP mode in this case:

image

Ideally, the error should be intercepted by the application and it should render the error natively, with maybe a button to retry.

On iOS a blank page is displayed forever.

To repro:
- enter airplane mode
- start a call

bmarty:
## Content

Ensure that Element Call is not rendering an error webpage when it cannot be loaded.

If Element Call does not start and the WebView emit an error, display a dialog, hide the webview. When the dialog is close, hangup, so finish the Activity.

If Element Call starts (the app receive a message from it), we ignore further error, because Element Call can recover by itself (when netword is back from instance).

## Motivation and context

Closes #3506 

## Screenshots / GIFs

|Before|After|
|-|-|
|image|(coming)|

## Tests

- Enter airplane mode
- Start a Call
- Observe the error dialog. It's not possible to enter PiP mode from this state.
- Exit airplane mode
- start a call
- the call starts
- enter airplane mode
- the call freezes
- exit airplane mode
- the call recovers, no error dialog have been displayed

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
@@ -17,6 +17,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
+import androidx.compose.ui.res.stringResource
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedFactory
 import dagger.assisted.AssistedInject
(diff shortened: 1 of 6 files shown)


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
```diff
@@ -163,11 +167,18 @@ class CallScreenPresenter @AssistedInject constructor(
                 is CallScreenEvents.SetupMessageChannels -> {
                     messageInterceptor.value = event.widgetMessageInterceptor
                 }
+                is CallScreenEvents.OnWebViewError -> {
+                    if (!ignoreWebViewError) {
+                        webViewError = event.description.orEmpty()
```
ganfra, line 172: Should this `webViewError` be refreshed if `ignoreWebViewError` becomes true after?

bmarty: There is no mechanism currently to load again the page if the network is back. So not null `webViewError` is sort of a final state.

ganfra (approved):

PR #3537, opened 2024-09-25
Fix image viewer glitch

ganfra:
- Fix glitch when transitioning from thumbnail to content
- Open avatars using overlay mode instead of backstack


bmarty (approved):
Thanks. Tested OK. As a bonus I have fixed the 2 ignored tests on this View.

Issue #2775, opened 2024-04-30
Jumping to a message may not bring it into view


frebib:
Jumping to a message seems to bring the bottom of the following message in line with the bottom of the timeline viewport. This works fine in most cases, however if the message you want to jump to is very long, it never actually scrolls up far enough to bring it into view. A video demonstration

https://github.com/element-hq/element-x-android/assets/775104/e11cb9c1-a93f-4cc4-8caf-b61f2e6a2b5a

A link to the message in the above video: https://matrix.to/#/!ping-v11:maunium.net/$WWJk26zwdtCIEVQSlGiv0v3F8kZszKN-rIQhIcdlc3k?via=maunium.net&via=matrix.org&via=envs.net

Edit: it's true that the inverse behaves kinda strange too. Jumping to a one-line message that is followed by another one-line message can put it very close to the bottom of the screen (depending on your display size/scale, I suppose). I would expect it to be nearer the centre of the screen, not 80% of the way down

PR #3556, opened 2024-09-27
Timeline improvements when navigating with permalinks.

ganfra:
## Content

This PR tries to improve how the timeline behave when there is a click to permalink : 
- waits for indexer to be processed at least once, so we avoid calling focusTimeline if the items is already cached
- tries to center the focused event as much as possible
- avoid focusLive being called too early and weird jumps

This is still not perfect, but works better than previously.

## Motivation and context
Better UX when navigating with permalink.

## Tests

- Open a room
- Click on permalink (try with a non loading event, so you can have the forward pagination).

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenterTest.kt
```diff
@@ -670,7 +670,7 @@ import kotlin.time.Duration.Companion.seconds
         timelineItemIndexer: TimelineItemIndexer = TimelineItemIndexer(),
     ): TimelinePresenter {
         return TimelinePresenter(
-            timelineItemsFactoryCreator = aTimelineItemsFactoryCreator(),
+            timelineItemsFactoryCreator = aTimelineItemsFactoryCreator(timelineItemIndexer),
```
bmarty, line 673: ❤️

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/utils/LazyListState.kt
```diff
@@ -35,3 +37,33 @@ fun LazyListState.isScrollingUp(): Boolean {
...
+    }
+
+    // await for the first layout.
+    scroll { }
+    layoutInfo.resolveItemOffsetToCenter(index)?.let {
+        // Item is already visible, just scroll to center.
+        animateScrollToItem(index, it)
+        return
+    }
+    // Item is not visible, jump to to it centered.
```
bmarty, line 62:
```suggestion
    // Item is not visible, jump to it centered.
```

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
```diff
@@ -135,6 +136,12 @@ class TimelinePresenter @AssistedInject constructor(
                         )
                     } else {
                         newEventState.value = NewEventState.None
+                        if (event.firstIndex == 0) {
+                            val forwardPaginationStatus = timelineController.currentPaginationStatus(Timeline.PaginationDirection.FORWARDS)
+                            if (!forwardPaginationStatus.hasMoreToLoad) {
+                                timelineController.focusOnLive()
+                            }
```
bmarty, line 143: It's weird to get data from the controller, to then ask the controller to do the action.

Just suggesting: maybe the added method `fun currentPaginationStatus` can be removed and this `if` block could be replaced by:
```kotlin
timelineController.maybeFocusOnLive(event.firstIndex)
```

bmarty (approved):
Thanks. The fix for the pinned event navigation when it's the last event is working as expected. No more loader in this case.

frebib:
I tested to see if this fixes #2775 and it's definitely different, but it seems to still behave unintuitively

PR #3562, opened 2024-09-30
Centralise the DI code generation logic

jmartinesp:
## Content

Centralise the DI setup in a single `setupAnvil()` function.

## Motivation and context

- It's easier to maintain.
- It should make the transition to other implementations easier.

## Tests

Build the app & check it runs I guess? 😅 

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/appnav/build.gradle.kts
+++ b/appnav/build.gradle.kts
@@ -8,11 +8,10 @@
 @file:Suppress("UnstableApiUsage")
 
 import extension.allFeaturesApi
+import extension.setupAnvil
 
 plugins {
     id("io.element.android-compose-library")
-    alias(libs.plugins.anvil)
-    alias(libs.plugins.kapt)
     id("kotlin-parcelize")
 }
 
@@ -20,12 +19,9 @@ android {
     namespace = "io.element.android.appnav"
 }
 
-dependencies {
-    implementation(projects.anvilannotations)
-    anvil(projects.anvilcodegen)
-    implementation(libs.dagger)
-    kapt(libs.dagger.compiler)
+setupAnvil()
 
+dependencies {
     allFeaturesApi(rootDir, logger)
 
     implementation(projects.libraries.core)
--- a/plugins/src/main/kotlin/extension/AnvilExtensions.kt
+++ b/plugins/src/main/kotlin/extension/AnvilExtensions.kt
@@ -9,24 +9,48 @@ package extension
 
 import org.gradle.api.Project
 import com.squareup.anvil.plugin.AnvilExtension
+import org.gradle.accessors.dm.LibrariesForLibs
+import org.gradle.kotlin.dsl.the
 
 /**
  * Setup Anvil plugin with the given configuration.
- * @param generateDaggerFactories whether to generate Dagger factories using Anvil
+ * @param generateDaggerCode whether to enable general Dagger code generation using Kapt
+ * @param generateDaggerFactoriesUsingAnvil whether to generate Dagger factories using Anvil instead of Kapt
  */
-fun Project.setupAnvil(generateDaggerFactories: Boolean = true) {
-    plugins.apply("com.squareup.anvil")
-    project.pluginManager.withPlugin("com.squareup.anvil") {
-        extensions.configure(AnvilExtension::class.java) {
-            this.generateDaggerFactories.set(generateDaggerFactories)
+fun Project.setupAnvil(
+    generateDaggerCode: Boolean = false,
+    generateDaggerFactoriesUsingAnvil: Boolean = true,
+) {
+    val libs = the<LibrariesForLibs>()
+    // Apply plugins and dependencies
+    applyPluginIfNeeded("com.squareup.anvil")
 
-            // These dependencies are only needed for compose libraries
-            if (project.pluginManager.hasPlugin("io.element.android-compose-library")) {
-                // Annotations to generate DI code for Appyx nodes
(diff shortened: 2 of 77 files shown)


appnav/build.gradle.kts
```diff
@@ -8,24 +8,20 @@
...
 android {
     namespace = "io.element.android.appnav"
 }
 
-dependencies {
-    implementation(projects.anvilannotations)
-    anvil(projects.anvilcodegen)
-    implementation(libs.dagger)
-    kapt(libs.dagger.compiler)
+setupAnvil()
```
bmarty, line 22: I think you need to pass `generateDaggerCode = true` and `generateDaggerFactoriesUsingAnvil = false` here (because `generateDaggerFactories.set(true)` was not present in the current version of the file)

plugins/src/main/kotlin/extension/AnvilExtensions.kt
```diff
@@ -0,0 +1,56 @@
...
+    generateDaggerFactoriesUsingAnvil: Boolean = true,
+) {
+    val libs = the<LibrariesForLibs>()
+    // Apply plugins and dependencies
+    applyPluginIfNeeded("com.squareup.anvil")
+
+    if (generateDaggerCode) {
+        applyPluginIfNeeded("org.jetbrains.kotlin.kapt")
+        // Needed at the top level since dagger code should be generated at a single point for performance
+        dependencies.add("implementation", libs.dagger)
```
bmarty, line 31: Nit: can maybe use the form
```suggestion
        dependencies.implementation(libs.dagger)
```

plugins/src/main/kotlin/extension/AnvilExtensions.kt
```diff
@@ -0,0 +1,56 @@
...
+        dependencies.add("kapt", libs.dagger.compiler)
+    }
+
+    // These dependencies are only needed for compose library or application modules
+    if (project.pluginManager.hasPlugin("io.element.android-compose-library")
+        || project.pluginManager.hasPlugin("io.element.android-compose-application")) {
+        // Annotations to generate DI code for Appyx nodes
+        dependencies.add("implementation", project.project(":anvilannotations"))
```
bmarty, line 39: Nit: can maybe use the form
```suggestion
        dependencies.implementation(project.project(":anvilannotations"))
```

plugins/src/main/kotlin/extension/AnvilExtensions.kt
```diff
@@ -0,0 +1,56 @@
...
+ * @param generateDaggerCode whether to enable general Dagger code generation using Kapt
+ * @param generateDaggerFactoriesUsingAnvil whether to generate Dagger factories using Anvil instead of Kapt
+ */
+fun Project.setupAnvil(
+    generateDaggerCode: Boolean = false,
+    generateDaggerFactoriesUsingAnvil: Boolean = true,
+) {
+    val libs = the<LibrariesForLibs>()
+    // Apply plugins and dependencies
+    applyPluginIfNeeded("com.squareup.anvil")
```
bmarty, line 26:
```suggestion
    applyPluginIfNeeded(libs.plugins.anvil)
```

with change on `applyPluginIfNeeded` to

```kotlin
private fun Project.applyPluginIfNeeded(plugin: Provider<PluginDependency>) {
    val pluginId = plugin.get().pluginId
    if (!pluginManager.hasPlugin(pluginId)) {
        pluginManager.apply(pluginId)
    }
}
```

plugins/src/main/kotlin/extension/AnvilExtensions.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    if (generateDaggerCode) {
+        applyPluginIfNeeded("org.jetbrains.kotlin.kapt")
```
bmarty, line 29: If you decide to do the change above in you'll need to change here too to

`applyPluginIfNeeded(libs.plugins.kapt)`

bmarty (approved):
Thanks. Only minor remarks that you may ignore, especially because it's on part that will be updated soon.

You should maybe also update this file: https://github.com/element-hq/element-x-android/blob/develop/tools/templates/files/fileTemplates/Template%20Module%20Feature%20Build%20Gradle%20Impl.kts

appnav/build.gradle.kts
jmartinesp: Actually, I think this was a mistake and we were generating the code twice. I removed that and set all the code generation logic in `app` and it seems to be working fine.

plugins/src/main/kotlin/extension/AnvilExtensions.kt
jmartinesp: Actually, this requires us to use this kind of impot:

```
import gradle.kotlin.dsl.accessors._719af408c2a6c8acb7cff251b94eadcc.implementation
```

I think that's quite prone to breaking, a new Gradle version will probably generate a new internal id.

appnav/build.gradle.kts
bmarty: OK, thanks

plugins/src/main/kotlin/extension/AnvilExtensions.kt
bmarty: OK, I did not have to add such import during my test, but maybe I missed something.

Issue #2882, opened 2024-05-20
MSC2530 image captions are used as filenames


frebib:
### Steps to reproduce

1. Send an MSC2530 compatible event with a supported client (e.g. Beeper, or use #2600) with a pretty lengthy caption, enough to hit the "File name too long" error message. In my case it was sent with mautrix-whatsapp with MSC2530 functionality enabled.
2. Look in the app logs

#### What did you expect?
The `body` field` to not be used as part of the filename, but the `filename` field instead (as per MSC2530)

#### What happened instead?
```
2024-05-20T22:04:26.220575Z ERROR elementx: org.matrix.rustcomponents.sdk.ClientException$Generic: msg=File name too long (os error 36) at path "/data/user/0/io.element.android.x/cache/temp/media/.tmppY5zlJ/These words here are the body field of the m.image event.jpeg"
```

### Application version and app store

5d8cc80

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

frebib:
Hit this again today
```
2024-09-30T19:37:45.938083Z ERROR elementx: org.matrix.rustcomponents.sdk.ClientException$Generic: msg=File name too long (os error 36) at path "/data/user/0/io.element.android.x/cache/temp/media/.tmp0f8hs5/2992: **UK Coal**\nThe Watership Down rabbits removed an additional 0.1 nanometers constructing their warren, although that was mostly soil. British rabbits have historically mined very little coal; the sole rabbit-run coal plant was shut down in the 1990s."
        at org.matrix.rustcomponents.sdk.FfiConverterTypeClientError.read(SourceFile:26)
        at org.matrix.rustcomponents.sdk.FfiConverterRustBuffer$DefaultImpls.liftFromRustBuffer(SourceFile:13)
        at org.matrix.rustcomponents.sdk.FfiConverterTypeClientError.liftFromRustBuffer(SourceFile:1)
        at org.matrix.rustcomponents.sdk.FfiConverterRustBuffer$DefaultImpls.lift(SourceFile:6)
        at org.matrix.rustcomponents.sdk.ClientException$ErrorHandler.lift(SourceFile:8)
        at org.matrix.rustcomponents.sdk.Matrix_sdk_ffiKt.access$uniffiCheckCallStatus(SourceFile:123)
        at org.matrix.rustcomponents.sdk.Matrix_sdk_ffiKt.uniffiRustCallAsync(SourceFile:194)
[...]

PR #3567, opened 2024-09-30
Use MSC2530 filename when loading media

frebib:
## Content

In the case of captioned images, the `body` of the event is the caption, and is ill suited to be a filename. Often the caption can be prohibitively long, causing the media to fail to load in some cases.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/2882

## Tests

- Step 1: Try to open an image- receive `An error has occurred` as per https://github.com/element-hq/element-x-android/issues/2882#issuecomment-2384009582
- Step 2: Add this patch and note that the media now loads with no errors

To reproduce: Send `!xkcd 2992` to https://matrix.to/#/@xkcd:spritsail.io and observe that the media will fail to load

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 14

## Checklist

- [x] Changes have been tested on an Android device or Android emulator with API 23


bmarty (approved):
Thanks for the fix! I think there is some cleanup to do in the model, now that we may have a caption, but I can handle that later.

PR #3574, opened 2024-10-01
Clarify model for Event with attachment

bmarty:
## Content

Clarify model for Event with attachment. They must have a `filename: String` and they may have a `body :String?` (caption).

If filename is missing from the SDK model, the body is used instead and in this case the body is not mapped to our model.

## Motivation and context

Cleanup following change in #3567 
The changes here should reduce the risk of errors.

## Tests

- Should not have any impact.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesFlowNode.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesFlowNode.kt
@@ -50,7 +50,6 @@ import io.element.android.features.poll.api.create.CreatePollMode
 import io.element.android.libraries.architecture.BackstackWithOverlayBox
 import io.element.android.libraries.architecture.BaseFlowNode
 import io.element.android.libraries.architecture.createNode
-import io.element.android.libraries.architecture.inputs
 import io.element.android.libraries.architecture.overlay.Overlay
 import io.element.android.libraries.architecture.overlay.operation.show
 import io.element.android.libraries.di.RoomScope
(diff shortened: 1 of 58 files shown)


ganfra (commented):
I think this is the opposite, from reading the spec? 


`body | string | Required: If filename is not set or the value of both properties are identical, this is the filename of the original upload. Otherwise, this is a caption for the image.Changed in v1.10: This property can act as a caption for the image.`


`filename | string | The original filename of the uploaded file.Added in v1.10
`

The body is mandatory but the filename is not

bmarty:
We are just mapping what the SDK is providing to the application here. There is maybe a problem in the SDK then?

ganfra:
> We are just mapping what the SDK is providing to the application here. There is maybe a problem in the SDK then?

Mmm AFAICT, the SDK returns a non-nullable body and a nullable filename?

bmarty:
Yes, I have read it too fast, my bad

The mapping I have done [in the EventContentMapper](https://github.com/element-hq/element-x-android/pull/3574/files#diff-9b8993ef173a404ecbf23a0c01d42502c4e95379978611aa95c3e9fc7785b0a5R53-R54) is still correct to me, and the application needs a non-nullable filename.

So we always have a `filename`, which may be from the `body` and we have an optional `body` which is actually either the `filename`, or the `caption`. Do you see a problem with this approach?

ganfra:
I think we should keep the mapping as they come from sdk, otherwise it brings confusion.
I'd instead offer some methods to get the correct value (or computed val),
like `computedFilename()` and `computedCaption()`

bmarty:
OK, let's wait for the SDK to change, I think this is discussed in #2570.

pixlwave:
Heads up that this is now available in the SDK: https://github.com/matrix-org/matrix-rust-sdk/pull/4081

`mediaTimelineItem.filename` will always be the filename and `mediaTimelineItem.caption`/`formattedCaption` will only exist when there's a caption to show.

bmarty:
@ganfra PR updated with what is provided in https://github.com/matrix-org/matrix-rust-sdk/pull/4081

bmarty:
@ganfra do you love the update of this PR?

PR #3576, opened 2024-10-01
Prevent crash with duplicate room suggestion

frebib:
## Content

If two rooms have the same canonical alias, e.g. the recently tombstoned and upgraded #foundation-office:matrix.org, typing `#foun` into the composer could cause the app to crash. This avoids that by using the (hopefully) globally unique room id as the key instead.

 

## Tests

- Step 1: Join #element-x-android:matrix.org and #definitely-matrix-hq:maunium.net, both of which have the canonical alias #element-x-android:matrix.org
- Step 2: Type `#ele` into the composer and watch the app crash

With this patch, the app does not crash

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Whatever Pixel 8 runs

## Checklist


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/suggestions/SuggestionsPickerView.kt
```diff
@@ -60,7 +60,7 @@ fun SuggestionsPickerView(
                 when (suggestion) {
                     is ResolvedSuggestion.AtRoom -> "@room"
                     is ResolvedSuggestion.Member -> suggestion.roomMember.userId.value
-                    is ResolvedSuggestion.Alias -> suggestion.roomAlias.value
+                    is ResolvedSuggestion.Alias -> suggestion.roomSummary.roomId
```
bmarty, line 63: For parity with the `Member` case above, maybe use the `.value`? But we could also remove the `.value`, for the `Member` case. Anyway this is still OK!

bmarty (approved):
Thanks for the fix!
Worth noting that to observe the crash, user need to enable room suggestions in the developer settings, since we do not want to enable the feature if EXI does not support it.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/suggestions/SuggestionsPickerView.kt
frebib: Right, added `.value` to get a comparable key 👍🏼

bmarty (approved):

Issue #3605, opened 2024-10-04
Remove (or improve) slidinc sync proxy check


bmarty:
Now the the homeserver can support native sliding sync, a slidinc sync proxy is not required.

Need to remove (or adapt) the following check which may block users:

<img width="292" alt="image" src="https://github.com/user-attachments/assets/4f6ed6d3-cf0e-4f7d-b17b-64637bc1bb2d">

bmarty:
Handled in #3609

PR #3609, opened 2024-10-04
Remove supportSlidingSync boolean.

bmarty:
Note that this field was not used (read) anymore, so this is just cleaning up the code.

Context: #3605.


jmartinesp (approved):
I confirmed we're no longer using this value anywhere, so it should be safe to delete it, but just to double check: this is because the Rust SDK now handles this logic, right?

bmarty:
> I confirmed we're no longer using this value anywhere, so it should be safe to delete it, but just to double check: this is because the Rust SDK now handles this logic, right?

Yes, I guess the SDK is now handlong the logic, else case in #3605 could not happen anymore.

Previously the application was doing the check, I have not done any research to find when this code became dead code.

PR #3617, opened 2024-10-07
Ensure the CI compile and execute all the unit tests.

bmarty:
I have noticed that we have test compilation failure that was not detected by the CI.
The first commit should fix this and test execution should not be longer.
Second commit will fix the existing issue.


jmartinesp (approved):
LGTM!

PR #3621, opened 2024-10-08
Warn the user when unverified user has changed their identity

bmarty:
## Content

As per https://github.com/element-hq/element-meta/issues/2544

Using https://github.com/matrix-org/matrix-rust-sdk/pull/4022 and https://github.com/matrix-org/matrix-rust-sdk/pull/4068

Draft since the application is crashing for now when clicking on OK to pin the user. **EDIT** The crash has been fixed in https://github.com/matrix-org/matrix-rust-sdk/pull/4089 so an SDK release will be neeeded, but this is not blocking this PR anymore (pinning identity cannot be tested, but the PR can be reviewed).

## Motivation and context

https://github.com/element-hq/element-meta/issues/2544

## Tests

- Alice and Bob in a e2e room
- Alice has not verified Bob
- Messages can be sent
- Bob reset their identity
- Alice should see the banner, until they click on OK to pin the identity of Bob (but currently this is crashing)

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesNode.kt
```diff
@@ -135,16 +135,20 @@ class MessagesNode @AssistedInject constructor(
                 callbacks.forEach { it.onUserDataClick(permalink.userId) }
             }
             is PermalinkData.RoomLink -> {
-                handleRoomLinkClick(permalink, eventSink)
+                handleRoomLinkClick(activity, permalink, eventSink)
             }
             is PermalinkData.FallbackLink,
             is PermalinkData.RoomEmailInviteLink -> {
-                context.openUrlInExternalApp(url)
+                activity.openUrlInChromeCustomTab(null, darkTheme, url)
```
bmarty, line 142: Note: at some point, direct usage of `openUrlInExternalApp` should be forbidden, but this is out of scope of this PR.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/identity/IdentityChangeStatePresenter.kt
```diff
@@ -0,0 +1,102 @@
...
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
+import javax.inject.Inject
+
+class IdentityChangeStatePresenter @Inject constructor(
+    private val room: MatrixRoom,
+) : Presenter<IdentityChangeState> {
+    @Composable
+    override fun present(): IdentityChangeState {
+        val roomMemberIdentityStateChange = remember {
```
jmartinesp, line 42: Couldn't this just use `produceState`?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/identity/IdentityChangeStateView.kt
```diff
@@ -0,0 +1,78 @@
...
+import io.element.android.libraries.matrix.ui.model.getAvatarData
+import io.element.android.libraries.ui.strings.CommonStrings
+
+@Composable
+fun IdentityChangeStateView(
+    state: IdentityChangeState,
+    modifier: Modifier = Modifier,
+) {
+    // Pick the first identity change to PinViolation
+    val identityChange = state.roomMemberIdentityStateChanges.firstOrNull {
```
jmartinesp, line 35: Maybe rename it to something like `dangerousIdentityChange` or `pinViolationIdentityChange`?

jmartinesp (approved):
LGTM, just a couple of suggestions.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/identity/IdentityChangeStatePresenter.kt
ganfra: Yes this would make more sense. Also use `by` delegate when possible?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/identity/IdentityChangeStatePresenter.kt
```diff
@@ -0,0 +1,102 @@
...
+                is IdentityChangeEvent.Submit -> {
+                    ignoredUserIdChange.value += event.userId
+                    // TODO notify the SDK
+                }
+            }
+        }
+
+        return IdentityChangeState(
+            roomMemberIdentityStateChanges = roomMemberIdentityStateChange.value
+                .filter { it.roomMember.userId !in ignoredUserIdChange.value }
```
ganfra: I guess the list will never be large, but we are filtering on the main thread for every composition ^

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/identity/IdentityChangeState.kt
```diff
@@ -0,0 +1,22 @@
...
+import io.element.android.libraries.matrix.api.room.RoomMember
+import kotlinx.collections.immutable.ImmutableList
+
+data class IdentityChangeState(
+    val roomMemberIdentityStateChanges: ImmutableList<RoomMemberIdentityStateChange>,
+    val eventSink: (IdentityChangeEvent) -> Unit,
+)
+
+data class RoomMemberIdentityStateChange(
+    val roomMember: RoomMember,
```
ganfra, line 20: Could we create a specific data class instead of passing `RoomMember` if we only need a few fields?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/identity/IdentityChangeStatePresenter.kt
bmarty: This code does not exist anymore, are you reviewing an old version?

bmarty:
@ganfra @jmartinesp thanks for the fast review. I think I have handled all your remarks.
I took the opportunity to update the code regarding typing notification as it had similar problems.

PR #3635, opened 2024-10-09
When creating an account with MAS, change the `prompt` parameter to `create`

### LINKED ISSUE #3627: Send the user to oidc register page when creating account

Send the user directly to the register page on an oidc enabled homeserver when going through the create account flow

So the user doesn't have to take an extra step to create an account after already expressing their intent to create an account

Add `prompt=create` to the authorization url

_No response_

### Additional context

_No response_

No

bmarty:
## Content

When user is creating account, load the url page directly in the account creation mode by changing (or adding) the `prompt` url parameter to `create`.

@sandhose does it make sense to you?

## Motivation and context

Improve user experience with MAS and closes #3627

## Screenshots / GIFs

|Before|After|
|-|-|
|<video width="98" alt="image" src="https://github.com/user-attachments/assets/ff90a0aa-a776-43fd-ad15-eb2160c91e78" />|<video width="98" alt="image" src="https://github.com/user-attachments/assets/e7d8f605-9f18-45d6-bd29-69ee597f2272"/>|

## Tests

- Try creating an account on a server with MAS support, such as beta.matrix.org and observe that you're having a wonderful navigation experience.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/uri/UriExtensions.kt
+++ b/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/uri/UriExtensions.kt
@@ -12,3 +12,16 @@ import android.net.Uri
 const val IGNORED_SCHEMA = "ignored"
 
 fun createIgnoredUri(path: String): Uri = Uri.parse("$IGNORED_SCHEMA://$path")
+
+fun Uri.setQueryParameter(key: String, value: String): Uri {
+    val existingParams = queryParameterNames
+    return buildUpon().apply {
+        clearQuery()
+        existingParams.forEach { existingKey ->
+            if (existingKey != key) {
+                appendQueryParameter(existingKey, getQueryParameter(existingKey))
+            }
+        }
+        appendQueryParameter(key, value)
+    }.build()
+}
--- a/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/confirmaccountprovider/ConfirmAccountProviderPresenter.kt
+++ b/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/confirmaccountprovider/ConfirmAccountProviderPresenter.kt
@@ -15,6 +15,7 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
+import androidx.core.net.toUri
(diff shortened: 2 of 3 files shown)


libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/uri/UriExtensions.kt
```diff
@@ -12,3 +12,16 @@ import android.net.Uri
...
+    return buildUpon().apply {
+        clearQuery()
+        existingParams.forEach { existingKey ->
+            if (existingKey != key) {
+                appendQueryParameter(existingKey, getQueryParameter(existingKey))
+            }
+        }
+        appendQueryParameter(key, value)
+    }.build()
+}
```
bmarty, line 27: Only `appendQueryParameter` exists in `Uri.Builder`, so sadly, I had to create this extension.

sandhose (commented):
This is right, although I'm surprised the SDK doesn't surface this parameter?

There is a setter here: https://github.com/matrix-org/matrix-rust-sdk/blob/9c6413551cd6e15f01453e57178fe98e0d12200b/crates/matrix-sdk/src/oidc/auth_code_builder.rs#L77-L80

Which is being used here: https://github.com/matrix-org/matrix-rust-sdk/blob/9c6413551cd6e15f01453e57178fe98e0d12200b/crates/matrix-sdk/src/oidc/mod.rs#L473

And exposed in the bindings here: https://github.com/matrix-org/matrix-rust-sdk/blob/9c6413551cd6e15f01453e57178fe98e0d12200b/bindings/matrix-sdk-ffi/src/client.rs#L367

I feel like there should be an `url_for_oidc_registration` equivalent in the SDK?

bmarty:
OK so maybe we could improve the existing Rust API, to replace `urlForOidcLogin()` by `urlForOidc(prompt: String)`, or simply add an API `urlForOidcRegister()`

@jmartinesp do you think this would be possible?

jmartinesp:
From a very quick look at the code, it seems feasible, although I'd rather make the `Prompt` an enum in the FFI layer so the client can't just pass anything. What do you think?

bmarty:
Yes, an enum would be better than a String. Thanks!

bmarty:
Closing this we will get a better way to handle this thanks to Jorge's work in https://github.com/matrix-org/matrix-rust-sdk/pull/4131.

PR #3642, opened 2024-10-10
Improve avatar rendering

ganfra:
## Content
This PR takes care of using a single thumbnail request size, so caching works properly.
Indeed, according to the matrix specs : 

> Servers SHOULD produce thumbnails with the following dimensions and methods:
> 32x32, crop
> 96x96, crop
> 320x240, scale
> 640x480, scale
> 800x600, scale

For rendering avatar we could use crop, but it's not available in the rust sdk yet.
So instead, always use the smallest scale function ie. requesting with a size <= 240 pixels would makes sure we do.

This way, we'll be hitting the server only once per mxc url, and not for each size we've defined in our AvatarSize.

## Motivation and context
Improve rendering by removing useless network calls.

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


jmartinesp (approved):
LGTM, thanks!

PR #3645, opened 2024-10-10
Cleanup tests

### LINKED ISSUE #3643: Moderation: kicking (removing) member of a room is broken
Try to remove someone from a room.
The operation succeed, but the list of user is not refreshed so user can be removed again and this is leading to a infinite loading banner.

bmarty:
## Content

Cleanup test.
Make AsyncAction.Confirming having a field to be able to reduce the number of recomposition. This is impacting lots of file, but I did not find an easy way to let `C` be `Unit` by default.
Rework Moderation presenter to make it a bit more live.

## Motivation and context

Cleaner code.

## Tests

- Play with the room member moderation screen. Note that this does not fix #3643, I did not find a reliable way to fix that. We need a live room member list from the SDK.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/AsyncAction.kt
```diff
@@ -15,23 +15,27 @@ import kotlin.contracts.contract
 
 /**
  * Sealed type that allows to model an asynchronous operation triggered by the user.
+ * @param C the type of data required to confirm the operation.
+ * @param T the type of data returned by the operation.
  */
 @Stable
-sealed interface AsyncAction<out T> {
+sealed interface AsyncAction<out C, out T> {
```
ganfra, line 22: Maybe instead of adding new type everywhere, we could let the Confirming be an interface you can inherits only you have data to confirm? 


```
interface Confirming : AsyncAction<Nothing>
data object SimpleConfirmation : Confirming
```

```
val unbanUserAsyncAction: AsyncAction<Unit>
data class UnbanUserActionConfirmation(val member: RoomMember): Confirming
```

ganfra (commented):
This will change number of recomposition for tests, but in real life it won't change anything as it's not using the same composition clock between app and tests

bmarty:
I will split this random PR into smaller ones.

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/AsyncAction.kt
bmarty: Handled in #3667.

PR #3667, opened 2024-10-14
AsyncAction confirming with param

bmarty:
## Content

- Let AsyncAction.Confirming be an interface, with a `AsyncAction.ConfirmingNoParams` data object.
This will allow inheritance of `AsyncAction.Confirming` with parameter(s).
- AcceptDeclineInviteState: create ConfirmingDeclineInvite to host InviteData when confirming decline of invite.

Replacement for #3645 (other change from #3645 will be in separate PRs.)
Can be reviewed commit per commit

## Motivation and context

Improve code when confirmation need data to be rendered.

## Tests

- No behavior change.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


features/invite/impl/src/main/kotlin/io/element/android/features/invite/impl/response/AcceptDeclineInviteView.kt
```diff
@@ -45,13 +45,13 @@ fun AcceptDeclineInviteView(
             onErrorDismiss = {
                 state.eventSink(InternalAcceptDeclineInviteEvents.DismissDeclineError)
             },
-            confirmationDialog = {
-                val invite = state.invite.getOrNull()
-                if (invite != null) {
+            confirmationDialog = { confirming ->
+                // Note: confirming will always be of type ConfirmingDeclineInvite.
+                if (confirming is ConfirmingDeclineInvite) {
```
jmartinesp, line 50: Can't we just force cast it then?

jmartinesp (approved):
LGTM, just a comment.

features/invite/impl/src/main/kotlin/io/element/android/features/invite/impl/response/AcceptDeclineInviteView.kt
bmarty: We could but why take the risk of a crash?
Previous code was not `val invite = state.invite!!`, so I prefer to keep the code safe.

jmartinesp: > We could but why take the risk of a crash?

Can we then log an error or upload some error report to sentry? I'm worried this might just silently fail, although the tests should cover it.

bmarty: My [previous attempt](https://github.com/element-hq/element-x-android/pull/3645/commits/40d82d62a1c06cfd35992a2efdc3e048867f89e4) to fix this would have avoid this drawback. But it was too invasive, so we go with this approach. This is not a new issue, so I guess this is fine.

PR #3673, opened 2024-10-14
Improve preview provider name consistency

bmarty:
## Content

Rename some classes which implement PreviewParameterProvider according to the new naming convention.

`Provider` suffix is enough and more used than `PreviewProvider`, so let's make the codebase more consistent.

## Tests

- There should ne no change in the screenshot nor in the application behavior.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


ganfra (approved):
I think I'd prefer to use `PreviewProvider` instead, but as you want.

bmarty:
> I think I'd prefer to use `PreviewProvider` instead, but as you want.

Yes, that's probably better but it would require to change much more classes. Also those classes do not provide Preview but States, so naming them `...StateProvider` make also some sense...

PR #3650, opened 2024-10-10
Format RoomListRoomSummary timestamps in real time

jmartinesp:
## Content

- Make `RoomListRoomSummary.timestamp` a `Date?` instead of a `String?`.
- Add `RoomListContentState.Rooms.formatTimestamp` and `RoomListState.formatTimestamp` lambdas: these lambdas use `RoomLastMessageFormatter` under the hood to format any provided dates in an inexpensive way, and the same instance is reused, avoiding unnecessary expensive instantiations.
- Modify the timestamps provided for the previews.

## Motivation and context

There is a bug where in the room list you might be able to see rooms whose state hasn't changed since yesterday still displaying the time at which some event happened instead of 'Yesterday', making you think they happened at some time today and they're out of order. 

If you open one of these rooms, the room list item is also refreshed and the right 'Yesterday' formatted timestamp is used.

## Screenshots / GIFs

There are a few changes in the previews for room list items. I've tried to keep at least some with the same values.

## Tests

- Open the app until the room list loads.
- Change the system's date in the device settings to some other date or tomorrow, i.e.
- Go back to the app and check the timestamps have changed.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 23
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [ ] Pull request is based on the develop branch
- [ ] Pull request title will be used in the release note, it clearly define what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
- [ ] You've made a self review of your PR

PR #3683, opened 2024-10-15
Refresh room summaries when date or time changes in the device

jmartinesp:
## Content

- Create `DateTimeObserver`, which will register a receiver for date and time changes in the devices and emit updates.
- Have `RoomListDataSource` listen to those updates.
- Refresh the summary timestamps when an update is received.
- Modify the DI so we have a `TimezoneProvider` instead of injecting the `Timezone` used when the app was opened.

Alternative to https://github.com/element-hq/element-x-android/pull/3650.

## Motivation and context

There is a bug where in the room list you might be able to see rooms whose state hasn't changed since yesterday still displaying the time at which some event happened instead of 'Yesterday', making you think they happened at some time today and they're out of order.

If you open one of these rooms, the room list item is also refreshed and the right 'Yesterday' formatted timestamp is used.

The same happens with the time if you change the timezone of your device.

## Screenshots / GIFs

They should be the same.

## Tests

- Open the app and let the room list load completely until it doesn't receive new updates.
- Go to system settings -> change the day. Check the timestamps change in the room list.
- Restore the current value, check it changes back to what's expected.
- Now change the timezone and check the time also changes in the room list.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/roomlist/impl/src/test/kotlin/io/element/android/features/roomlist/impl/RoomListPresenterTest.kt
+++ b/features/roomlist/impl/src/test/kotlin/io/element/android/features/roomlist/impl/RoomListPresenterTest.kt
@@ -22,7 +22,7 @@ import io.element.android.features.logout.api.direct.aDirectLogoutState
 import io.element.android.features.networkmonitor.api.NetworkMonitor
 import io.element.android.features.networkmonitor.test.FakeNetworkMonitor
 import io.element.android.features.roomlist.impl.datasource.RoomListDataSource
-import io.element.android.features.roomlist.impl.datasource.RoomListRoomSummaryFactory
+import io.element.android.features.roomlist.impl.datasource.aRoomListRoomSummaryFactory
 import io.element.android.features.roomlist.impl.filters.RoomListFiltersState
 import io.element.android.features.roomlist.impl.filters.aRoomListFiltersState
 import io.element.android.features.roomlist.impl.model.createRoomListRoomSummary
@@ -651,7 +651,7 @@ class RoomListPresenterTest {
         leaveRoomPresenter = { leaveRoomState },
         roomListDataSource = RoomListDataSource(
             roomListService = client.roomListService,
-            roomListRoomSummaryFactory = RoomListRoomSummaryFactory(
+            roomListRoomSummaryFactory = aRoomListRoomSummaryFactory(
                 lastMessageTimestampFormatter = lastMessageTimestampFormatter,
                 roomLastMessageFormatter = roomLastMessageFormatter,
             ),
--- a/features/roomlist/impl/src/test/kotlin/io/element/android/features/roomlist/impl/RoomListPresenterTest.kt
+++ b/features/roomlist/impl/src/test/kotlin/io/element/android/features/roomlist/impl/RoomListPresenterTest.kt
@@ -29,6 +29,7 @@ import io.element.android.features.roomlist.impl.model.createRoomListRoomSummary
 import io.element.android.features.roomlist.impl.search.RoomListSearchEvents
 import io.element.android.features.roomlist.impl.search.RoomListSearchState
 import io.element.android.features.roomlist.impl.search.aRoomListSearchState
+import io.element.android.libraries.androidutils.system.DateTimeObserver
 import io.element.android.libraries.architecture.Presenter
 import io.element.android.libraries.dateformatter.api.LastMessageTimestampFormatter
 import io.element.android.libraries.dateformatter.test.A_FORMATTED_DATE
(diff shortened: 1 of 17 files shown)


libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/diff/DiffCache.kt
```diff
@@ -24,6 +24,7 @@ interface MutableDiffCache<E> : DiffCache<E> {
     fun removeAt(index: Int): E?
     fun add(index: Int, element: E?)
     operator fun set(index: Int, element: E?)
+    fun clear()
```
bmarty, line 27: Not used?

libraries/dateformatter/impl/src/main/kotlin/io/element/android/libraries/dateformatter/impl/di/DateFormatterModule.kt
```diff
@@ -25,5 +26,5 @@ object DateFormatterModule {
     fun providesLocale(): Locale = Locale.getDefault()
 
     @Provides
-    fun providesTimezone(): TimeZone = TimeZone.currentSystemDefault()
+    fun providesTimezone(): TimezoneProvider = TimezoneProvider { TimeZone.currentSystemDefault() }
```
bmarty, line 29: I am wondering if requesting `TimeZone.currentSystemDefault()` for all Events could come at a cost. We could maybe have a flow of  `TimeZone` that emit new item when the timezone change?

features/roomlist/impl/src/test/kotlin/io/element/android/features/roomlist/impl/RoomListPresenterTest.kt
```diff
@@ -672,3 +675,11 @@ class RoomListPresenterTest {
         logoutPresenter = { aDirectLogoutState() },
     )
 }
+
+class FakeDateTimeObserver : DateTimeObserver {
+    override val changes = MutableSharedFlow<DateTimeObserver.Event>()
+
+    fun given(event: DateTimeObserver.Event) {
```
bmarty, line 682: This fun is not used, maybe add a small test?

bmarty (commented):
A few remarks.

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/system/DateTimeObserver.kt
```diff
@@ -0,0 +1,59 @@
...
+    val changes: Flow<Event>
+
+    sealed interface Event {
+        data object TimeZoneChanged : Event
+        data class DateChanged(val previous: Instant, val new: Instant) : Event
+    }
+}
+
+@ContributesBinding(AppScope::class)
+class DefaultDateTimeObserver @Inject constructor(
```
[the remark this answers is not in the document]
jmartinesp: Nice catch, I completely forgot to add the `@SingleIn` annotation.

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/system/DateTimeObserver.kt
```diff
@@ -0,0 +1,59 @@
...
+        private var lastTime = Instant.now()
+
+        override fun onReceive(context: Context, intent: Intent) {
+            val newDate = Instant.now()
+            when (intent.action) {
+                Intent.ACTION_TIMEZONE_CHANGED -> changes.tryEmit(Event.TimeZoneChanged)
+                Intent.ACTION_DATE_CHANGED -> changes.tryEmit(Event.DateChanged(lastTime, newDate))
+                Intent.ACTION_TIME_CHANGED -> changes.tryEmit(Event.DateChanged(lastTime, newDate))
+            }
+            lastTime = newDate
```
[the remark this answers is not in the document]
jmartinesp: I added them in case we wanted to generate new timestamps only if `>= 1 day` had elapsed between them, but there's not really a need for that now that I think about it... they might not be worth keeping.

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/diff/DiffCache.kt
jmartinesp: It's a leftover from previous tests that I forgot to remove, thanks!

libraries/dateformatter/impl/src/main/kotlin/io/element/android/libraries/dateformatter/impl/di/DateFormatterModule.kt
jmartinesp: There might be some tiny overhead from instantiating the value, but I don't think it's noticeable at all. The underlying code is cloning an existing value AFAICT, then just wrapping it in a new object. Reusing the observer for this feels like an overkill, to be honest.

bmarty (approved):
Thanks for the update! I took the liberty to add a few commits to improve the test.

PR #3685, opened 2024-10-15
Call: ensure that the microphone is working when the application is backgrounded.

### LINKED ISSUE #3679: Bluetooth mic muted during call if the app is not in foreground and the screen is on
### Steps to reproduce

When calling using a Bluetooth headset I must keep Element-X in foreground and keep the screen on. If I turn it off or background the app my mic goes silent.

### Outcome

The mic is muted as soon as the screen is off or the app is in background without picture in picture.

### Your phone model

_No response_

### Operating system version

Android 14

0.6.4

### Homeserver

_No response_

### Will you send logs?

Yes

No

bmarty:
## Content

Request audio focus only when the call is actually started and start the foreground service as soon as the call is started.
Second commit change the service permission. We do not really need the permission `MANAGE_OWN_CALLS` but are rather interested by `FOREGROUND_SERVICE_MICROPHONE`. 

## Motivation and context

Close #3679 

## Tests

- Start a call
- put the app in BG
- Switch the screen of.

the call should stay active, with a working microphone

## Tested devices

- [x] Physical running Android 14.
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/services/CallForegroundService.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/services/CallForegroundService.kt
@@ -7,9 +7,11 @@
 
 package io.element.android.features.call.impl.services
 
+import android.Manifest
 import android.app.Service
 import android.content.Context
 import android.content.Intent
+import android.content.pm.PackageManager
 import android.content.pm.ServiceInfo
 import android.os.Build
 import android.os.IBinder
@@ -33,8 +35,12 @@ import timber.log.Timber
 class CallForegroundService : Service() {
     companion object {
         fun start(context: Context) {
-            val intent = Intent(context, CallForegroundService::class.java)
-            ContextCompat.startForegroundService(context, intent)
+            if (ContextCompat.checkSelfPermission(context, Manifest.permission.RECORD_AUDIO) == PackageManager.PERMISSION_GRANTED) {
+                val intent = Intent(context, CallForegroundService::class.java)
+                ContextCompat.startForegroundService(context, intent)
+            } else {
+                Timber.w("Microphone permission is not granted, cannot start the call foreground service")
+            }
         }
 
         fun stop(context: Context) {
@@ -67,8 +73,8 @@ class CallForegroundService : Service() {
             .setContentIntent(pendingIntent)
             .build()
         val notificationId = NotificationIdProvider.getForegroundServiceNotificationId(ForegroundServiceType.ONGOING_CALL)
-        val serviceType = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
-            ServiceInfo.FOREGROUND_SERVICE_TYPE_PHONE_CALL
+        val serviceType = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+            ServiceInfo.FOREGROUND_SERVICE_TYPE_MICROPHONE
         } else {
             0
         }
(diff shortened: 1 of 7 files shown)


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
```diff
@@ -115,6 +120,11 @@ class ElementCallActivity :
         }
     }
 
+    private fun setCallIsActive() {
+        requestAudioFocus()
+        CallForegroundService.start(this)
```
jmartinesp, line 125: Don't we ever stop the foreground service?

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
```diff
@@ -180,6 +180,7 @@ class CallScreenPresenter @AssistedInject constructor(
             urlState = urlState.value,
             webViewError = webViewError,
             userAgent = userAgent,
+            isCallActive = isJoinedCall,
```
jmartinesp, line 183: Have you checked this also works for URL-based calls? I'm not sure `isJoinedCall` is ever set to `true` for those.

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/services/CallForegroundService.kt
```diff
@@ -33,8 +35,12 @@ import timber.log.Timber
 class CallForegroundService : Service() {
     companion object {
         fun start(context: Context) {
-            val intent = Intent(context, CallForegroundService::class.java)
-            ContextCompat.startForegroundService(context, intent)
+            if (ContextCompat.checkSelfPermission(context, Manifest.permission.RECORD_AUDIO) == PackageManager.PERMISSION_GRANTED) {
```
jmartinesp, line 38: Really weird corner case, but what happens if the user didn't enable the microphone permission and joined the room anyway? Will the audio from the other side still be heard, even when the device is locked? I assume it would work for the app just being in background since the PiP window is alive.

jmartinesp (commented):
There are a few things I'd like to double check.

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
bmarty: Yes, I kept the existing code in `onDestroy()`

jmartinesp: Oh, I didn't see it in the diff, so I assumed it wasn't there 🤦 . Thanks!

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
bmarty: https://github.com/element-hq/element-x-android/pull/3685/commits/cf80b3b804fe99363d5a988f2295f92ddb04f56a

jmartinesp (approved):
Thanks for the changes, the feature now seems a lot more stable and useful!

jmartinesp:
cc @toger5 since he was especially interested in this. He might also be a good early tester 😛 .

davidegirardi:
I tested the latest nightly. It failed on the very first call but then worked with no issues, even if I revoked the picture in picture permissions.

PR #3694, opened 2024-10-16
OIDC prompt

bmarty:
## Content

SDK 0.1.54 and fix all the API break.

Then last commit fixes #3635 (so this PR replaces #3635)

## Motivation and context

Use latest SDK and fix a UX issue.

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/timeline/item/event/EventTimelineItem.kt
```diff
@@ -46,10 +45,7 @@ data class EventTimelineItem(
     }
 }
 
-fun interface EventDebugInfoProvider {
-    fun get(): TimelineItemDebugInfo
-}
-
-fun interface EventShieldsProvider {
```
jmartinesp, line 53: We could still keep these, wrapping the lazy provider. That way we could also use the lambdas for the fakes.

bmarty: Good point, let me give it a try

bmarty: 0db54df877a80fc90ee1d910a9dfee2d60f7553a

pixlwave:
Question: Have we exposed the well-known check on the auth server to know which `prompt` values are accepted? AIUI Not all servers will accept `Create` and so we should show an error in the registration flow at this stage.

bmarty:
> Question: Have we exposed the well-known check on the auth server to know which `prompt` values are accepted? AIUI Not all servers will accept `Create` and so we should show an error in the registration flow at this stage.

I guess in this case the webpage will render a sign in page?

pixlwave:
> I guess in this case the webpage will render a sign in page?

I'm unsure, presumably element.io should have registration disabled, might be visible if you try it on there.

bmarty:
On element.io I got an url like this:

https://mas.element.io/authorize?response_type=code&client_id=01JACN03EKTTDEMNR5RMGFHM74&redirect_uri=io.element%3A%2Fcallback&scope=openid+urn%3Amatrix%3Aorg.matrix.msc2967.client%3Aapi%3A*+urn%3Amatrix%3Aorg.matrix.msc2967.client%3Adevice%3A1YBo5FBngj&state=EUJALoTFcMFPQIt7&nonce=XFXnEgUpteENotlr&prompt=create&code_challenge_method=S256&code_challenge=jzS7e6gU-q3xavcM2Jhq4wyoiGd5881L1d8G-qa4Z7o

And it is rendered like this on mobile:

image

So I guess the prompt param (set to `create`) is correctly "ignored".

So I think it's safe to merge the PR. Thanks for making me do the extra test though!

Issue #3684, opened 2024-10-15
[Task] Show in user's profile when they have been verified


mxandreas:
Please see and discuss the details in the meta issue.

PR #3700, opened 2024-10-17
Remove duplicated code regarding user (room member and user profile) screens

bmarty:
## Content

Let `RoomMemberDetailsPresenter` delegates the computation of profile state to `UserProfilePresenter`, which allow to reduce code duplication.

## Motivation and context

Less code duplication (production and test).

Preparatory work for #3684

## Tests

There should be no change.
- Navigate to some room member details page
- Open matrix user profile, with member not in room (can be done using adb)

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


features/userprofile/impl/src/main/kotlin/io/element/android/features/userprofile/impl/DefaultUserProfilePresenterFactory.kt
```diff
@@ -0,0 +1,24 @@
...
+import io.element.android.libraries.di.SessionScope
+import io.element.android.libraries.matrix.api.core.UserId
+import javax.inject.Inject
+
+@ContributesBinding(SessionScope::class)
+class DefaultUserProfilePresenterFactory @Inject constructor(
+    private val factory: UserProfilePresenter.Factory,
+) : UserProfilePresenterFactory {
+    override fun create(userId: UserId): Presenter<UserProfileState> = factory.create(userId)
+}
```
bmarty, line 24: This double factory level is a bit weird, there is maybe a better way to handle this. Any ideas @ganfra @jmartinesp ?

jmartinesp: I can't see any, sorry... I was going to suggest trying to reuse `UserProfilePresenter.Factory` but that's not possible since the `AssistedFactory` needs to return a concrete type.

The best I could get was making `UserProfilePresenter.Factory` implement `UserProfilePresenterFactory`, then use it in `DefaultUserProfilePresenterFactory` like this:

```kotlin
@ContributesBinding(SessionScope::class)
class DefaultUserProfilePresenterFactory @Inject constructor(
    private val factory: UserProfilePresenter.Factory,
) : UserProfilePresenterFactory by factory
```

jmartinesp (approved):
LGTM, it's a shame we can't simplify it anymore.

features/userprofile/impl/src/main/kotlin/io/element/android/features/userprofile/impl/DefaultUserProfilePresenterFactory.kt
bmarty: Thanks for looking into it. Let's keep it like this then!

features/userprofile/impl/src/main/kotlin/io/element/android/features/userprofile/impl/root/UserProfilePresenter.kt
```diff
@@ -47,10 +49,23 @@ class UserProfilePresenter @AssistedInject constructor(
         fun create(userId: UserId): UserProfilePresenter
     }
 
-    private val userProfilePresenterHelper = UserProfilePresenterHelper(
-        userId = userId,
-        client = client,
-    )
+    @Composable
+    fun getDmRoomId(): State<RoomId?> {
```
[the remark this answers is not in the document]
bmarty: Yes, done in the latest commit: e46503ae821816b98d44cf67d8295cd8c02a961e

PR #3705, opened 2024-10-18
Rename some function to avoid name clash

bmarty:
First commit is printing out warning during the generation of the gallery, such as:

```bash
Check for screenshots name duplication...
Duplicated screenshot name: SheetContent
    - features.messages.impl.timeline.components.reactionsummary_SheetContent
    - features.messages.impl.actionlist_SheetContent
Warning: 1 duplicated screenshot name(s) found
```

Second commit fix the existing issue.

The problem is visible (for now) in the gallery at https://element-hq.github.io/element-x-android/?a=1&w=100&l=de%2Cfr#SheetContent. (make sure to check `Show all screenshots`).

2 unrelated screens was in the same table section:

image

This change will also give better title to those 2 screens in the gallery


jmartinesp (approved):
I wish we had a better solution than renaming the component, but at least this way we'd have a way to get notified of these issues. Thanks!

PR #3751, opened 2024-10-29
Change wording to "Verify identity"

bmarty:
See the discussion here: https://www.figma.com/design/0oUTCbYhROEDprSp7bdWhx?node-id=896-11141#1004391159

image


features/userprofile/shared/src/main/kotlin/io/element/android/features/userprofile/shared/UserProfileView.kt
```diff
@@ -106,7 +106,7 @@ fun UserProfileView(
 private fun VerifyUserSection(state: UserProfileState) {
     if (state.isVerified.dataOrNull() == false) {
         ListItem(
-            headlineContent = { Text(stringResource(R.string.screen_room_member_details_verify_button_title, state.userName ?: state.userId)) },
+            headlineContent = { Text(stringResource(CommonStrings.common_verify_identity)) },
```
pixlwave, line 109: Should this be `action.verify_identity` as verify is a verb here?

PR #3776, opened 2024-10-31
Target api 35

bmarty:
## Content

Change compile and target to 35 and fix new warnings in the code. I do not think that we need to take special action for the change about foreground service (see https://developer.android.com/about/versions/15/behavior-changes-15#fgs-hardening). I do not see in https://developer.android.com/about/versions/15/behavior-changes-15 other actions that we should handle.

Also improve code quality on the `Versions` object.

## Motivation and context

Be up to date and allow dependencies to be upgraded (see for instance https://github.com/element-hq/element-x-android/pull/3766)

## Tests

- Run the app in API 35 emulator and see that it's working fine.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -48,9 +48,9 @@ android {
         } else {
             "io.element.android.x"
         }
-        targetSdk = Versions.targetSdk
-        versionCode = Versions.versionCode
-        versionName = Versions.versionName
+        targetSdk = Versions.TARGET_SDK
+        versionCode = Versions.VERSION_CODE
+        versionName = Versions.VERSION_NAME
 
         // Keep abiFilter for the universalApk
         ndk {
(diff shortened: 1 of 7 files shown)


jmartinesp:
We may have to check:

> Changes to language code handling: When using the [Locale](https://developer.android.com/reference/java/util/Locale) API, language codes for Hebrew, Yiddish, and Indonesian are no longer converted to their obsolete forms (Hebrew: iw, Yiddish: ji, and Indonesian: in). When specifying the language code for one of these locales, use the codes from ISO 639-1 instead (Hebrew: he, Yiddish: yi, and Indonesian: id).

Could this allow us to remove some localazy post-processing?

> Change PendingIntent creators to [block background activity launches](https://developer.android.com/guide/components/activities/background-starts) by default. This helps prevent apps from accidentally creating a PendingIntent that could be abused by malicious actors.

> Intents must have actions: Intents without an action will no longer match any intent-filters. This means that intents used to start activities or services must have a clearly defined action.

The last 2 may be closely tied to the Element Call integration.

jmartinesp (approved):
I took another look, tested it a bit and I *think* we're handling all the changes I mentioned above.

PR #3806, opened 2024-11-05
Update project status

mxandreas:
## Content

Update the project status to reflect that we're now in a phase of rollout and migrating.

## Motivation and context

To reflect the true phase of the project.


bmarty (approved):
Thanks!

PR #3803, opened 2024-11-04
Send caption with image and video

bmarty:
## Content

Add a composer in the media preview to be able to add a caption to image and video.

## Motivation and context

Closes #2573 

## Screenshots / GIFs

https://github.com/user-attachments/assets/27a826a1-21f8-4e27-b7df-a2f2e0d8c4e4

## Tests

- Sent an image or a video to a room
- observe that you can optionally write a caption, which will be then rendered in the timeline.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 23
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [ ] Pull request is based on the develop branch
- [ ] Pull request title will be used in the release note, it clearly define what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
- [ ] You've made a self review of your PR

PR #3816, opened 2024-11-06
Iteration on caption

bmarty:
## Content

This PR is following #3803 .
- Improve the last message preview in the room list for media. Either use the caption, or use the filename, as per EX on iOS.
- Ensure that filename of voice message are not rendered to the user.
- Render the caption in the notification, below the image.
- Other small bug fixes.

## Motivation and context

Polish the application

## Tests

- Alice send media with caption to Bob
- Bob can see the caption in the notification, and in the room list

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
@@ -107,6 +107,7 @@ class DefaultNotifiableEventResolver @Inject constructor(
                     senderDisambiguatedDisplayName = senderDisambiguatedDisplayName,
                     body = messageBody,
                     imageUriString = content.fetchImageIfPresent(client)?.toString(),
+                    imageMimeType = content.getImageMimetype(),
                     roomName = roomDisplayName,
                     roomIsDm = isDm,
                     roomAvatarPath = roomAvatarUrl,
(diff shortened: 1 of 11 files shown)


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/model/NotifiableMessageEvent.kt
```diff
@@ -31,7 +31,8 @@ data class NotifiableMessageEvent(
     val body: String?,
     // We cannot use Uri? type here, as that could trigger a
     // NotSerializableException when persisting this to storage
-    val imageUriString: String?,
+    private val imageUriString: String?,
```
jmartinesp, line 34: Not needed for this PR, but since we're no longer persisting these (I think) we might be able to use Uri again.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
```diff
@@ -316,6 +317,17 @@ class DefaultNotifiableEventResolver @Inject constructor(
             }
             .getOrNull()
     }
+
+    private suspend fun NotificationContent.MessageLike.RoomMessage.getImageMimetype(): String? {
```
jmartinesp, line 321: Couldn't this be reused in the previous function (`fetchImageIfPresent`)?

jmartinesp (approved):
LGTM, thanks!

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/model/NotifiableMessageEvent.kt
bmarty: A yes, true!

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
bmarty: Yes, maybe. Also I do not see a better way to avoid the code duplication here. I do not like using `Pair`.

PR #3337, opened 2024-08-24
Fix login navigation getting stuck when the app was compiled with no-op analytics provider

SpiritCroc:
Currently, the FTUE flow requires the `analyticsService.didAskUserConsent()` flow to update in order to evaluate `isAnyStepIncomplete()` and thus to dismiss the onboarding flow. With the no-op analytics provider, `didAskUserConsent()` is always true, such that no `updateState()` call is triggered after accepting or denying notification permission, leaving the user stuck at that screen until re-opening the app.

## Tests

[Compile with no-op analytics provider](https://github.com/SchildiChat/schildichat-android-next/commit/f8a36f2b5ecfa456afefee571e6c1f1cbefc0104) and do a fresh login.

## Tested devices

- [x] Physical / only tested with SchildiChat fork so far
- [ ] Emulator
- OS version(s): 13

## Checklist

- [x] Changes have been tested on an Android device or Android emulator with API 23
- [x] UI change has been tested on both light and dark themes
- [x] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly define what will change for the user
- [x] Pull request includes screenshots or videos if containing UI changes
- [x] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
- [x] You've made a self review of your PR

## Sign-off

Signed-off-by: Tobias Büttner <dev@spiritcroc.de>

PR #3830, opened 2024-11-07
Fix verification failed issue, simplify verification logic

jmartinesp:
## Content

- Avoid unnecessary `initVerificationControllerIfNeeded` call which caused the automatic verification flow failure.
- Reuse Rust `Client` instances created on the login process so we don't need to restore one right before the session verification.
- Remove unnecessary sources of verification state updates.
- Add an intermediate FTUE flow step which will display an indeterminate progress indicator instead of a blank screen.

## Motivation and context

This should fix the issues we've seen where the authentication failed automatically due to a race condition.

Keeping the Client instances created on login would also make our setup more similar to iOS and avoid unnecessary work.

## Tests

- Login and verify your device (self-verification).
- Try verifying some other device from this one.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/verification/RustSessionVerificationService.kt
```diff
@@ -237,37 +233,20 @@ class RustSessionVerificationService(
...
-                    VerificationState.VERIFIED -> SessionVerifiedStatus.Verified
-                    VerificationState.UNVERIFIED -> SessionVerifiedStatus.NotVerified
-                }
-                Timber.d("New verification status: ${_sessionVerifiedStatus.value}")
-            }
-        } else {
-            // Otherwise, just check the current verification status from the session verification controller instead
-            Timber.d("Updating verification status: flow is pending or was finished some time ago")
-            runCatching {
-                initVerificationControllerIfNeeded()
```
jmartinesp, line 259: This line here is what caused the automatic verification failure if it was called before the verification service was ready.

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/FtueFlowNode.kt
```diff
@@ -99,15 +107,15 @@ class FtueFlowNode @AssistedInject constructor(
             NavTarget.SessionVerification -> {
                 val callback = object : FtueSessionVerificationFlowNode.Callback {
                     override fun onDone() {
-                        lifecycleScope.launch { moveToNextStep() }
+                        lifecycleScope.launch { moveToNextStepIfNeeded() }
```
bmarty, line 110: Remove `lifecycleScope.launch` here and in the case below? It seems that `moveToNextStepIfNeeded` is handling that (and this is not new).

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/state/DefaultFtueService.kt
```diff
@@ -90,34 +105,19 @@ class DefaultFtueService @Inject constructor(
             } else {
                 getNextStep(FtueStep.AnalyticsOptIn)
             }
-            FtueStep.AnalyticsOptIn -> {
-                updateState()
```
bmarty, line 94: This was a fix for application without analytics (see https://github.com/element-hq/element-x-android/pull/3337). So maybe restore?

bmarty (commented):
A first few remarks.

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/state/DefaultFtueService.kt
jmartinesp: In any case, I'll leave it there just in case, since this might be caused by something else triggering the `moveToNextStepIfNeeded()` function and we might remove that calling site in the future and come back to the bug described there.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/auth/ClientAuthenticationObserver.kt
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ * Please see LICENSE in the repository root for full details.
+ */
+
+package io.element.android.libraries.matrix.api.auth
+
+import io.element.android.libraries.matrix.api.MatrixClient
+
+fun interface ClientAuthenticationObserver {
```
ganfra, line 12: expose this as a `MatrixAuthenticationService.Listener` and maybe adds a way to add and remove listeners?

jmartinesp: I'm not sure how to design that, since we need:

- `RustMatrixAuthenticationService` to have access to it so it can call `onClientAuthenticationSucceeded`.
- `MatrixClientsHolder` to somehow be able to be notified from this listener.

But `MatrixClientsHolder` itself can't implement the listener, since its constructor holds a `MatrixAuthenticationService` and it would be a cyclic dependency: `Service -> Holder -> Service`. This wasn't fun to avoid...

Also, if I rename it to `Listener`, it would be weird if a `Listener` had listeners itself... And I'm not sure how useful it would be to be able to add/remove them when we only have one. If we need to make it more flexible we can iterate in the future, IMO.

bmarty: I propose this implementation:

[listenToNewMatrixClient.patch](https://github.com/user-attachments/files/17677680/listenToNewMatrixClient.patch)

Tested OK on my side, let me know what you think.

Note: the patch does not include on tests.

bmarty (commented):
Some remarks, still testing on my side

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
```diff
@@ -155,8 +157,8 @@ class RustMatrixAuthenticationService @Inject constructor(
                         passphrase = pendingPassphrase,
                         sessionPaths = currentSessionPaths,
                     )
-                clear()
                 sessionStore.storeData(sessionData)
+                authenticationObserver.onClientAuthenticationSucceeded(rustMatrixClientFactory.create(client))
```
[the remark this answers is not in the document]
jmartinesp: I don't really mind, and it seems like it could avoid some rare race conditions. Let's do it!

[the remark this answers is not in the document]
bmarty: Yes, since the insertion in the session store will trigger the switch of Appyx Node, it's probably better that the `matrixClient` is ready to be used before.

appnav/src/test/kotlin/io/element/android/appnav/di/MatrixClientsHolderTest.kt
```diff
@@ -81,4 +81,17 @@ class MatrixClientsHolderTest {
         matrixClientsHolder.restoreWithSavedState(savedStateMap)
         assertThat(matrixClientsHolder.getOrNull(A_SESSION_ID)).isEqualTo(fakeMatrixClient)
     }
+
+    @Test
+    fun `test ClientAuthenticationObserver emits a value and we save it`() = runTest {
```
bmarty, line 86: Maybe rename this test now that `ClientAuthenticationObserver` has been deleted.

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/state/DefaultFtueService.kt
```diff
@@ -90,34 +101,19 @@ class DefaultFtueService @Inject constructor(
...
-            .filter { it != SessionVerifiedStatus.Unknown }
-            // This is not ideal, but there are some very rare cases when reading the flow seems to get stuck
-            .timeout(5.seconds)
-            .catch {
-                Timber.e(it, "Failed to get session verification status, assume it's not verified")
-                emit(SessionVerifiedStatus.NotVerified)
-            }
-            .first()
-        return readyVerifiedSessionStatus == SessionVerifiedStatus.NotVerified && !canSkipVerification()
+        // Wait until the session verification status is known
```
bmarty, line 113: Neat: the `@OptIn` can be removed.

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/state/DefaultFtueService.kt
```diff
@@ -50,6 +47,15 @@ class DefaultFtueService @Inject constructor(
 ) : FtueService {
     override val state = MutableStateFlow<FtueState>(FtueState.Unknown)
 
+    /**
+     * This flow emits true when the FTUE flow is ready to be displayed.
+     * In this case, the FTUE flow is ready when the session verification status is known.
+     */
+    @OptIn(FlowPreview::class)
+    val isVerificationStatusKnown = sessionVerificationService.sessionVerifiedStatus
```
bmarty, line 55: Neat: the `@OptIn` can be removed (twice in this file).

features/ftue/impl/src/test/kotlin/io/element/android/features/ftue/impl/DefaultFtueServiceTest.kt
```diff
@@ -73,6 +74,27 @@ class DefaultFtueServiceTest {
...
+        val analyticsService = NoopAnalyticsService()
+        val sessionVerificationService = FakeSessionVerificationService()
+        val permissionStateProvider = FakePermissionStateProvider(permissionGranted = true)
+        val lockScreenService = FakeLockScreenService()
+        val service = createDefaultFtueService(
+            sessionVerificationService = sessionVerificationService,
+            analyticsService = analyticsService,
+            permissionStateProvider = permissionStateProvider,
+            lockScreenService = lockScreenService,
+        )
```
bmarty, line 88: Thanks for this new test!

bmarty (approved):
I'll merge develop on it and do more test, but it looks OK!

bmarty (approved):
Thanks for the update.
I have done more tests and this seems to be fine.
We will need to handle the problem when Element Web is accepting an incoming verification request even if it does not know the private key. In this case, the Element X session will not become verified.

This is tracked by https://github.com/element-hq/element-web/issues/27655

PR #3829, opened 2024-11-07
Verification UI / UX iteration

bmarty:
## Content

Iterate on a few screens of the verification flows

## Motivation and context

Closes #3824

## Screenshots / GIFs

See the recorded screenshots

## Tests

- Perform a login followed by a session verification using another device
- From an other device, perform a session verification flow.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 23
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [ ] Pull request is based on the develop branch
- [ ] Pull request title will be used in the release note, it clearly define what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
- [ ] You've made a self review of your PR

PR #3832, opened 2024-11-07
Disable button during the "verifying" step.

bmarty:
A point missing in #3829 
Figma: https://www.figma.com/design/pDlJZGBsri47FNTXMnEdXB/Compound-Android-Templates?node-id=41-10333

image

(see the recorded screenshot)


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/Button.kt
```diff
@@ -118,6 +119,14 @@ fun TextButton(
     leadingIcon = leadingIcon
 )
 
+@Composable
+fun InvisibleButton(
```
ganfra, line 123: I think it's a bit weird to have an `InvisibleButton` api?

bmarty: I can't think of a better idea, let's keep this for now.

ganfra (approved):

PR #3833, opened 2024-11-07
ElementCall: allow user to switch to another call.

bmarty:
## Content

When the user is in a call, starting (or joining) another call had no effect.

With the change in this PR, if the Activity get a new Intent with other call data, the current call is terminated and the new call starts. Actually the Activity is recreated as it was way simpler than trying to update all the internal state.

## Motivation and context

Same behavior than on the iOS app.

## Tests

- In room A start a call
- Enter pip mode
- Navigate to room B
- In room B start a call

Call A should be terminated and Call B should be started

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
@@ -45,12 +45,13 @@ import io.element.android.features.call.impl.pip.PipView
 import io.element.android.features.call.impl.services.CallForegroundService
 import io.element.android.features.call.impl.utils.CallIntentDataParser
 import io.element.android.libraries.architecture.bindings
+import io.element.android.libraries.core.log.logger.LoggerTag
 import io.element.android.libraries.designsystem.theme.ElementThemeApp
 import io.element.android.libraries.preferences.api.store.AppPreferencesStore
 import timber.log.Timber
 import javax.inject.Inject
 
-private const val loggerTag = "ElementCallActivity"
+private val loggerTag = LoggerTag("ElementCallActivity")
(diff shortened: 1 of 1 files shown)


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
```diff
@@ -185,23 +188,23 @@ class ElementCallActivity :
...
-            // Coming back from notification, do nothing
-            else -> return
+        val currentCallType = webViewTarget.value
+        if (currentCallType == null && callType == null) {
+            Timber.tag(loggerTag).d("Re-opened the activity but we have no url to load or a cached one, finish the activity")
+            finish()
+        } else if (currentCallType == null) {
+            Timber.tag(loggerTag).d("Set the call type and create the presenter")
+            webViewTarget.value = callType
+            presenter = presenterFactory.create(callType!!, this)
```
bmarty, line 201: Sadly i had to use `!!` here, the compiler is not clever enough (or is it me?)

jmartinesp (approved):

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
jmartinesp: You could also explicitly add `&& callType != null` to the if clause.

bmarty: Yes, but `callType != null` is always `true` in this case, so I think it's worse.

PR #3878, opened 2024-11-15
Hide the keyboard when navigating from the chat room screen

jmartinesp:
## Content

Hide the keyboard when the composer is in use when:

- We tap on the room details screen and navigate to it.
- We tap on any room member details (avatar, display name) and navigate to their profiles.

## Motivation and context

It seems like this became a regression.

## Tests

- Open the keyboard by tapping on the composer in a chat room.
- Tap on the room header so the room details screen is opened. The keyboard should hide.
- Go back.
- Now tap on any user's info in the timeline to open their profile. The keyboard should hide here too.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
@@ -142,6 +142,11 @@ fun MessagesView(
     // This is needed because the composer is inside an AndroidView that can't be affected by the FocusManager in Compose
     val localView = LocalView.current
 
+    fun hidingKeyboard(block: () -> Unit) {
+        localView.hideKeyboard()
+        block()
+    }
+
     fun onContentClick(event: TimelineItem.Event) {
         Timber.v("onMessageClick= ${event.id}")
         val hideKeyboard = onEventContentClick(event)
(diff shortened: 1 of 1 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
```diff
@@ -142,6 +142,11 @@ fun MessagesView(
     // This is needed because the composer is inside an AndroidView that can't be affected by the FocusManager in Compose
     val localView = LocalView.current
 
+    fun hidingKeyboard(block: () -> Unit) {
+        localView.hideKeyboard()
+        block()
+    }
```
bmarty, line 148: Nice pattern!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
```diff
@@ -189,13 +195,8 @@ fun MessagesView(
...
                     roomCallState = state.roomCallState,
-                    onBackClick = {
-                        // Since the textfield is now based on an Android view, this is no longer done automatically.
-                        // We need to hide the keyboard when navigating out of this screen.
-                        localView.hideKeyboard()
-                        onBackClick()
-                    },
-                    onRoomDetailsClick = onRoomDetailsClick,
+                    onBackClick = { hidingKeyboard { onBackClick() } },
+                    onRoomDetailsClick = { hidingKeyboard { onRoomDetailsClick() } },
```
bmarty, line 199: I was thinking about also wrapping `onJoinCallClick` below but since it will start another Activity, I think it's fine to keep the keyboard opened.

bmarty (approved):
LGTM, thanks!

PR #3883, opened 2024-11-18
Sync Strings - add translations to Finnish

ElementBot:


jmartinesp (approved):
LGTM as long as the CI passes.

Issue #3896, opened 2024-11-19
Element Call links cause a "401" error modal and the call does not load


benparsons:
### Steps to reproduce

1. click a link in Google Calendar to eg call.element.io/....
2. EXA opens
3. it previously would open the call, now it just gives a bare 401 error popup

1. in EXA, click a link to eg call.element.io/....
2. same popup appears

#### What did you expect?

This did previously work

#### What happened instead?

401 error

### Your phone model

Samsung S20 FE

### Operating system version

Android 13

### Application version and app store

0.7.4-nightly

### Homeserver

element.io

### Will you send logs?

Yes

NB logs sent include opening an external link from another, then opening the same link from inside a room

### Are you willing to provide a PR?

No

benparsons:
![image](https://github.com/user-attachments/assets/526ad235-b5fe-4dbf-8d37-d90d3b424187)

hughns:
The is a fix in progress in https://github.com/element-hq/element-x-android/pull/3962

PR #3962, opened 2024-11-27
Element Call: display error dialog only when loading the main URL

jmartinesp:
## Content

Only propagate 'webview errors' to display the error dialog when the errors happen while trying to load the webview's main URL. At the moment this happened when trying to load any asset or external URL too.

## Motivation and context

Fix https://github.com/element-hq/element-x-android/issues/3896.

Note this *may* not work properly on URL redirections, although I hope a redirection means the webview's URL has changed and the check is still valid. In any case, it would be a rare corner case.

## Tests

- Open an Element Call external URL.
- If it load properly, it's fixed.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


toger5 (approved):

bmarty (approved):
OK for the change, but we should probably give more info in the `onError` lambda and let the presenter decide what to do.

Also there is another call to `onError` in `onReceivedSslError`, maybe also protect this call hence I do not think it is necessary to fix the issue.

PR #3963, opened 2024-11-28
Add timeline action item to copy caption

bmarty:
## Content

Add timeline action item to copy caption if available
Reorder timeline action item and improve the preview
Other rework on the codebase.

## Motivation and context

Follow order in https://www.figma.com/design/ux3tYoZV9WghC7hHT9Fhk0/Compound-iOS-Components?node-id=1786-66258

## Tests

- Send a image with a caption
- Long click and see the new item "Copy caption"

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/actionlist/model/TimelineItemActionComparator.kt
```diff
@@ -0,0 +1,37 @@
...
+        TimelineItemAction.RemoveCaption,
+        TimelineItemAction.ViewSource,
+        TimelineItemAction.ReportContent,
+        TimelineItemAction.Redact,
+    )
+
+    override fun compare(o1: TimelineItemAction, o2: TimelineItemAction): Int {
+        val index1 = orderedList.indexOf(o1)
+        val index2 = orderedList.indexOf(o2)
+        return index1.compareTo(index2)
```
jmartinesp, line 35: I guess items not added to the list above by mistake will appear at the top? Index -1 will be lower than anything else.

Do we have some way to ensure the sorting is exhaustive? Using a `when` and returning index values seems like a PITA, but I can't think of anything else.

jmartinesp (approved):
Cautiously LGTM: I'm a bit concerned about the new sorter not taking into account new actions if we forget to add them and us not realising it until it's too late.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/actionlist/model/TimelineItemActionComparator.kt
bmarty: I wrote a `when` at the beginning, but yes, it's painful, and we have to give arbitrary index. What we could do is let the app crashes if `index == -1`? I guess that if a developer is adding an action, they will test it at least once :)

Issue #3572, opened 2024-10-01
Click on notification to open chat, type on keyboard not working


0verEngineer:
### Steps to reproduce

1. Click on a notification to open the chat
2. The Keyboard is visible
3. Typing on the Keyboard does not work because the cursor is not set in the message field

#### What did you expect?
If i do this i expect the cursor to be in the message field

#### What happened instead?
See above

### Your phone model

Google Pixel 7 Pro

### Operating system version

Android 14

### Application version and app store

0.6.4 from Github releases

### Homeserver

Synapse

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

Devenda:
I too have this issue.
Additionally, but probably unrelated, the text box needs a rather precise touch to actually select it and start typing.

voc0der:
Subbed, also having this issue.

voc0der:
Hello, this is not fixed in the latest version. It seems like it should've been based on the release notes.

If someone sends me a message, I cannot reply in the notification, and further, if I click the notification, it does not automatically prompt the keyboard open. I can't tell what changed? 

Samsung S24 here. Both users on new ver.

voc0der:
@ganfra Hello, this is not fixed in the latest version. It seems like it should've been based on the release notes.

If someone sends me a message, I cannot reply in the notification, and further, if I click the notification, it does not automatically prompt the keyboard open.

I have to click the text input box to bring up a keyboard. The problem is the 2 extra steps. The notification thing I dont know if you'll fix because of e2e, but otherwise...

I can't tell what changed?

Samsung S24 here. Both users on new ver.

I don't think we need a new ticket, this is the right description still.

Issue #3098, opened 2024-06-26
After opening multiple notifications, going back from room view does not navigate to rooms list view


nathanael-h:
### Steps to reproduce

1. Ensure you received multiple notifications for different rooms
2. Open notification for room1
3. Room1 is opened
4. Open notification for room2
5. Room2 is opened
6. Open notification for room3
7. Room3 is opened
8. Touch back arrow on top left
9. App navigate from room3 to room2 
10. Touch back arrow
11. App navigate from room2 to room1
12. Touch back arrow
13. App navigate from room1 to room list view

#### What did you expect?
I expected that the back arrow navigate directly to the room list view
#### What happened instead?
It seems that opening a room notification push the room view on the navigation stack and that the back arrow simply pop the latest view from the navigation stack
I think that opening a room notification should clear the navigation stack before pushing a new view on it. 

### Your phone model

Fairphone 3+ 

### Operating system version

Android 13, /e/OS 2.1

### Application version and app store

Element X Android 0.4.14

### Homeserver

matrix.isidorus.fr / yunohost / synapse version 1.109.0 / sliding sync v0.99.18

### Will you send logs?

No

### Are you willing to provide a PR?

No

frebib:
In the very least, if a room is in the backstack more than once, I would expect that the most recent copy of that room would remain open (especially now that we have draft support) but the older duplicates would be removed, otherwise you could end up revisiting the same room many many times trying to get out of the backstack from a ton of notifications

nathanael-h:
Hello, I kindly share with you that this annoying issue drove me to test another Matrix client and finally use it daily (Fluffychat). Don't get me wrong, I share with you this end-user feedback so that you and the product and UX team can use it. I do not want to pressure. 

To explain more what is annoying: after a few hours away from phone (and also from any other matrix client), I might have messages in multiple matrix rooms (6 for example), with many new notifications. I open notification one after another. And at the end when I want to get back to the rooms list, I need to hit back arrow as many times as I got notifications (at least 6, sometimes more).

atomdmac:
@bmarty  What are the cons to implementing this change?  What the use cases where the current behavior is desirable?

titulebolide:
I really do agree with @nathanael-h. After a long time opening successively many notifications, this is quite surprising having to go back several time.
I think this would also be consistent with the UX of other messaging app.

0verEngineer:
[...]

PR #3968, opened 2024-11-29
Fix navigation to rooms from notifications

jmartinesp:
## Content

When opening a room from a notification:

- Makes sure not to add a new room screen to the backstack if the same room is already displayed.
- Reset the navigation to the room list level. This is the behaviour in most IM apps, and it seems more natural (tested this on WhatsApp, Telegram and also EXI).
- Make sure the keyboard is closed when navigating out of the chat screen even if the navigation is triggered externally (should fix https://github.com/element-hq/element-x-android/issues/3572).

## Motivation and context

Fix https://github.com/element-hq/element-x-android/issues/3098 and https://github.com/element-hq/element-x-android/issues/3572.

## Tests

- Open some chat room (let's call it, 'Chat A'), tap on the composer to open the keyboard.
- Go to the home screen of the device without closing the keyboard in-app.
- Receive a notification from another room (Chat B).
- Open the notification.
- Chat B should be displayed, the keyboard should be dismissed and if you tap on the composer you can type normally.
- Go to the home screen again.
- Receive and open another notification from Chat B.
- The same room screen should be reused, no new chat screen should be pushed to the backstack.
- Tapping on back should take you to the room list, not the previous room.
- Send a message with a link to some room in Chat B.
- Follow the link, a new chat screen for that room should be opened.
- If you go back you're in the previous room, as expected, not to Chat A as it happened before.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
@@ -472,9 +472,9 @@ class LoggedInFlowNode @AssistedInject constructor(
             navTarget is NavTarget.RoomList
         }
 
-        val isRoomAlreadyDisplayed = backstack.elements.value.any {
-            it.key.navTarget is NavTarget.Room && (it.key.navTarget as NavTarget.Room).roomIdOrAlias == roomIdOrAlias
-        }
+        val isRoomAlreadyDisplayed = backstack.active?.key?.navTarget?.let { target ->
+            target is NavTarget.Room && target.roomIdOrAlias == roomIdOrAlias
+        } == true
 
         if (!isRoomAlreadyDisplayed) {
             attachChild<RoomFlowNode> {
--- a/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
@@ -17,6 +17,7 @@ import androidx.compose.ui.Modifier
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.lifecycleScope
 import androidx.lifecycle.repeatOnLifecycle
+import com.bumble.appyx.Appyx
 import com.bumble.appyx.core.composable.PermanentChild
 import com.bumble.appyx.core.lifecycle.subscribe
 import com.bumble.appyx.core.modality.BuildContext
@@ -25,6 +26,7 @@ import com.bumble.appyx.core.node.Node
 import com.bumble.appyx.core.plugin.Plugin
 import com.bumble.appyx.core.plugin.plugins
 import com.bumble.appyx.navmodel.backstack.BackStack
+import com.bumble.appyx.navmodel.backstack.active
 import com.bumble.appyx.navmodel.backstack.operation.pop
 import com.bumble.appyx.navmodel.backstack.operation.push
 import com.bumble.appyx.navmodel.backstack.operation.replace
@@ -135,8 +137,13 @@ class LoggedInFlowNode @AssistedInject constructor(
         }
     }
 
+    private val navigateBackToRoomObserverPlugin = NavigateBackToRoomObserverPlugin()
+
     override fun onBuilt() {
         super.onBuilt()
+
+        Appyx.globalPlugins = listOf(navigateBackToRoomObserverPlugin)
+
         lifecycle.subscribe(
             onCreate = {
                 appNavigationStateService.onNavigateToSession(id, matrixClient.sessionId)
@@ -476,7 +483,10 @@ class LoggedInFlowNode @AssistedInject constructor(
             target is NavTarget.Room && target.roomIdOrAlias == roomIdOrAlias
         } == true
 
-        if (!isRoomAlreadyDisplayed) {
+        if (isRoomAlreadyDisplayed) {
+            // Ask the active backstack to navigate back to the room, if needed
+            navigateBackToRoomObserverPlugin.navigateBackToRoom(roomIdOrAlias)
+        } else {
             attachChild<RoomFlowNode> {
                 // Remove existing path from the room list target
                 backstack.singleTop(NavTarget.RoomList)
--- a/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
@@ -471,17 +471,28 @@ class LoggedInFlowNode @AssistedInject constructor(
         waitForNavTargetAttached { navTarget ->
             navTarget is NavTarget.RoomList
         }
-        attachChild<RoomFlowNode> {
-            backstack.push(
-                NavTarget.Room(
-                    roomIdOrAlias = roomIdOrAlias,
-                    serverNames = serverNames,
-                    trigger = trigger,
-                    initialElement = RoomNavigationTarget.Messages(
-                        focusedEventId = eventId
+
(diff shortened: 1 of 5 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
```diff
@@ -247,6 +248,13 @@ fun MessagesView(
         onUserDataClick = onUserDataClick,
     )
     ReinviteDialog(state = state)
+
+    // Make sure we hide the keyboard when the view is disposed
+    DisposableEffect(Unit) {
```
ganfra, line 253: Should we not find a solution directly on the FlowNode for this? :/
Also using `SoftwareKeyboardController` doesn't work?

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -471,17 +471,28 @@ class LoggedInFlowNode @AssistedInject constructor(
...
-        attachChild<RoomFlowNode> {
-            backstack.push(
-                NavTarget.Room(
-                    roomIdOrAlias = roomIdOrAlias,
-                    serverNames = serverNames,
-                    trigger = trigger,
-                    initialElement = RoomNavigationTarget.Messages(
-                        focusedEventId = eventId
+
+        val isRoomAlreadyDisplayed = backstack.elements.value.any {
```
ganfra, line 475: I think we should only check the active element, not if the room is in the backstack?

mxandreas:
Looks good to me.

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
jmartinesp: Yes, you're right. I was using this because I thought the backstack was shared with the one in `RoomFlowNode`, but with the current implementation we only need the active element check.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
jmartinesp: I can try to find a better solution to apply at the `BaseFlowNode` level, instead of hiding the keyboard when a node is disposed, maybe it should be done when a new transition is about to start so it's more immediate.

About `SoftwareKeyboardController` I'm not sure, because the underlying UI component for the composer is an `EditText` and it might not handle focus the same way as compose components.

ganfra: Lets do this for now, and will investigate a bit more later

jmartinesp: I think I actually have a generic implementation ready, based on the `backstack.elements` value. It seems to work on any screen and it's immediate.

ganfra: Even better then!

ganfra: How does this react with screen we want the keyboard to open right away?

jmartinesp: Good question, I guess it'd try to dismiss the keyboard before the action to display it on focus happens, but I'd have to double check. Can you give me an example of one of these screens?

ganfra: Actually I think we don't have except the search on the room list, but it's not a new node

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -135,8 +137,13 @@ class LoggedInFlowNode @AssistedInject constructor(
         }
     }
 
+    private val navigateBackToRoomObserverPlugin = NavigateBackToRoomObserverPlugin()
+
     override fun onBuilt() {
         super.onBuilt()
+
+        Appyx.globalPlugins = listOf(navigateBackToRoomObserverPlugin)
```
bmarty, line 145: Maybe cleaner to do
```
Appyx.globalPlugins += navigateBackToRoomObserverPlugin
```
in the onCreate block below 
and remove the plugin in the `onDestroy` with
```
Appyx.globalPlugins -= navigateBackToRoomObserverPlugin
```

bmarty: `Appyx.globalPlugins += navigateBackToRoomObserverPlugin` was compiling when I tested locally (to be more precise: the IDE did not complain), but yes, it's worth checking again that this is still working as expected

jmartinesp: Oh, I think you're right. I guess it's creating a new list with the previous contents, since the signature is `val globalPlugins: List<Plugin>`.

ganfra: Yes it should word, because the globalPlugins is a var, so it'll copy the value and add the new one.
But I find it weird to use a globalPlugin for that...

jmartinesp: It's either that or passing down the plugin until it reaches the required flow node, and that has bitten us in the past. I can do it like that though, if you think it's best.

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -471,17 +478,31 @@ class LoggedInFlowNode @AssistedInject constructor(
...
-                    serverNames = serverNames,
-                    trigger = trigger,
-                    initialElement = RoomNavigationTarget.Messages(
-                        focusedEventId = eventId
+
+        val isRoomAlreadyDisplayed = backstack.active?.key?.navTarget?.let { target ->
+            target is NavTarget.Room && target.roomIdOrAlias == roomIdOrAlias
+        } == true
+
+        if (isRoomAlreadyDisplayed) {
```
ganfra, line 486: Maybe we should always replace the backstack with RoomList+Room no matter what is currently in the backstack

ganfra: exemple : if you are following some permalinks, you can have several rooms in the backstack, if you click on a notif I think we want the behavior to always be the same

spaetz:
Just wanted to add that this is not only useful for notifications. I have home screen shortcuts that open specific room views. If I click one twice, the same happens. I have to exit (left-swipe) a room twice before I get back to the room list view.

bmarty:
We will revisit the PR to ensure that when coming from a notification (or a shortcut) we always end up with a stack with the room list and a single room.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
ganfra: I've tested it more, and I don't think we should use this. The keyboard not closing should only concern screen where we use a view based text field, otherwise the keyboard is closed when the field is removed from composition. 
If we try to open a keyboard quickly on a newly pushed screen, the `SoftwareKeyboardController.hide` will be triggered and so the keyboard will appear and disappear right away.

ganfra:
Replaced by 2 separate PRs.

PR #3978, opened 2024-12-02
Do not auto-play videos.

bmarty:
Following what has been done on iOS and avoiding user to get loud audio when clicking on a video in the timeline, the application will not auto play video when open in full screen.

### DIFF
--- a/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/local/video/MediaVideoView.kt
+++ b/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/local/video/MediaVideoView.kt
@@ -80,11 +80,10 @@ private fun ExoPlayerMediaVideoView(
     localMedia: LocalMedia?,
     modifier: Modifier = Modifier,
 ) {
-    val isControllerVisibleByDefault = LocalInspectionMode.current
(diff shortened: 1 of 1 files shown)


libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/local/video/MediaVideoView.kt
```diff
@@ -238,7 +235,8 @@ private fun ExoPlayerMediaVideoView(
 
     OnLifecycleEvent { _, event ->
         when (event) {
-            Lifecycle.Event.ON_RESUME -> exoPlayer.play()
+            Lifecycle.Event.ON_CREATE -> exoPlayer.addListener(playerListener)
```
ganfra, line 238: How it wasn't necessary before?

bmarty: The listener was set in a LaunchedEffect block (previous version line 132, see the diff)

ganfra: Ok, I didn't see it!

ganfra (approved):

PR #3991, opened 2024-12-04
increase ringing timeout from 15 seconds to 90 seconds

fkwp:
Bugs around ringing have been resolved and it's considered generally reliable now so it's a good time to increase the timeout.


ganfra (approved):

escix:
The ringing still timeout at 30s.

escix:
@ganfra  The ringing still timeout at 30s.

Issue #3571, opened 2024-10-01
UnifiedPush not working with Conversations


blaueente:
### Steps to reproduce

1. Select "Conversations" as UnifiedPush provider
2. Run notifications Troubleshoot tests

#### What did you expect?
Working notifications (they work with other demo programs)
####  What happened instead?
Error  
"Fail to check the gateway
https://up.conversations.im/_matrix/push/v1/notify: HTTP 404 Not Found"

Apparently the URL for using conversations as a push gateway is wrong. Maybe it's the "_matrix"?  or is "v1" a too old version? Anyway, testing with https://f-droid.org/packages/org.unifiedpush.example/ works.

![Screenshot_20241001-130537](https://github.com/user-attachments/assets/89c0ae59-2418-4cf2-855e-6bfead31a360)

### Your phone model

Pixe

### Operating system version

Android 14 AP2A.240905.003.F1

### Application version and app store

Element X Version 0.6.3 (from playstore)

### Homeserver

matrix.org

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

spaetz:
This is because ntfy includes a matrix gateway (https://github.com/binwiederhier/ntfy/blob/630f2957deb670dcacfe0a338091d7561f176b9c/server/server_matrix.go#L16) that is, it receives calls to  /_matrix/push/v1/notify and sends them out via UnifiedPush.

conversations (https://codeberg.org/iNPUTmice/up) does not contain a matrix push gateway, and that is needed for EXA. So it won't work. Either conversations' up needs to include a Matrix gateway, or the Matrix homeserver needs to be taught to directly speak to a UnifiedPush server.

amalon:
> conversations (https://codeberg.org/iNPUTmice/up) does not contain a matrix push gateway, and that is needed for EXA. So it won't work. Either conversations' up needs to include a Matrix gateway, or the Matrix homeserver needs to be taught to directly speak to a UnifiedPush server.

What is EXA? I see the same issue with NextPush, which works fine as a UnifiedPush provider with normal Element on android, so I'm confused why NextPush would need matrix-specific changes for ElementX to work?

spaetz:
Looking at the code, it should fall back to use the standard matrix.org gateway if the UnifiedPush server does not provide one. Not sure what is happening there, I am not a developer myself.
P.S. EXA is short for Element-x-android

ccgauvin94:
[...]

PR #4040, opened 2024-12-14
Fix push notifications for UnifiedPush distributors without provided matrix gateway

SpiritCroc:
Fixes https://github.com/element-hq/element-x-android/issues/3571

 
## Content

We need to account for UnifiedPush distributors not providing a Matrix gateway, to properly use our fallback gateway.
[To quote p1gp1g](https://github.com/SchildiChat/schildichat-android-next/issues/38#issuecomment-2540793863), we need to:

> * return the customGateway if the check returns 200 and `discoveryResponse.unifiedpush.gateway == "matrix"`
> * return the default gateway if the check returns 200 and there isn't `discoveryResponse.unifiedpush.gateway == "matrix"`
> * return the default gateway if the check returns 404
> * return the previous gateway if it fails      
> * if there isn't a previous gateway, we can use one or the other, it doesn't really matter

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/3571

## Tests

- Set up a distributor which does feature a matrix gateway, such as ntfy
- Select ntfy as push distributor
- Check "Troubleshoot notifications" that the custom gateway is used and push works
- Set up a distributor which does not feature a matrix gateway, such as gCompat
- Select "gCompat UP-Distributor" as push distributor
- Check "Troubleshoot notifications" that the fallback gateway is used and push works

## Tested devices

NOTE: I only tested on [SchildiChat Next](https://github.com/SchildiChat/schildichat-android-next/commit/330cb384cf3a77277c757ad84cc55d8ce82e2bc9)
- [x] Physical device
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushGatewayResolver.kt
+++ b/libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushGatewayResolver.kt
@@ -14,6 +14,7 @@ import io.element.android.libraries.di.AppScope
 import kotlinx.coroutines.withContext
 import retrofit2.HttpException
 import timber.log.Timber
+import java.net.HttpURLConnection
 import java.net.URL
 import javax.inject.Inject
 
@@ -55,7 +56,7 @@ class DefaultUnifiedPushGatewayResolver @Inject constructor(
                         UnifiedPushConfig.DEFAULT_PUSH_GATEWAY_HTTP_URL
                     }
                 } catch (exception: HttpException) {
-                    if (exception.code() == 404) {
+                    if (exception.code() == HttpURLConnection.HTTP_NOT_FOUND) {
                         logger.i("Checking for UnifiedPush endpoint yielded 404, using fallback")
                         UnifiedPushConfig.DEFAULT_PUSH_GATEWAY_HTTP_URL
                     } else {
--- a/libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushGatewayResolver.kt
+++ b/libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushGatewayResolver.kt
@@ -12,12 +12,13 @@ import io.element.android.libraries.core.coroutine.CoroutineDispatchers
 import io.element.android.libraries.core.data.tryOrNull
 import io.element.android.libraries.di.AppScope
 import kotlinx.coroutines.withContext
+import retrofit2.HttpException
 import timber.log.Timber
 import java.net.URL
 import javax.inject.Inject
 
 interface UnifiedPushGatewayResolver {
-    suspend fun getGateway(endpoint: String): String
+    suspend fun getGateway(endpoint: String, previousGateway: String?): String
 }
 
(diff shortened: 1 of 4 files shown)


libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushGatewayResolver.kt
```diff
@@ -47,14 +48,24 @@ class DefaultUnifiedPushGatewayResolver @Inject constructor(
...
                         logger.d("The endpoint seems to be a valid UnifiedPush gateway")
+                        customUrl
                     } else {
-                        logger.e("The endpoint does not seem to be a valid UnifiedPush gateway")
+                        // The endpoint returned a 200 OK but didn't promote an actual matrix gateway, which means it doesn't have any
+                        logger.w("The endpoint does not seem to be a valid UnifiedPush gateway, using fallback")
+                        UnifiedPushConfig.DEFAULT_PUSH_GATEWAY_HTTP_URL
+                    }
+                } catch (exception: HttpException) {
+                    if (exception.code() == 404) {
```
bmarty, line 58: Could use `HttpURLConnection.HTTP_NOT_FOUND`.

bmarty (approved):
Thanks for the change. I do not really like to provide the previous value as a parameter for `UnifiedPushGatewayResolver.getGateway()` but I do not see any better solution. Returning `String?` could be an alternative, but it may not fit all the needs.

SpiritCroc:
Yeah agreed, passing `previousGateway` isn't very elegant. Seems to be the most straightforward approach to fulfil the spec I got without over-engineering it though. I assume just always returning `null` from `getGateway()` and falling back outside to the previous value would work except for the "if there isn't a previous gateway, we can use one or the other, it doesn't really matter" case. If "it doesn't matter" is to be interpreted that we can also fallback to no gateway at all, then that's probably the most elegant way, I don't know right now how relevant this is in practice.

bmarty:
OK, let's be pragmatic and merge those changes.

There are 2 failing tests, can you handle that please? Else let me know, I can take the point

```
VectorUnifiedPushMessagingReceiverTest > onNewEndpoint, if registration fails, the endpoint should not be stored FAILED
    java.lang.AssertionError at VectorUnifiedPushMessagingReceiverTest.kt:131

VectorUnifiedPushMessagingReceiverTest > onNewEndpoint run the expected tasks FAILED
    java.lang.AssertionError at VectorUnifiedPushMessagingReceiverTest.kt:91
```

SpiritCroc:
> OK, let's be pragmatic and merge those changes.
> 
> There are 2 failing tests, can you handle that please? Else let me know, I can take the point
> 
> ```
> VectorUnifiedPushMessagingReceiverTest > onNewEndpoint, if registration fails, the endpoint should not be stored FAILED
>     java.lang.AssertionError at VectorUnifiedPushMessagingReceiverTest.kt:131
> 
> VectorUnifiedPushMessagingReceiverTest > onNewEndpoint run the expected tasks FAILED
>     java.lang.AssertionError at VectorUnifiedPushMessagingReceiverTest.kt:91
> ```

I probably can look at those again earliest next weekend. I had a quick look before but didn't see immediately what the appropriate fix for these two would be. So if you can fix them earlier that would be great, thanks!

bmarty:
I propose an alternative fix in https://github.com/element-hq/element-x-android/pull/4101. Let me know what you think. Thanks!

PR #4047, opened 2024-12-16
Fixed presentation of checked / unchecked state for checkboxes and switches and selected state for radio buttons when used with a screen reader

pvagner:
## Content

Checkboxes, switches and radio buttons are implemented in a way that icon, title, the interactive control and optional description is encapsulated in a parent composable which receives onClick event. Checkboxes, switches and radio buttons them selves have no click handler. When considering universal user experience it's very nice approach as it increases touch target for these UI controls. However when considering screen reader accessibility screen readers are not receiving role and state information for these controls. Thus when using the app with a screen reader I can't find out if a check box is checked or not, if a switch is turned on or not and which of the radio buttons is selected.
I am trying to address this by adding toggleable and selectable modifiers where appropriate.

## Motivation and context

I am using this app daily, I used to experiment and guess the state of these checkboxes, switches and radio buttons, ask my friends until I have found I can try to improve it.

## Screenshots / GIFs

I believe appearance is unchanged. Also I have opted not to create and work with screenshots at all since I am unable to.

## Tests

Tested on a real device.

* Pressed both the volume buttons simultaneously to enable talkback screen reader
* Opened the app
* Touched any of the rooms on the screen
* Performed the double tap and hold gesture to bring up the context menu
* Used single finger flick gestures to navigate to the favorite the selected room.
* I have repeated the similar procedure in the user settings.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 14

## Checklist

- [x] Changes have been tested on an Android device or Android emulator with API 23

### DIFF
--- a/features/analytics/api/src/main/kotlin/io/element/android/features/analytics/api/preferences/AnalyticsPreferencesView.kt
+++ b/features/analytics/api/src/main/kotlin/io/element/android/features/analytics/api/preferences/AnalyticsPreferencesView.kt
@@ -8,9 +8,11 @@
 package io.element.android.features.analytics.api.preferences
 
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.selection.toggleable
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.semantics.Role
 import androidx.compose.ui.tooling.preview.PreviewParameter
 import io.element.android.features.analytics.api.AnalyticsOptInEvents
 import io.element.android.features.analytics.api.R
@@ -43,6 +45,13 @@ fun AnalyticsPreferencesView(
     )
     Column(modifier) {
         ListItem(
+            modifier = Modifier
+                .toggleable(
+                    value = state.isEnabled,
+                    role = Role.Checkbox,
+                    enabled = true,
+                    onValueChange = { onEnabledChanged(!state.isEnabled) }
+                ),
             headlineContent = {
                 Text(stringResource(id = R.string.screen_analytics_settings_share_data))
             },
@@ -52,10 +61,7 @@ fun AnalyticsPreferencesView(
             leadingContent = null,
             trailingContent = ListItemContent.Switch(
                 checked = state.isEnabled,
-            ),
-            onClick = {
-                onEnabledChanged(!state.isEnabled)
-            }
+            )
         )
         ListSupportingText(annotatedString = linkText)
     }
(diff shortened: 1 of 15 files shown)


features/analytics/api/src/main/kotlin/io/element/android/features/analytics/api/preferences/AnalyticsPreferencesView.kt
```diff
@@ -43,6 +45,13 @@ fun AnalyticsPreferencesView(
     )
     Column(modifier) {
         ListItem(
+            modifier = Modifier
+                .toggleable(
+                    value = state.isEnabled,
+                    role = Role.Checkbox,
+                    enabled = true,
+                    onValueChange = { onEnabledChanged(!state.isEnabled) }
+                ),
```
bmarty, line 54: I am wondering if all these changes could be handle at a single place: in the `ListItem` implementation. It will reduce the number of change and will have the advantage of always use `toggleable`/`selectable` so that developer will not forget about it.

Here is a patch with a proposal:
[ListItemModifier.patch](https://github.com/user-attachments/files/18151809/ListItemModifier.patch)

Note: not tested and may not always work, this patch is just to give some guidance about how it could be implemented. More work is necessary to make this patch an acceptable pull request.

bmarty (commented):
Thanks for the improvement! I have submitted a patch with a proposal, let me know hat you think.

features/analytics/api/src/main/kotlin/io/element/android/features/analytics/api/preferences/AnalyticsPreferencesView.kt
pvagner: Hello again,
The patch is working as it should with no modifications from me.
It correctly applies modifiers as expected.
However I still have to go through the places I have modified and verify if the list item is really used in a way this implementation is expecting it.
For example:
* If there is an onClick handler on the switch or check box it-self it's being presented twice with some screen readers. Having click handler on both list item and switch / checkbox should therefore be avoided if we wish to make the screen reader experience fluent. Currently at least the favourite item in the room list long click menu is implemented like this.
* Some radio buttons are not working and I need to find out why for example when going to settings -> Advanced, there is a setting that affects app theme, it can be System, Light or Dark. And once the dialog is showing the radio buttons don't currently have the modifier applied as expected.

So if you are okay with the outcome and you do have enough bandwidth for doing it, please consider pushing your patch, in favour of this pull request and we should handle edge cases as follow up PRs. It's great improvement and it enables screen reader users to adjust settings on their own if they wish.

pvagner: Dear @bmarty 
I have noticed you are leading efforts on improving accessibility these days.
I have found out what you are proposing for trailingContent is working fine however we should add the same thing for leadingContent on the ListItems to fix radio button presentation for screen reader users.
Can you please include this or would you like me to take your patch and edit this PR accordingly?

I'd appreciate if this got fixed in the mean time.

bmarty: Thanks @pvagner , I have created a replacement PR https://github.com/element-hq/element-x-android/pull/4626

Let me check what's need to be done for `leadingContent`.

bmarty:
Replaced by https://github.com/element-hq/element-x-android/pull/4626

PR #4080, opened 2024-12-20
Improve gallery loading state

bmarty:
- Add a background (see Figma)
- Render full loading state when there is only a loading item in the timeline to avoid glitch between the full loading UI and the single timeline loading item


libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/gallery/MediaGalleryView.kt
```diff
@@ -196,33 +196,47 @@ private fun MediaGalleryPage(
...
-                    imagesAndVideos = groupedMediaItems.data.imageAndVideoItems,
-                    eventSink = state.eventSink,
-                    onItemClick = onItemClick,
-                )
-                MediaGalleryMode.Files -> MediaGalleryFiles(
-                    files = groupedMediaItems.data.fileItems,
-                    eventSink = state.eventSink,
-                    onItemClick = onItemClick,
+    val groupedMediaItems = state.groupedMediaItems
+    if (groupedMediaItems.isLoadingItems(mode)) {
```
bmarty, line 200: Need to render `LoadingContent` in a single place to avoid resetting the CircularProgressIndicator.

ganfra (approved):

PR #4089, opened 2024-12-23
Close the media preview screen ASAP with sending queue enabled

jmartinesp:
## Content

- When the media file is done processing and we've pressed the 'send' button, just dismiss the attachment preview instead of waiting for the file to be sent.
- For this, get the `@SessionCoroutineScope` from the `SessionMatrixModule` and use it to send the media even if the preview screen or the chat room screen are dismissed.
- As soon as the sending process begins in background, dismiss the attachment preview screen.

## Motivation and context

We decided to have this behaviour to improve the UX.

## Screenshots / GIFs

https://github.com/user-attachments/assets/a8461209-c7da-4830-ac21-8cad4ac16798

## Tests

- Enable the 'Media upload through send queue' developer option.
- Send a large media file, press the send button quickly.
- There may still be some progress dialog while the thumbnail is being generated and the media is compressed, if enabled.
- As seen in the video above, the preview screen should be dismissed and a local echo of the media file should appear.
- After a few seconds, the media should be marked as sent.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
```diff
@@ -240,7 +251,6 @@ class AttachmentsPreviewPresenter @AssistedInject constructor(
             cleanUp(mediaUploadInfo)
             // Reset the sendActionState to ensure that dialog is closed before the screen
             sendActionState.value = SendActionState.Done
-            onDoneListener()
```
bmarty, line 243: We still need to invoke this in the case of `sendOnBackground` is false, i.e. `FeatureFlags.MediaUploadOnSendQueue` is disabled.

Confirmed by test: the screen does not close

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
```diff
@@ -92,13 +94,22 @@ class AttachmentsPreviewPresenter @AssistedInject constructor(
...
+                        val sendOnBackground = featureFlagService.isFeatureEnabled(FeatureFlags.MediaUploadOnSendQueue)
+                        val sendMediaScope = if (sendOnBackground) sessionCoroutineScope else coroutineScope
+                        ongoingSendAttachmentJob.value = sendMediaScope.launch {
                             sendPreProcessedMedia(
                                 mediaUploadInfo = mediaUploadInfo.data,
                                 caption = caption,
                                 sendActionState = sendActionState,
                             )
                         }
+
```
bmarty, line 108: Just a remark here: when the send queue is enabled, the duration of the invokation of `sendPreProcessedMedia` stay quite short (but not negligible). I have measure that it can take up to 1s for big files. The processing of the thumbnail creation and video shrinking is much longer. So this PR is OK, but the gain is not that big.

Also note that if the user close the screen during the media pre-processing, it will cancel the media sending as well. So maybe use the session scope also for the media pre-processing? In this case the local echo will take longer to appear and killing the app during the processing will stop the sending.

bmarty (commented):
Thanks for the improvement. I suspect that the gain is not as big as expected by the product, please read my comments for more details.

app/src/main/kotlin/io/element/android/x/di/RoomComponent.kt
```diff
@@ -15,9 +15,10 @@ import io.element.android.libraries.di.RoomScope
 import io.element.android.libraries.di.SessionScope
 import io.element.android.libraries.di.SingleIn
 import io.element.android.libraries.matrix.api.room.MatrixRoom
+import io.element.android.libraries.matrix.impl.di.SessionMatrixModule
 
 @SingleIn(RoomScope::class)
-@MergeSubcomponent(RoomScope::class)
+@MergeSubcomponent(RoomScope::class, modules = [SessionMatrixModule::class])
```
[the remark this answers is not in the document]
jmartinesp: Either it is or the build cache was causing some issues locally for me, since I couldn't get the app to build until I explicitly added that. I think I also tried removing it later and came back to the same error where it couldn't find the `@SessionCoroutineScope` annotated coroutine scope, but I can double check just in case.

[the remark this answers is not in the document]
jmartinesp: It was the build cache ¯\\_(ツ)\_/¯.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
jmartinesp: > Also note that if the user close the screen during the media pre-processing, it will cancel the media sending as well.

But that's OK I think? The media pre-processing should be done before sending starts, and in the meantime the screen won't be dismissed since we'll have the progress dialog present: dismissing this dialog cancels the enqueued sending process, so AFAICT there's no way to close the screen before the preprocessing is done while expecting the sending to still work.

bmarty: > Also note that if the user close the screen during the media pre-processing, it will cancel the media sending as well. So maybe use the session scope also for the media pre-processing? In this case the local echo will take longer to appear and killing the app during the processing will stop the sending.

You can ignore this, I initially though that the PR will make the screen be closed as soon as the user press the send button (even if the pre-processing is not over), but this is not the case. Sorry about this!

bmarty (approved):
LGTM, thanks for the update!

PR #4092, opened 2024-12-23
Test using Maestro CLI + emulator instead of Cloud

jmartinesp:
## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


.github/workflows/maestro-local.yml
```diff
@@ -74,23 +73,53 @@ jobs:
...
+        with:
+          api-level: ${{ env.API_LEVEL }}
+          arch: ${{ env.ARCH }}
+          profile: ${{ env.DEVICE }}
+          target: ${{ env.TARGET }}
+          emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+          disable-animations: true
+          disk-size: 3G
+          script: |
+            chmod +x .github/workflows/scripts/maestro/maestro-local-with-screen-recording.sh
```
bmarty, line 115: Is it necessary to set `+x` each time? The file access right can be committed IIRC.

.github/workflows/maestro-local.yml
```diff
@@ -1,40 +1,39 @@
...
+  DEVICE: Nexus 6
+  API_LEVEL: 35
+  TARGET: google_apis
 
 jobs:
   build-apk:
     name: Build APK
     runs-on: ubuntu-latest
-    if: github.event_name == 'workflow_dispatch' || github.event.label.name == 'Run-Maestro'
+    if: github.event_name == 'workflow_dispatch'
```
bmarty, line 24: I believe that this whole job `build-apk` can be removed...

.github/workflows/maestro-local.yml
```diff
@@ -62,8 +60,9 @@ jobs:
   maestro-cloud:
     name: Maestro test suite
     runs-on: ubuntu-latest
-    needs: build-apk
-    if: github.event_name == 'workflow_dispatch' || github.event.label.name == 'Run-Maestro'
+    needs: [build-apk]
```
bmarty, line 63: ...but I am confused with this needs statement.

Does this means that this `yml` is also able to build an APK on-demand (using `workflow_dispatch`)?

.github/workflows/maestro-local.yml
```diff
@@ -74,23 +73,53 @@ jobs:
...
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+      - name: Install maestro
+        run: curl -fsSL "https://get.maestro.mobile.dev" | bash
+      - name: Run Maestro tests in emulator
+        uses: reactivecircus/android-emulator-runner@v2
+        env:
+          MAESTRO_USERNAME: maestroelement
+          MAESTRO_PASSWORD: ${{ secrets.MATRIX_MAESTRO_ACCOUNT_PASSWORD }}
+          MAESTRO_RECOVERY_KEY: ${{ secrets.MATRIX_MAESTRO_ACCOUNT_RECOVERY_KEY }}
```
bmarty, line 101: Note: I see that the recovery key is visible in the recorded video, but I guess this is not a very sensitive data for this test account. The password is not visible.

bmarty (approved):
Very nice, thanks!
I have added some comments. There is no rush to handle this before next year :)

.github/workflows/maestro-local.yml
jmartinesp: Oh, true... although I'm not sure how private those artifacts containing the recordings are, I tried downloading some from both a logged out browser and an account without permissions and I couldn't 🤔 .

.github/workflows/maestro-local.yml
bmarty: OK, I did not think about running the workflow on a random branch. Not sure we will ever need to do that, but OK!

.github/workflows/maestro-local.yml
bmarty: Interesting, thanks. So I guess we're fine!

PR #4101, opened 2024-12-30
Unified push gateway resolver improvement

bmarty:
Replacement for #4040.

See #4040 for more details.

I tested switching between `ntfy` `Firebase` and `gCompat UP-Distributor` without any issue.


ganfra (approved):
Sounds sane, thanks!

PR #4103, opened 2024-12-31
Fix dm avatar rtl

### LINKED ISSUE #4096: Bug in showing profile image in RTL languages
### Steps to reproduce

1. Set your language to an RTL language.
2. Open a chat.
3. Tap on the top bar.
4. In the view which shows your and the contact's profile photos, you see a gap in the top right part of your profile's image which is not visible in LTR languages because of images' positions.

#### What did you expect?
The contact's image should be top right of user's own image even in RTL languages.

### Your phone model

Samsung J7

### Operating system version

Android 10

Element X 0.7.5 from f-droid

### Homeserver

_No response_

### Will you send logs?

No

No

bmarty:
Closes #4096


ganfra (approved):
Nice thanks!

PR #4105, opened 2025-01-02
misc(send queue) : do not disable send queue when Network is Offline

ganfra:
## Content
Just align with the behaviour from iOS.
Will probably iterate later when we'll rework how we detect network conditions.

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/loggedin/SendQueues.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/loggedin/SendQueues.kt
@@ -13,9 +13,9 @@ import io.element.android.features.networkmonitor.api.NetworkStatus
 import io.element.android.libraries.di.SessionScope
 import io.element.android.libraries.di.SingleIn
 import io.element.android.libraries.matrix.api.MatrixClient
-import io.element.android.libraries.matrix.api.core.RoomId
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.debounce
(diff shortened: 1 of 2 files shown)


appnav/src/main/kotlin/io/element/android/appnav/loggedin/SendQueues.kt
```diff
@@ -29,18 +29,19 @@ class SendQueues @Inject constructor(
...
-            .onEach { networkStatus ->
-                matrixClient.setAllSendQueuesEnabled(enabled = networkStatus == NetworkStatus.Online)
-            }
-            .launchIn(coroutineScope)
-
-        @OptIn(FlowPreview::class)
-        matrixClient.sendQueueDisabledFlow()
+        combine(
+            networkMonitor.connectivity,
+            matrixClient.sendQueueDisabledFlow(),
```
bmarty, line 40: Why do we need to listen to `matrixClient.sendQueueDisabledFlow()` here?

bmarty (approved):
LGTM, just one non-blocking question.

appnav/src/main/kotlin/io/element/android/appnav/loggedin/SendQueues.kt
ganfra: If the send queue get disabled for some reason, we want to be aware so we can enable it back (could be the case for exemple when restarting the app)

bmarty: OK, thanks, you can merge!

PR #4121, opened 2025-01-07
Make Maestro run for each PR push

jmartinesp:
## Content

Follow up of https://github.com/element-hq/element-x-android/pull/4092, since it wasn't working as expected.

- Remove the `workflow_run` configuration for the Maestro job.
- Make the 'build apk' job always run again too.
- Make sure the test results are uploaded when the Maestro job fails too (this broke somehow).

## Motivation and context

Ideally we wanted to run Maestro once the 'Build APK' flow was finished and reuse its output so we don't have to build the APK twice, but this has proven difficult to achieve, since the 2 options GH actions gives us for that are:

- `workflow_run`, which can set up dependencies in a 'when X flow finishes, run flow Y' way, and it's technically what we want, but has the downside that it'll always use the `.yml` contents of the `develop` branch, which makes it difficult to maintain or change (as proven by https://github.com/element-hq/element-x-android/pull/4092 not really working after being merged).
- `workflow_call`,  which allows a flow to explicitly start another: it worked as seen in [this commit](https://github.com/element-hq/element-x-android/pull/4121/commits/a99a0e3eda087b6b610839f5ef5e6e90406866df), but ended up in either the `Maestro` flow being run and displayed as part of the `Build APK` one or viceversa, and in both cases it was quite awkward since you couldn't see the results directly in the 'checks' UI.

So in the end we're just re-building the APK until a better option appears.

## Tests

There should be a working 'Maestro (local)' flow in the list of actions run.

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ env:
   CI_GRADLE_ARG_PROPERTIES: --stacktrace -Dsonar.gradle.skipCompile=true --no-configuration-cache
 
 jobs:
-  debug:
+  build:
     name: Build APKs
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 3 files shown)


bmarty (approved):
Thanks for the follow-up!

PR #4148, opened 2025-01-14
Remove or replace unnecessary `BackHandler` calls

jmartinesp:
## Content

- For `UserProfileView`, remove the redundant `BackHandler` -> `navigateUp()` call.
- ~~For `QrCodeErrorView` and `QrCodeConfirmationView`, remove the external `BackHandler` and move the callback to `FlowStepPage.onBackClick`~~. Restored, since they were used to prevent the back button arrow icon from appearing in the nav bar.
- For `SetupBiometricView`, remove the `enabled = true` parameter, as this is the default value.

## Motivation and context

I noticed these while investigating some issues about StackOverflowError happening when navigating back from some screens, although I don't think they're related.

## Tests

The back button/gestures should still work the same on the affected screens.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/userprofile/shared/src/main/kotlin/io/element/android/features/userprofile/shared/UserProfileView.kt
+++ b/features/userprofile/shared/src/main/kotlin/io/element/android/features/userprofile/shared/UserProfileView.kt
@@ -7,7 +7,6 @@
 
 package io.element.android.features.userprofile.shared
 
-import androidx.activity.compose.BackHandler
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.consumeWindowInsets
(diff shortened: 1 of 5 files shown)


features/userprofile/shared/src/main/kotlin/io/element/android/features/userprofile/shared/UserProfileView.kt
```diff
@@ -51,7 +50,6 @@ fun UserProfileView(
     openAvatarPreview: (username: String, url: String) -> Unit,
     modifier: Modifier = Modifier,
 ) {
-    BackHandler { goBack() }
```
bmarty, line 54: I have checked that this removal does not break the "physical" back button press and this is fine because this Node is added to a `BackStack`.

bmarty (approved):
Thanks for the cleanup!

Issue #4150, opened 2025-01-15
ANRs in `SyncService.start/stop`


jmartinesp:
We recently became aware of an abnormal growth of ANR (application not responding) issues, and the most common one seems to be one with the next stacktrace:

<details><summary>Stacktrace</summary>
<pre><code>
  #00  pc 0x000000000004dc5c  /apex/com.android.runtime/lib64/bionic/libc.so (syscall+28)
  #01  pc 0x000000000381f610  /data/app/~~kNW33dJeKDR24qAYlClfVw==/io.element.android.x-PLPFGHZP_N-S7DxV4AkKdQ==/split_config.arm64_v8a.apk
  #02  pc 0x00000000028a5f00  /data/app/~~kNW33dJeKDR24qAYlClfVw==/io.element.android.x-PLPFGHZP_N-S7DxV4AkKdQ==/split_config.arm64_v8a.apk
  #03  pc 0x0000000001b2897c  /data/app/~~kNW33dJeKDR24qAYlClfVw==/io.element.android.x-PLPFGHZP_N-S7DxV4AkKdQ==/split_config.arm64_v8a.apk (ffi_matrix_sdk_ffi_rust_future_poll_u32+124)
  #04  pc 0x000000000001404c  /data/app/~~kNW33dJeKDR24qAYlClfVw==/io.element.android.x-PLPFGHZP_N-S7DxV4AkKdQ==/split_config.arm64_v8a.apk (BuildId: ddeb5c7dcafc8a4ed4ba180f840e1967d8207a8b)
  #05  pc 0x0000000000010a78  /data/app/~~kNW33dJeKDR24qAYlClfVw==/io.element.android.x-PLPFGHZP_N-S7DxV4AkKdQ==/split_config.arm64_v8a.apk (BuildId: ddeb5c7dcafc8a4ed4ba180f840e1967d8207a8b)
  #06  pc 0x00000000000063c8  /data/app/~~kNW33dJeKDR24qAYlClfVw==/io.element.android.x-PLPFGHZP_N-S7DxV4AkKdQ==/split_config.arm64_v8a.apk (BuildId: ddeb5c7dcafc8a4ed4ba180f840e1967d8207a8b)
[...]

PR #4176, opened 2025-01-20
Create `SyncOrchestrator`

jmartinesp:
## Content

Centralise the start/stop sync logic in a single component of the app:

- Create `SyncOrchestrator`, which will act as the single point where sync start/stop can happen, based on several inputs (sync state, network available, app in foreground, app in call, app needing to sync an event for a notification).
- Add the extra cases (app in call, syncing an event) to `AppForegroundStateService`.
- Remove the start/stop sync logic from several parts of the code, replace them with toggling the inputs for the `SyncOrchestrator`.

## Motivation and context

Having too many points where the sync could be started and stopped could cause issues where the sync was unexpectedly stopped when it should instead be running or a deadlock may happen for several start/stop calls.

It should help with https://github.com/element-hq/element-x-android/issues/4150

## Tests

Just play with the sync state with network online/offline and app foreground state, receiving notifications, joining calls, etc.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/SyncOnNotifiableEvent.kt
```diff
@@ -45,14 +43,13 @@ class SyncOnNotifiableEvent @Inject constructor(
...
             // If the app is in foreground, sync is already running, so just add the subscription.
             if (!appForegroundStateService.isInForeground.value) {
-                val syncService = client.syncService()
-                syncService.startSyncIfNeeded()
                 if (isRingingCallEvent) {
                     room.waitsUntilUserIsInTheCall(timeout = 60.seconds)
                 } else {
+                    appForegroundStateService.updateIsInCallState(true)
                     room.waitsUntilEventIsKnown(eventId = notifiableEvent.eventId, timeout = 10.seconds)
+                    appForegroundStateService.updateIsInCallState(false)
```
bmarty, line 51: Using the same api may lead to wrong state if the app get a push during a call?

appnav/src/main/kotlin/io/element/android/appnav/SyncObserver.kt
```diff
@@ -0,0 +1,135 @@
...
+import timber.log.Timber
+import javax.inject.Inject
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.seconds
+
+/**
+ * Observes the app state and network state to start/stop the sync service.
+ */
+@SingleIn(SessionScope::class)
+class SyncObserver @Inject constructor(
```
jmartinesp, line 39: I don't like this name (it's not observing the sync, but the app state), but I can't come up with a better one 🫤 . Any ideas?

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -147,6 +138,9 @@ class LoggedInFlowNode @AssistedInject constructor(
 
     override fun onBuilt() {
         super.onBuilt()
+
+        syncOrchestrator.start()
```
bmarty, line 142: I think this code will not be run if the app is killed and a Push is received. In this case the sync will not start?

bmarty (commented):
Thanks! Please see my comments.

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
jmartinesp: I think you're right, and I completely overlooked this... This complicates things, as we would have to use `SyncOrchestrator` in the `AppScope` and that means holding as many as `MatrixClient`s there are in `MatrixClientsHolder`, maybe following a similar approach 🫤 .

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/sync/SyncOrchestrator.kt
```diff
@@ -0,0 +1,16 @@
...
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.matrix.api.sync
+
+/**
+ * Observes the app state and network state to start/stop the sync service.
+ */
+interface SyncOrchestrator {
```
ganfra, line 13: Not sure this class belongs the the matrix api, it's more an app thing to me

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/sync/SyncOrchestratorProvider.kt
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+package io.element.android.libraries.matrix.api.sync
+
+import io.element.android.libraries.matrix.api.core.SessionId
+
+/**
+ * Provides a [SyncOrchestrator] for a given [SessionId].
+ */
+fun interface SyncOrchestratorProvider {
```
ganfra, line 15: Same ^

appnav/src/main/kotlin/io/element/android/appnav/di/MatrixClientsHolder.kt
```diff
@@ -29,13 +30,14 @@ private const val SAVE_INSTANCE_KEY = "io.element.android.x.di.MatrixClientsHold
 @ContributesBinding(AppScope::class)
 class MatrixClientsHolder @Inject constructor(
```
ganfra, line 31: Maybe we should rename this class so it's a bit clearer it's not just about `MatrixClients`?
Also we might want to create a class instead of using Pair? 

I don't think we need the `DefaultSyncOrchestratorProvider`, we should let `MatrixClientsHolder` (with a new name :P)  implements the `SyncOrchestratorProvider`

ganfra (commented):
Some remarks

jmartinesp:
@ganfra all the points should have been fixed in the latest commit.

appnav/src/main/kotlin/io/element/android/appnav/di/MatrixSessionCache.kt
```diff
@@ -25,17 +27,27 @@ import javax.inject.Inject
...
+/**
+ * In-memory cache for logged in Matrix sessions.
+ *
+ * This component contains both the [MatrixClient] and the [SyncOrchestrator] for each session.
+ */
 @SingleIn(AppScope::class)
-@ContributesBinding(AppScope::class)
-class MatrixClientsHolder @Inject constructor(
+@ContributesBinding(AppScope::class, boundType = MatrixClientProvider::class)
+@ContributesBinding(AppScope::class, boundType = SyncOrchestratorProvider::class)
```
bmarty, line 37: TIL we can have multiple `@ContributesBinding` annotations

appnav/src/main/kotlin/io/element/android/appnav/di/MatrixSessionCache.kt
```diff
@@ -25,17 +27,27 @@ import javax.inject.Inject
...
+class MatrixSessionCache @Inject constructor(
     private val authenticationService: MatrixAuthenticationService,
-) : MatrixClientProvider {
-    private val sessionIdsToMatrixClient = ConcurrentHashMap<SessionId, MatrixClient>()
+    private val syncOrchestratorFactory: DefaultSyncOrchestrator.Factory,
+) : MatrixClientProvider, SyncOrchestratorProvider {
+    private val sessionIdsToMatrixClient = ConcurrentHashMap<SessionId, InMemoryMatrixSession>()
```
bmarty, line 42: Maybe rename this val now that it will contain `InMemoryMatrixSession`?

appnav/src/main/kotlin/io/element/android/appnav/di/DefaultSyncOrchestrator.kt
```diff
@@ -0,0 +1,148 @@
...
+                        SyncStateAction.NoOp -> Unit
+                    }
+                }
+        }
+    }
+
+    /**
+     * Stop observing the app state and network state.
+     */
+    override fun stop() {
```
bmarty, line 133: Since this method is only used for test, you should remove it.
The background coroutine scope can be used from the TestScope, and can be provided in the constructor parameter.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/SyncOnNotifiableEvent.kt
```diff
@@ -13,46 +13,50 @@ import io.element.android.libraries.featureflag.api.FeatureFlags
...
     suspend operator fun invoke(notifiableEvent: NotifiableEvent) = withContext(dispatchers.io) {
         val isRingingCallEvent = notifiableEvent is NotifiableRingingCallEvent
         if (!featureFlagService.isFeatureEnabled(FeatureFlags.SyncOnPush) && !isRingingCallEvent) {
             return@withContext
         }
         val client = matrixClientProvider.getOrRestore(notifiableEvent.sessionId).getOrNull() ?: return@withContext
+
+        // Start the sync if it wasn't already started
+        syncOrchestratorProvider.getSyncOrchestrator(notifiableEvent.sessionId)?.start() ?: return@withContext
+
```
bmarty, line 44: I am wondering if `matrixClientProvider.getOrRestore` could also take care of starting the syncOrchestrator. Maybe the sync orchestrator should always be started? In this case the interface `SyncOrchestrator` could just be private?

appnav/src/main/kotlin/io/element/android/appnav/di/DefaultSyncOrchestrator.kt
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        Timber.tag(tag).d("start observing the app and network state")
+
+        if (syncService.syncState.value != SyncState.Running) {
+            Timber.tag(tag).d("initial startSync")
+            baseCoroutineScope.launch(dispatchers.io) {
+                try {
+                    initialSyncMutex.lock()
+                    syncService.startSync()
```
bmarty, line 77: Why do we need a first `startSync` here? The combine below is not enough to decide what to do depending on the status of all the observed flows?

appnav/src/main/kotlin/io/element/android/appnav/di/DefaultSyncOrchestrator.kt
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        coroutineScope = CoroutineScope(baseCoroutineScope.coroutineContext + CoroutineName(tag) + dispatchers.io)
+
+        coroutineScope?.launch {
+            // Wait until the initial sync is done, either successfully or failing
+            initialSyncMutex.lock()
+
+            combine(
+                // small debounce to avoid spamming startSync when the state is changing quickly in case of error.
```
bmarty, line 94: Maybe the debounced can be moved after the combine block (before .distinctUntilChanged())?

services/appnavstate/test/src/main/kotlin/io/element/android/services/appnavstate/test/FakeAppForegroundStateService.kt
```diff
@@ -9,19 +9,29 @@ package io.element.android.services.appnavstate.test
...
+    override fun startObservingForeground() {
         // No-op
     }
 
     fun givenIsInForeground(isInForeground: Boolean) {
-        state.value = isInForeground
+        this.isInForeground.value = isInForeground
+    }
+
+    override fun updateIsInCallState(isInCall: Boolean) {
```
bmarty, line 30: Only used in test, can you double check please?

bmarty (changes requested):
A new bunch of comments, I have not run the code yet

appnav/src/main/kotlin/io/element/android/appnav/di/DefaultSyncOrchestrator.kt
jmartinesp: I was mimicking what we had before in `LoggedInFlowNode` for the airgapped case. After the changes in the SDK it 's probably not needed anymore.

services/appnavstate/test/src/main/kotlin/io/element/android/services/appnavstate/test/FakeAppForegroundStateService.kt
jmartinesp: It's not though?

image

You're looking at the fake implementation, so it's normal that one will only be used in tests.

appnav/src/main/kotlin/io/element/android/appnav/di/DefaultSyncOrchestrator.kt
jmartinesp: This causes issues in the tests, since it's preventing some intermediate states from being generated. I'd rather keep it were it is, if possible.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/SyncOnNotifiableEvent.kt
jmartinesp: Good call, this simplified the code quite a bit!

appnav/src/main/kotlin/io/element/android/appnav/di/MatrixSessionCache.kt
jmartinesp: Well, we don't have them anymore after the latest changes 😅 .

appnav/src/main/kotlin/io/element/android/appnav/di/DefaultSyncOrchestrator.kt
```diff
@@ -49,10 +48,10 @@ class DefaultSyncOrchestrator @AssistedInject constructor(
 
     private val initialSyncMutex = Mutex()
 
-    private var coroutineScope: CoroutineScope? = null
-
     private val tag = "SyncOrchestrator"
 
+    private val coroutineScope = CoroutineScope(baseCoroutineScope.coroutineContext + CoroutineName(tag) + dispatchers.io)
+
```
bmarty: I guess this scope is cancelled when the baseCoroutineScope is cancelled.

Have you tested what happen when the session is logged out?

It should work as expected, but maybe worth double checking.

appnav/src/test/kotlin/io/element/android/appnav/SyncOrchestratorTest.kt
```diff
@@ -0,0 +1,312 @@
...
+
+    @Test
+    fun `when the app goes to background and the sync was running, it will be stopped after a delay`() = runTest {
+        val stopSyncRecorder = lambdaRecorder<Result<Unit>> { Result.success(Unit) }
+        val syncService = FakeSyncService(initialSyncState = SyncState.Running).apply {
+            stopSyncLambda = stopSyncRecorder
+        }
+        val networkMonitor = FakeNetworkMonitor(initialStatus = NetworkStatus.Connected)
+        val appForegroundStateService = FakeAppForegroundStateService(initialForegroundValue = true)
+        val coroutineScope = CoroutineScope(coroutineContext + SupervisorJob())
```
bmarty, line 45: Why not using `backgroundScope` from `TestScope` here? `backgroundScope` is cancelled automatically at the end of the test.

appnav/src/test/kotlin/io/element/android/appnav/di/MatrixSessionCacheTest.kt
```diff
@@ -0,0 +1,155 @@
...
+        val loginSucceeded = fakeAuthenticationService.login("user", "pass")
+
+        assertThat(loginSucceeded.isSuccess).isTrue()
+        assertThat(matrixSessionCache.getOrNull(A_SESSION_ID)).isNotNull()
+
+        coroutineScope.cancel()
+    }
+
+    private fun TestScope.createSyncOrchestratorFactory(
+        baseCoroutineScope: CoroutineScope = this,
```
bmarty, line 143: Same remark about using the `backgroundScope`. This parameter can be removed.

appnav/src/main/kotlin/io/element/android/appnav/di/MatrixSessionCache.kt
```diff
@@ -25,45 +26,61 @@ import javax.inject.Inject
...
 
+/**
+ * In-memory cache for logged in Matrix sessions.
+ *
+ * This component contains both the [MatrixClient] and the [SyncOrchestrator] for each session.
+ */
 @SingleIn(AppScope::class)
-@ContributesBinding(AppScope::class)
-class MatrixClientsHolder @Inject constructor(
+@ContributesBinding(AppScope::class, boundType = MatrixClientProvider::class)
```
bmarty, line 35: Nit: the `boundType` can be removed, this class now implement only 1 interface :)

appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
```diff
@@ -0,0 +1,109 @@
...
+                appForegroundStateService.isInCall,
+                appForegroundStateService.isSyncingNotificationEvent,
+            ) { syncState, networkState, isInForeground, isInCall, isSyncingNotificationEvent ->
+                val isAppActive = isInForeground || isInCall || isSyncingNotificationEvent
+                val isNetworkAvailable = networkState == NetworkStatus.Connected
+
+                Timber.tag(tag).d("isAppActive=$isAppActive, isNetworkAvailable=$isNetworkAvailable")
+                if (syncState == SyncState.Running && !isAppActive) {
+                    // Don't stop the sync immediately, wait a bit to avoid starting/stopping the sync too often
+                    delay(3.seconds)
```
bmarty, line 81: Are you sure this delay will act as a mechanism for `SyncStateAction.StopSync` not being emitted? I think for this to work we need to use `flatmapLatest` and add a delay when the value is `SyncStateAction.StopSync`.

Could be something like

```
                .flatMapLatest {
                    when(it) {
                        SyncStateAction.StartSync -> flowOf(SyncStateAction.StartSync)
                        SyncStateAction.StopSync -> {
                            // Don't stop the sync immediately, wait a bit to avoid starting/stopping the sync too often
                            delay(3.seconds)
                            flowOf(SyncStateAction.StopSync)
                        }
                        SyncStateAction.NoOp -> emptyFlow()
                    }
                }
```

bmarty (changes requested):
Some more remarks

appnav/src/main/kotlin/io/element/android/appnav/di/DefaultSyncOrchestrator.kt
jmartinesp: Good call, thanks to this we found a critical error. Patch incoming!

appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
jmartinesp: I can confirm this works as expected, I tested this manually, but in any case if you modify the `advanceTimeBy(10.seconds)` to <3s the tests that check this behaviour won't pass anymore. FWIW I've added a new test to double check this, although I think there was already some other test that checked a similar case.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
```diff
@@ -495,31 +495,32 @@ class RustMatrixClient(
...
+//                    }
+//                }
+//            }
+//            close()
+//
+//            deleteSessionDirectory(deleteCryptoDb = true)
+//            if (userInitiated) {
+//                sessionStore.removeSession(sessionId.value)
+//            }
+//        }
```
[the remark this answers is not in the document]
jmartinesp: Yes, I forgot to uncomment this after my tests. Thanks!

services/appnavstate/api/src/main/kotlin/io/element/android/services/appnavstate/api/AppForegroundStateService.kt
```diff
@@ -18,8 +18,28 @@ interface AppForegroundStateService {
...
+
+    /**
+     * Update the in-call state.
+     */
+    fun updateIsInCallState(isInCall: Boolean)
+
+    /**
+     * Update the active state for the syncing notification event flow.
+     */
+    fun updateIsSyncingNotificationEvent(isSyncingNotificationEvent: Boolean)
```
bmarty, line 44: It's weird to add all those new methods to the interface `AppForegroundService`.

Would it be possible to have a interface in the module `:libraries:architecture`

```
interface SyncRequester {
    val needSync: StateFlow<Boolean>
}
```

that can be implemented by any module who need the sync to start using `ContributesMultibinding` and then let the `SyncOrchestrator` collect all the states?

It will suit more our architecture patterns and will simplify I think the implementation.

I am trying to poc this.

bmarty: I see a problem which is maybe more important if one day we support multi session, I do not think that this will behave as expected, i.e., I think that if a call start on one session, all the session will resume the sync. I am thinking out loud, it's maybe not the case, but we need to check that. Maybe by providing the sessionId to all those new API to be able to do some post filtering.

jmartinesp: > I see a problem which is maybe more important if one day we support multi session

I had the same thought, but IMO we should cross that bridge when we come to it, I think we'd just overcomplicate things for no reason if we did the change now.

appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
jmartinesp: I don't think so: not only the tests work, but also, in coroutines, `combine` has the behaviour of `combineLatest` in other Rx implementations, so it should cancel the previous iteration when a new one is emitted.

bmarty: TIL !

services/appnavstate/api/src/main/kotlin/io/element/android/services/appnavstate/api/AppForegroundStateService.kt
bmarty: OK

bmarty (approved):
Thanks for the update.
I think the tests could be simplified a bit, now that we do not start an initial sync.
Then I believe that the PR can be merged to be in the latest nightly.

bmarty (changes requested):
I have found something, please wait before merging, I will comment

appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
bmarty: Actually there is a pb in the test, that why it's passing, but this is not OK, please see my other comment.

appnav/src/test/kotlin/io/element/android/appnav/SyncOrchestratorTest.kt
```diff
@@ -0,0 +1,320 @@
...
+        stopSyncRecorder.assertions().isNeverCalled()
+        advanceTimeBy(1.seconds)
+        appForegroundStateService.isInForeground.value = true
+        advanceTimeBy(1.seconds)
+
+        // Ensure the stop action wasn't called yet either, since we didn't give it enough time to emit after the expected delay
+        stopSyncRecorder.assertions().isNeverCalled()
+
+        // Now change it again and wait for enough time
+        appForegroundStateService.isInForeground.value = false
```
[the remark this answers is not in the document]
jmartinesp: Yes, seems that way. Which is weird, because I can see the debugger stops at the `delay(3.seconds)`, so the flow is suspended, but then advancing the clock in the scheduler for any amount (1ms even) skips the delay.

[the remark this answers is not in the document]
bmarty: it skips the delay, but it does not cancel the flow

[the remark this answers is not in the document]
jmartinesp: That's what I mean, advancing the clock by 1ms should not skip a delay of 3 seconds. I guess I had the wrong idea about how a delay call would work inside these operators.

appnav/src/test/kotlin/io/element/android/appnav/SyncOrchestratorTest.kt
```diff
@@ -0,0 +1,320 @@
...
+        }
+        val networkMonitor = FakeNetworkMonitor(initialStatus = NetworkStatus.Connected)
+        val appForegroundStateService = FakeAppForegroundStateService(initialForegroundValue = true)
+        val syncOrchestrator = createSyncOrchestrator(
+            syncService = syncService,
+            networkMonitor = networkMonitor,
+            appForegroundStateService = appForegroundStateService,
+        )
+
+        // We start observing, we skip the initial sync attempt since the state is running
```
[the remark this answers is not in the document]
jmartinesp: Actually, not entirely. Since we have the debounce in the `syncState`, removing this would mean checking the orchestrator state when it still hasn't had time to emit anything. So we'd need at least an initial delay of 100ms.

appnav/src/test/kotlin/io/element/android/appnav/SyncOrchestratorTest.kt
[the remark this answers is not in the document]
bmarty: do delays have effect in unit test?

appnav/src/main/kotlin/io/element/android/appnav/di/DefaultSyncOrchestrator.kt
bmarty: Can you try with this patch?

[moveDebounce.patch](https://github.com/user-attachments/files/18691006/moveDebounce.patch)

appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
bmarty: I think it's fixed now, with the debounce {} usage. Thanks!

bmarty (approved):

PR #4187, opened 2025-01-23
Migrate to CalVer like versioning

bmarty:
## Content

In order to simplify the versioning, we are migrating both Element X applications to a CalVer like system.
4 digits for the year, 2 digits for the month and 2 digits for the release number of the month.

This PR update the current value and update the release script.
The release script will update the version during the release and will no more set the next version after the release is done.

## Motivation and context

(issue is coming)

## Screenshots / GIFs

NA

## Tests

- Start a release and see what is the proposed release version

We will test that everything is fine during the next release

- Try to upgrade the application using a build with this new system. It should work, i.e. the new versionCode will always be higher that any previous versionCode.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

If you navigate to the settings:

image

## Checklist

### DIFF
--- a/plugins/src/main/kotlin/Versions.kt
+++ b/plugins/src/main/kotlin/Versions.kt
@@ -9,49 +9,30 @@ import org.gradle.api.JavaVersion
 import org.gradle.jvm.toolchain.JavaLanguageVersion
 
 /**
- * Version codes are quite sensitive, because there is a mix between bundle and APKs, and we have to take into
- * account the future upgrade of Element Android.
+ * Version codes are quite sensitive, because there is a mix between bundle and APKs.
  * Max versionCode allowed by the PlayStore (for information):
  * 2_100_000_000
- * Current version code of EAx on the PlayStore, for the first uploaded beta (we cannot go below):
- * ----1_001_000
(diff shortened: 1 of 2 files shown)


ganfra (approved):
Looking good, thanks!

plugins/src/main/kotlin/Versions.kt
```diff
@@ -9,49 +9,30 @@ import org.gradle.api.JavaVersion
...
- * ---40_001_020
- * Nightly version:
- * ---40_001_020
- * Debug version:
- * ---40_010_200
+ * We are using a CalVer-like approach to version the application. The version code is calculated as follows:
+ * - 4 digits for the year
+ * - 2 digits for the month
+ * - 2 digits for the release number
+ * So for instance, the first release of Jan 2025 will have the version code: 20250100 (20_250_100)
```
jmartinesp, line 24: Won't this go below the `40_XXX_YYY` versions we have at the moment?

bmarty: This version code will be multiplied by 10 to add a digit for the arch.

PR #4192, opened 2025-01-23
CalVer: use 2 digits for the year and 2 digits for the month.

bmarty:
Complement for #4187, following requirements from https://github.com/element-hq/element-meta/issues/2664

This is how it will appear in the app:

image


jmartinesp (approved):

bmarty:
Thanks for the review Jorge, I think I have to restore the code that increase the release number after the release to be fully compliant with https://github.com/element-hq/element-meta/issues/2664
It's not blocking for the next release, so let's do that later...

PR #4198, opened 2025-01-26
Open chat links in regular browser tabs

### LINKED ISSUE #3885: Links open in in-app browser
### Steps to reproduce

Click on a link in the chat

#### What did you expect?

Link opens in external browser

#### What happened instead?

Link opens in some sort of in-app web view. This is okayyyyy (if ugly), but the real problem is that there is no way to keep the web view open if I want to go back to chatting.

### Your phone model

Samsung S22

### Operating system version

Android 14

0.7.3-nithgly 666cedd6

### Homeserver

_No response_

### Will you send logs?

No

No

cbs228:
## Content

Open web hyperlinks in chat messages in the user's regular, preferred web browser. Fixes bug introduced in v0.6.5.

Closes #3885

## Motivation and context

In 5baefd479f0045110918b5abbcb7a21fbf28c01e, a link to a "help" web page was added to the `MessagesView`. Since this page is essentially part of the app, it should and does open in an "in-app" [Chrome Custom Tab](https://developer.chrome.com/docs/android/custom-tabs/guide-get-started#opening_a_custom_tab
), where supported.

Unfortunately, this change affects *every* link in the messages view, including links sent in chat messages. As of v0.6.5, links in chat messages now open in a Custom Tab instead of the user's configured "full" web browser. This is a UX nuisance.

Separate "regular" links from "custom tab" links with a new parameter `onLinkClick(..., customTab)`. If true, the link opens in a custom tab.

Alternatively, the first patch in this PR removes the Custom Tab behavior from `MessagesView` entirely. This is likely not the preferred end result, however.

## Screenshots / GIFs

|Before|After|
|-|-|
| Screenshot_20250125-211827_Fennec |Screenshot_20250125-211852_Fennec |

## Tests

1. Build F-Droid version with
   ```
   export JAVA_HOME=/usr/lib/jvm/java-21-openjdk-amd64/
   export GRADLE_OPTS="-Dorg.gradle.jvmargs=-Xmx9g -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8 -XX:+HeapDumpOnOutOfMemoryError -XX:+UseG1GC -Dkotlin.daemon.jvm.options=-Xmx4g"
   ./gradlew app:assembleFDroidDebug
   ```

2. Install on device.

3. Sign in to Element.

4. Open a chat room. Find a message which includes an `https://` web link.

5. Tap on the link.

6. Observe that the link is opened in the "full" web browser with the full browser UI and other tabs. (See screenshots.)

I have not tried to force an identity change to test that the identity change help still loads in a Chrome Custom Tab. I'm not sure what the best way to test that is.

This PR should not introduce any new Android APIs. If API 24 emulator testing is required, can the project recommend an emulator?

## Tested devices

- [x] Physical: Android 14 / LineageOS 21, with f-droid build and the Fennec F-Droid web browser
- [ ] Emulator

## Checklist

- [x] UI change has been tested on both light and dark themes (N/A, but tested anyway)

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
@@ -111,7 +111,7 @@ fun MessagesView(
     onRoomDetailsClick: () -> Unit,
     onEventContentClick: (event: TimelineItem.Event) -> Boolean,
     onUserDataClick: (UserId) -> Unit,
-    onLinkClick: (String) -> Unit,
+    onLinkClick: (String, Boolean) -> Unit,
     onSendLocationClick: () -> Unit,
     onCreatePollClick: () -> Unit,
     onJoinCallClick: () -> Unit,
@@ -273,7 +273,7 @@ private fun MessagesViewContent(
     state: MessagesState,
     onContentClick: (TimelineItem.Event) -> Unit,
     onUserDataClick: (UserId) -> Unit,
-    onLinkClick: (String) -> Unit,
+    onLinkClick: (String, Boolean) -> Unit,
     onReactionClick: (key: String, TimelineItem.Event) -> Unit,
     onReactionLongClick: (key: String, TimelineItem.Event) -> Unit,
     onMoreReactionsClick: (TimelineItem.Event) -> Unit,
@@ -347,7 +347,7 @@ private fun MessagesViewContent(
                         state = state.timelineState,
                         timelineProtectionState = state.timelineProtectionState,
                         onUserDataClick = onUserDataClick,
-                        onLinkClick = onLinkClick,
+                        onLinkClick = { url -> onLinkClick(url, false) },
                         onContentClick = onContentClick,
                         onMessageLongClick = onMessageLongClick,
                         onSwipeToReply = onSwipeToReply,
@@ -396,7 +396,7 @@ private fun MessagesViewContent(
 private fun MessagesViewComposerBottomSheetContents(
     subcomposing: Boolean,
     state: MessagesState,
-    onLinkClick: (String) -> Unit,
+    onLinkClick: (String, Boolean) -> Unit,
 ) {
     if (state.userEventPermissions.canSendMessage) {
         Column(modifier = Modifier.fillMaxWidth()) {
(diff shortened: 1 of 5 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
```diff
@@ -112,6 +112,7 @@ fun MessagesView(
     onEventContentClick: (event: TimelineItem.Event) -> Boolean,
     onUserDataClick: (UserId) -> Unit,
     onLinkClick: (String) -> Unit,
+    onHelpLinkClick: (String) -> Unit,
```
bmarty, line 115: It was maybe simpler to add a Boolean parameter to the lambda `onLinkClick: (String) -> Unit,` to allow the usage of CustomTab if true.

bmarty (changes requested):
Thanks for the proposal. I think it makes sense to open links using the default browser since we do not have specific action like sharing like the X app may have.

Can you handle my remark about the lambda parameter? I believe it will do the same thing but with less change on the codebase.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
cbs228: We definitely save a few lines of code this way. It might be more expressive to use an enum here—it's kind of hard to tell what the Boolean does when you're deep in the call tree—but that would definitely add complexity.

Ready for re-review.

cbs228:
@bmarty, I think I've worked through all the comments and checklists. I don't have permission to run CI. Is there anything more I need to do for this PR?

bmarty (approved):
Thanks for the update, it looks good to me. I have triggered the CI.

bmarty (approved):

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
bmarty: Thanks for the update, I have merged your PR.
FYI it's possible to give more info about the parameter of the lambda by providing names, for instance:
```kotlin
onLinkClick: (url: String, useCustomTab: Boolean) -> Unit,
```
In this case I think it's OK to keep it like this.
thanks!

richvdh:
Thank you for the fix!

PR #4201, opened 2025-01-27
Media Viewer: show snackbar when reaching end of timeline.

bmarty:
## Content

Media Viewer: show snackbar when reaching end of timeline.

When the loading page is displayed and there is no more media to load, the loading page vanished (this is handled automatically by the HorizontalPager - without animation (is does not seem to be supported yet by the library) and a Snackbar is displayed. This PR handle the displaying of the snackbar.

## Motivation and context

Follow design from Figma https://www.figma.com/design/Ni6Ii8YKtmXCKYNE90cC67/Timeline-(new)?node-id=2660-29699&m=dev 

## Tests

- Open a room with at least one media
- navigate to "Media and files" and open the media, then quickly swipe to right to see the loader
- wait for the pagination request to reach the beginning of the room and observe that the media viewer navigate to the previous item and a snckbar is displayed.

## Tested devices

- [ ] Physical
- [x] Emulator using hack code to simulate the scenario above.
- OS version(s):

## Checklist

### DIFF
--- a/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerPresenter.kt
+++ b/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerPresenter.kt
@@ -10,12 +10,15 @@ package io.element.android.libraries.mediaviewer.impl.viewer
 import android.content.ActivityNotFoundException
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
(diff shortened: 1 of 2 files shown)


libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerPresenter.kt
```diff
@@ -64,6 +72,37 @@ class MediaViewerPresenter @AssistedInject constructor(
         var currentIndex by remember { mutableIntStateOf(searchIndex(data, inputs.eventId)) }
         val snackbarMessage by snackbarDispatcher.collectSnackbarMessageAsState()
 
+        val isRenderingLoadingBackward by remember {
```
ganfra, line 75: This will cause the the Present method to recompose, I think you should try to manage this only using snapshot flows inside the **LaunchedEffect**. 
Maybe also extract that into a separate method.

bmarty: I have extracted to sub method. For the recomposition issue can we handle it later?

Issue #4208, opened 2025-01-28
Linkified text is too greedy and gobbles punctuation


edent:
### Steps to reproduce

![Image](https://github.com/user-attachments/assets/9ee74f06-58ad-4f7a-90b5-2a095d739b2f)

#### What did you expect?

Punctuation at the end of a URl shouldn't be included in that URl.

#### What happened instead?

The link goes to a 404.

I suspect this is related to this Android bug - https://issuetracker.google.com/issues/161727315

### Your phone model

Google Pixel 8Pro

### Operating system version

Android 15

### Application version and app store

Element X 0.7.6 Play Store

### Homeserver

matrix.org

### Will you send logs?

No

### Are you willing to provide a PR?

No

codesalatdev:
Not to be _that_ guy, but parenthesis and punctuation are perfectly valid parts of a well-formed URI as described by [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A)

While I personally agree with the fact that it's _really_ annoying to type, the RFC also states that:

> In practice, URIs are delimited in a variety of ways, but usually
   within double-quotes "http://example.com/", angle brackets
   <http://example.com/>, or just by using whitespace

Take the search query "symbol )" on duckduckgo as an example: `https://duckduckgo.com/?t=h_&q=symbol+)`. Not including the parenthesis in the URI would alter the search results significantly.

edent:
I don't experience this behaviour in Slack or Teams for Android.

You can user-blame all you want, but I don't think that will attract more participants to your platform. Your response makes me a lot less likely to use or recommend Element.

JadedBlueEyes:
> Not to be _that_ guy, but parenthesis and punctuation are perfectly valid parts of a well-formed URI as described by [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A)
> 
> While I personally agree with the fact that it's _really_ annoying to type, the RFC also states that:
> 
> > In practice, URIs are delimited in a variety of ways, but usually
>    within double-quotes "http://example.com/", angle brackets
>    \<http://example.com/&gt;, or just by using whitespace
> 
> Take the search query "symbol )" on duckduckgo as an example: `https://duckduckgo.com/?t=h_&q=symbol+)`. Not including the parenthesis in the URI would alter the search results significantly.

Auto-linkification is a convenience feature. It doesn't need to follow a spec, it should correctly highlight as many links as possible. 

[...]

PR #4214, opened 2025-01-29
Try avoiding trailing punctuation inside linkified URLs

jmartinesp:
## Content

When an URL is linkified automatically by `LinkifyCompat`, try removing any trailing punctuation and other mismatched symbols from the link span.

This solution is not great and might fail in some cases, but it should cover the most common ones, since it seems like a 'smart' linkification library is hard to find as a replacement.

## Motivation and context

It turns out the `Linkify` component from Google isn't as smart as we thought: as soon as a `/` char is found inside the URL, it'll greedily fetch as much text as possible. This is probably because a different regex is used internally than for a simple domain.

Should fix https://github.com/element-hq/element-x-android/issues/4208.

## Screenshots / GIFs

|Before|After|
|-|-|
|image|image|

## Tests

You can play with the `ClickableLinkTextWithParenthesisPreview` composable preview, or send messages with URLs and problematic formats to be linkified.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/androidutils/src/test/kotlin/io/element/android/libraries/androidutils/text/LinkifierHelperTest.kt
+++ b/libraries/androidutils/src/test/kotlin/io/element/android/libraries/androidutils/text/LinkifierHelperTest.kt
@@ -1,0 +1,124 @@
+/*
+ * Copyright 2025 New Vector Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.androidutils.text
+
+import android.telephony.TelephonyManager
+import android.text.style.URLSpan
+import androidx.core.text.getSpans
+import androidx.core.text.toSpannable
+import com.google.common.truth.Truth.assertThat
+import io.element.android.tests.testutils.WarmUpRule
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.Shadows.shadowOf
+import org.robolectric.annotation.Config
+import org.robolectric.shadow.api.Shadow.newInstanceOf
+
+@RunWith(RobolectricTestRunner::class)
+class LinkifierHelperTest {
+    @get:Rule
+    val warmUpRule = WarmUpRule()
+
+    @Test
+    fun `linkification finds URL`() {
+        val text = "A url https://matrix.org"
+        val result = LinkifyHelper.linkify(text)
+        val urlSpans = result.toSpannable().getSpans<URLSpan>()
+        assertThat(urlSpans.size).isEqualTo(1)
+        assertThat(urlSpans.first().url).isEqualTo("https://matrix.org")
+    }
+
+    @Test
+    fun `linkification finds partial URL`() {
+        val text = "A partial url matrix.org/test"
+        val result = LinkifyHelper.linkify(text)
+        val urlSpans = result.toSpannable().getSpans<URLSpan>()
+        assertThat(urlSpans.size).isEqualTo(1)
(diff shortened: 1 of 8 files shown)


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/ClickableLinkText.kt
```diff
@@ -174,3 +203,15 @@ internal fun ClickableLinkTextPreview() = ElementThemedPreview {
         interactionSource = remember { MutableInteractionSource() },
     )
 }
+
+@Preview(group = PreviewGroup.Text)
+@Composable
+internal fun ClickableLinkTextWithParenthesisPreview() = ElementThemedPreview {
+    ClickableLinkText(
+        annotatedString = AnnotatedString("Does this work (url: github.com/element-hq/element-x-android/README?)?!.", ParagraphStyle()),
```
ganfra, line 211: Maybe add exemple where parenthesis is part of the url?

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/ClickableLinkText.kt
```diff
@@ -143,10 +143,19 @@ fun AnnotatedString.linkify(linkStyle: SpanStyle): AnnotatedString {
...
                 // Prevent linkifying domains in user or room handles (@user:domain.com, #room:domain.com)
                 if (start > 0 && !spannable[start - 1].isWhitespace()) continue
+
+                // Try to avoid including trailing punctuation in the link.
+                // Since this might fail in some edge cases, we catch the exception and just use the original end index.
+                runCatching {
+                    end = adjustEndIndex(spannable, start, end)
+                }.onFailure {
+                     Timber.e(it, "Failed to adjust end index for link span")
+                }
```
bmarty, line 157: This is adjusting the `end` of the span, so this is fixing the rendering issue, but it does this affect the URL itself, i.e. the target of the link.

I think the line `annotation = span.url,` below needs to be updated.

bmarty (changes requested):
Thanks, I have tested and the target link need to be updated as well.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/ClickableLinkText.kt
bmarty: It's maybe because it did not work with the example I tried:

image

bmarty: Does the Event renderer use `ClickableLinkText`? It seems it's not the case 😅

jmartinesp:
This should be ready to review again: it's now affecting the intended texts (🤦 ) and has some extra unit tests for the new component.

libraries/androidutils/src/test/kotlin/io/element/android/libraries/androidutils/text/LinkifierHelperTest.kt
```diff
@@ -0,0 +1,124 @@
...
+
+    @Test
+    fun `linkification handles parenthesis in URL`() {
+        val text = "A url: (https://github.com/element-hq/element-android/READ(ME))"
+        val result = LinkifyHelper.linkify(text)
+        val urlSpans = result.toSpannable().getSpans<URLSpan>(start = 0, end = result.length)
+        assertThat(urlSpans.size).isEqualTo(1)
+        assertThat(urlSpans.first().url).isEqualTo("https://github.com/element-hq/element-android/READ(ME)")
+    }
+}
```
bmarty, line 124: Thanks for all those tests!

bmarty (approved):
Thanks and sorry for the late review. Good to be merged, I believe that I wrote only non-blocking remarks.

PR #4215, opened 2025-01-29
Let the SDK provide the "network is available information" 

bmarty:
## Content

We may have some trouble with the current network detection, since it rely on the ability to reach random server (not to name Google's one) and this can be broken when the application is run in air-gapped environment.

This PR move the network detection to the SDK providing the ability to perform sync request or not.

Need https://github.com/matrix-org/matrix-rust-sdk/pull/4592, which is not merged yet.

## Motivation and context

Improve network detection and bind it to the ability to communicate with the homeserver.

This will fix the "no network" banner which was displayed when the application was run in air-gapped mode.

## Tests

- Open the app and see the room list
- set airplane mode on
- the no network banner is displayed
- set airplane mode off
- the no network banner go away and the app can be used normally

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


features/networkmonitor/impl/src/main/kotlin/io/element/android/features/networkmonitor/impl/DefaultNetworkMonitor.kt
```diff
@@ -5,97 +5,53 @@
...
+                .value
+                .toNetworkStatus()
+        )
+
+    private fun SyncState.toNetworkStatus(): NetworkStatus = when (this) {
+        SyncState.Idle,
+        SyncState.Running,
+        SyncState.Error,
+        SyncState.Terminated -> NetworkStatus.Online
+        SyncState.Offline -> NetworkStatus.Offline
```
jmartinesp, line 55: Shouldn't only `Running` be `Online`?

bmarty: Since `NetworkStatus.Offline` will make the "no network" banner to be shown, I think it's better to be optimistic here.

features/networkmonitor/impl/src/main/kotlin/io/element/android/features/networkmonitor/impl/DefaultNetworkMonitor.kt
```diff
@@ -5,97 +5,53 @@
...
-    }
-
-    private fun NetworkCapabilities.getNetworkStatus(): NetworkStatus {
-        val hasInternet = hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)
-        return if (hasInternet) {
-            NetworkStatus.Online
-        } else {
-            NetworkStatus.Offline
-        }
+        .stateIn(
```
ganfra, line 41: I think when we discussed at the time, that the network monitor was still supposed to detect network changes, so we can re-trigger a sync as soon as possible (to avoid waiting for the auto retry timeout in those cases).

Also, I think we can remove the `StateFlow` from here and just expose an extension method on the `SyncState` like `isConnected()`

bmarty:
@ganfra @jmartinesp I have reworked the PR, it's working fine on my side. Let me know what you think, then once merged we can do the release.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/sync/SyncState.kt
```diff
@@ -14,3 +14,11 @@ enum class SyncState {
     Terminated,
     Offline,
 }
+
+fun SyncState.isConnected() = when (this) {
```
ganfra, line 18: I know the name was from me initially, but maybe we should instead call it `isOffline`?
`fun SyncState.isOffline() = this is SyncState.Offline`.

Or even an extension method on the `SyncService`
`fun SyncService.isOffline(): StateFlow<Boolean> = syncState.mapState { it == SyncState.Offline }`

ganfra (approved):
One remark regarding SyncState/SyncService.

Also, I think we need to change the `observeSyncStateAndNetworkStatus` in `LoggedInFlowNode` : we don't need to listen to the sync state now, we can just listen to the network monitor state.
Note : it'll be changed anyway by Jorge [PR](https://github.com/element-hq/element-x-android/pull/4176)

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/sync/SyncState.kt
bmarty: Can we do the opposite
`fun SyncService.isOnline(): StateFlow<Boolean> = syncState.mapState { it != SyncState.Offline }`
it will be simpler to use.

PR #4235, opened 2025-02-05
Fix nightly reports

bmarty:
First commit is just moving test classes to the correct package.

Second commit fix the test in release mode: `./gradlew :features:roomdetails:impl:testReleaseUnitTest` was failing before this change because RoomAlias does not fail in release mode when the room alias is invalid.

### DIFF
--- a/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/MatrixClient.kt
+++ b/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/MatrixClient.kt
@@ -7,7 +7,9 @@
 
 package io.element.android.libraries.matrix.api
 
+import io.element.android.libraries.core.data.tryOrNull
 import io.element.android.libraries.matrix.api.core.DeviceId
+import io.element.android.libraries.matrix.api.core.MatrixPatterns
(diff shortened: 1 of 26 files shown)


jmartinesp (approved):
Thanks!

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/MatrixClient.kt
```diff
@@ -172,11 +174,12 @@ fun MatrixClient.getRoomInfoFlow(roomIdOrAlias: RoomIdOrAlias): Flow<Optional<Ma
...
  */
-fun MatrixClient.roomAliasFromName(name: String): Result<RoomAlias> {
-    return runCatching {
-        RoomAlias("#$name:${userIdServerName()}")
-    }
+fun MatrixClient.roomAliasFromName(name: String): RoomAlias? {
+    return name.takeIf { it.isNotEmpty() }
+        ?.let { "#$it:${userIdServerName()}" }
+        ?.takeIf { MatrixPatterns.isRoomAlias(it) }
+        ?.let { tryOrNull { RoomAlias(it) } }
```
bmarty, line 184: @ganfra do you agree with this change?

PR #4239, opened 2025-02-06
Fix nightly reports - next step

bmarty:
Following #4235 

Add some tests to fix this violation:

```
* What went wrong:
Execution failed for task ':app:koverVerifyStates'.
> Rule 'Check code coverage of states' violated:
    instructions covered percentage for class 'io.element.android.features.lockscreen.impl.unlock.PinUnlockState' is 79.245300, but expected minimum is 90
    instructions covered percentage for class 'io.element.android.features.securebackup.impl.root.SecureBackupRootState' is 65.625000, but expected minimum is 90
    instructions covered percentage for class 'io.element.android.libraries.mediaviewer.impl.local.player.MediaPlayerControllerState' is 0.000000, but expected minimum is 90
    instructions covered percentage for class 'io.element.android.libraries.textcomposer.components.FormattingOptionState' is 0.000000, but expected minimum is 90
```

See https://github.com/element-hq/element-x-android/actions/runs/13176799102/job/36777908312


bmarty:
One more thing to fix and we will be green

image

PR #4242, opened 2025-02-07
SyncOrchestrator: restore the initial sync step

jmartinesp:
## Content

Restore the initial sync in the `SyncOrchestrator.start()` method. However, this is now split into a private `observeStates` method, that purely observers the different app states that lead to starting/stopping the sync service, and the `start` one that has the initial sync too. This allows us to test both behaviours separately, as well has writing tests for the transition between them.

## Motivation and context

Without this initial sync, the `SyncState` will be `Idle`, which means 'not started', not 'offline', so some UI didn't react accordingly (the `ConnectivityIndicatorView`, i.e.), giving the false impression of the app having network connectivity when it was started *without* network connectivity. If we do a initial sync, it'll fail and return `SyncState.Offline`, which will be correctly represented in the UI.

## Tests

- Start the app with no network connectivity.
- The offline indicator should be displayed.
- Everything else should keep working as expected.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
@@ -21,9 +21,8 @@ import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.debounce
 import kotlinx.coroutines.flow.distinctUntilChanged
-import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onCompletion
-import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.launch
 import timber.log.Timber
 import java.util.concurrent.atomic.AtomicBoolean
 import kotlin.time.Duration.Companion.milliseconds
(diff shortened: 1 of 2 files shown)


appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
```diff
@@ -52,56 +51,70 @@ class SyncOrchestrator @AssistedInject constructor(
...
      * Before observing the state, a first attempt at starting the sync service will happen if it's not already running.
+     *
+     * @param performInitialSync If `true`, the initial sync will be performed before starting observing the state. This is the default behaviour.
+     *
+     * **IMPORTANT: this should only be `false` in tests.**
      */
     @OptIn(FlowPreview::class)
-    fun start() {
+    fun start(
+        performInitialSync: Boolean = true,
```
bmarty, line 61: About the naming, `Initial Sync` was a concept for sync v2, so it may be confusing. Maybe rename to `performFirstSync`, `alwaysCallStartSync` or something else?

That said, if this parameter is just to ease how tests work, I think it should just be removed and the tests need to be updated. A private method could be used in the test class to wait for this first sync to be done.

jmartinesp: I can split this into 2 methods: the public `start()` one, performing the forced first sync and an internal one that does the observation, called from this start one, and which only observes the state. This way we can check both what happens with the forced sync and without it in the tests, making them simpler as you suggested in the previous PR.

bmarty: I do not follow here, ideally we should not write production code for the tests, but write test for the production code. Do I miss anything?

bmarty: (if the production needs the force sync, there is no reason to write test for code which does not perform the force sync.)

jmartinesp: We should write code that's easy to test, and I'm suggesting splitting the implementation here into:

- `@VisibleForTesting internal fun observeStates()` or something similar, which contains the combine the observing part and triggers starts/stops of syncs. It's not private because that would make it impossible to call from tests.
- `fun start()`, which does the first sync, then calls the `observeStates` one.

This way we can have tests that care about the first sync and tests which only check the change in states, avoiding the side effects of this first sync. Which is what you wanted if I'm not mistaken when you suggested removing the first sync and rewriting the tests to simplify them.

ganfra: I do agree that code should be easy to test, but I don't think it's hard to wait for the first sync to happen like Benoit suggested?
Anyway, I'm ok with both as long as the `start` method doesn't have a param.

ganfra (approved):
Not tested but LGTM!

jmartinesp:
> Not tested but LGTM!

#YOLO 🫠 

I'll trust my previous tests and the fact that we can check how this works in nightlies then 😅 .

Issue #4243, opened 2025-02-07
Videos fail to send


Destinyg133:
### Steps to reproduce

Some videos fail to send
I will include log, hopefully it can help 
They can be sent using element-android normally but with EXA it just fails or goes into neverending loop

[logs.2025-02-07-11.log](https://github.com/user-attachments/files/18705560/logs.2025-02-07-11.log)
[logcat.log](https://github.com/user-attachments/files/18705559/logcat.log)

#### What did you expect?
To send video files normally

#### What happened instead?
Neverendinf loop or failed to send message

### Your phone model

mi10t

### Application version and app store

synapse 1.123

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

jmartinesp:
I see:

```
matrix_sdk::http_client::native: Sending request num_attempt=1 | crates/matrix-sdk/src/http_client/native.rs:55 | spans: send_attachment > send_attachment > prepare_and_send_attachment > send{server_versions=[V1_0, V1_1, V1_2, V1_3, V1_4, V1_5, V1_6, V1_7, V1_8, V1_9, V1_10, V1_11] config=RequestConfig { timeout: 532s } request_id="REQ-51" method=POST uri="https://domain.tld/_matrix/media/v3/upload" request_size="63.4 MiB" status=502 response_size="0 B" status=502 response_size="0 B" status=502 response_size="0 B" status=502 response_size="0 B" status=502 response_size="0 B" status=502 response_size="0 B" status=502 response_size="0 B"}
```

So you're getting a 502 status code from the homeserver. This status code is usually returned for errors with federation, could you confirm whether the room you're uploading the videos too is in the same HS as you or on a different one? There might be some issue in the SDK preventing uploading media through federation from working properly.

Another option is the 502 could mean the file is too large, but AFAICT this is only used in the `/download` endpoint when HS can't serve a media that large, not for uploading media.

jmartinesp:
After checking the related synapse code and confirming nothing should be returning a `502` status code I asked the synapse backend team about this issue and they said it would most likely be caused by your reverse proxy limiting the size of uploaded files and returning the `502` http status code. 

Could you double check if that's the case? Maybe it's working on EA because the video is compressed with higher compression and passes this check?

Destinyg133:
[...]

PR #4257, opened 2025-02-10
When transcoding a video fails, send it as a file

jmartinesp:
## Content

Changes:

- Make sure the 'at most' video dimensions are 1080p for non-compressed media and 720p for compressed media, matching iOS.
- Try avoiding transcoding when possible, and when it's not try matching as much properties of the original video as possible.
- If transcoding fails, send the original video as a file.

While reviewing you might want to skip/not thoroughly review commit 64ba34c12355745991a3b04f9398aa2bf8466565 since it's reworked shortly after.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/4243, or at least adds a workaround.

## Tests

I was able to make the transcoding fail with [this video](https://drive.google.com/file/d/1cgJ7hVYNQP1AT5pkbQjRRHCVaCYTOCbx/view?usp=sharing) in a Xiaomi Mi 9T with Android 14.

- Try sending the video as an attachment in some room. Processing and transcoding the video will fail.
- Send the video. It will be sent as is, without a thumbnail, but it should still be playable on the other side. For some reason it's send as an `m.video` message instead of a `m.file` one, but it seems like this might need to be changed in the SDK.
- Try some other MP4 video, larger than 1080p if you have disabled media optimisation or 720p if it's enabled, the video will be resized.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/AndroidMediaPreProcessor.kt
+++ b/libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/AndroidMediaPreProcessor.kt
@@ -188,13 +188,17 @@ class AndroidMediaPreProcessor @Inject constructor(
     }
 
     private suspend fun processVideo(uri: Uri, mimeType: String?, shouldBeCompressed: Boolean): MediaUploadInfo {
-        val resultFile = videoCompressor.compress(uri, shouldBeCompressed)
-            .onEach {
-                // TODO handle progress
-            }
-            .filterIsInstance<VideoTranscodingEvent.Completed>()
-            .first()
-            .file
+        val resultFile = runCatching {
+            videoCompressor.compress(uri, shouldBeCompressed)
+                .onEach {
+                    // TODO handle progress
+                }
+                .filterIsInstance<VideoTranscodingEvent.Completed>()
+                .first()
+                .file
+        }
+            // If the video could not be compressed, just copy use the original one by copying to a temporary file
+            .getOrNull() ?: copyToTmpFile(uri)
         val thumbnailInfo = thumbnailFactory.createVideoThumbnail(resultFile)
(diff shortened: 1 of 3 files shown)


jmartinesp:
The title is a bit outdated, this PR was just added so we could test solutions to fix an issue when trying to transcode videos before sending them (the linked one [above](https://github.com/element-hq/element-x-android/issues/4243)) in some Android versions (MIUI for Android 12?).

The original error said:

> java.lang.IllegalStateException: Timestamps must be monotonically increasing: 0, 0

So I tried to fix that by using an interpolator that ensures the video timestamps always go forward, to see if that could solve the issue (it doesn't). This is either a problem with the codecs in the OS shipped by problematic vendors or an issue with the lib we're using to simplify the transcoding process.

So I guess our options are either allowing videos to be sent as is if transcoding fails (either as a file or as the actual media) or using a different lib or transcoder for this. I'm waiting for a decision from @mxandreas , but the latest commit tries to implement the first solution.

mxandreas:
> videos to be sent as is if transcoding fails (either as a file or as the actual media)

If "as is" means that the video size/quality will not be affected, but we can still remove the sensitive pieces (like geolocation tags), then I think that's fine. However, I would send them as "file" to have some indication that the transcoding was not applied.

ganfra:
@jmartinesp what are we doing with that?

jmartinesp:
> @jmartinesp what are we doing with that?

It's now ready for review (I just remembered we had to add a mime-type change). If the video can't be sent as a video it'll be sent as a file, as the description says. I think it matches what Andreas said above.

ganfra (approved):
LGTM, maybe we should add some test on the `VideoCompressor` class?

jmartinesp:
> VideoCompressor

Actually, that's going to be difficult since that's where we're actually performing the video transcoding. Maybe I can extract the transcoding configuration logic somewhere else and test that instead.

Issue #4260, opened 2025-02-12
Room avatar fallbacks don't support Multi-Codepoint ZWJ emoji


HarHarLinks:
### Steps to reproduce

1. Name a room starting with a multi codepoints ZWJ emoji, such as many of the flags 🏳️‍🌈

#### What did you expect?
Avatar fallback is the flag 🏳️‍🌈
Like on web

#### What happened instead?
Avatar fallback is only the first emoji of a combined emoji sequence 🏳️

### Will you send logs?

No

### Are you willing to provide a PR?

No

PR #4277, opened 2025-02-18
When an emoji is used as the 'initial' for an avatar, use the whole emoji

jmartinesp:
## Content

When an 'initials avatar' is generated for a user or room based on either its name or id and it begins with an emoji, make sure we retrieve the whole emoji as the char to display in the avatar.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/4260.

## Screenshots / GIFs

|Before|After|
|-|-|
|image|image|

## Tests

- Create a room with a complex emoji, such as the flag of some country/group, or an emoji with some skin tone or gender, as the first character in its name, and no avatar image.
- Check the generated 'initials' avatar is using the right emoji.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

### DIFF
--- a/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/avatar/AvatarData.kt
+++ b/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/avatar/AvatarData.kt
@@ -28,19 +28,30 @@ data class AvatarData(
                 }
 
                 var length = 1
-                var first = dn[startIndex]
+                var next = dn[startIndex]
 
                 // LEFT-TO-RIGHT MARK
-                if (dn.length >= 2 && 0x200e == first.code) {
+                if (dn.length >= 2 && 0x200e == next.code) {
                     startIndex++
-                    first = dn[startIndex]
+                    next = dn[startIndex]
+                }
+
+                while (next.isWhitespace()) {
+                    if (dn.length > startIndex + 1) {
+                        startIndex++
+                        next = dn[startIndex]
+                    } else {
+                        break
+                    }
                 }
 
                 // check if it’s the start of a surrogate pair
-                if (first.code in 0xD800..0xDBFF && dn.length > startIndex + 1) {
-                    val second = dn[startIndex + 1]
-                    if (second.code in 0xDC00..0xDFFF) {
-                        length++
+                while (isPartOfEmoji(next) && dn.length > startIndex + length) {
+                    length++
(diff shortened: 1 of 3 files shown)


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/avatar/AvatarData.kt
```diff
@@ -27,24 +28,28 @@ data class AvatarData(
...
+                        break
                     }
                 }
 
-                dn.substring(startIndex, startIndex + length)
+                val fullCharacterIterator = BreakIterator.getCharacterInstance()
+                fullCharacterIterator.setText(dn)
+                val glyphBoundary = fullCharacterIterator.following(startIndex)
+
+                dn.substring(startIndex, glyphBoundary)
```
bmarty, line 52: `glyphBoundary` can be equal to `-1`, `substring` will not like it.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/avatar/AvatarData.kt
```diff
@@ -27,24 +28,28 @@ data class AvatarData(
...
+                    if (dn.length > startIndex + 1) {
+                        startIndex++
+                        next = dn[startIndex]
+                    } else {
+                        break
                     }
                 }
 
-                dn.substring(startIndex, startIndex + length)
+                val fullCharacterIterator = BreakIterator.getCharacterInstance()
```
bmarty, line 48: This will return a `BreakIterator` for the default locale, I am wondering what will happen if dn contains chars for another locale.

bmarty (approved):
I have some remark about robustness, but I am failing to find a unit test that will make the code throw an exception. I guess we can accept this code and see if we got crashes...

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/avatar/AvatarData.kt
jmartinesp: I'm not sure if we should have different results for different locales, to be honest, it would be weird if an avatar looked different in devices with different locales 🫤 .

jmartinesp:
The screenshot tests revealed a possible issue, nice 🎉 !

Issue #4291, opened 2025-02-21
Google Login Not Working when default browser is not chrome


msusman1:
### Steps to reproduce

1. Clone the latest main branch
2. open the app and change account provider that supports social login like: chat.the-revivalists.org
3. Click continue to start social login

#### What did you expect?
it should start the Google login flow

#### What happened instead?
it gives an error: access blocked with error code: 403 (disallowed_useragent)

#### Reason of the error:
we are using web view in [LoginFlowNode](https://github.com/element-hq/element-x-android/blob/develop/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/LoginFlowNode.kt)
`            is NavTarget.OidcView -> {
                oidcEntryPoint.createFallbackWebViewNode(this, buildContext, navTarget.oidcDetails.url)
            }`

#### Way to fix?
 if the default browser is not Chrome we are currently using Webview which google no longer supported, we should open the auth link in an external browser instead

![Image](https://github.com/user-attachments/assets/8c5d9315-b89f-4cc0-ae60-2047066a5b9a)

### Your phone model

Pixel 6

### Operating system version

Android 15

### Application version and app store

Build from develop branch

### Homeserver

chat.the-revivalists.org

### Will you send logs?

Yes

### Are you willing to provide a PR?

Yes

msusman1:
[



](url)

bmarty:
I think that we can close this issue now that https://github.com/element-hq/element-x-android/pull/4808 has been merged.

PR #4293, opened 2025-02-21
Use external app for OIDC fallback instead of webview

msusman1:
## Motivation and context
if the default browser is not Chrome we are currently using Webview which google no longer supports, and we should open the auth link in an external browser instead

access blocked
 
## Tests

- Change the default Browser to other than Chrome
- change the home server
- try login by tapping continue
-  google login URL opened in external browser

## Tested devices

- [ ] Physical
- [X] Emulator
- OS version(Android 15):

## Checklist

### DIFF
--- a/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/LoginFlowNode.kt
+++ b/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/LoginFlowNode.kt
@@ -139,8 +139,9 @@ class LoginFlowNode @AssistedInject constructor(
                                 oidcEntryPoint.openUrlInCustomTab(it, darkTheme, oidcDetails.url)
                             }
                         } else {
-                            // Fallback to WebView mode
-                            backstack.push(NavTarget.OidcView(oidcDetails))
+                            activity?.let {
+                                oidcEntryPoint.openUrlInExternalApp(it, oidcDetails.url)
+                            }
                         }
                     }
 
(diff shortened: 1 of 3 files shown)


msusman1:
@bmarty please have a look at this
[Reference Issue](https://github.com/element-hq/element-x-android/issues/4291#issue-2868760023)

iamtalhaasghar:
Yes embedded web-view based login when OIDC upstream provider is Google doesn't work and it isn't [recommended either](https://developers.googleblog.com/en/upcoming-security-changes-to-googles-oauth-20-authorization-endpoint-in-embedded-webviews/#instructions). This PR fixes this issue. We were having this problem with our community hence we decided to contributed the patch upstream. Please let us know if there is something else that we need to do. 

We would love to have this PR merged.

iamtalhaasghar (approved):
Yes embedded web-view based login when OIDC upstream provider is Google doesn't work and it isn't [recommended either](https://developers.googleblog.com/en/upcoming-security-changes-to-googles-oauth-20-authorization-endpoint-in-embedded-webviews/#instructions). This PR fixes this issue. We were having this problem with our community hence we decided to contributed the patch upstream. Please let us know if there is something else that we need to do. 

We would love to have this PR merged.

bmarty:
@sandhose do you have any concern with this changes?

On my side I think that some actions that was handled by the [OidcView](https://github.com/element-hq/element-x-android/blob/develop/libraries/oidc/impl/src/main/kotlin/io/element/android/libraries/oidc/impl/webview/OidcView.kt#L46) may not be handled correctly, but I haven't tested this patch yet.

sandhose:
I think it's indeed better to fallback to the system browser rather than a WebView, even though that means switching apps. We do need to make sure that the callback to the app works as intended

bmarty:
Replaced by #4808 which contains more cleanup

Issue #4241, opened 2025-02-07
[Task] Emoji based user verification


stefanceriu:
Please see and discuss the details in the meta issue.

PR #4294, opened 2025-02-21
Implement user verification

jmartinesp:
## Content

- Moves the 'select key recovery method' screen out of the `VerifySelfSessionNode` to its own navigation Node so we can have the same `Initial` state for both session verification and user verification.
- Add 'Profile' list item for DMs so you can access the other user's profile.
- Make the 'verify' list item in the user profile actually trigger a user verification.
- Implement user verification request and the whole verification flow.

## Motivation and context

Close https://github.com/element-hq/element-x-android/issues/4241.

## Screenshots / GIFs

There are a few new screenshots in the PR.

## Tests

You'll need 2 accounts that haven't verified each other yet.

- With one of the accounts, go to the other user's profile (maybe through a DM -> Settings -> Profile, to test this works) and start the verification.
- The other user should receive a verification request and it should be displayed. Verification requests received while the app was in background should also be displayed soon after re-opening the app.
- If you start a user verification request, you should be able to verify each other. Also, the verification should be cancellable and restartable.
- Once the users have verified each other, opening each others' profiles should display a 'verified' badge.

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): 14, 15.

## Checklist


tests/uitests/src/test/snapshots/images/features.userprofile.shared_UserProfileView_Night_1_en.png
bmarty, line 1: [Figma](https://www.figma.com/design/lMrKOhS8BEb75GXVq7FnNI/ER-96%253A-User-Verification-by-Emoji?node-id=116-41939) says that the wording should be "Verify user"

tests/uitests/src/test/snapshots/images/features.verifysession.impl.incoming_IncomingVerificationView_Day_5_en.png
bmarty, line 1: This does not match Figma, the top icon should be "Reaction solid"

image

tests/uitests/src/test/snapshots/images/features.verifysession.impl.outgoing_VerifySelfSessionView_Day_13_en.png
bmarty, line 1: Empty screen?

tests/uitests/src/test/snapshots/images/features.verifysession.impl.outgoing_VerifySelfSessionView_Day_1_en.png
bmarty, line 1: The top icon should be "lock solid"

image

features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/ui/VerificationUserProfileContent.kt
```diff
@@ -0,0 +1,72 @@
...
+    ) {
+        Avatar(avatarData)
+
+        Spacer(modifier = Modifier.padding(12.dp))
+
+        Column {
+            Text(text = displayName ?: userId.value, style = ElementTheme.typography.fontBodyLgMedium, color = ElementTheme.colors.textPrimary)
+
+            if (displayName != null) {
+                Text(text = userId.value, style = ElementTheme.typography.fontBodyMdRegular, color = ElementTheme.colors.textSecondary)
```
bmarty, line 58: In Figma there is a space of 2 with the text above. Add a `verticalArrangement = Arrangement.spacedBy(2.dp)` to the `Column`?

bmarty (commented):
Some first remarks on the UI. I'll will continue the review.

I need to click twice on "Start verification" for something to happen. It seems that the screen is added twice on the backstack since at the end I need to press twice on the back arrow. See below:


https://github.com/user-attachments/assets/67198cce-ad58-40da-9b08-9df5fa963e7a



One question, the "Verified" decoration (green dot with check) for verified user in timeline top bar and room list item will come later?

image

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -127,8 +131,18 @@ class LoggedInFlowNode @AssistedInject constructor(
...
+            // Without this launch the rendering and actual state of this Appyx node's children gets out of sync, resulting in a crash.
+            // This might be because this method is called back from Rust in a background thread.
+            MainScope().launch {
+                // Wait until the app is in foreground to display the incoming verification request
+                appNavigationStateService.appNavigationState.first { it.isInForeground }
+
+                // Wait for the UI to be ready
+                delay(500.milliseconds)
+
+                backstack.singleTop(NavTarget.IncomingVerificationRequest(verificationRequest))
```
bmarty, line 144: We should probably ignore new incoming request if there is already a verification in progress. But maybe this is handled by the SDK?

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -127,8 +131,18 @@ class LoggedInFlowNode @AssistedInject constructor(
     )
 
     private val verificationListener = object : SessionVerificationServiceListener {
-        override fun onIncomingSessionRequest(sessionVerificationRequestDetails: SessionVerificationRequestDetails) {
-            backstack.singleTop(NavTarget.IncomingVerificationRequest(sessionVerificationRequestDetails))
+        override fun onIncomingSessionRequest(verificationRequest: VerificationRequest.Incoming) {
+            // Without this launch the rendering and actual state of this Appyx node's children gets out of sync, resulting in a crash.
+            // This might be because this method is called back from Rust in a background thread.
+            MainScope().launch {
```
bmarty, line 137: Can `RustSessionVerificationService` ensure that `onIncomingSessionRequest` is invoked on the same thread that was used when registering the listener?

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -127,8 +131,18 @@ class LoggedInFlowNode @AssistedInject constructor(
...
+                // Wait until the app is in foreground to display the incoming verification request
+                appNavigationStateService.appNavigationState.first { it.isInForeground }
```
bmarty, line 139: Maybe add a timeout?

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/di/FtueModule.kt
```diff
@@ -0,0 +1,23 @@
...
+import io.element.android.features.ftue.impl.sessionverification.choosemode.ChooseSelfVerificationModePresenter
+import io.element.android.features.ftue.impl.sessionverification.choosemode.ChooseSelfVerificationModeState
+import io.element.android.libraries.architecture.Presenter
+import io.element.android.libraries.di.SessionScope
+
+@ContributesTo(SessionScope::class)
+@Module
+interface FtueModule {
+    @Binds
+    fun bindChooseVerificationMethodPresenter(presenter: ChooseSelfVerificationModePresenter): Presenter<ChooseSelfVerificationModeState>
```
bmarty, line 22: Not a big deal, but the name does not match :)

```suggestion
    fun bindChooseSelfVerificationModePresenter(presenter: ChooseSelfVerificationModePresenter): Presenter<ChooseSelfVerificationModeState>
```

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/FtueSessionVerificationFlowNode.kt
```diff
@@ -37,15 +44,18 @@ class FtueSessionVerificationFlowNode @AssistedInject constructor(
...
     buildContext = buildContext,
     plugins = plugins,
 ) {
     sealed interface NavTarget : Parcelable {
         @Parcelize
-        data class Root(val showDeviceVerifiedScreen: Boolean) : NavTarget
+        data object Root : NavTarget
+
+        @Parcelize
+        data class UseAnotherDevice(val showDeviceVerifiedScreen: Boolean) : NavTarget
```
bmarty, line 58: The value for `showDeviceVerifiedScreen` is always `true`, so I guess this parameter can be removed.

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/choosemode/ChooseSelfVerificationModePresenter.kt
```diff
@@ -0,0 +1,58 @@
...
+
+data class ChooseSelfVerificationModeState(
+    val isLastDevice: Boolean,
+    val canEnterRecoveryKey: Boolean,
+    val directLogoutState: DirectLogoutState,
+    val eventSink: (ChooseSelfVerificationModeEvent) -> Unit,
+)
+
+sealed interface ChooseSelfVerificationModeEvent {
+    object SignOut : ChooseSelfVerificationModeEvent
```
bmarty, line 57:
```suggestion
    data object SignOut : ChooseSelfVerificationModeEvent
```

bmarty (commented):
More remarks. Doing more test now.

tests/uitests/src/test/snapshots/images/features.verifysession.impl.outgoing_VerifySelfSessionView_Day_13_en.png
jmartinesp: Yes, this is the 'Exit' state, it shouldn't display any UI.

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
jmartinesp: Yep, I can't remember the behaviour, but it was either new requests cancelled previous ones or were ignored, according to the docs.

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
jmartinesp: I don't think so, this is due to how JNA works, it'll create a separate thread for this callback and use it when receiving data from Rust. Maybe we could somehow retain the thread instance and force the callback to somehow run some runnable in it, but it seems like an overkill.

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
jmartinesp: Maybe, but I'm not sure what a sensible value would be. Maybe instead of the callback we should try to check if the request hasn't been cancelled.

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/choosemode/ChooseSelfVerificationModeView.kt
```diff
@@ -0,0 +1,127 @@
...
+    state: ChooseSelfVerificationModeState,
+    onUseAnotherDevice: () -> Unit,
+    onUseRecoveryKey: () -> Unit,
+    onResetKey: () -> Unit,
+    onLearnMore: () -> Unit,
+    modifier: Modifier = Modifier
+) {
+    val activity = LocalActivity.current
+    BackHandler {
+        activity?.finish()
```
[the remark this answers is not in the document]
jmartinesp: This screen is displayed after logging in, or if you haven't verified your session and close and reopen the app. It can't go to a previous state, so I think closing the app is ok, but I can check why I did it like this because just pressing back should have the same effect through the usual navigation flow. Something tells me there was something weird with said navigation flow 😅 .

jmartinesp:
> I need to click twice on "Start verification" for something to happen. It seems that the screen is added twice on the backstack since at the end I need to press twice on the back arrow.

I can't reproduce it 🫤 :

https://github.com/user-attachments/assets/d661b5f0-ed18-4c12-87b1-aaf9b4dc9ee0

> One question, the "Verified" decoration (green dot with check) for verified user in timeline top bar and room list item will come later?

All the status badges will come in a different PR. I'm not sure if the room list one is feasible though.

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/choosemode/ChooseSelfVerificationModeView.kt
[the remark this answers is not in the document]
jmartinesp: I see, it's because the `NoOpBackstackHandlerStrategy` used in `FtueFlowNode` prevents the user from going back. I think we can remove that instead: AFAICT, pressing back here will close the app, but then again, the state should be restored when reopened.

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
bmarty: > retain the thread instance and force the callback to somehow run some runnable in it

This is what I had in mind, but OK to keep it like this :)

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/choosemode/ChooseSelfVerificationModeView.kt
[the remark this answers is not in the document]
bmarty: I understand that the user cannot go back, but a dialog would help to understand that verification is mandatory to go further maybe. Here when clicking back, user may think that the application just crashed.

Not blocking the PR though.

I confirm that the state is restored. I guess it's OK then!

bmarty:
Thanks. It's strange that on second click I can go further and actually do the verification.

jmartinesp:
> Thanks. It's strange that on second click I can go further and actually do the verification.

Yeah, I think it's because we need the change I mentioned above, otherwise the UI and the state machine states differ and we have that issue.

jmartinesp:
If the rest of the issues are fixed, can we move this PR forward even if it has the issue with the unverified sessions canceling the flow? iOS has the same problem, and the feature was delivered weeks ago. Also, it needs product/design feedback so it'll take a while, blocking other PRs related to user verification.

bmarty (approved):
I guess it's fine, thanks!

PR #4305, opened 2025-02-25
Upgrade SDK version to 25.02.26

jmartinesp:
## Content

Changes:

- Upgrade the Rust SDK bindings to `v25.02.26`.
- Remove logic that opened the OIDC callback URL on logout as it's no longer needed, the SDK already handles this.
- Improve how we handle retrieving and destroying SDK Room instances so we don't run into issues where the timeline is not reset when reloading a room because it's still being held somewhere, or the opposite where the timeline was destroyed eagerly and now the methods associated with it won't work.
- Disable the `Typos` lint check that was making `app:debugLint` fail.

## Motivation and context

Upgrade to the latest version of the SDK.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist


features/logout/impl/src/main/kotlin/io/element/android/features/logout/impl/DefaultLogoutUseCase.kt
```diff
@@ -19,7 +19,7 @@ class DefaultLogoutUseCase @Inject constructor(
     private val authenticationService: MatrixAuthenticationService,
     private val matrixClientProvider: MatrixClientProvider,
 ) : LogoutUseCase {
-    override suspend fun logout(ignoreSdkError: Boolean): String? {
+    override suspend fun logout(ignoreSdkError: Boolean) {
         val currentSession = authenticationService.getLatestSessionId()
         return if (currentSession != null) {
```
bmarty, line 24: I think that the `return` word can be removed.

libraries/ui-strings/build.gradle.kts
```diff
@@ -11,4 +11,8 @@ plugins {
 
 android {
     namespace = "io.element.android.libraries.ui.strings"
+
+    lint {
+        disable += "Typos"
+    }
```
bmarty, line 17: I would prefer that we fix the typos in Localazy when spotted. Was there some false positive?

bmarty (approved):
LGTM, thanks!

libraries/ui-strings/build.gradle.kts
bmarty: OK

jmartinesp: I'll address this in a separate PR, no worries.

PR #4309, opened 2025-02-26
runQualityChecks task shouldn't fail fast

jmartinesp:
## Content

Changes:

- Make sure we run every task it depends on so we get a list with all the issues instead of having to retry several times.
- Tweak `runQualityChecks` task to make sure all Android lint tasks run.
- Fix the CI flow so it works the same as when we run it locally to make sure the results match.

## Motivation and context

It's a pain to have to run `runQualityChecks` several times just to get the next list of issues once we've fixed the current one.

## Tests

- Add some issue for the `runQualityChecks` task to discover. Check finding the error doesn't stop the rest of the tasks from running too.

## Checklist


bmarty (approved):
Thanks!

PR #4310, opened 2025-02-26
Use `Settings.System.DEFAULT_RINGTONE_URI` for ringing notifications

jmartinesp:
## Content

This replaces `RingtoneManager.getActualDefaultRingtoneUri`, it should get the same audio file and avoid some reported issues about not having permission to load the audio file.

## Motivation and context

Fix crashes seen in rageshakes.

## Tests

- Uninstall the app.
- Set a custom ringtone audio picked from your device's storage.
- Install the app and enable notifications, no crash should happen.
- Receive a ringing call notification (start a call with another account in a DM), the custom audio should be used.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/notifications/RingingCallNotificationCreator.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/notifications/RingingCallNotificationCreator.kt
@@ -11,7 +11,7 @@ import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
 import android.media.AudioManager
-import android.media.RingtoneManager
+import android.provider.Settings
 import androidx.core.app.NotificationCompat
 import androidx.core.app.PendingIntentCompat
 import androidx.core.app.Person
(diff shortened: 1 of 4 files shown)


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/notifications/RingingCallNotificationCreator.kt
```diff
@@ -126,9 +124,7 @@ class RingingCallNotificationCreator @Inject constructor(
                     setContentText(textContent)
                     // Else the content text is set by the style (will be "Incoming call")
                 }
-                if (ringtoneUri != null) {
-                    setSound(ringtoneUri, AudioManager.STREAM_RING)
-                }
+                setSound(Settings.System.DEFAULT_RINGTONE_URI, AudioManager.STREAM_RING)
```
bmarty, line 127: Neat: the `setSound` could be moved out of the `apply` block.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/channels/NotificationChannels.kt
```diff
@@ -163,16 +158,14 @@ class DefaultNotificationChannels @Inject constructor(
...
-                        setSound(
-                            ringtoneUri,
-                            AudioAttributes.Builder()
-                                .setContentType(AudioAttributes.CONTENT_TYPE_SONIFICATION)
-                                .setLegacyStreamType(AudioManager.STREAM_RING)
-                                .setUsage(AudioAttributes.USAGE_NOTIFICATION_RINGTONE)
-                                .build()
-                        )
-                    }
+                    setSound(
```
bmarty, line 161: Neat: no need for an `apply` block anymore.

bmarty (approved):
Thanks!

Issue #4326, opened 2025-02-27
Appyx Nodes and dependencies are being held in memory


jmartinesp:
We received an issue recently about the `Client` instance of a logged out session wasn't immediately being cleaned up and that started and investigation.

Apparently, there were several places where a `MatrixClient` wrapping this Rust `Client` instance had a strong reference and a path to the GC root and could cause it to be retained more timed than we'd want it to, but it also uncovered a larger issue: Appyx nodes are being retained in memory for some time even when they're no longer in the navigation stack.

Why is this happening? There are 2 main reasons:

1. Appyx keeps a `Map` of navigation keys linked to `ChildEntry` instances to represent the navigation state at each `ParentNode`. However, this representation of the navigation state is used in several places of the navigation library with an operator that effectively caches the previous state to diff it with the next one, so we're keeping a reference to a map of ChildEntries, which also keep references to their associated Nodes, and these to their dependencies (Presenters, services, etc.). Several of these hold a direct or transitive reference to a `MatrixClient`.
2. For some reason, something similar happens with `Compose`: anything referenced in the `Presenter.present` composable function will be held in memory even when the composable is no longer in use (maybe [this issue](https://issuetracker.google.com/issues/399397350)?), and we use a pattern for relaying events back to the `Presenter` using `eventSink` lambdas that sometimes leak a reference to the presenters through extension functions, or referencing some of the presenter's dependencies directly. We believe this behaviour might have been introduced recently in Compose, and given a heap dump, it looks like this:

<details>
<img width="1155" alt="Image" src="https://github.com/user-attachments/assets/5d8ff9e3-8998-4b27-bafe-f272ab34fb7a" />
</details>

Apparently, both cases can be solved if several navigation steps happen, clearing some kind of internal caches.

PR #4333, opened 2025-02-27
Restore manual `Client` cleanup on session logout

jmartinesp:
## Content

- Restore `Client` instances and related dependencies cleanup when the user logs out.
- Try to minimise the number of strong references to `Client`/`MatrixClient`.
- Also try to wrap more calls to the SDK with `runCatching` to prevent issues if some component is called when its Rust reference has been destroyed.

## Motivation and context

This is done as a workaround for https://github.com/element-hq/element-x-android/issues/4326.

## Tests

Log out, check the `Client` is not running in the background anymore.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
```diff
@@ -101,7 +109,7 @@ class RustClientSessionDelegate(
                 } else {
                     clientLog.d("No session data found.")
                 }
-                client?.logout(userInitiated = false, ignoreSdkError = true)
+                client.get()?.logout(userInitiated = false, ignoreSdkError = true)
```
bmarty, line 112: This is not new, but I am wondering why we are not using `currentClient` here (instead of `client.get()?`).

bmarty (approved):
LGTM, thanks!

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustClientSessionDelegate.kt
jmartinesp: No reason, it's better to use `currentClient` indeed!

PR #4347, opened 2025-03-03
Migrate to coil3

bmarty:
## Content

Migrate from coil2 to coil3

Ref: https://coil-kt.github.io/coil/upgrading_to_coil3/

## Motivation and context

Use up to date dependencies.

## Tests

- Check that images (avatar, media in rooms, etc.) are correctly loaded
- Signout and signin using another account. Images should be loaded correctly.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/media/ImageLoaderFactories.kt
+++ b/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/media/ImageLoaderFactories.kt
@@ -9,10 +9,10 @@ package io.element.android.libraries.matrix.ui.media
 
 import android.content.Context
 import android.os.Build
-import coil.ImageLoader
-import coil.ImageLoaderFactory
-import coil.decode.GifDecoder
-import coil.decode.ImageDecoderDecoder
+import coil3.ImageLoader
+import coil3.gif.AnimatedImageDecoder
+import coil3.gif.GifDecoder
+import coil3.network.okhttp.OkHttpNetworkFetcherFactory
 import com.squareup.anvil.annotations.ContributesBinding
 import io.element.android.libraries.di.AppScope
 import io.element.android.libraries.di.ApplicationContext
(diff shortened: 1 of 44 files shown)


libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/media/ImageLoaderFactories.kt
```diff
@@ -33,11 +33,17 @@ class DefaultLoggedInImageLoaderFactory @Inject constructor(
     override fun newImageLoader(matrixClient: MatrixClient): ImageLoader {
         return ImageLoader
             .Builder(context)
-            .okHttpClient { okHttpClient.get() }
             .components {
+                add(
+                    OkHttpNetworkFetcherFactory(
+                        callFactory = {
+                            okHttpClient.get()
```
bmarty, line 40: This does not seem to be used, not sure what's happening here.

bmarty: I set a breakpoint on `okHttpClient.get()` and starts the application in debug mode (from AS). Images are loaded but the code never breaks.

ganfra: This sounds wrong then :/ 
According to the documentation, should be enough

bmarty: I have added logs and confirm that this code is not used. Still digging

ganfra (approved):

libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/media/ImageLoaderFactories.kt
bmarty: Actually this is fine. The `OkHttpClient` is not used to load images from the homeserver (avatars, images in room, etc.), we use the SDK API to retrieve them 🤦‍♂️.

The `OkHttpClient` is only used to load images for the map (location sharing), and in this case, the `CallFactory` is correctly invoked.

PR #4350, opened 2025-03-04
Prevent PRs with the X-Blocked label from being merged

robintown:
Just a small CI suggestion: X-Blocked is useful for requesting review on a change while ensuring it won't be merged until all external dependencies are accounted for, and it'd be nice to have this enforced more consistently across the org.
Adapted from matrix-js-sdk (https://github.com/matrix-org/matrix-js-sdk/blob/develop/.github/workflows/pull_request.yaml)
For https://github.com/element-hq/voip-internal/issues/303

### DIFF
--- a/.github/workflows/blocked.yml
+++ b/.github/workflows/blocked.yml
@@ -1,0 +1,17 @@
+name: Prevent blocked
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


bmarty (approved):
Thanks, it should work!

PR #4358, opened 2025-03-04
Update UnifiedPush library

bmarty:
## Content

Update UnifiedPush library

## Motivation and context

Be up to date and let Renovate handle this in the future.

We will also benefit from improvement from the library, in particular the support for [RAISE_TO_FOREGROUND](https://unifiedpush.org/developers/spec/android/#service-to-raise-to-the-foreground) added in https://github.com/UnifiedPush/android-connector/commit/7438c52c3f0ef2edba4c908c2cf796297aeb888c .

## Tests

- Ensure that UnifiedPush notification are still working (still under test on my side)

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


libraries/pushproviders/unifiedpush/build.gradle.kts
```diff
@@ -41,7 +41,13 @@ dependencies {
     implementation(libs.serialization.json)
 
     // UnifiedPush library
-    api(libs.unifiedpush)
+    implementation(libs.unifiedpush) {
+        // Exclude package com.google.crypto.tink
+        // Duplicate classes between
+        // tink-1.16.0.jar -> tink-1.16.0 (com.google.crypto.tink:tink:1.16.0)
+        // tink-android-1.8.0.jar tink-android-1.8.0 (com.google.crypto.tink:tink-android:1.8.0)
+        exclude(group = "com.google.crypto.tink", module = "tink")
```
bmarty, line 49: Hello @p1gp1g , do you know why I need to add this line?

It seems to be related to https://github.com/UnifiedPush/android-connector/blob/cf347ae931ea8b6e6c773cd566abea940d1297f1/connector/build.gradle#L58 and the related PR https://github.com/tink-crypto/tink-java-apps/pull/5, but maybe I miss something?

bmarty: Thanks for the fast and complete response. Maybe the unified push library should use `tink-android` and not `tink` and that would fix the conflicts? (even if I agree that we have another dependency (not checked which one yet) which has a dependency on an outdated version `1.8.0`).

p1gp1g: You would have the same issue if 2 dependencies rely on tink-android, isn't it ?

I have opened [an issue](https://codeberg.org/UnifiedPush/android-connector/issues/5) to migrate to tink-android, there are some pro/con to do the migration there

jmartinesp: > (even if I agree that we have another dependency (not checked which one yet) which has a dependency on an outdated version 1.8.0)

I think it's this one:

```
androidx.security:security-crypto:1.1.0-alpha06
+--- androidx.annotation:annotation:1.1.0 -> 1.9.1 (*)
+--- androidx.annotation:annotation:1.2.0 -> 1.9.1 (*)
+--- androidx.collection:collection:1.1.0 -> 1.4.4 (*)
\--- com.google.crypto.tink:tink-android:1.8.0
```

bmarty: Related: https://stackoverflow.com/questions/78362124 ...

bmarty:
I am testing that the upgrade do not break anything, and sadly I do not receive anymore push from ntfy, even when not using a matrix.org account.

The troubleshoot fails at the "Test push loopback":

image

Logs:

```
POST https://ntfy.sh/_matrix/push/v1/notify 
{"notification":{"event_id":"$THIS_IS_A_FAKE_EVENT_ID","room_id":"!room:domain","devices":[{"app_id":"im.vector.app.android","pushkey":"https://ntfy.sh/upMQ1D9W3MlzAD?up=1"}]}}
<-- 507 https://ntfy.sh/_matrix/push/v1/notify
{"code":50701,"http":507,"error":"cannot publish to UnifiedPush topic without previously active subscriber"}

```

Seems related: https://github.com/binwiederhier/ntfy/issues/649

p1gp1g:
This may be because your ntfy didn't have any registration when it has received elemen-x request. And ntfy is a bit buggy during this state. This would be fixed with https://github.com/binwiederhier/ntfy-android/pull/97

You can try [Sunup](https://codeberg.org/Sunup/android), it will use the default gateway (matrix.gateway.unifiedpush.org)

Eskuero:
I selfhost both ntfy and synapse

I built ntfy-android with https://github.com/binwiederhier/ntfy-android/pull/97 and https://github.com/binwiederhier/ntfy-android/pull/98
I built Element X with this patch based on the latest tag

1. Upon installing I ran the the troubleshooting with my existing setup and it passed without issues as before
2. Closed both apps from the foreground and sent a message, received instantly.
3. Force closed Element X and cleared it's cache
4. Sent another message, got the notification instantly too.
5. Killed Element X and removed the topic from the ntfy android app
6. Tried sending another message and obviously got nothing, but synapse logs still claimed to be trying to sending them to ntfy server (I guess this is a synapse/nfty server issue who are not communicating deletions and out of scope here)
7. Opened Element X again and it automatically registered a new topic without any input.
8. Run a troubleshoot without errores.
9. Killed Element X again
10. Sent another message, received without issues.

Eskuero:
Is there anything else missing or that I can test ? Been running the patch for a month without issues on notification receipts so on my clueless knowledge it seems ready to go.

libraries/pushproviders/unifiedpush/build.gradle.kts
```diff
@@ -42,7 +42,13 @@ dependencies {
...
```
jmartinesp, line 50: I think this has to be removed now the dependency with the transitivity dependency to tink is gone.

Eskuero:
I have the app already built with the patches in my fdroid repo
https://repo.fromshado.ws/repo/
https://repo.fromshado.ws/repo/ntfy-android-fdroidrelease-fdroid.apk

If you would rather give another try at building the clean patchfile to apply is also here:
https://github.com/Eskuero/nino_samples/tree/master/ntfy-android

And also is true that probably the ntfy.sh instance is putting some ratelimits which might harm testing. If you want I can give you access to my ntfy instance. Ping me at @eskuero:matrix.fromshado.ws

p1gp1g:
> With Sunup the notifications arrive, but they're severely delayed (30s) and for some reason they appear twice in the list of notifications 😓 .

This is probably not the best place to debug setups. There are a lot of reasons why a notification could be delayed (depending on the load on the server, if the client have history to fetch, if it can get the decryption keys, etc.). It also depends on your OS configuration, if it is in doze mode etc. When doing tests, you may figure out that sometimes the web client get its notification after the android client, sometimes the opposite. You 'd see the same thing with FCM too

libraries/pushproviders/unifiedpush/build.gradle.kts
bmarty: Done in [57283a8](https://github.com/element-hq/element-x-android/pull/4358/commits/57283a89c2ac878c53183cb75a1546218cee80a3)

bmarty: it seems to be still necessary

jmartinesp: Weird, it wouldn't let me build the app locally with this exclusion 🫤 .

gradle/libs.versions.toml
```diff
@@ -177,7 +177,7 @@ sqldelight-driver-jvm = { module = "app.cash.sqldelight:sqlite-driver", version.
 sqldelight-coroutines = { module = "app.cash.sqldelight:coroutines-extensions", version.ref = "sqldelight" }
 sqlcipher = "net.zetetic:android-database-sqlcipher:4.5.4"
 sqlite = "androidx.sqlite:sqlite-ktx:2.5.0"
-unifiedpush = "com.github.UnifiedPush:android-connector:2.4.0"
+unifiedpush = "org.unifiedpush.android:connector:3.0.4"
```
bmarty, line 180:
```suggestion
unifiedpush = "org.unifiedpush.android:connector:3.0.8"
```

bmarty:
> Can you try with version 3.0.9 ?

Sure. Is there any bug fixed on the new version that may help with the encountered problem?

gradle/libs.versions.toml
```diff
@@ -177,7 +177,7 @@ sqldelight-driver-jvm = { module = "app.cash.sqldelight:sqlite-driver", version.
 sqldelight-coroutines = { module = "app.cash.sqldelight:coroutines-extensions", version.ref = "sqldelight" }
 sqlcipher = "net.zetetic:android-database-sqlcipher:4.5.4"
 sqlite = "androidx.sqlite:sqlite-ktx:2.5.0"
-unifiedpush = "com.github.UnifiedPush:android-connector:2.4.0"
+unifiedpush = "org.unifiedpush.android:connector:3.0.8"
```
bmarty, line 180:
```suggestion
unifiedpush = "org.unifiedpush.android:connector:3.0.9"
```

p1gp1g:
> > Can you try with version 3.0.9 ?
> 
> Sure. Is there any bug fixed on the new version that may help with the encountered problem?

There was a bug that doesn't apply here, sorry for the ping. For the details, PushService can be used to replace the MessagingReceiver, but isn't used here and it didn't work correctly with SDK<33

p1gp1g:
By the way, what is _the encountered problem_ ? Is it the 507 response ?

jmartinesp:
Just for the record, I was able to get reliable push notifications working in NTFY by doing:

1. Deleting every single topic in ntfy.
2. Changing the push server to [ntfy.adminforge.de](http://ntfy.adminforge.de/).
3 .Switching to firebase, then back to ntfy in the notification settings.
4. Checking the new server is used in ntfy (a new topic is created with the server name being part of it).

Issue #4240, opened 2025-02-07
[Task] User verification state indicators


stefanceriu:
Please see and discuss the details in the meta issue.

PR #4392, opened 2025-03-11
Add user verification and verification state violation badges

jmartinesp:
## Content

- Move `observerRoomMemberIdentityStateChanges` and associated classes from `MessageComposer` to `:libraries:matrixui` so we can reuse it in several places.
- Add user verified badge to the top app bar of a DM room.
- Add verification violation badge in the 'People' list item in the room details screen.
- Add verified and violation badge to the 'Profile' list item in the DM details screen.
- Add both badges to the room member list too.
- Display verification state only for room member details screen, not in general user profile ones. Also, update this info in real time and make sure we can withdraw the verification state of a user with a verification violation state in this screen.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/4240

## Screenshots / GIFs

There are quite a few in the PR.

## Tests

Just test the changes mentioned above, if possible.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
```diff
@@ -183,6 +182,18 @@ class MessagesPresenter @AssistedInject constructor(
             enableVoiceMessages = featureFlagsService.isFeatureEnabled(FeatureFlags.VoiceMessages)
         }
 
+        var dmUserVerificationState by remember { mutableStateOf<IdentityState?>(null) }
+
+        LifecycleResumeEffect(Unit) {
+            if (room.isDm && room.isEncrypted) {
```
bmarty, line 188: Maybe use `room.isDm` and `room.isEncrypted` as keys of `LifecycleResumeEffect`?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
```diff
@@ -183,6 +182,18 @@ class MessagesPresenter @AssistedInject constructor(
...
+                localCoroutineScope.launch {
+                    val dmUserId = room.getMembers().getOrNull()?.find { it.userId != room.sessionId }?.userId
```
bmarty, line 190: I am wondering if getMembers will always returns the whole member list. Maybe for DM this is OK.

Else you may want to use
```kotlin
val membersState by room.membersStateFlow.collectAsState()
val otherMember = room.getDirectRoomMember(membersState)
```
instead?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
```diff
@@ -188,6 +190,7 @@ fun MessagesView(
                     roomAvatar = state.roomAvatar.dataOrNull(),
                     heroes = state.heroes,
                     roomCallState = state.roomCallState,
+                    isDmUserVerified = state.dmUserVerificationState == IdentityState.Verified,
```
bmarty, line 193: I am not confortable with this comparison because `isDmUserVerified` will be false when the room is not a DM. If later we decide to render something else when this is a DM and the other user is not verified, we will have some trouble.

Since `dmUserVerificationState` is nullable, can you change this to 

```suggestion
                    isDmUserVerified = state.dmUserVerificationState?.let { it == IdentityState.Verified },
```

`isDmUserVerified` will need to be nullable as well.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
```diff
@@ -463,22 +467,31 @@ private fun MessagesViewTopBar(
...
+                } else {
+                    IconTitlePlaceholdersRowMolecule(
+                        iconSize = AvatarSize.TimelineRoom.dp,
+                        modifier = titleModifier
+                    )
+                }
+
+                if (isDmUserVerified) {
+                    Icon(imageVector = CompoundIcons.Verified(), tint = ElementTheme.colors.iconSuccessPrimary, contentDescription = null)
+                }
```
bmarty, line 494: This icon will not be visible if the room name is too long:

image

Or its size will be reduced.

image

tests/testutils/src/main/kotlin/io/element/android/tests/testutils/WithFakeLifecycleOwner.kt
```diff
@@ -9,31 +9,54 @@ package io.element.android.tests.testutils
...
-        state = block()
-    }
-    return state!!
+fun <T> withFakeLifecycleOwner(
+    lifecycleOwner: FakeLifecycleOwner = FakeLifecycleOwner(),
+    block: @Composable () -> T
+): T {
+    currentComposer.startProvider(LocalLifecycleOwner provides lifecycleOwner)
+    val state = block()
+    currentComposer.endProviders()
```
bmarty, line 36:
```suggestion
    currentComposer.endProvider()
```

Seems to be more appropriate. `endProviders()` has to be used when `startProviders()` has been called. This is probably equivalent though.

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListView.kt
```diff
@@ -284,34 +289,61 @@ private fun LazyListScope.roomMemberListSection(
...
+        matrixUser = roomMemberWithIdentity.roomMember.toMatrixUser(),
         avatarSize = AvatarSize.UserListItem,
-        trailingContent = roleText?.let {
-            @Composable {
-                Text(
-                    text = it,
-                    style = ElementTheme.typography.fontBodySmRegular,
-                    color = ElementTheme.colors.textSecondary,
-                )
+        trailingContent = {
```
bmarty, line 314: This is not new, but I think we will need to add some padding at the start of the trailing content:

image

bmarty (commented):
Nice work, thanks!
I have some first remarks. I have not tested the PR yet.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
jmartinesp: I think you're right, I just noticed the warning in its docs:

> Get the members of the room. Note: generally this should not be used, please use [membersStateFlow] and [updateMembers] instead.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
jmartinesp: Ok, I mean if it's not a DM for sure the DM user is not verified (since there isn't one 😅 ) but maybe with the null value it'll be clearer.

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListView.kt
jmartinesp: I think some padding is added by default:

image

tests/testutils/src/main/kotlin/io/element/android/tests/testutils/WithFakeLifecycleOwner.kt
```diff
@@ -33,10 +38,15 @@ fun <T> withFakeLifecycleOwner(
...
-    currentComposer.endProviders()
+    currentComposer.endProvider()
     return state
 }
 
+/**
+ * Test a [Presenter] with a fake [LifecycleOwner].
+ *
+ * **WARNING: DO NOT USE OUTSIDE TESTS.**
+ */
```
bmarty, line 49: Since this module is a test module, no prod module should have a dependency on it. Being able to use this method from a production module will reveal that there is an issue in the module dependencies. But it's probably safer to keep this warning anyway.

bmarty (approved):
Thanks for the update!

tests/uitests/src/test/snapshots/images/features.messages.impl_MessagesView_Day_12_en.png
bmarty, line 1: A small start padding on the icon should be nice maybe.

jmartinesp: I need to double check, but when I was testing this with a preview earlier I'm pretty sure there was some spacing even though the title was ellipsized and full width... maybe it's some screenshot recording issue.

PR #4393, opened 2025-03-11
"Core Team" is a team of matrix-org. Use team "Vector Core" instead.

bmarty:
Fix CI giving label community to PR from us.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -36,7 +36,7 @@ jobs:
           username: ${{ github.event.pull_request.user.login }}
           organization: element-hq
           team: Vector Core
-          GITHUB_TOKEN: ${{ secrets.ELEMENT_BOT_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.ELEMENT_BOT_TOKEN_READ_ORG }}
       - name: Add label
         if: steps.teams.outputs.isTeamMember == 'false'
         uses: actions/github-script@v7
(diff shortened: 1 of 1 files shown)


ganfra (approved):

ganfra:
> This does not work:
> 
> image
> Checking again

Doesn't it need to be merged to be active?

bmarty:
The run seems to use the correct new value:

image

guillaumevillemont:
Can you try using `ELEMENT_BOT_TOKEN_READ_ORG` as the token instead ? 

I believe `ELEMENT_BOT_TOKEN` only has repositories access. I created a new token as I didn't want to edit a token potentially used in other workflow.

guillaumevillemont:
@bmarty I failed I ticking the right repo name :man_facepalming: Should be good now

bmarty:
OK, I have changed the token. Let's see what happening on the next created PR.

bmarty:
The workflow do no add the label for our PR, which is great, but do not do it neither for PRs from the community. See an example here: https://github.com/element-hq/element-x-android/actions/runs/13846033519/job/38768297382?pr=4410

It because the fork cannot access our secrets: `GITHUB_TOKEN: ${{ secrets.ELEMENT_BOT_TOKEN_READ_ORG }}`.

guillaumevillemont:
uh :disappointed:  I believe this never work then (Since the previous team was a public team, no tokens were required)
I don't think moving the secret from org to repo level would help. 

So at some point, anyone should be able to check who's part of the team (from their fork's CI at least). Does it make sense to keep the team private on Github then ?
If we keep the team private (which is sensible), I don't see how that can work :thinking:

PR #4396, opened 2025-03-12
Clear SDK cache properly

bmarty:
## Content

First commit is changing calls from `destroy` to `close`, since the `close` methods have the `@Synchronized` annotation. In practice it should not change anything.

When the event cache feature flag is enabled, use the new method `clearCaches` from the SDK to clear its cache, instead of deleting all its files.

## Motivation and context

Use new API.

## Tests

- From the developer settings, enable the Event Cache feature
- From the developer settings, clear the cache

Observe that the application is using less disk space.

## Tested devices

- [ ] Physical
- [x] Emulator

### DIFF
--- a/features/preferences/impl/src/test/kotlin/io/element/android/features/preferences/impl/tasks/DefaultClearCacheUseCaseTest.kt
+++ b/features/preferences/impl/src/test/kotlin/io/element/android/features/preferences/impl/tasks/DefaultClearCacheUseCaseTest.kt
@@ -27,7 +27,7 @@ import org.robolectric.RobolectricTestRunner
 @RunWith(RobolectricTestRunner::class)
 class DefaultClearCacheUseCaseTest {
     @Test
-    fun `execute clear cache should do all the expected tasks no event cache`() = runTest {
+    fun `execute clear cache should do all the expected tasks`() = runTest {
         val clearCacheLambda = lambdaRecorder<Unit> { }
         val matrixClient = FakeMatrixClient(
             clearCacheLambda = clearCacheLambda,


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
```diff
@@ -486,20 +486,26 @@ class RustMatrixClient(
...
 
     override suspend fun getCacheSize(): Long {
         return baseDirectory.getCacheSize()
     }
 
-    override suspend fun clearCache() {
-        close()
-        deleteSessionDirectory(deleteCryptoDb = false)
+    override suspend fun clearCache(isEventCacheEnabled: Boolean) {
+        if (isEventCacheEnabled) {
```
ganfra, line 501: Are you sure we need to check for event cache? iOS is not

ganfra: I'd just remove our manual clear cache and use only the `clear_cache` from sdk.

bmarty: OK, done in https://github.com/element-hq/element-x-android/pull/4396/commits/c857a3b6f48a67fbda8060137d84661db957ce88

features/preferences/impl/src/test/kotlin/io/element/android/features/preferences/impl/tasks/DefaultClearCacheUseCaseTest.kt
```diff
@@ -27,7 +27,7 @@ import org.robolectric.RobolectricTestRunner
 @RunWith(RobolectricTestRunner::class)
 class DefaultClearCacheUseCaseTest {
     @Test
-    fun `execute clear cache should do all the expected tasks`() = runTest {
+    fun `execute clear cache should do all the expected tasks no event cache`() = runTest {
```
bmarty, line 30:
```suggestion
    fun `execute clear cache should do all the expected tasks`() = runTest {
```

ganfra (approved):

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
jmartinesp: > [clear_caches()](https://github.com/matrix-org/matrix-rust-sdk/blob/5a22944f5298a622c0f8d9b2a64d254d6312a07d/bindings/matrix-sdk-ffi/src/client.rs#L1192) is doing a better job than our "manual" clear cache, that just deletes everything. I guess we can keep it like this and when the flag about the event cache will be removed, just keep the call to `clear_caches`. Do you agree?

Actually, I think this is what was done on iOS and suggested by the SDK team (sorry for the late feedback, I just saw the discussion).

Issue #4391, opened 2025-03-11
Timeline (pagination) performance is noticeably slower than on Element X iOS


frebib:
Since the event cache has been enabled by default, timeline performance has improved quite a bit. In many cases though, it still feels a lot slower than on iOS. This is in part [due to uniffi](https://matrix.to/#/!aCvOWKNDeXXpPgquKq:matrix.org/$hAHyR8bxFW3YBRT6SvqomlNOw7jv2DEbRIMk_IlXhKw?via=matrix.org&via=envs.net&via=tchncs.de)

As a user, I shouldn't ever really be able to get to the top of a loading timeline (as is on iOS) when in reality I hit the loading spinner almost every time. Pagination should happen so fast that the user never sees the unloaded portion of the timeline.

jmartinesp:
FWIW I think we're only actively paginating when the load more indicator is displayed (or prefetched), and this can be way too late. We should probably paginate at 10-20 items before the last one. I'll draft a PR for this.

PR #4399, opened 2025-03-12
Add timeline item prefetching

jmartinesp:
## Content

This should trigger when getting close the start of the loaded timeline, making scrolling back smoother, specially when combined with the persistent event cache.

## Motivation and context

Try to make scrolling a bit smoother.

Fixes https://github.com/element-hq/element-x-android/issues/4391 (or at least, it should help).

## Tests

Ideally with the persistent event cache enabled, open a room where you've already paginated for a while and start scrolling back.

The scrolling should be smoother than before as we start prefetching the data way sooner. There may still be some 'pauses' as loading new items is not instant.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
@@ -76,11 +76,16 @@ import io.element.android.libraries.testtags.TestTags
 import io.element.android.libraries.testtags.testTag
 import io.element.android.libraries.ui.strings.CommonStrings
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.conflate
 import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.transform
 import kotlinx.coroutines.launch
 import timber.log.Timber
+import kotlin.time.Duration.Companion.milliseconds
 
 @Composable
 fun TimelineView(
@@ -139,6 +144,10 @@ fun TimelineView(
         )
     }
 
+    fun prefetchMoreItems() {
+        state.eventSink(TimelineEvents.LoadMore(Timeline.PaginationDirection.BACKWARDS))
+    }
+
     // Animate alpha when timeline is first displayed, to avoid flashes or glitching when viewing rooms
     AnimatedVisibility(visible = true, enter = fadeIn()) {
         Box(modifier) {
@@ -185,9 +194,10 @@ fun TimelineView(
                 onClearFocusRequestState = ::clearFocusRequestState
             )
 
-            TimelinePrefetchingHelper(lazyListState = lazyListState) {
-                state.eventSink(TimelineEvents.LoadMore(Timeline.PaginationDirection.BACKWARDS))
-            }
+            TimelinePrefetchingHelper(
+                lazyListState = lazyListState,
+                prefetch = ::prefetchMoreItems
+            )
 
             TimelineScrollHelper(
                 hasAnyEvent = state.hasAnyEvent,
@@ -217,7 +227,7 @@ private fun MessageShieldDialog(state: TimelineState) {
     )
 }
 
-@OptIn(ExperimentalCoroutinesApi::class)
+@OptIn(ExperimentalCoroutinesApi::class, FlowPreview::class)
 @Composable
 private fun TimelinePrefetchingHelper(
     lazyListState: LazyListState,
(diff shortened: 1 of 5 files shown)


jmartinesp:
Note scrolling is still choppy in debug builds, so ideally this should be tested in nightly/release builds.

jmartinesp:
> I tried this and I'm not sure it made much/any difference? I even tried increasing this number here to 50 and it felt about the same again

Oh, that's weird, there was definitely a difference for me in a 5 y.o. device, it wasn't always smooth when scrolling back, but it was most of the time, while it was definitely a lot more janky before. Are you enabling the persistent event cache for the tests? 

Also, I think it doesn't work as well as it should in encrypted rooms because the encryption layer does extra work that slows down delivering new items from the SDK to the app. I tried it in some non-encrypted rooms and I didn't see a single indicator while scrolling up, while I still saw a few in some small encrypted ones.

> Curiously I'm also not seeing the new debug log lines in my release build- do they get "compiled out" or something in release builds?

You need to enable those by changing [this value here](https://github.com/element-hq/element-x-android/blob/7d27e6581be745fd9ebaf1f5e276c279e25b72ae/app/src/main/kotlin/io/element/android/x/initializer/TracingInitializer.kt#L34) to `true`. We're discussing

ganfra:
> As I'm understanding it, it will attempt to request pagination at most one time (although in practice it's usually zero unless you scroll in just the right way) and then stop until the user scrolls again. What I would probably expect as a user (and what I'm pretty sure most of the other "mainstream" chat apps do is paginate repeatedly until a desired number of messages are ready but off-screen. Is that a desirable behaviour here? I understand that it's a tradeoff between responsiveness and conservative resource use.

Yes, but we'll get more improvements later from the rust sdk with probably some automatic pagination!

frebib:
Out of my own curiosity, I had a poke at trying to make this do what I would like, and came up with this adaptation of Jorge's change from this PR https://github.com/frebib/element-x-android/commit/6e3b4796f76f3da717550220fa8fc74df39c513f
I'm far from a UX expert/app developer, but I'm gonna roll with this for a while. At least with this patch on my phone with my homeserver, it's very hard/nearly impossible to hit the loading spinner unless in an encrypted room (where pagination seems a lot slower for some reason?). When you open the room, the app does this, then stops:
```
03-12 20:42:27.180  6436  6436 D org.matrix.rust.sdk: elementx: Prefetching pagination with 0 off-screen (0 total) items |
03-12 20:42:27.312  6436  6436 D org.matrix.rust.sdk: elementx: Prefetching pagination with 21 off-screen (15 total) items |
03-12 20:42:29.342  6436  6436 D org.matrix.rust.sdk: elementx: Prefetching pagination with 59 off-screen (53 total) items |
03-12 20:42:31.803  6436  6436 D org.matrix.rust.sdk: elementx: Prefetching pagination with 91 off-screen (85 total) items |
03-12 20:42:33.823  6436  6436 D org.matrix.rust.sdk: elementx: Prefetching pagination with 124 off-screen (118 total) items |
03-12 20:42:35.450  6436  6436 D org.matrix.rust.sdk: elementx: Prefetching pagination with 152 off-screen (146 total) items |
03-12 20:42:37.770  6436  6436 D org.matrix.rust.sdk: elementx: Prefetching pagination with 186 off-screen (180 total) items |
```

ganfra:
@jmartinesp what are we doing with this then? Trying to improve or merge as is?

jmartinesp:
I think we may want to tweak it a bit more before merging it. I'll try to get some time today to look at this again.

jmartinesp:
I think I found a sweet spot prefetching at -40 items *and* using `snapshotFlow` so we don't have to wait for a recomposition to trigger the 'load more' action. I'll experiment with this a bit more, but I just loaded the whole previously loaded history of an encrypted room (~1000 items) at full speed without a single loading indicator.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -218,6 +214,41 @@ private fun MessageShieldDialog(state: TimelineState) {
...
 }
 
+@OptIn(ExperimentalCoroutinesApi::class)
+@Composable
+private fun TimelinePrefetchingHelper(
+    lazyListState: LazyListState,
+    prefetch: () -> Unit,
+) {
+    // We're using snapshot flows for these because using `LaunchedEffect` with `derivedState` doesn't seem to be responsive enough
+    val firstVisibleItemIndexFlow = snapshotFlow {
```
ganfra, line 229: This looks wrong, I don't think `snapshotFlow` should be created in the `Composable` scope, but are intended to be created in the CoroutineScope they are used. I think here you are recreating flows each time

jmartinesp: You're right, it should be used inside the `LaunchedEffect`. Now I'm worried that this was actually working and the tests passed 😓 .

ganfra (approved):
Lets merge and see how it goes!

PR #4405, opened 2025-03-13
Add Google Tink dependency, replacing `androidx.security.crypto` 

jmartinesp:
## Content

- Remove the deprecated [androidx.security.crypto](https://developer.android.com/reference/androidx/security/crypto/) library.
- Replace it with a simplified but backwards compatible implementation using [Tink](https://developers.google.com/tink). There are some constants that are now hardcoded, but those were previously in the replaced library and accessible to anyone, so it shouldn't be a security concern.

## Motivation and context

The library is being deprecated and was causing some conflicts related to it using an outdated Tink version. Maybe @BillCarsonFr should also take a quick look?

## Tests

Start the app with a logged in account, if it works fine the replacement works.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


libraries/encrypted-db/src/main/kotlin/io/element/encrypteddb/crypto/EncryptedFile.kt
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * It uses hardcoded constants that are used in that library, for backwards compatibility reasons.
+ */
+internal class EncryptedFile(
+    private val context: Context,
+    private val file: File
+) {
+    private val androidKeysetManager by lazy {
+        val keysetManagerBuilder = AndroidKeysetManager.Builder()
+            .withKeyTemplate(KeyTemplates.get("AES256_GCM_HKDF_4KB"))
```
ganfra, line 33: Can you extract those constants? 
I don't really like to have to use all those androidx constants by default :/

jmartinesp: It's not great, but using our own constants means adding a migration system for somethings that's not supposed to change ever (unless we wanted to change something like the encryption scheme 😓 ).

ganfra (approved):

PR #4406, opened 2025-03-13
Update SDK version to `25.03.13` and fix breaking changes

jmartinesp:
## Content

Changes:

- Upgrade SDK to `v25.03.13`.
- Fix OIDC prompts used for login, remove not supported ones that weren't being used anyway.
- Make `MatrixRoom.isEncrypted` optional to take into account the unknown state, added `MatrixRoom.getUpdatedIsEncrypted()` to fetch the latest value.
- Make usages of `isEncrypted` async or default to considering it `false` if no value is present.
- Turn `TracingInitializer` into `PlatformInitializer`, since it'll start not only the tracing service but also set up the amount of threads the SDK should use.
- Fixed some more Rust APIs becoming async (such as `Client::get_profile` or `Client::get_notification_settings`).
- Moved all room data to `RoomInfo`, made sure `RustMatrixRoom` always has an initial room info instance we can check.

## Motivation and context

Upgrading the SDK version was needed and there were lots of API changes.

## Tests

- Check OIDC login still works.
- Check the UI still matches the DM / encryption info in several screens.
- Logs should still work as they used to.
- Notification settings should also be updated without issues.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


ganfra (approved):
Looking good overall, thanks for the changes over room info!
2 questions : 
- should we not make sure we call the `getUpdatedIsEncrypted` in the `RoomFlowNode`?
- Maybe we should keep the `EncryptionState.Unknown` in our layer? at least there are some places where we assert that `isEncrypted == encryptionState.ENCRYPTED` and so we can display wrong information (like the "room non encrypted" badge).

jmartinesp:
> _Edit_ : I've checked [the rust PR](https://github.com/matrix-org/matrix-rust-sdk/pull/4778) and as far as I understand we can't get `Unknown` anymore. If this is true, then forget my previous questions :P

I've added some code to fetch the encryption state if it's still somehow unknown for some reason (I don't really know which SDK state could make this happen, but better safe than sorry).

ganfra (approved):
Thanks for the changes, lets merge then!

PR #4421, opened 2025-03-17
Sync Strings

ElementBot:


screenshots/de/features.ftue.impl.sessionverification.choosemode_ChooseSelfVerificationModeView_Day_2_de.png
jmartinesp, line 1: You can't read the whole text in the two buttons at the bottom, maybe we should review the German translations or try to find an alternative way to display the text: auto-sizing the font in the button, using several lines, displaying some toast with the full text when long pressed...

bmarty: Yes, that's annoying. The recorded screenshots are performed on a rather small screen. I am not saying that we should not work to improve that, but this limit the occurrences of such issues in real life.

PR #4432, opened 2025-03-18
Run the 'prevent blocked' workflow even if PR has conflicts

robintown:
See the description of `pull_request` versus `pull_request_target` documented at https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request.


bmarty (approved):
Thanks!

PR #4434, opened 2025-03-19
Add option to hide deleted messages

### LINKED ISSUE #4417: Add option to hide redacted (deleted) messages

Add an option in the advanced settings to hide deleted messages instead of showing "message deleted". This feature exists in the old Element Android.

Give users to option to hide unnecessary "message deleted" placeholders.

- Add option to the advanced settings. I'm not sure whether the hiding behavior should be the default. In the old Android app, turning the setting off (default I think) hides the messages
- filter redacted events in the timeline, also hide date headers if all messages under that header are redacted

_No response_

### Additional context

_No response_

Yes

Xant3s:
## Content

Adds the option to hide redacted (deleted) messages from the timeline. The user can enable or disable this feature in the advanced settings. 

I was unsure whether to name this setting "hide deleted messages" or "show placeholders for deleted messages" and which behavior should be default. Right now it's called "hide deleted messages" and is set to false by default.

## Motivation and context

Closes #4417

## Screenshots / GIFs

|Before|After|
|-|-|
|image|image|
|image|Placeholder message is gone|

## Tests

- Write a message in any room
- Delete that message
- Enable the setting "hide deleted messages" in the advanced settings and navigate back to your test room
- Disable the setting "hide deleted messages" in the advanced settings and navigate back to your test room to see the difference

## Tested devices

- [ ] Physical
- [X] Emulator
- OS version(s): API 24, API 34

## Checklist

- [X] UI change has been tested on both light and dark themes
- [X] Pull request title will be used in the release note, it clearly define what will change for the user
- [X] Pull request includes screenshots or videos if containing UI changes

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
@@ -28,6 +28,9 @@ import io.element.android.features.messages.impl.timeline.factories.TimelineItem
 import io.element.android.features.messages.impl.timeline.factories.TimelineItemsFactoryConfig
 import io.element.android.features.messages.impl.timeline.model.NewEventState
 import io.element.android.features.messages.impl.timeline.model.TimelineItem
+import io.element.android.features.messages.impl.timeline.model.event.isRedacted
+import io.element.android.features.messages.impl.timeline.model.virtual.TimelineItemDaySeparatorModel
+import io.element.android.features.messages.impl.timeline.model.virtual.TimelineItemTypingNotificationModel
 import io.element.android.features.messages.impl.typing.TypingNotificationState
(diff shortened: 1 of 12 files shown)


jmartinesp:
Hi, thank you for your contribution. The team has discussed whether we want this behaviour or not, and these are the conclusions:
- On one side, it would be nice to have feature parity with EW and legacy EA, but maybe instead of having the same behaviour we'd want to be a bit smarter about it, i.e. hiding deleted messages in DMs, or if it's the sender of the message the one who redacted it too, but displaying it if some moderator/admin redacted the event instead.
- Also, given we're modifying the timeline this can have side effects as the ones you tried to manage in your PR: virtual items as day dividers have to be taken into account and the same thing happens with read receipts and other markers. To avoid inconsistencies, forgotten corner cases, etc. it seems like this is a change that belongs in the Rust SDK, since it's the source of the data to display.

So unfortunately we can't accept this PR, sorry. We may work on this in the future, but at the moment I'm sorry to say it's not a priority, since the Rust SDK team has plenty of already planned work for the next months.

PR #4449, opened 2025-03-21
Make verification screens scrollable and emoji labels multiline

jmartinesp:
## Content

- Make the labels in the emoji verification items have multiple lines.
- Make both verification screens scrollable, as iOS does (only header + content).
- To achieve this, change how `HeaderFooterPage` scrolls. The `LazyColumn` approach was removed to simplify the layout, now a scrollable `Column` is used with some tweaks.
- Renamed some variables in `HeaderFooterPage` so they're easier to tell apart.
- Made the top app bar in the verification screens transparent so content can scroll through it, tweaked the paddings so this is possible.
- Changed the icon used  while verifying the emojis in `VerifySelfSessionView` to match the one in the incoming verification (the right one).
- Removed extra info from `VerificationEmoji` which wasn't used anywhere.

## Motivation and context

Fixes https://github.com/element-hq/customer-success/issues/450?issue=element-hq%7Celement-meta%7C2761

## Screenshots / GIFs

There are a few screenshots in the PR. This has the side effect of some buttons moving to the bottom now.

## Tests

If you're using a device with a small screen or large font size, the emoji verification screen should be scrollable (and usable) now.

User initiated verification should display the right icon in the header.

Other screens using `HeaderFooterPage` should still work as they used to.

Alternatively, you can test this with compose previews.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


ganfra (approved):
LGTM like this! 
Have you checked all the screens where we use this HeaderFooterPage? I guess screenshot tests should be safe to trust :D

jmartinesp:
> LGTM like this! Have you checked all the screens where we use this HeaderFooterPage? I guess screenshot tests should be safe to trust :D

I didn't test every single one, but I did check a few of them and didn't see much change.

PR #4454, opened 2025-03-21
Sync Strings - tweaks to identity change messages

andybalaam:
## Content

Tweak the strings in the identity change messages as per https://github.com/element-hq/element-meta/issues/2792

## Motivation and context

The existing strings were confusing, and users did not like clicking "Ok" when they did not understand, so this is changed to "Dismiss".

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [x] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly define what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #4461, opened 2025-03-24
Sync Strings

ElementBot:


jmartinesp:
This needs a patch like the following, added for https://github.com/element-hq/element-x-android/pull/4454:

[sync-strings-orig.patch](https://github.com/user-attachments/files/19417546/sync-strings-orig.patch)

The only change not included from the original patch is the one for `action_ok` -> `action_dismiss`,  as that was changed only in Andy's branch.

@andybalaam you might want to wait until this PR is merged before merging yours.

bmarty (approved):

PR #4470, opened 2025-03-25
Use embedded version of Element Call

jmartinesp:
## Content

- For room calls, the app will use an embedded version of Element Call shipped with the app instead of using an external service.
- We no longer use `element.json` file for customising the EC instance used by the app.
- We allow setting a custom base URL for testing purposes.

## Motivation and context

Embed Element Call inside the app to avoid relying on an external service for VOIP calls.

This is better reviewer commit-by-commit.

## Tests

- Join a call in a room, check everything works as it used to.
- Now open a call from `https://call.element.io` or some other endpoint in the app, this should also work.
- Go to developer settings and change the EC base url to something like `https://call.element.dev/room` (note the `/room` suffix **is mandatory**). Joining a room call should still work, but using the external URL just provided.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewWidgetMessageInterceptor.kt
```diff
@@ -117,6 +123,15 @@ class WebViewWidgetMessageInterceptor(
...
                 super.onReceivedSslError(view, handler, error)
             }
+
+            override fun shouldInterceptRequest(view: WebView?, request: WebResourceRequest): WebResourceResponse? {
+                println("Intercepting: ${request.url}")
+                return assetLoader.shouldInterceptRequest(request.url)
+            }
+
+            override fun shouldInterceptRequest(view: WebView?, url: String): WebResourceResponse? {
+                return assetLoader.shouldInterceptRequest(Uri.parse(url))
```
ElementBot, line 133: - :warning: Use the KTX extension function `String.toUri` instead?

services/analytics/api/src/main/kotlin/io/element/android/services/analytics/api/AnalyticsStore.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import kotlinx.coroutines.flow.Flow
+
+/**
+ * Local storage for:
+ * - user consent (Boolean);
+ * - did ask user consent (Boolean);
+ * - analytics Id (String).
+ */
+interface AnalyticsStore {
```
bmarty, line 18: I do not understand why this interface seems to be added by this PR. It seems that it already removed from `DefaultAnalyticsStore.kt`, but not from this PR.

jmartinesp: Ah, I now see this wasn't needed, I could have achieved the same with `AnalyticsService.didAskUserConsent()` 🫤 , I missed that one.

bmarty: yes, the store must stay internal. But I do not see the removing of previous `AnalyticsStore` in the diff. Do you see it on your side?

bmarty: It's been moved in https://github.com/element-hq/element-x-android/commit/3597774dd245d6dac7c86d0d533ed970c4645926

bmarty: OK, there is 2 interfaces `AnalyticsStore` :).
There need to be moved back to the `impl` module.

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewWidgetMessageInterceptor.kt
ElementBot, line 133: - :warning: Use the KTX extension function `String.toUri` instead?

bmarty, line 133: Will this work when the Element Call url is tweaked by the user (`getCustomElementCallBaseUrlFlow`)?

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/developer/DeveloperSettingsView.kt
```diff
@@ -136,22 +136,20 @@ private fun ElementCallCategory(
...
+        val supportingText = if (callUrlState.baseUrl.isNullOrEmpty()) {
             stringResource(R.string.screen_advanced_settings_element_call_base_url_description)
         } else {
             callUrlState.baseUrl
         }
         PreferenceTextField(
             headline = stringResource(R.string.screen_advanced_settings_element_call_base_url),
-            value = callUrlState.baseUrl ?: callUrlState.defaultUrl,
+            value = callUrlState.baseUrl,
+            placeholder = "https://.../room",
```
bmarty, line 148: Why a `/room` in the placeholder?

gradle/libs.versions.toml
```diff
@@ -211,6 +211,9 @@ dagger_compiler = { module = "com.google.dagger:dagger-compiler", version.ref =
 anvil_compiler_api = { module = "dev.zacsweers.anvil:compiler-api", version.ref = "anvil" }
 anvil_compiler_utils = { module = "dev.zacsweers.anvil:compiler-utils", version.ref = "anvil" }
 
+# Element Call
+element_call_embedded = "io.element.android:element-call-embedded:0.9.0-rc.2"
```
bmarty, line 215: 🎉

bmarty (approved):
Nice work!
I managed to start a call 🎉 

One thing, when starting a call in airplane mode, there is no error, since the code is loaded locally. So we're not helping the user here. See #3527

**EDIT**

At the end, I get an error like this:

image

So I guess it's fine!

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewWidgetMessageInterceptor.kt
jmartinesp: Yes, it will return `null`, which will tell the `WebView` there's no need to intercept the request.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/developer/DeveloperSettingsView.kt
jmartinesp: The new EC URL path for in-app calls (room calls) uses the `/room` suffix.

bmarty: Not sure to understand, as it is the url of the service hosted somewhere (no?). So I do not see the constraint on the URL path. But OK.

libraries/matrix/impl/build.gradle.kts
```diff
@@ -33,8 +33,6 @@ dependencies {
     implementation(projects.libraries.network)
     implementation(projects.libraries.preferences.api)
     implementation(projects.services.analytics.api)
-    implementation(projects.services.analyticsproviders.posthog)
-    implementation(projects.services.analyticsproviders.sentry)
```
bmarty, line 37: Thanks!

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/developer/DeveloperSettingsView.kt
hughns: @bmarty you are correct that the URL path isn't constrained. So, I would recommend not having the placeholder value at all. This is consistent with the [iOS implementation](https://github.com/element-hq/element-x-ios/pull/3939).

bmarty: OK, thanks. Tracked by https://github.com/element-hq/element-x-android/issues/4475

escix:
I am guessing this only works for joining call in a room (where a call is already started).
Because, without the developer option with element-call URL, I was not able to make a call but was able to join a call.

escix:
> > I am guessing this only works for joining call in a room (where a call is already started). Because, without the developer option with element-call URL, I was not able to make a call but was able to join a call.
> 
> This should be used for both joining and starting new calls in a room. If I understood correctly, you tried starting a new call in some room and it failed? Could you create an issue if that's the case?

Sorry, it is my fault.
I just found out the well-known MatrixRTC foci needs to be in a array format even for single url. That fixed the issue.

Issue #4389, opened 2025-03-11
Element Call Ringing does not stop


fkwp:
### Steps to reproduce

1. The application is running in the background
2. Call ringing is happening due an incoming call in a DM room
3. The caller stops the call OR I pickup the call with another device
4. Ringing is still in progress rather than stopping

#### What did you expect?
Ringing stops as soon as
- the caller stops the call
- or I did pickup the call with my other device

#### What happened instead?
it keeps ringing until a timeout

### Will you send logs?

Yes

### Are you willing to provide a PR?

Yes

manuroe:
Investigations for new (or remaining) problems will happen in https://github.com/element-hq/element-x-android/issues/4604.

PR #4478, opened 2025-03-26
Add `WakeLock` to dismiss ringing call screen when call is cancelled

jmartinesp:
## Content

Add a wakelock to keep the app alive while a ringing incoming call notification/screen is displayed.

## Motivation and context

We already had some checks in place to automatically cancel a ringing call notification/screen when the call was no longer active, but the `RoomInfo` updates weren't being processed because the app was 'paused'.

The partial wakelock should ensure these room info updates are handled, at least it did in my local tests.

Should fix https://github.com/element-hq/element-x-android/issues/4389.

## Tests

Having 2 devices with different accounts and a DM between them:

- Lock one of the devices.
- Start a call in the DM room from the other one. The first device should display a ringing incoming call screen.
- Hang up from the calling device.
- The ringing incoming call screen should be dismissed shortly after.

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): 14, 16

## Checklist

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/DefaultElementCallEntryPoint.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/DefaultElementCallEntryPoint.kt
@@ -18,15 +18,12 @@ import io.element.android.libraries.di.AppScope
 import io.element.android.libraries.di.ApplicationContext
 import io.element.android.libraries.matrix.api.core.EventId
 import io.element.android.libraries.matrix.api.core.UserId
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.launch
 import javax.inject.Inject
 
 @ContributesBinding(AppScope::class)
 class DefaultElementCallEntryPoint @Inject constructor(
     @ApplicationContext private val context: Context,
     private val activeCallManager: ActiveCallManager,
-    private val coroutineScope: CoroutineScope,
 ) : ElementCallEntryPoint {
     companion object {
         const val EXTRA_CALL_TYPE = "EXTRA_CALL_TYPE"
@@ -37,7 +34,7 @@ class DefaultElementCallEntryPoint @Inject constructor(
         context.startActivity(IntentProvider.createIntent(context, callType))
     }
 
-    override fun handleIncomingCall(
+    override suspend fun handleIncomingCall(
         callType: CallType.RoomCall,
         eventId: EventId,
         senderId: UserId,
@@ -60,8 +57,6 @@ class DefaultElementCallEntryPoint @Inject constructor(
             notificationChannelId = notificationChannelId,
             textContent = textContent,
         )
-        coroutineScope.launch {
-            activeCallManager.registerIncomingCall(notificationData = incomingCallNotificationData)
-        }
+        activeCallManager.registerIncomingCall(notificationData = incomingCallNotificationData)
     }
 }
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/DefaultElementCallEntryPoint.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/DefaultElementCallEntryPoint.kt
@@ -18,12 +18,15 @@ import io.element.android.libraries.di.AppScope
 import io.element.android.libraries.di.ApplicationContext
 import io.element.android.libraries.matrix.api.core.EventId
 import io.element.android.libraries.matrix.api.core.UserId
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
 import javax.inject.Inject
 
 @ContributesBinding(AppScope::class)
 class DefaultElementCallEntryPoint @Inject constructor(
     @ApplicationContext private val context: Context,
     private val activeCallManager: ActiveCallManager,
+    private val coroutineScope: CoroutineScope,
 ) : ElementCallEntryPoint {
     companion object {
         const val EXTRA_CALL_TYPE = "EXTRA_CALL_TYPE"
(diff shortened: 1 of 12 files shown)


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/ActiveCallManager.kt
```diff
@@ -111,6 +130,12 @@ class DefaultActiveCallManager @Inject constructor(
             delay(ElementCallConfig.RINGING_CALL_DURATION_SECONDS.seconds)
             incomingCallTimedOut(displayMissedCallNotification = true)
         }
+
+        // Acquire a wake lock to keep the device awake during the incoming call, so we can process the room info data
+        if (activeWakeLock?.isHeld == true) {
+            activeWakeLock.release()
+        }
+        activeWakeLock?.acquire(ElementCallConfig.RINGING_CALL_DURATION_SECONDS * 1000L)
```
ganfra, line 138: Can we use the non timeout version here and keep the one from before?
Also, it doesn't seem to be thread safe, not sure on which thread those methods are called

jmartinesp: Actually, if I remove it I get this lint warning:

> Provide a timeout when requesting a wakelock with PowerManager. Wakelock. acquire(long timeout). This will ensure the OS will cleanup any wakelocks that last longer than you intend, and will save your user's battery

Which I think makes sense, and it's safer with the timeout.

About the thread safety, do we want to just wrap every function with a mutex lock?

ganfra: If we have suspending method everywhere, yes I guess

jmartinesp: Actually I tried that and the result is not great, it leads to adding new dependencies, makes the tests more complex... 🫤

But I'm not sure if using `@Synchronized` or a synchronized lock would be a good solution either.

jmartinesp: In the end I used `Mutex`. Let me know if it's better.

jmartinesp:
Rebased so we have the fix for the incoming verification flaky tests in the branch.

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
```diff
@@ -187,7 +188,7 @@ class CallScreenPresenter @AssistedInject constructor(
         )
     }
 
-    private suspend fun loadUrl(
+    private suspend fun getRoomCallUrl(
```
bmarty, line 191: A bit weird to have a method with `get` prefix that return `Unit`.

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/receivers/DeclineCallBroadcastReceiver.kt
```diff
@@ -27,10 +29,16 @@ class DeclineCallBroadcastReceiver : BroadcastReceiver() {
     }
     @Inject
     lateinit var activeCallManager: ActiveCallManager
+
+    @Inject
+    lateinit var coroutineScope: CoroutineScope
```
bmarty, line 34: Maybe stick to the `appCoroutineScope` naming?

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/DefaultElementCallEntryPoint.kt
```diff
@@ -18,12 +18,15 @@ import io.element.android.libraries.di.AppScope
...
 import io.element.android.libraries.matrix.api.core.UserId
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
 import javax.inject.Inject
 
 @ContributesBinding(AppScope::class)
 class DefaultElementCallEntryPoint @Inject constructor(
     @ApplicationContext private val context: Context,
     private val activeCallManager: ActiveCallManager,
+    private val coroutineScope: CoroutineScope,
```
bmarty, line 29: Maybe stick to the `appCoroutineScope` naming?

But for this particular case, maybe we do not need a scope and `handleIncomingCall` could be `suspend`?

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/ActiveCallManager.kt
```diff
@@ -55,25 +61,26 @@ interface ActiveCallManager {
...
      * @param callType The type of call that the user joined, either an external url one or a room one.
      */
-    fun joinedCall(callType: CallType)
+    suspend fun joinedCall(callType: CallType)
 }
 
 @SingleIn(AppScope::class)
 @ContributesBinding(AppScope::class)
 class DefaultActiveCallManager @Inject constructor(
+    @ApplicationContext private val context: Context,
```
bmarty, line 83: Does not need to be a class member.

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/ActiveCallManager.kt
```diff
@@ -83,44 +90,70 @@ class DefaultActiveCallManager @Inject constructor(
...
+
+        if (powerManager?.isWakeLockLevelSupported(PowerManager.PARTIAL_WAKE_LOCK) == true) {
+            powerManager.newWakeLock(
+                PowerManager.PARTIAL_WAKE_LOCK,
+                "${context.packageName}:IncomingCallWakeLock",
+            )
+        } else {
+            null
+        }
+    }
```
bmarty, line 105: Could be simplified as:
```kotlin
    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
    internal val activeWakeLock: PowerManager.WakeLock? = context.getSystemService<PowerManager>()
        ?.takeIf { it.isWakeLockLevelSupported(PowerManager.PARTIAL_WAKE_LOCK) }
        ?.newWakeLock(PowerManager.PARTIAL_WAKE_LOCK, "${context.packageName}:IncomingCallWakeLock")
```

bmarty (approved):
LGTM thanks! A few minor remarks.

Issue #4480, opened 2025-03-26
Using the volume rocker during calls is not affecting the call audio volume.


toger5:
fixed by: https://github.com/element-hq/element-x-android/pull/4481

This is because element X outputs audio over the media os channel but while in a call the volume rocker (hardware buttons) will affect the in-call volume.

A simple solution would be to make the volume rocker affect the media output:
https://developer.android.com/media/platform/output
`setVolumeControlStream(AudioManager.STREAM_MUSIC)` (or another `AudioManager.`)

The more proper solution would be to make the web-view output to the in-call "audio channel/sink"

PR #4481, opened 2025-03-26
Use in-call volume and mode for EC

jmartinesp:
## Content

When the call starts, get the list of audio devices and try to set the 'most likely to be used' device as audio input + output. This also changes the audio mode to 'in communication', making the volume rocker change the 'in-call' audio stream.

This works as a workaround for not being able to pick the audio output in the EC settings, since this is not supported by the Android WebView (and it doesn't seem like it'll be supported any time soon).

Note that adding a new audio device while in-call won't work, the workaround can't make that work too, sadly. We'll see if we can do something in the future.

## Motivation and context

Should fix https://github.com/element-hq/element-x-android/issues/4480.

## Tests

- Before starting a call, connect either some wired or wireless headphones, with microphone, if possible.
- Start/join the call, check the audio input and output of the device is used.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenView.kt
```diff
@@ -161,6 +167,10 @@ private fun CallWebView(
                 }
             },
             onRelease = { webView ->
+                // Reset audio mode
+                val audioManager = webView.context.getSystemService<AudioManager>()
+                audioManager?.mode = AudioManager.MODE_NORMAL
```
bmarty, line 172: Maybe store the previous mode and restore it here?

jmartinesp: From what I saw in some issues (this is poorly documented, AFAICT) the app can be 'blacklisted' if it doesn't play any audio in the 6s after enabling this mode or doesn't return to the `MODE_NORMAL` when it's done playing.

bmarty: OK, thanks.

toger5:
I tested it and it works. I can change the output volume while in a call with the hardware buttons (without using androids per sink volume control ui)
The output device selection in EC does not work however. It will always defaults to the loudspeaker.

I think this makes sense to become its own issue?
It will need more work on the product to design how we want to control the output devices in EC.

It will would be interesting to have an experiment if it event works?
Since we send widget actions on mute/unmute i am wondering if we can throw together a short experiment with this pr:
 - muting the camera switches the android media device to the head piece
 - unmuting the camera switches to the loudspeaker
 ?

toger5:
Another issue I observed with this is, that the min volume is still very audible (there is no 0 volume option for in call audio)

Which is very nice for the user so they will never be stuck with no sound and also what a normal call does.

It is still sth we need to be aware of, since we might want to be able to silence the phone if in a conference when someone else comes by in the real world.

I think this is sth we would like to solve in EC where we provide a mute output button or similar.

toger5:
@jmartinesp
It is okay to never be able to mute the call: https://github.com/element-hq/element-call/issues/3137
So using the in-call output is the way to go.

toger5:
> @toger5 I think you already did, but can you confirm that even though the changes don't fix all the issues, we can merge this as is and iterate later if needed? Thanks!

Yes when i tested it this was already an imporvement and worked as expected (volume rocker changed call sound)

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/compat/AudioManager.kt
```diff
@@ -0,0 +1,52 @@
...
+            AudioDeviceInfo.TYPE_USB_HEADSET,
+            AudioDeviceInfo.TYPE_USB_DEVICE,
+            AudioDeviceInfo.TYPE_USB_ACCESSORY,
+            // Wired audio devices
+            AudioDeviceInfo.TYPE_WIRED_HEADSET,
+            AudioDeviceInfo.TYPE_WIRED_HEADPHONES,
+            // The built-in earpiece of the device
+            AudioDeviceInfo.TYPE_BUILTIN_EARPIECE,
+            // The built-in speaker of the device
+            AudioDeviceInfo.TYPE_BUILTIN_SPEAKER,
```
bmarty, line 30: For Video Call, I guess TYPE_BUILTIN_SPEAKER should have a higher priority than TYPE_BUILTIN_EARPIECE, no?

jmartinesp: Oh, nice catch.

PR #4489, opened 2025-03-27
Upgrade embedded EC version to `v0.9.0-rc.4`

jmartinesp:
## Content

What the title says, the VOIP team released a new version of the embedded EC library.

## Motivation and context

This should fix an issue with sending rageshakes from EC.

## Checklist

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -212,7 +212,7 @@ anvil_compiler_api = { module = "dev.zacsweers.anvil:compiler-api", version.ref 
 anvil_compiler_utils = { module = "dev.zacsweers.anvil:compiler-utils", version.ref = "anvil" }
 
 # Element Call
-element_call_embedded = "io.element.android:element-call-embedded:0.9.0-rc.2"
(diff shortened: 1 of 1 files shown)


bmarty (approved):
Thanks.
The CI does not help a lot to check the validity of this upgrade, we should maybe consider adding a Maestro test to start a call?

bmarty:
If we remove the dependency of this library, the CI will still be happy :).
Starting a call will lead to a network error (ofc):

image

We should maybe add a small API in the library that we invoke (like getVersion()) or something, that we invoke to ensure that the library is there! There is also maybe a risk that R8 is shrinking it (but low since we should already have some issue if it was case).

hughns:
> If we remove the dependency of this library, the CI will still be happy :). Starting a call will lead to a network error (ofc):
> 
> image
> We should maybe add a small API in the library that we invoke (like getVersion()) or something, that we invoke to ensure that the library is there! There is also maybe a risk that R8 is shrinking it (but low since we should already have some issue if it was case).

@BillCarsonFr has recently been working on testing of Element Call within an Element Web instance by spinning up a homeserver/SFU stack. Perhaps that work could also be ported to work with EX Android?

Issue #3772, opened 2024-10-31
Replying to a very long message crashes the app


everypizza1:
### Steps to reproduce

1. Send a very long message 
2. Attempt to reply to it

#### What did you expect?
App shouldn't crash
#### What happened instead?
App does crash 

### Your phone model

Google Pixel 8a

### Operating system version

Android 15

### Application version and app store

Element X 0.7.2 Google Play

### Homeserver

sulian.eu/Synapse 1.118

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

jmartinesp:
Can you provide some more info? I tried replying to a lorem ipsum of 22594 characters and it worked just fine.

In the stacktrace I see:

> java.lang.IllegalArgumentException: Can't represent a width of 302655 and height of 0 in Constraints

Could it be that all of the text was in a single line?

jmartinesp:
> Could it be that all of the text was in a single line?

Yes, that did it. I believe this is a jetpack compose issue (the UI library in modern Android apps) since there were some previous crashes with similar causes that were fixed. I'm not sure if we can work around this somehow.

PR #4491, opened 2025-03-27
Limit the text length in the 'in reply to' preview

jmartinesp:
## Content

What the title says.

## Motivation and context

Otherwise, very long texts with no line breaks can cause a Compose crash.

Fixes https://github.com/element-hq/element-x-android/issues/3772

## Screenshots / GIFs

There shouldn't be any visual change, unless you have a huge screen that can display in reply to previews of > 512 chars.

## Tests

- Generate a very long text, > 20k characters, maybe using a lorem ipsum and *make sure there are no line breaks*.
- Send the text.
- Try to reply to the message.

If the app doesn't crash, we worked around the Compose bug.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/reply/InReplyToView.kt
+++ b/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/reply/InReplyToView.kt
@@ -31,7 +31,6 @@ import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.tooling.preview.PreviewParameter
 import androidx.compose.ui.unit.dp
 import io.element.android.compound.theme.ElementTheme
-import io.element.android.libraries.core.extensions.ellipsize
 import io.element.android.libraries.core.extensions.toSafeLength
 import io.element.android.libraries.designsystem.atomic.atoms.PlaceholderAtom
 import io.element.android.libraries.designsystem.icons.CompoundDrawables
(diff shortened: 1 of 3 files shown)


libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/reply/InReplyToView.kt
```diff
@@ -152,8 +153,10 @@ private fun ReplyToContentText(metadata: InReplyToMetadata?) {
     val text = when (metadata) {
         InReplyToMetadata.Redacted -> stringResource(id = CommonStrings.common_message_removed)
         InReplyToMetadata.UnableToDecrypt -> stringResource(id = CommonStrings.common_waiting_for_decryption_key)
-        is InReplyToMetadata.Text -> metadata.text
-        is InReplyToMetadata.Thumbnail -> metadata.text
+        // Add a limit to the text length to avoid a crash in Compose
+        is InReplyToMetadata.Text -> metadata.text.ellipsize(512)
+        // Add a limit to the text length to avoid a crash in Compose
+        is InReplyToMetadata.Thumbnail -> metadata.text.ellipsize(512)
```
bmarty, line 159: I remember that we already have a similar fix (limiting the size of text to avoid a Compose crash), but I can find it. Maybe I dreamt it?
Else maybe create a const for this `512` (not a big deal)?

bmarty (approved):
Thanks!

libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/reply/InReplyToView.kt
jmartinesp: I had the same issue, I know it's there, I just don't know where 😅 .

bmarty: > I've tried unifying both in an extension function in a new commit.

That's also why I wanted to find where it was done before. Thanks for the update!

bmarty (approved):

PR #4499, opened 2025-03-28
Add video autoplay to media gallery

jmartinesp:
## Content

- Add video autoplay to any video the user explicitly opens (either from the timeline or the media gallery). Scrolling between videos in the gallery shouldn't result in autoplaying any other videos.
- Make sure reloading the media gallery items doesn't result in the video playback being paused then shortly resumed again.

## Motivation and context

Fixes https://github.com/element-hq/element-meta/issues/2797, achieving parity with iOS.

## Tests

- Open a video from the timeline, it should be displayed in the media gallery of the room you're in and autoplayed.
- If you scroll to some other video, this one should not autoplay.
- There are no accidental play/pause changes when new items are added to the media gallery.
- Scrubbing the video, moving to a different timestamp, should keep working as expected.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/datasource/MediaGalleryDataSource.kt
+++ b/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/datasource/MediaGalleryDataSource.kt
@@ -19,6 +19,7 @@ import io.element.android.libraries.mediaviewer.impl.model.GroupedMediaItems
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.launchIn
@@ -85,7 +86,9 @@ class TimelineMediaGalleryDataSource @Inject constructor(
             }
         }.flatMapLatest {
             timelineMediaItemsFactory.timelineItems
-        }.map { timelineItems ->
+        }
+        .distinctUntilChanged()
+        .map { timelineItems ->
             mediaItemsPostProcessor.process(mediaItems = timelineItems)
         }.map {
             mediaTimeline.orCache(it)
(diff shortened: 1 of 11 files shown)


jmartinesp:
Oops, there is some leftover code from experiments that should not be there anymore, sorry. I'll fix that ASAP.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/datasource/MediaGalleryDataSource.kt
```diff
@@ -85,11 +86,13 @@ class TimelineMediaGalleryDataSource @Inject constructor(
...
-        }.map {
-            mediaTimeline.orCache(it)
-        }.onEach { groupedMediaItems ->
+        }
+        .distinctUntilChanged()
+        .map { timelineItems ->
+            val groupedItems = mediaItemsPostProcessor.process(mediaItems = timelineItems)
+            mediaTimeline.orCache(groupedItems)
+        }
+        .onEach { groupedMediaItems ->
```
jmartinesp, line 95: No actual changes here, but Detekt complained about this so I made some changes to make it happy.

bmarty (commented):
This is working (thanks for the last fix).
However I see one point that could be improved:
- opening a video;
- auto play starts;
- swipe to another media;
- the video is automatically paused;
- swipe back to the video.

The video auto start again but from the beginning, even if the user has manually paused it.

I believe that previously the media was paused and the player position was not lost, except if the user is swiping to much (in this case the View is disposed).

Maybe the video should not auto-play when swiping back to it?

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/local/video/MediaVideoView.kt
```diff
@@ -164,10 +175,19 @@ private fun ExoPlayerMediaVideoView(
...
+    LaunchedEffect(needsAutoPlay, isDisplayed, mediaPlayerControllerState.isReady) {
+        val isReadyAndNotPlaying = mediaPlayerControllerState.isReady && !mediaPlayerControllerState.isPlaying
+        if (needsAutoPlay && isDisplayed && isReadyAndNotPlaying) {
+            // When displayed, start autoplaying from the beginning
+            exoPlayer.play()
+            needsAutoPlay = false
+        } else if (!isDisplayed && mediaPlayerControllerState.isPlaying) {
+            // If not displayed, make sure to pause the video
             exoPlayer.pause()
+            exoPlayer.seekTo(0)
```
bmarty, line 190: Why resetting to player position here?

bmarty (approved):
Thanks for the update!

I see that `LocalMediaView` is also used for the media attachment preview (see code in `DefaultLocalMediaRenderer`). But I think that it's not a good idea to auto-play in this case, so the current implementation is fine.

One last non-blocking question about resetting the player location when swiping between media.

PR #4512, opened 2025-04-01
Improve TextFieldDialog

bmarty:
Follow up of #4507 , in particular [this comment](https://github.com/element-hq/element-x-android/pull/4507#discussion_r2022464402).


jmartinesp (approved):
Thanks!

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/dialogs/TextFieldDialog.kt
```diff
@@ -0,0 +1,154 @@
...
+                },
+                error = error,
+                keyboardOptions = keyboardOptions,
+                keyboardActions = KeyboardActions(onAny = {
+                    if (validation(textFieldContents.text)) {
+                        onSubmit(textFieldContents.text)
+                    }
+                }),
+                maxLines = maxLines,
+                modifier = Modifier.focusRequester(focusRequester),
```
jmartinesp, line 98: Actually, shouldn't this be `Modifier.fillMaxWidth().focus...` ? I just tried replacing the EC base URL component with this dialog and the text field doesn't reach the end.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/list/TextFieldListItem.kt
jmartinesp, line 1: This is not new, but I just realised we're using the wrong text field component. Instead of the M3 `OutlinedTextField` we should use our design's library `TextField`. This explains why the inlined `label` and the `placeholder` are used, while in our component the label is outside the text field.

jmartinesp (commented):
Actually, a couple of improvements if you have the time.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/dialogs/TextFieldDialog.kt
bmarty: Yes, maybe. Not that this code is not new. I'll check it and add more commit

bmarty: Actually, this PR is touching unrelated code (room member moderation), let's handle that in a dedicated one.

PR #4513, opened 2025-04-01
Make `RustMatrixClient.close` asynchronous

jmartinesp:
## Content

Make `fun RustMatrixClient.close()` suspend, removing the `Closeable` interface implementation from `MatrixRoom`.

Since this interface is no longer implemented, rename `close` method to `destroy` to match other implementations.

## Motivation and context

This is a safer way to destroy the Rust instances associated to a client, previously we were destroying some of those in a coroutine while others (which may depend on those) were done synchronously.

## Tests

Clear cache and/or logout. If there aren't any unexpected errors, it's working as expected.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
@@ -476,7 +476,7 @@ class RustMatrixClient(
 
     override fun roomDirectoryService(): RoomDirectoryService = roomDirectoryService
 
-    suspend fun destroy() {
+    internal suspend fun destroy() {
         innerNotificationClient.close()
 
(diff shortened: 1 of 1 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClient.kt
```diff
@@ -476,16 +476,13 @@ class RustMatrixClient(
 
     override fun roomDirectoryService(): RoomDirectoryService = roomDirectoryService
 
-    override fun close() {
+    suspend fun destroy() {
```
bmarty, line 479: Could be only visible for test.

bmarty (approved):
I am feeling quite confortable with this changes, thanks!

PR #4514, opened 2025-04-01
Enable Rust trace log packs

jmartinesp:
## Content

Adds new section in developer options to allow uploading trace logs for some specific features of the SDK.

## Motivation and context

This is a way to forcefully enable trace logs only for a few Rust crates in a safe way, it was added in the previous upgrade of the Rust SDK bindings.

## Screenshots / GIFs

image

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/developer/DeveloperSettingsPresenter.kt
+++ b/features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/developer/DeveloperSettingsPresenter.kt
@@ -77,6 +77,7 @@ class DeveloperSettingsPresenter @Inject constructor(
             appPreferencesStore.getTracingLogLevelFlow().map { AsyncData.Success(it.toLogLevelItem()) }
         }
         val tracingLogLevel by tracingLogLevelFlow.collectAsState(initial = AsyncData.Uninitialized)
+        val tracingLogPacks by appPreferencesStore.getTracingLogPacksFlow().collectAsState(emptySet())
 
         LaunchedEffect(Unit) {
             FeatureFlags.entries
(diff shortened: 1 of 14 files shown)


features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/developer/DeveloperSettingsPresenter.kt
```diff
@@ -77,6 +77,7 @@ class DeveloperSettingsPresenter @Inject constructor(
             appPreferencesStore.getTracingLogLevelFlow().map { AsyncData.Success(it.toLogLevelItem()) }
         }
         val tracingLogLevel by tracingLogLevelFlow.collectAsState(initial = AsyncData.Uninitialized)
+        val tracingLogPacks by appPreferencesStore.getTracingLogPacksFlow().collectAsState(emptySet())
```
bmarty, line 80: There is only 3 packs for now, but maybe better to use a sorted alphabetically list instead of a set?

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/tracing/TraceLogPack.kt
```diff
@@ -0,0 +1,22 @@
...
+        override val title: String = "Event Cache"
+    },
+    SEND_QUEUE("send_queue") {
+        override val title: String = "Send Queue"
+    },
+    TIMELINE("timeline") {
+        override val title: String = "Timeline"
+    };
+
+    abstract val title: String
```
bmarty, line 21: Why treating diffentely the `key` and the `title`? Is it for (future?) i18n support?

bmarty (approved):
Thanks. Maybe also add the subtitle `Requires app reboot` somewhere around the new toggles?

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/tracing/TraceLogPack.kt
jmartinesp: Yeah, mostly so if we add a typo by mistake it wouldn't save/load it as a different value.

bmarty: I see. The enum is serialized and stored. That a sufficient reason!

PR #4515, opened 2025-04-01
Make sure the live timeline is destroyed before clearing a room's cache

jmartinesp:
## Content

- Destroy live timeline instantiated right before clearing the room's cache. In a future refactor we should avoid creating the timeline at all, but at the moment this is the best we can do.
- Destroy all room references before evicting them from the LRU cache when destroying `RustMatrixClient` on logout / clear cache. This *should* also destroy their live timelines.

## Motivation and context

According to the SDK team, having a live timeline while clearing the cache could lead to DB issues.

## Tests

Clear the cache from the developer settings or the room's context menu. If it still works as expected, the change didn't cause new issues, and *should* improve existing ones.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


bmarty (approved):
OK, thanks!

PR #4533, opened 2025-04-04
Remember flows

bmarty:
## Content

Ensure that Flows are not created again when a recomposition occurred. They should be remembered, except if they are hold in a val. So this PR:
- add a Konsist test to detect potential issue with this
- add `remember` around flow function creator
- change `fun` which return flow to `val`, so that it's clear the the flow is not created at each usage.

## Motivation and context

Tiny performance improvement and Compose best practice.

## Screenshots / GIFs

No expected change

## Tests

- No expected change

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt
@@ -7,6 +7,7 @@
 
 package io.element.android.libraries.matrix.impl.sync
 
+import io.element.android.libraries.core.coroutine.mapState
 import io.element.android.libraries.matrix.api.sync.SyncService
 import io.element.android.libraries.matrix.api.sync.SyncState
 import kotlinx.coroutines.CoroutineDispatcher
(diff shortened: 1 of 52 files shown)


tests/konsist/src/test/kotlin/io/element/android/tests/konsist/KonsistFlowTest.kt
```diff
@@ -24,7 +24,12 @@ class KonsistFlowTest {
         val regex = "(.*)\\)(\n\\s*)*\\.collectAsState".toRegex()
 
         val allowedMethods = listOf(
-            "isOnline()",
+            "syncService.isOnline()",
+            "userListDataStore.selectedUsers()",
+            "accountProviderDataSource.flow()",
+            "syncService.isOnline()",
+            "timeline.paginationStatus",
+            "analyticsService.getUserConsent()",
```
jmartinesp: Maybe using an annotation to suppress the warning would be better, but I don't know if that's possible/easy. Just something we could keep in mind for future usages of Konsist.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt
```diff
@@ -73,4 +74,6 @@ class RustSyncService(
             }
             .distinctUntilChanged()
             .stateIn(sessionCoroutineScope, SharingStarted.Eagerly, SyncState.Idle)
+
+    override val isOnline: StateFlow<Boolean> = syncState.mapState { it != SyncState.Offline }
```
jmartinesp, line 78: TIL about our `mapState` helper function 👀 .

jmartinesp (approved):
LGTM, thanks!

PR #4577, opened 2025-04-11
Ensure that we have only one single instance of SeenInviteStore per session

bmarty:
## Content

Ensure that we have only one single instance of SeenInviteStore per session by using a SeenInviteStoreFactory.

## Motivation and context

Fix crash:
java.lang.IllegalStateException: There are multiple DataStores active for the same file: /data/user/0/io.element.android.x/files/datastore/session_0ebb139587b6d940_seen-invites.preferences_pb. You should either maintain your DataStore as a singleton or confirm that there is no two DataStore's active on the same file (by confirming that the scope is cancelled).

Fixes #4558

## Tests

- Run the app
- From an external app (for instance a photo app), share something to Element X
- If the app does not crash, the bug is fixed.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/invite/impl/src/main/kotlin/io/element/android/features/invite/impl/response/AcceptDeclineInvitePresenter.kt
+++ b/features/invite/impl/src/main/kotlin/io/element/android/features/invite/impl/response/AcceptDeclineInvitePresenter.kt
@@ -13,7 +13,7 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import im.vector.app.features.analytics.plan.JoinedRoom
-import io.element.android.features.invite.api.SeenInvitesStore
+import io.element.android.features.invite.api.SeenInvitesStoreFactory
 import io.element.android.features.invite.api.response.AcceptDeclineInviteEvents
 import io.element.android.features.invite.api.response.AcceptDeclineInviteState
 import io.element.android.features.invite.api.response.ConfirmingDeclineInvite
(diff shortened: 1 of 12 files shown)


features/invite/impl/src/main/kotlin/io/element/android/features/invite/impl/response/AcceptDeclineInvitePresenter.kt
```diff
@@ -35,8 +35,13 @@ class AcceptDeclineInvitePresenter @Inject constructor(
     private val client: MatrixClient,
     private val joinRoom: JoinRoom,
     private val notificationCleaner: NotificationCleaner,
-    private val seenInvitesStore: SeenInvitesStore,
+    seenInvitesStoreFactory: SeenInvitesStoreFactory,
 ) : Presenter<AcceptDeclineInviteState> {
+    private val seenInvitesStore = seenInvitesStoreFactory.getOrCreate(
+        sessionId = client.sessionId,
+        sessionCoroutineScope = client.sessionCoroutineScope,
+    )
```
jmartinesp, line 43: I don't mind using this, but I think you could also create a `@Provides` method for `SessionScope` that would return the cached store if needed:

```kotlin
@Provides
fun providesSeenInvitesStore(factory: SeenInvitesStoreFactory, sessionId: SessionId, @SessionCoroutineScope sessionCoroutineScope: CoroutineScope) {
    return factory.getOrCreate(sessionId, sesionCoroutineScope)
 }
 ```
 
 And then you can just inject the result in the session scope.

jmartinesp (approved):
Thanks I can confirm it fixes the issue for me!

features/invite/impl/src/main/kotlin/io/element/android/features/invite/impl/response/AcceptDeclineInvitePresenter.kt
bmarty: Ah that's a bit cleaner, will give it a try

bmarty: 80156681aeafdd0b4457ed894c046d94b492e194

PR #4580, opened 2025-04-11
Upgrade Rust bindings to `v25.04.11`

jmartinesp:
## Content

Upgrade the SDK bindings and fix what the SDK changes broke.

## Motivation and context

Keeping the app up to date.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/exception/ClientException.kt
+++ b/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/exception/ClientException.kt
@@ -7,10 +7,10 @@
 
 package io.element.android.libraries.matrix.api.exception
 
-sealed class ClientException(message: String) : Exception(message) {
-    class Generic(message: String) : ClientException(message)
-    class MatrixApi(val kind: ErrorKind, val code: String, message: String) : ClientException(message)
-    class Other(message: String) : ClientException(message)
+sealed class ClientException(message: String, val details: String?) : Exception(message) {
+    class Generic(message: String, details: String?) : ClientException(message, details)
+    class MatrixApi(val kind: ErrorKind, val code: String, message: String, details: String?) : ClientException(message, details)
+    class Other(message: String) : ClientException(message, null)
 }
 
 fun ClientException.isNetworkError(): Boolean {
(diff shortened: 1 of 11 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/item/event/EventMessageMapper.kt
```diff
@@ -23,29 +23,30 @@ import io.element.android.libraries.matrix.api.timeline.item.event.VideoMessageT
...
-            isEdited = it.isEdited,
-            isThreaded = it.threadRoot != null,
+            isEdited = it.content.isEdited,
+            isThreaded = isThreaded,
             type = type
         )
     }
 
-    fun mapMessageType(type: RustMessageType) = when (type) {
+    fun mapMessageKind(type: RustMessageType) = when (type) {
```
bmarty, line 49: Why this renaming?

jmartinesp: Oops, I can rollback that. At first I was using the new `MsgLikeKind` at this point, but I was lacking some info from the type so I restored that but I forgot to also restore the method name.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/exception/ClientException.kt
```diff
@@ -7,10 +7,10 @@
 
 package io.element.android.libraries.matrix.api.exception
 
-sealed class ClientException(message: String) : Exception(message) {
-    class Generic(message: String) : ClientException(message)
-    class MatrixApi(val kind: ErrorKind, val code: String, message: String) : ClientException(message)
-    class Other(message: String) : ClientException(message)
+sealed class ClientException(message: String, val details: String?) : Exception(message) {
```
bmarty, line 10: I am wondering if we should exploit this `details` at the application level.

Maybe something like:

```kotlin
sealed class ClientException(message: String, details: String?) : Exception(
    buildString {
        append(message)
        if (details != null) {
            append(" (")
            append(details)
            append(")")
        }
    }
) {
```

Not a blocker though.

bmarty (approved):
Thanks!

PR #4589, opened 2025-04-13
Refactor account provider

lex-neufeld:
## Content
UI: String for description of matrix.org changed in some places (now consistent and localized everywhere). Future UI development can use the resourceID from the AccountProvider or use something else if they want.

AccountProvider.subtitle: String is now AccountProvider.descriptionResourceId: Int

NOT YET IMPLEMENTED
AccountProvider renamed to ....? In addition to not using the keyword "Provider", it might make more sense to name it what it is, and not what it is being used for in the loginFlow. Homeserver appears to be the correct name in the Matrix universe, but it seems it is already used elsewhere in this project to refer only to Homeservers that come down the ElementEnterprise pipeline. Host? Server? ChatHost? ChatServer? People probably have feelings and ideas, but please not "Account" and not "Provider". 
PLEASE COMMENT

Deleted AccountProvider.isValid as it wasn't being used anywhere.

created AccountProviderDataSource.accountProvidersList to replace list creation in the UI and support having more than one item in the list and always include Matrix.Org

AccountProvider.isMatrixOrg is now AccountProvider.isMatrixOrg() (yes, in a data class. Kotlin allows it. Please comment.)

Cleaner code in Views, Presenters, Tests

AccountProviderDataSource.enterpriseService is now nullable and AccountProviderDataSource can be constructed with no parameters.

NOT FULLY IMPLEMENTED
Better unit testing of AccountProvider and login flow

## Motivation and context

1. AccoundProvider info for matrix.org was being hard-coded in many different places. Sometimes English strings, sometimes stringResrouces being used. Sometimes isValid = true, sometimes isValid = false. Logic in Views that probably shouldn't be there. Hard coded lists in the presentation layer. Lots of duplicated logic to set a few booleans that really only need to be set once at creation.
2. The name "Provider" is already used as a keyword for something else in many classes in this project. We end up with brain-cancer-inducing names like AccountProviderStateProvider. AccountProvider is not a UI Provider for Account, and actually isn't an account at all, but a server address and description. This adds unnecessary cognitive overhead to navigating the code.
3. "subtitle" is a keyword used by our UI elements. This adds unnecessary cognitive overhead to navigating the code.
4. nearly all uses of isMatrixOrg were work-arounds, and I saw a codebase where they just set MATRIX_ORG_URL to something else and element-x-android then just sets isMatrixOrg to true.
5. duplicate logic and duplicate hard coded values in Views and Presenters
6. Some places were hurting for quick and easy access to AccountProviderDataSource
7. Desire to preserve the sanctity of Matrix.Org and its inclusion in the UI even if the defaultAccountProvider is changed.
8. Increase centralization and encapsulation of AccountProviders.
9. Improve UnitTests for login flow

## Tests

1. manual testing of login flow using virtual device Medium Phone API 36 light and dark mode with matrix.org (logout, change, search, login), invalid server, other valid server (search, login attempt, wrong password)
2. ./tools/quality/check.sh (I think all are passing, needs to be run again)
3. ./gradlew ktlintFormat
4. ./gradlew knitCheck
5. ./gradlew test (MEDIA UPLOAD IS FAILING. I am 99% certain I didn't touch media upload in any way. Need to check)

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): macOS Sequoia 15.3.1, Android Studio Meerkat | 2024.3.1, Medium Phone API 36

## Checklist

- [x] Changes have been tested on an Android device or Android emulator with API 24 (IT WAS HORRIBLE, but it ran. Probably nothing related to my changes.)

### DIFF
--- a/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/accountprovider/AccountProvider.kt
+++ b/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/accountprovider/AccountProvider.kt
@@ -7,11 +7,15 @@
 
 package io.element.android.features.login.impl.accountprovider
 
+import io.element.android.appconfig.AuthenticationConfig
+
 data class AccountProvider(
     val url: String,
     val title: String = url.removePrefix("https://").removePrefix("http://"),
-    val subtitle: String? = null,
+    val descriptionResourceId: Int? = null,
     val isPublic: Boolean = false,
-    val isMatrixOrg: Boolean = false,
-    val isValid: Boolean = false,
-)
+) {
+    fun isMatrixOrg(): Boolean {
+        return url == AuthenticationConfig.MATRIX_ORG_URL
+    }
+}
--- a/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/accountprovider/AccountProviderDataSource.kt
+++ b/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/accountprovider/AccountProviderDataSource.kt
@@ -9,6 +9,7 @@ package io.element.android.features.login.impl.accountprovider
 
 import io.element.android.appconfig.AuthenticationConfig
 import io.element.android.features.enterprise.api.EnterpriseService
+import io.element.android.features.login.impl.R
 import io.element.android.libraries.di.AppScope
 import io.element.android.libraries.di.SingleIn
 import kotlinx.coroutines.flow.MutableStateFlow
@@ -18,20 +19,32 @@ import javax.inject.Inject
 
 @SingleIn(AppScope::class)
 class AccountProviderDataSource @Inject constructor(
-    enterpriseService: EnterpriseService,
+    enterpriseService: EnterpriseService? = null,
 ) {
-    private val defaultAccountProvider = (enterpriseService.defaultHomeserver() ?: AuthenticationConfig.MATRIX_ORG_URL).let { url ->
+    val matrixOrgAccountProvider = AccountProvider(
+        url = AuthenticationConfig.MATRIX_ORG_URL,
+        descriptionResourceId = R.string.screen_change_account_provider_matrix_org_subtitle,
+        isPublic = true,
+    )
+    // add more hard coded AccountProvider here if you need them
+
+    private val defaultAccountProvider = if (enterpriseService?.defaultHomeserver() != null &&
+        enterpriseService.defaultHomeserver() != AuthenticationConfig.MATRIX_ORG_URL) {
         AccountProvider(
-            url = url,
-            subtitle = null,
-            isPublic = url == AuthenticationConfig.MATRIX_ORG_URL,
-            isMatrixOrg = url == AuthenticationConfig.MATRIX_ORG_URL,
+            enterpriseService.defaultHomeserver()!!,
         )
+    } else {
+        matrixOrgAccountProvider
     }
 
     private val accountProvider: MutableStateFlow<AccountProvider> = MutableStateFlow(
         defaultAccountProvider
     )
+
+    val accountProvidersList = listOf(
+        defaultAccountProvider,
+        matrixOrgAccountProvider
+    ).distinct()
 
     val flow: StateFlow<AccountProvider> = accountProvider.asStateFlow()
 
--- a/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/accountprovider/AccountProviderProvider.kt
+++ b/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/accountprovider/AccountProviderProvider.kt
@@ -9,22 +9,17 @@ package io.element.android.features.login.impl.accountprovider
 
 import androidx.compose.ui.tooling.preview.PreviewParameterProvider
 import io.element.android.appconfig.AuthenticationConfig
+import io.element.android.features.login.impl.R
 
 open class AccountProviderProvider : PreviewParameterProvider<AccountProvider> {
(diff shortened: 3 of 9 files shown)


lex-neufeld:
Manually fixed repository issues with the last PR. Need to re-run tests.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/accountprovider/AccountProvider.kt
```diff
@@ -7,11 +7,15 @@
 
 package io.element.android.features.login.impl.accountprovider
 
+import io.element.android.appconfig.AuthenticationConfig
+
 data class AccountProvider(
     val url: String,
     val title: String = url.removePrefix("https://").removePrefix("http://"),
-    val subtitle: String? = null,
+    val descriptionResourceId: Int? = null,
```
bmarty, line 15: Better to keep using a `String` here, the description could come from the homeserver itself, or from an internal configuration.
The rest of the change in this data class is OK though.

`isValid` was used at the time where we were checking for the availability of a sliding sync proxy. This is not necessary anymore, and some cleanup has been missed.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/accountprovider/AccountProviderDataSource.kt
```diff
@@ -18,21 +19,33 @@ import javax.inject.Inject
 
 @SingleIn(AppScope::class)
 class AccountProviderDataSource @Inject constructor(
-    enterpriseService: EnterpriseService,
+    enterpriseService: EnterpriseService? = null,
```
bmarty, line 22: Better to avoid using `null` in injected constructor parameters. This is not a common practice, even if it seems to compile/work fine.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/accountprovider/AccountProviderDataSource.kt
```diff
@@ -18,21 +19,33 @@ import javax.inject.Inject
...
+    // add more hard coded AccountProvider here if you need them
+
+    private val defaultAccountProvider = if (enterpriseService?.defaultHomeserver() != null &&
+        enterpriseService.defaultHomeserver() != AuthenticationConfig.MATRIX_ORG_URL) {
         AccountProvider(
-            url = url,
-            subtitle = null,
-            isPublic = url == AuthenticationConfig.MATRIX_ORG_URL,
-            isMatrixOrg = url == AuthenticationConfig.MATRIX_ORG_URL,
+            enterpriseService.defaultHomeserver()!!,
```
bmarty, line 34: We try to avoid using `!!` in the codebase. You could use `enterpriseService.defaultHomeserver()?.let { ... }` instead.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/accountprovider/AccountProviderDataSource.kt
```diff
@@ -18,21 +19,33 @@ import javax.inject.Inject
...
     }
 
     private val accountProvider: MutableStateFlow<AccountProvider> = MutableStateFlow(
         defaultAccountProvider
     )
 
+    val accountProvidersList = listOf(
+        defaultAccountProvider,
+        matrixOrgAccountProvider
+    ).distinct()
```
bmarty, line 47: When the `defaultAccountProvider` is not `matrixOrgAccountProvider`, we do not want to propose matrix.org to the user.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/accountprovider/AccountProviderProvider.kt
```diff
@@ -9,22 +9,17 @@ package io.element.android.features.login.impl.accountprovider
...
-            // Add other state here
+            defaultAccountProvider
+            defaultAccountProvider.copy(descriptionResourceId = null),
+            longAccountProvider,
+            longAccountProvider.copy(title = "custom title"),
+            defaultAccountProvider.copy(descriptionResourceId = null, title = "Other", isPublic = false),
         )
 }
-
-fun anAccountProvider() = AccountProvider(
```
bmarty, line 24: This code should be keep, but could be rework to follow the current codebase coding pattern.

- `AccountProvider` constructor should not have default values
- `anAccountProvider` function should have all the parameters to build an `AccountProvider` instance and those parameter can have default values.

bmarty (commented):
Thanks for the draft PR! I have made some first remarks.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/accountprovider/AccountProvider.kt
lex-neufeld: "the description could come from the homeserver itself" is this currently implemented? And how is localization then handled?

Can we have both? If an AccountProvider has a descriptionResourceID: Int, that can be used that to set description: String . AccountProvider objects are in memory for such a short time, that I think localization would be effectively identical as this commit. If not using a resourceID, we can put whatever the homeserver wants in the description and leave descriptionResourceId = null. Then the Views always use the AccountProvider.description without any additional logic and never use AccountProvider.descriptionResourceID. This would actually make the Views even cleaner than what is currently in this PR and should support all use cases. AccountProvider would be slightly bigger and more complicated, but should still be understandable.

bmarty:
Closing as stale PR.

PR #4602, opened 2025-04-16
Fix audio output selection for Element Call

bmarty:
## Content

First commit ensure that the highest priority device is selected when Element Call starts.
Second commit ensure that the audio selection algorithm is run again when new devices are connected or device are disconnected.

## Motivation and context

Improve Element Call experience.

## Tests

- Start a call using Element Call
- Audio should play on the phone speaker, and not the earpiece
- Connect a Bluetooth device
- Audio is now played on the Bluetooth device
- Disconnect the Bluetooth device
- Audio is now played on the phone speaker

If the Bluetooth device is already connected when the call starts, the audio should be routed to the Bluetooth device.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenView.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenView.kt
@@ -8,6 +8,7 @@
 package io.element.android.features.call.impl.ui
 
 import android.annotation.SuppressLint
+import android.content.Context
 import android.media.AudioDeviceCallback
 import android.media.AudioDeviceInfo
 import android.media.AudioManager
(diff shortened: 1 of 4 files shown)


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenView.kt
```diff
@@ -151,14 +157,28 @@ private fun CallWebView(
...
 
-                audioManager?.enableExternalAudioDevice()
+                        override fun onAudioDevicesRemoved(removedDevices: Array<out AudioDeviceInfo>?) {
+                            Timber.d("Audio devices removed")
+                            audioManager.enableExternalAudioDevice()
+                        }
+                    }
+
+                    audioManager.registerAudioDeviceCallback(audioDeviceCallback, null)
+                }
```
jmartinesp, line 181: Could we move this to an external function? Also, maybe we should check if the selected audio device was actually set after selecting it and retry if it's not. I'm not sure if there's some error/success callback for `setCommunicationDevice`.

jmartinesp (approved):
LGTM, just a suggestion.

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenView.kt
bmarty: `setCommunicationDevice` returns a boolean and can also throw Exception, so yes, maybe there is some improvement to do here. Let me check.

Issue #3437, opened 2024-09-10
Pinned Messages QA


kongo09:
@daniellekirkwood is moving the actions to other tickets so we can break this down into tasks for the team. Updated each section with the associated action. Questions in comments :) 

### Steps to reproduce

I spent some time with pinned messages to QA them.

### Outcome

- [x] **Developer Options Wording** Not relevant, this flag will disappear

- Currently: Pinned Events
- Should this turn into a user setting, the term _Events_ is more associated with a meeting event than with anything that can be pinned.
- Recommendation: Don't turn this into a user setting but make it default. If we use it as a setting, call it _Pinned Messages_

- [x] **Power Level** - feedback logged, https://github.com/element-hq/element-internal/issues/618

- Currently: requires Admin level
- this feels unfamiliar, as in WhatsApp anyone can pin a message in a group, not only an admin
- Recommendation: Would it make sense that the Admin setting would be something along the lines of "Only Admins can pin messages" with the default of "off"?

- [ ] **Visibility** - needs retesting, can't reproduce

- Currently: on EAX 0.5.3-nightly after getting admin rights and pinning the first message, the message is not visible in the room.
- Expected: message should be visible right away
- Workaround: leave the room and come back. Pinned message shows up then

- [x] **Show All** - verified in comments, it now works

- Currently: on EAX 0.5.3-nightl the "Show All" button doesn't work and does nothing
- Expected: list is shown
- Workaround: go to the room details and enter list there
- Update: now seems to work? Maybe something changed? Maybe flaky?
- Currently: the text for "show all" is taking a lot of attention and space in the pinned messages bar, which makes the pinned messages preview even less useful
- Recommendation: can we replace this text with a simple icon (like an eye or a zoom out or whatever)?

- [x] **Pinned Messages List Senders** - verified in comments, it works as expected, feedback logged https://github.com/element-hq/element-internal/issues/618

- Currently: Messages don't show sender
- Expected: Messages are displayed pretty much the same way they show in the timeline with a sender

- [x] **Pinned Messages List Action** Feedback logged https://github.com/element-hq/element-internal/issues/618

- Currently: Single tap on pinned messages does nothing
- Expected: View in timeline
- Workaround: Long tap to get context menu and tap "view in timeline" there

- [ ] **View in timeline Action** - needs further discussion (in Element)

- Currently: App jumps to timeline and scrolls around but not to the actual pinned message
- Expected: The pinned message is seen as the top message in the timeline
- Workaround: none

[...]

PR #4606, opened 2025-04-18
Ensure that pinning an event makes the pinned messages banner appear

bmarty:
## Content

When a user pin or unpin a message, make sure the the pinned messages banner is displayed.

Note that the banner will also be made visible when another user update the list of pinned event, but I do not think this is a problem, as it should not occur a lot. The banner is always visible, except if the user has scrolled to the top.

## Motivation and context

Part of #3437

## Tests

- Navigate to a room where the user is allowed to pin event
- scroll to the top a bit
- pin an event
- observe that the banner is visible (new)
- scroll to the top a bit again
- the banner is automatically hidden
- pin another event
- observe that the banner is visible (new)

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


ganfra (approved):
Feels a bit hacky, but why not

Issue #4604, opened 2025-04-17
Element Call Ringing does not stop


fkwp:
### Steps to reproduce

1. The application is running in the background
2. Call ringing is happening due an incoming call in a DM room
3. The caller stops the call OR I pickup the call with another device
4. Ringing is still in progress rather than stopping

#### What did you expect?
Ringing stops as soon as
- the caller stops the call
- or I did pickup the call with my other device

#### What happened instead?
it keeps ringing until a timeout

### Will you send logs?

Yes

### Are you willing to provide a PR?

Yes

manuroe:
It was mw who asked @fkwp to create a new issue (thanks for it!). Code changes we made for in #4389 make new rageshakes unrelated to the old ones.
So, let's continue with this new one. @jmartinesp can you investigate the new problem or the scenario we did not cover last time?

jmartinesp:
I think I found what was going on when the call is not cancelled, we were accidentally stopping the sync service so the room info's `hasRoomCall` value was never updated. I'll create a PR with this.

About ringing calls not notifying while the device is locked, I found this after some testing:

- I started a call with EW.
- The `m.call.notify` event was received by the locked device 2s later, still in the 10s threshold to notify with a ringing notification.
- The client asks the SDK to give it the event for the notification.
- 12s elapse, with the `io.element.call.encryption_keys` event somehow blocking the event fetching.
- After all this time is elapsed, the `m.call.notify` event is finally processed, but > 10s have elapsed so it no longer rings.

```
### This is the m.call.encryption_keys event ###

12:01:56.044 org.matrix.rust.sdk     io.element.android.x.debug           W  elementx: [Push/VectorFirebaseMessagingService] New Firebase message. Priority: 1/1 | VectorFirebaseMessagingService.kt:41
12:01:58.569 org.matrix.rust.sdk     io.element.android.x.debug           D  elementx: [Push/PushHandler] ## handling pushData: !doHASiVaRpfQUfZYXq:matrix.org/$59bT0kU-tQbmtu7lVP9dYxWGt9Et19lLCqGcLVYuMik | DefaultPushHandler.kt:62

### This is the m.call.notify event ###

12:01:58.637 org.matrix.rust.sdk     io.element.android.x.debug           W  elementx: [Push/VectorFirebaseMessagingService] New Firebase message. Priority: 1/1 | VectorFirebaseMessagingService.kt:41
[...]

PR #4613, opened 2025-04-21
Fix ringing calls not stopping when the other user cancels the call

jmartinesp:
## Content

- Makes sure the sync service is running when screening an incoming ringing call.
- Add extra logs to help debugging ringing call issues.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/4604#event-17320922870.

## Tests

With notifications enabled:

- In a DM, start a call from another device.
- The current device should receive a ringing call notification.
- Now cancel the call from the initiator device.
- The ringing should stop shortly after in the current device.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
```diff
@@ -78,15 +78,21 @@ class SyncOrchestrator @AssistedInject constructor(
     internal fun observeStates() = coroutineScope.launch {
         Timber.tag(tag).d("start observing the app and network state")
 
+        val appForegroundState = combine(
```
ganfra, line 81: rename as `isAppActive `? appForegroundState doesn't sound like a bool

ganfra (approved):

PR #4623, opened 2025-04-22
OIDC configuration

bmarty:
## Content

Let OIDC configuration be read from BuildConfig file.
Also improve the error dialog when there is an error in the configuration (see screenshot below). Since this is generally build issue, there is no need to translate such error message.

## Motivation and context

Make Oidc configuration configurable at build time.

## Screenshots / GIFs

|Before|After|
|-|-|
|OidcError_before|OidcError_after|

## Tests

- There should be no impact at all on the Element X Android application, but reviewer should check that login/registration still works as expected on matrix.org or any other homeserver which supports MAS.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/auth/OidcConfig.kt
+++ b/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/auth/OidcConfig.kt
@@ -7,24 +7,26 @@
 
 package io.element.android.libraries.matrix.api.auth
 
+import io.element.android.libraries.matrix.api.BuildConfig
+
 object OidcConfig {
-    const val CLIENT_URI = "https://element.io"
+    const val CLIENT_URI = BuildConfig.CLIENT_URI
 
     // Notes:
     // 1. the scheme must match the value declared in the AndroidManifest.xml
     // 2. the scheme must be the reverse of the host of CLIENT_URI
-    const val REDIRECT_URI = "io.element:/callback"
+    const val REDIRECT_URI = BuildConfig.REDIRECT_URI
 
     // Note: host must match with the host of CLIENT_URI
-    const val LOGO_URI = "https://element.io/mobile-icon.png"
+    const val LOGO_URI = BuildConfig.LOGO_URI
 
     // Note: host must match with the host of CLIENT_URI
-    const val TOS_URI = "https://element.io/acceptable-use-policy-terms"
+    const val TOS_URI = BuildConfig.TOS_URI
 
     // Note: host must match with the host of CLIENT_URI
-    const val POLICY_URI = "https://element.io/privacy"
+    const val POLICY_URI = BuildConfig.POLICY_URI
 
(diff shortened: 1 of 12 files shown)


libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/auth/OidcConfig.kt
```diff
@@ -7,6 +7,27 @@
...
+    // Note: host must match with the host of CLIENT_URI
+    const val TOS_URI = BuildConfig.TOS_URI
+
+    // Note: host must match with the host of CLIENT_URI
+    const val POLICY_URI = BuildConfig.POLICY_URI
+
+    // Some homeservers/auth issuers don't support dynamic client registration, and have to be registered manually
+    val STATIC_REGISTRATIONS = mapOf(
+        "https://id.thirdroom.io/realms/thirdroom" to "elementx",
+    )
```
bmarty, line 32: @pixlwave this map is not "externally" configurable, do you know if we should do something about it?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/AuthenticationException.kt
```diff
@@ -24,6 +25,13 @@ fun Throwable.mapAuthenticationException(): AuthenticationException {
             is ClientBuildException.WellKnownLookupFailed -> AuthenticationException.Generic(message)
             is ClientBuildException.EventCache -> AuthenticationException.Generic(message)
         }
+        is OidcException -> when (this) {
+            is OidcException.Generic -> AuthenticationException.Oidc(message)
+            is OidcException.CallbackUrlInvalid -> AuthenticationException.Oidc(message)
+            is OidcException.Cancelled -> AuthenticationException.Oidc(message)
+            is OidcException.MetadataInvalid -> AuthenticationException.Oidc(message)
+            is OidcException.NotSupported -> AuthenticationException.Oidc(message)
+        }
```
jmartinesp, line 34: Since we're mapping them to the same type, are we sure the messages are clear enough to track the root cause of the issue?

jmartinesp (approved):
LGTM, thanks!

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/AuthenticationException.kt
bmarty: Good point, for what I have seen, it seems to be clear enough. We may want to improve the mapping if it's not the case.

Previously the homeserver reachability was designed as the cause of the failure, so I guess this is already much better.

pixlwave:
@wrenix It sounds like your MAS is out of date as this requirement was [removed in September](https://github.com/element-hq/matrix-authentication-service/pull/3202). Updating your server should fix this issue.

PR #4626, opened 2025-04-23
Accessibility: improve behavior of list items

bmarty:
## Content

Improve accessibility for ListItem.

One example, when long press on a room and select the "Favorite" item, so in this screen:

image

The current read text is:
> Favorite favorite double tap to activate

so quite wrong and do not give the state of the switch. With this changes, if the switch is on, the read text is:
> on favorite switch, double tap to toggle

and if the switch if off, the read text is:
> off favorite switch, double tap to toggle

and the switch itself is not focusable, so the next focusable item is the "Settings" item.

Remove callback duplication, that can lead to screen reader focused on the switch itself, which has no added value if the behavior is the same.

## Motivation and context

Improve accessibility.

Replacement for #4047 

## Tests

- Enable screen reader and check the behavior of the List items used in bottom sheet, settings...

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


tests/uitests/src/test/snapshots/images/features.roomdetails.impl.securityandprivacy_SecurityAndPrivacyViewDark_0_en.png
bmarty, line 1: Waiting for a design decision about this change:

image

(asked in https://www.figma.com/design/7TqjqdMBaPpm3IavKsMYu6?node-id=5235-51684&m=dev#1228513902)

bmarty:
This is breaking click on ListItem, needs to push a fix.

features/poll/impl/src/main/kotlin/io/element/android/features/poll/impl/create/CreatePollView.kt
```diff
@@ -178,8 +178,14 @@ fun CreatePollView(
                         supportingContent = { Text(text = stringResource(id = R.string.screen_create_poll_anonymous_desc)) },
                         trailingContent = ListItemContent.Switch(
                             checked = state.pollKind == PollKind.Undisclosed,
-                            onChange = { state.eventSink(CreatePollEvents.SetPollKind(if (it) PollKind.Undisclosed else PollKind.Disclosed)) },
                         ),
+                        onClick = {
+                            state.eventSink(
+                                CreatePollEvents.SetPollKind(
+                                    if (state.pollKind == PollKind.Disclosed) PollKind.Undisclosed else PollKind.Disclosed
```
ganfra, line 185: I think it would be better to let the Presenter handle this, but as it's not new, feel free to ignore.

ganfra (approved):

PR #4651, opened 2025-04-28
Fix wrong member count in join room screen for invitation

bmarty:
## Content

It appears that the value from the room info is not correct. For invitation in DM, I have observe that the room member count can be 0 or 2 (if the user has previously left the room). With this fix, the value is 1, which is OK.

No change for group rooms AFAICT.

## Motivation and context

Fix issue.

## Screenshots / GIFs

No impact

## Tests

- Get an invite and observe the room member count in the room invite screen

image

image

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/joinroom/impl/src/main/kotlin/io/element/android/features/joinroom/impl/JoinRoomPresenter.kt
+++ b/features/joinroom/impl/src/main/kotlin/io/element/android/features/joinroom/impl/JoinRoomPresenter.kt
@@ -121,7 +121,13 @@ class JoinRoomPresenter @AssistedInject constructor(
                         }
                         else -> null to null
                     }
-                    value = roomInfo.get().toContentState(sender, reason)
+                    val preview = matrixClient.getRoomPreview(roomIdOrAlias, serverNames)
+                    val joinedMembersCountOverride = preview.getOrNull()?.previewInfo?.numberOfJoinedMembers
+                    value = roomInfo.get().toContentState(
+                        membershipSender = sender,
+                        joinedMembersCountOverride = joinedMembersCountOverride,
+                        reason = reason,
+                    )
                 }
                 roomDescription.isPresent -> {
                     value = roomDescription.get().toContentState()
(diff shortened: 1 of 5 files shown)


features/joinroom/impl/src/test/kotlin/io/element/android/features/joinroom/impl/JoinRoomViewTest.kt
```diff
@@ -25,7 +25,7 @@ import org.junit.rules.TestRule
 import org.junit.runner.RunWith
 
 @RunWith(AndroidJUnit4::class)
-class JoinBaseRoomViewTest {
```
jmartinesp, line 28: Agh, I missed this accidental rename when I fixed the others. Thanks.

features/joinroom/impl/src/main/kotlin/io/element/android/features/joinroom/impl/JoinRoomPresenter.kt
```diff
@@ -113,15 +113,21 @@ class JoinRoomPresenter @AssistedInject constructor(
...
+                            val membershipDetails = preview.getOrNull()?.membershipDetails()?.getOrNull()
+                            membershipDetails?.senderMember to membershipDetails?.currentUserMember?.membershipChangeReason
                         }
                         CurrentUserMembership.INVITED -> {
                             roomInfo.get().inviter to null
                         }
                         else -> null to null
                     }
-                    value = roomInfo.get().toContentState(sender, reason)
+                    val preview = matrixClient.getRoomPreview(roomIdOrAlias, serverNames)
```
jmartinesp, line 124: If the user is banned we have 2 get room preview calls. Maybe the newly added one could be before this block and reuse it for the banned state?

bmarty: Ah true, good catch. Updated in 10d79ab5d4734f7404e3d506ae5c08c73d4f95d9

Issue #3625, opened 2024-10-08
Element Call: unable to select any sound source other than loud speaker


daedric7:
### Steps to reproduce

1. Start a call
2. Open sound settings
3. Notice several options for microphone (including bluetooth) but non for Speaker

#### What did you expect?

A list of available sound sources

#### What happened instead?

Only a fixed option:

![11695b9c-96e5-4018-8c64-98ec51cc5a15](https://github.com/user-attachments/assets/8931e2e8-149c-4ee6-b50e-077825f4829a)

### Your phone model

Pixel 8a

### Operating system version

Android 14 stock

### Application version and app store

Latest nightly as of this writing

### Homeserver

Synapse 1.116 with slefhosted Element Call

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

davidegirardi:
I noticed that selecting the Bluetooth mic input automatically switches to Bluetooth output automatically.

archerallstars:
In my case, regardless of the mic I choose, the sound won't come out of my bluetooth headset.

Android 11 stock. Lenovo Tab P11.

jmartinesp:
Sorry for the very late reply, but I investigated this today and it's a limitation of the Android OS, and it doesn't seem like it'll get fixed any time soon as it was marked by Google as 'wont't fix': https://issues.chromium.org/issues/41276355.

Basically, the Android OS won't let the webview select a different audio output device, it'll just reuse whatever the OS decided was the right device to use for the communication audio mode. There is ongoing work to improve which device is selected and make it default to external devices if possible, so choosing a device still won't be possible but it's as much as we can do.

bmarty:
We can still offer the possibility for the user to select an audio output device, but this will have to be done on the native code, not in the webview. But since we have a bridge between the two world, there is probably a solution for this?

bmarty:
CC @fkwp to get Element Call POV on my last comment.

fkwp:
> We can still offer the possibility for the user to select an audio output device, but this will have to be done on the native code, not in the webview. But since we have a bridge between the two worlds, there is probably a solution for this?

that is also my current understanding of "fixing" the issue. We tackle it with our next initiative -- which starts soonTM

PR #4663, opened 2025-04-30
Element Call: Add audio output selector handled by Android

jmartinesp:
## Content

- Add code and callbacks to pass the list of available audio devices and the selected on to Element Call and retrieve the 'on device selected' action from its UI.
- When using an earpiece device, enable the proximity sensor for screen on/off.
- Simplify the window flags in `ElementCallActivity`.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/3625.

## Screenshots / GIFs

https://github.com/user-attachments/assets/7ce4a292-5387-44d0-9fa9-615fdcba3e11

## Tests

- Make sure the call works between EW and EXA or maybe EXA and EXI.
- Select one of the audio devices. Both the audio input and output should switch to it, i.e. selecting 'earpiece' should use the earpiece on the top of your screen and the front microphone, speaker should use the loudspeaker and the camera microphone, and a BT device should be used for both input and output.
- Connect and disconnect a BT headset if you have one to check it's selected by default on connection and gets removed on disconnection.

It would be specially interesting to test this on Android 11, since I don't own a device with this OS version and the APIs are very different. Also testing it on Android 12 would be good, since there is a workaround needed for that version related to the AudioManager's comms mode.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
```diff
@@ -132,6 +147,26 @@ class ElementCallActivity :
         }
     }
 
+    override fun onStart() {
+        super.onStart()
+
+        if (proximitySensorWakeLock?.isHeld == false) {
+            val proximitySensorEnabled = proximitySensorWakeLock?.acquire()
```
ElementBot, line 154: - :warning: Provide a timeout when requesting a wakelock with `PowerManager.Wakelock.acquire(long timeout)`. This will ensure the OS will cleanup any wakelocks that last longer than you intend, and will save your user's battery.

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
```diff
@@ -125,6 +140,26 @@ class ElementCallActivity :
...
```
ElementBot, line 147: - :warning: Provide a timeout when requesting a wakelock with `PowerManager.Wakelock.acquire(long timeout)`. This will ensure the OS will cleanup any wakelocks that last longer than you intend, and will save your user's battery.

jmartinesp:
After https://github.com/element-hq/element-x-android/pull/4743 this should now be ready to review.

bmarty:
I have tested this and I am loosing the audio output when selecting `Default (Built-in speaker)`.

Actually it was due to the fact that this is not the same volume mapping. For earpiece, the `In-call volume` is used, for `Default`, this is the `Media` volume. I am not sure if we can improve this?

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewAudioManager.kt
```diff
@@ -0,0 +1,413 @@
...
+    val typePart = when (type) {
+        AudioDeviceInfo.TYPE_BLUETOOTH_SCO -> "Bluetooth"
+        AudioDeviceInfo.TYPE_USB_ACCESSORY -> "USB accessory"
+        AudioDeviceInfo.TYPE_USB_DEVICE -> "USB device"
+        AudioDeviceInfo.TYPE_USB_HEADSET -> "USB headset"
+        AudioDeviceInfo.TYPE_WIRED_HEADSET -> "Wired headset"
+        AudioDeviceInfo.TYPE_WIRED_HEADPHONES -> "Wired headphones"
+        AudioDeviceInfo.TYPE_BUILTIN_SPEAKER -> "Built-in speaker"
+        AudioDeviceInfo.TYPE_BUILTIN_EARPIECE -> "Built-in earpiece"
+        else -> "Unknown:"
```
bmarty:
```suggestion
        else -> "Unknown"
```

(I am not sure)

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewAudioManager.kt
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    /**
+     * Assigns the callback in the WebView to be called when the user selects an audio device.
+     *
+     * It should be called with some delay after [registerWebViewDeviceSelectedCallback].
+     */
+    private fun setWebViewOnAudioDeviceSelectedCallback() {
+        Timber.d("Adding callback in controls.onOutputDeviceSelect")
+        webView.evaluateJavascript("controls.onOutputDeviceSelect = (id) => { onAudioDeviceSelectedCallback.setOutputDevice(id); };", null)
```
bmarty: Maybe add the ref `https://github.com/element-hq/element-call/blob/livekit/docs/controls.md#audio-devices` to the Kdoc or at the top of the class?

jmartinesp:
> Actually it was due to the fact that this is not the same volume mapping. For earpiece, the `In-call volume` is used, for `Default`, this is the `Media` volume. I am not sure if we can improve this?

Oh, that's tricky. I'm not sure if we can make they use the same volume rocker or have some other way to make this transition seamless..

jmartinesp:
> But it seems that the audio input (in EX) was not working. I could not hear any voice coming to Element Web (sound alert was working so this is not a volume/mute problem on the browser)

Did you have video output or was that also failing? I've had one call this morning where the encryption keys exchange seemed to fail and that resulted in the 2 users seeing each other as present in the call but no video or audio getting through to the other.

bmarty:
I have tested it again and this is the status for a call with EW and EX and EW does not render incoming video nor audio. The other side is working (EX render video and play audio)

jmartinesp:
> I have tested it again and this is the status for a call with EW and EX and EW does not render incoming video nor audio. The other side is working (EX render video and play audio)

Could you check you don't have set a custom Element Call URL? On EW you can use`/devtools` for that.

> Actually it was due to the fact that this is not the same volume mapping. For earpiece, the In-call volume is used, for Default, this is the Media volume. I am not sure if we can improve this?

So I have good news and bad news after checking other apps and how they handle this issue: 
- The good news is I've found a way to make the audio use always the 'in call' stream and it would simplify the code.
- The bad news is it's based on always using the deprecated APIs (the < API 12 we already have in place in this PR), which I'd rather not use and could be removed in the not-so-far future.

I'll check if there's any alternative to this, but I'm not really hopeful...

bmarty:
I have not set any custom url on EW. For the volume I think it is acceptable if it's an OS limitation... But can be misleading for user (like me) thinking that the audio is not working when switching the output. There is maybe a way to read the volume level and warn the user that the volume is probably too low (I think I have seen other apps doing that, a long time ago, I do not remember which one).

jmartinesp:
@bmarty could you retry with the latest changes?

jmartinesp:
[The latest changes](https://github.com/element-hq/element-x-android/pull/4663/commits/cddcf61bb840e894d12d05f4d334f317f66ea657) should fix the issue with the audio for the speaker initially not following the right volume rocker, but now we need https://github.com/element-hq/element-call/pull/3309.

jmartinesp:
> [The latest changes](https://github.com/element-hq/element-x-android/pull/4663/commits/cddcf61bb840e894d12d05f4d334f317f66ea657) should fix the issue with the audio for the speaker initially not following the right volume rocker, but now we need [element-hq/element-call#3309](https://github.com/element-hq/element-call/pull/3309).

Rebased after upgrading the Element Call version in https://github.com/element-hq/element-x-android/pull/4832, which contains the changes needed to fix the audio for the device selected by default.

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewAudioManager.kt
```diff
@@ -0,0 +1,446 @@
...
+     */
+    private fun registerWebViewDeviceSelectedCallback() {
+        val webViewAudioDeviceSelectedCallback = AndroidWebViewAudioBridge(
+            onAudioDeviceSelected = { selectedDeviceId ->
+                Timber.d("Audio device selected in webview, id: $selectedDeviceId")
+                previousSelectedDevice = listAudioDevices().find { it.id.toString() == selectedDeviceId }
+                audioManager.selectAudioDevice(selectedDeviceId)
+            },
+            onAudioPlaybackStarted = {
+                MainScope().launch {
```
bmarty: Provide a CoroutineScope in the constructor instead of using the MainScope?

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewAudioManager.kt
```diff
@@ -0,0 +1,446 @@
...
+        }
+
+        setWebViewAndroidNativeBridge()
+    }
+
+    /**
+     * Call this method when the call stops to disable in-call audio mode.
+     *
+     * It's the counterpart of [onCallStarted], and should be called as a pair with it once the call has ended.
+     */
```
bmarty, line 180: Kudos for all the comments, it helps to understand the code a lot!

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewAudioManager.kt
jmartinesp: I think we'll still need to make sure it's called on `Dispatchers.Main`, but good idea. This way everything can stop as soon as the UI is destroyed.

PR #4664, opened 2025-04-30
Improve the callback uri format and customization.

bmarty:
## Content

The PR aims to improve how the redirection is handled during a OIDC authentication, to filter between different installed applications, by providing a different scheme for all the different version (debug/nightly/release)

Current version:
image

New version:

image

This is for the debug version.

For the prod version, the text will be: "Element X at io.element.android:/ wants to access your account."
For the nightly version, the text will be: "Element X nightly at io.element.android.nightly:/ wants to access your account."

This will have the effect to avoid this screen to be displayed when the user clicks on "Continue":
image

Note:
- this is probably solving a problem only for developers and advanced users.
- this will help Mas to distinguish between the client. Element desktop uses io.element.desktop for instance (see https://github.com/element-hq/element-desktop/pull/2285)
- it should also solve the problem if both `Element Pro` and `Element X` are installed, since we are using the scheme `io.element` for Element Pro. The scheme `io.element.debug` will be for Element Pro debug, and the scheme `io.element.nightly` will be for Element Pro nightly.

## Motivation and context

Better UI from a user POV.

## Tests

- Login to matrix.org or any OIDC homeserver and observe that the user story is more straight forward.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -106,14 +106,25 @@ android {
     logger.warnInBox("Building ${defaultConfig.applicationId} ($baseAppName)")
 
     buildTypes {
+        val oidcRedirectSchemeBase = BuildTimeConfig.METADATA_HOST_REVERSED ?: "io.element.android"
         getByName("debug") {
             resValue("string", "app_name", "$baseAppName dbg")
+            resValue(
+                "string",
+                "login_redirect_scheme",
+                "$oidcRedirectSchemeBase.debug",
+            )
             applicationIdSuffix = ".debug"
             signingConfig = signingConfigs.getByName("debug")
         }
 
         getByName("release") {
             resValue("string", "app_name", baseAppName)
+            resValue(
+                "string",
+                "login_redirect_scheme",
+                oidcRedirectSchemeBase,
+            )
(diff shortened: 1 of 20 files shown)


bmarty:
@sandhose told me that the spec says you should not have a hostname in native callbacks.

So we may use `io.element.[identifier]:/callback` instead. I am updating the PR.

bmarty:
> > So we may use `io.element.[identifier]:/callback` instead. I am updating the PR.
> 
> Will this essentially match `[applicationID]:/callback`? If it does, then maybe using `applicationID` directly would also help with the Element Pro/Element X issue (but remembering to still honour the enterprise override if it has been set).

Good point. I guess I can use `io.element.application` for Element X, and so Element Pro will use `io.element`.
The problem with the applicationId is that it is `io.element.android.x` and I believe that at some point the x should vanish, at least from a marketing POV, the `applicationId` cannot be changed.

bmarty:
Update:
- Element X Android will use `io.element.android` as a base for Oidc Url redirect scheme.
- Element Pro Android will use `io.element.android.pro` as a base for Oidc Url redirect scheme.

PR #4677, opened 2025-05-05
Sync Strings

ElementBot:


bmarty:
(Just for the record I had to re-run the action because of a Localazy server error https://github.com/element-hq/element-x-android/actions/runs/14826559288)

jmartinesp (approved):
Looks ok after a cursory review.

PR #4679, opened 2025-05-05
Fix read receipt behavior when the timeline is opened.

### LINKED ISSUE #4662: "Seen" status only populate when target user exit and re-enter room not if they are already in the room
### Steps to reproduce

send a message and wait for the message to be seen.

if the target user is already in the room (i.e. app open and in the room and can see the messages coming through) the seen status is not displayed on the origin device.

But is the target user go out of the room and re-enter the room the seen status in the origin device populate properly.

#### What did you expect?
the seen status to be updated in the origin device if the target device is on and the user is already in the room

#### What happened instead?
seen status not updating even through the target user saw the message

### Your phone model

Samsung Galaxy

### Operating system version

_No response_

25.04.03

### Homeserver

1.128.0

### Will you send logs?

No

No

bmarty:
## Content

Ensure that ReadReceipt is sent when new messages are coming in and user is on the timeline.

## Motivation and context

Closes #4662 

## Tests

- Alice opens a room R she shares with Bob
- Bob open the same room
- Bob sees that Alice RR is at the bottom of the timeline
- Bob sends a new message
- Alice views the message from Bob and does not do anything else particular
- Bob sees Alice's RR moving down as expected. This PR fixes this, previously Alice RR did not move.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
@@ -131,7 +131,7 @@ class TimelinePresenter @AssistedInject constructor(
                         if (event.firstIndex == 0) {
                             newEventState.value = NewEventState.None
                         }
-                        println("## sendReadReceiptIfNeeded firstVisibleIndex: ${event.firstIndex}")
+                        Timber.d("## sendReadReceiptIfNeeded firstVisibleIndex: ${event.firstIndex}")
                         appScope.sendReadReceiptIfNeeded(
                             firstVisibleIndex = event.firstIndex,
                             timelineItems = timelineItems,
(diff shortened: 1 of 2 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
```diff
@@ -278,7 +279,10 @@ class TimelinePresenter @AssistedInject constructor(
         if (newEventState.value == NewEventState.FromMe) {
             return@withContext
         }
-        val newMostRecentItem = timelineItems.firstOrNull()
+        val newMostRecentItem = timelineItems.firstOrNull {
+            // Ignore typing item
+            (it as? TimelineItem.Virtual)?.model !is TimelineItemTypingNotificationModel
+        }
```
jmartinesp, line 285: Oh, good catch 👌 .

jmartinesp (approved):

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
bmarty: Yes, I could also have dropped the first item, but I guess it's more accurate like this.

bmarty:
@jmartinesp I have updated the fix to fix regression on tests.

I have tested again and the fix is still working (there is no tests covering the use case). Can you have a second look please? Thanks!

jmartinesp (approved):

PR #4683, opened 2025-05-06
Disable mutliple click (parallel or serial) on a room

### LINKED ISSUE #4619: It's possible to open a room view twice by tapping it super quickly
### Steps to reproduce

From the room list, tap the same room name twice super quickly.

#### What did you expect?

The room view to open only once.

#### What happened instead?

The room view has opened twice: if I press return/back, then I get back to the same room view, instead of the room list.

### Your phone model

Motorola Razr

### Operating system version

_No response_

ElementX beta from the app store

### Homeserver

_No response_

### Will you send logs?

No

No

bmarty:
## Content

Add minimum delay between two click handling on the room list.
Protect:
- opening the setting
- opening a room
- opening the create room screen

## Motivation and context

Fixes #4619

## Tests

- Open the room list
- Tap on several rooms simultaneously
- Observe that only one room is opened by going back to the room list
- Quickly tap on a room
- Observe that only one room is opened by going back to the room list
- Tap on a room and on the user avatar icon
- Observe that the settings screen is opened or the room is opened but not both by going back to the room list
- Tap on a room and on the floating action button to create a room
- Observe that the settings to create a room is opened or the room is opened but not both by going back to the room list

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/throttler/FirstThrottler.kt
+++ b/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/throttler/FirstThrottler.kt
@@ -38,4 +38,8 @@ class FirstThrottler(private val minimumInterval: Long = 800) {
         // Too early
         return CanHandleResult.No(minimumInterval - delaySinceLast)
     }
+
+    fun canHandleAsBoolean(): Boolean {
+        return canHandle() == CanHandleResult.Yes
+    }
 }
--- a/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/throttler/FirstThrottler.kt
+++ b/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/throttler/FirstThrottler.kt
@@ -6,40 +6,29 @@
  */
 package io.element.android.libraries.androidutils.throttler
 
-import android.os.SystemClock
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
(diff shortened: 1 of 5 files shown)


libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/throttler/FirstThrottler.kt
```diff
@@ -38,4 +38,8 @@ class FirstThrottler(private val minimumInterval: Long = 800) {
         // Too early
         return CanHandleResult.No(minimumInterval - delaySinceLast)
     }
+
+    fun canHandleAsBoolean(): Boolean {
```
ganfra, line 42: I'd call this `canHandle` and the other one could be `getResult`?

ganfra (commented):
Maybe this should be handled directly in the compose layer? Would be simpler to handle this with a custom `Modifier` so it can easily be reused (and tested?)

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/throttler/FirstThrottler.kt
bmarty: I have rewritten this class.

ganfra (approved):
Would be handy to create a reusable `Modifier` so we can use this everywhere more easily, but we can do later I guess.

bmarty:
> Would be handy to create a reusable `Modifier` so we can use this everywhere more easily, but we can do later I guess.

I have thought about it, but I am not sure how to implement this, as the FirstThrottle is shared between several click handlers.

PR #4707, opened 2025-05-12
Improve audio focus management

### LINKED ISSUE #4617: Voice messages don't integrate nicely with the Android media player

When I'm playing a music track in e.g. Spotify/Deezer, and start listening to an audio message in ElementX, I'd like the music to stop. Ideally, it'd stop only for the duration of the audio message, and resume afterwards; but it's OK if it just stops it.

Right now, it's not doing that, and it will play on top of the music track, which makes it super hard to listen to.

I can't hear to multiple tracks / audios at the same time, and I assume most people can't either.

Bribing ElementX Android developers into implementing it, by offering them pain au chocolat, or something along those lines.

_No response_

### Additional context

_No response_

No

bmarty:
## Content

- When the user starts playing a voice message / an audio file / a video: request audio focus to the system so that any other application currently playing audio will pause. This was done for Element Call but not for those use cases. First commit extract the code from Element Call to reuse it at other places
- When Element X is currently playing voice message / audio or video file and the user starts an audio stream from another app (ex: resume Spotify playback from the notification drawer), audio focus is lost and in this case, the application pauses the current playback. For Element Call, nothing is done when the audio focus is lost.

## Motivation and context

Closes #4617 and better integration with the system.

## Screenshots / GIFs

No change in the UI

## Tests

- This is a challenge: try to have an audio output from Element X in the same time than audio output from every other (decent) applications. It should not be possible anymore.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/audio/impl/src/main/kotlin/io/element/android/libraries/audio/impl/DefaultAudioFocus.kt
+++ b/libraries/audio/impl/src/main/kotlin/io/element/android/libraries/audio/impl/DefaultAudioFocus.kt
@@ -33,7 +33,7 @@ class DefaultAudioFocus @Inject constructor(
     override fun requestAudioFocus(mode: AudioFocusRequester) {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
             val audioAttributes = AudioAttributes.Builder()
-                .setUsage(AudioAttributes.USAGE_VOICE_COMMUNICATION)
+                .setUsage(mode.toAudioUsage())
                 .build()
             val request = AudioFocusRequest.Builder(AudioManager.AUDIOFOCUS_GAIN)
                 .setAudioAttributes(audioAttributes)
@@ -44,7 +44,7 @@ class DefaultAudioFocus @Inject constructor(
             val listener = AudioManager.OnAudioFocusChangeListener { }
             audioManager.requestAudioFocus(
                 listener,
-                AudioManager.STREAM_VOICE_CALL,
+                mode.toAudioStream(),
                 AudioManager.AUDIOFOCUS_GAIN_TRANSIENT_EXCLUSIVE,
             )
             audioFocusChangeListener = listener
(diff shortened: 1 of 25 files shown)


libraries/audio/impl/src/main/kotlin/io/element/android/libraries/audio/impl/DefaultAudioFocus.kt
```diff
@@ -30,18 +30,34 @@ class DefaultAudioFocus @Inject constructor(
...
+        onFocusLost: () -> Unit,
+    ) {
+        val listener = AudioManager.OnAudioFocusChangeListener {
+            when (it) {
+                AudioManager.AUDIOFOCUS_GAIN -> {
+                    // Do nothing
+                }
+                AudioManager.AUDIOFOCUS_LOSS,
+                AudioManager.AUDIOFOCUS_LOSS_TRANSIENT,
+                AudioManager.AUDIOFOCUS_LOSS_TRANSIENT_CAN_DUCK -> {
```
jmartinesp, line 44: Maybe ignore it for the `_CAN_DUCK` variant if the requester is the media viewer? https://developer.android.com/media/optimize/audio-focus#automatic-ducking

Apparently the OS will automatically duck the audio in that case, but we shouldn't do it for voice-related audio (messages, call).

jmartinesp (approved):
Thanks, it was relatively easy to test using split screen with EXA an another media app.

bmarty:
> Thanks, it was relatively easy to test using split screen with EXA an another media app.

Yes, it's also possible to perform such tests using a video app (Youtube, Netflix, etc.) in Pip mode.

libraries/audio/impl/src/main/kotlin/io/element/android/libraries/audio/impl/DefaultAudioFocus.kt
bmarty: Reading the doc again, I think I can improve the code behavior yes. I will add more commits.

bmarty: OK, so I was about to check the result of the audio focus requester to prevent the start of the play back, but it seems that (well, on my phone at least) when there is an active call, the audio is ducked automatically by the system. Other apps does not always prevent playback during a phone call, for instance:
- Netflix: cannot play
- Youtube and Spotify: can start playback, but ducked.

So without doing anything, Element will have the same behavior than Youtube and Spotify, which is probably acceptable.

PR #4722, opened 2025-05-14
Notification events resolving and rendering in batches

jmartinesp:
## Content

- Added `NotificationResolverQueue` type that enqueues notification fetching requests received in a certain time window, and resolves them in batches, emitting the results when done.
- `DefaultPushHandler` now subscribes to these results and also processes them in batches, categorising the items and redirecting them to their callbacks (`onRedactedEventReceived`, `handleRingingCallEvent`, etc.).
- `DefaultNotificationDrawerManager` now has a `onNotifiableEventsReceived` to render several notifications at the same time.

## Motivation and context

The latest SDK version allows clients to fetch several notifications as a group so we need to run a single sliding sync request for them. However, since the items were returned at the same time and immediately sent to the `NotificationManager` so they can be displayed, the OS thought we were spamming the notifications and 'muted' (or discarded) some of them. Instead, we had to rework how we process the resolved notifications.

## Tests

- Send several messages in a very small time window in a room. If done correctly, both messages should be added to the notification at the same time.
- Redact some of those messages, it should display as 'Redacted message' or something similar in the notification drawer.
- Start a call in a DM, you should receive the ringing call notification with no issues.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationResolverQueue.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationResolverQueue.kt
@@ -16,6 +16,7 @@ import io.element.android.libraries.push.impl.notifications.model.ResolvedPushEv
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableSharedFlow
@@ -65,7 +66,7 @@ class NotificationResolverQueue @Inject constructor(
         Timber.d("Starting processing job for request: $request")
     }
 
-    private fun processQueue() = appCoroutineScope.launch {
+    private fun processQueue() = appCoroutineScope.launch(SupervisorJob()) {
         delay(BATCH_WINDOW_MS.milliseconds)
 
         // If this job is still active (so this is the latest job), we launch a separate one that won't be cancelled when enqueueing new items


bmarty (commented):
Thanks for improving this part!
Some early remarks.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
```diff
@@ -61,10 +61,14 @@ private val loggerTag = LoggerTag("DefaultNotifiableEventResolver", LoggerTag.No
  * this pattern allow decoupling between the object responsible of displaying notifications and the matrix sdk.
  */
 interface NotifiableEventResolver {
-    suspend fun resolveEvent(sessionId: SessionId, roomId: RoomId, eventId: EventId): Result<ResolvedPushEvent>
+    suspend fun resolveEvents(
+        sessionId: SessionId,
+        notificationEventRequests: List<NotificationEventRequest>
+    ): Result<Map<EventId, ResolvedPushEvent?>>
```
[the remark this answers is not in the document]
jmartinesp: https://github.com/element-hq/element-x-android/pull/4722/commits/6b517e169dafaff4e9d79112c91f51b724b9f947

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultOnMissedCallNotificationHandler.kt
```diff
@@ -30,8 +30,9 @@ class DefaultOnMissedCallNotificationHandler @Inject constructor(
         // Resolve the event and add a notification for it, at this point it should no longer be a ringing one
         val notificationData = matrixClientProvider.getOrRestore(sessionId).getOrNull()
             ?.notificationService()
-            ?.getNotification(roomId, eventId)
+            ?.getNotifications(sessionId, mapOf(roomId to listOf(eventId)))
```
bmarty, line 33: It looks weird to have to provide a `sessionId` here, since we are using a method from a service of a matrix client.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationResolverQueue.kt
```diff
@@ -0,0 +1,92 @@
...
+     * A flow that emits pairs of a list of notification event requests and a map of the resolved events.
+     * The map contains the original request as the key and the resolved event as the value.
+     */
+    val results: SharedFlow<Pair<List<NotificationEventRequest>, Map<NotificationEventRequest, Result<ResolvedPushEvent>>>> = MutableSharedFlow()
+
+    init {
+        coroutineScope.launch {
+            while (coroutineScope.isActive) {
+                // Wait for a batch of requests to be received in a specified time window
+                delay(BATCH_WINDOW_MS.milliseconds)
```
bmarty, line 53: Are you sure this code will behave as a first throttler?
If we get a Push just before the delay is ended, it will trigger the request without waiting for any other potential push. I have not checked in practice how much time we have between mutliple pending push received from Firebase though.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
```diff
@@ -178,7 +180,11 @@ class DefaultNotificationCreator @Inject constructor(
             // 'importance' which is set in the NotificationChannel. The integers representing
             // 'priority' are different from 'importance', so make sure you don't mix them.
             .apply {
-                if (roomInfo.shouldBing) {
+                val notificationManager = NotificationManagerCompat.from(context)
+                val previousNotification = notificationManager.activeNotifications.find { it.tag == roomInfo.roomId.value }
+                val elapsed = System.currentTimeMillis() - (previousNotification?.postTime ?: 0L)
+                Timber.d("Creating noisy notification for room ${roomInfo.roomId.value} with elapsed time $elapsed")
```
bmarty, line 186: Move this log into the if block (not sure)

bmarty (commented):
More comments. I will test the code on a real device.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
jmartinesp: Ooops, this should actually be gone. It was done for an experiment and I removed part of it, but apparently I kept this by mistake.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultOnMissedCallNotificationHandler.kt
jmartinesp: True, but `RustNotificationService` doesn't have a reference to the matrix client. Maybe I should just add a `SessionId` as a constructor param for it. WDYT?

bmarty: > Maybe I should just add a `SessionId` as a constructor param for it. WDYT?

Yes please. We're doing it for `RustRoomFactory` already for instance.

jmartinesp: `RustRoomFactory` doesn't use injection though. Adding it here would mean creating a factory abstraction.

jmartinesp: https://github.com/element-hq/element-x-android/pull/4722/commits/b5ece84b52dc9730af6f5a1041e9717761766c39

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationResolverQueue.kt
```diff
@@ -40,47 +41,59 @@ class NotificationResolverQueue @Inject constructor(
...
+        currentProcessingJob = processQueue()
+        Timber.d("Starting processing job for request: $request")
+    }
+
+    private fun processQueue() = coroutineScope.launch {
+        if (!isActive) return@launch
+
+        delay(BATCH_WINDOW_MS.milliseconds)
+
+        if (!isActive) return@launch
```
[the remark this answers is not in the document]
jmartinesp: You're right, it's not needed. `delay` will check if the context is alive for the whole duration of the function being suspended.

PR #4744, opened 2025-05-19
fix: Ignore global proxy settings if system thinks there's none

### LINKED ISSUE #3447: I can't login to any homeserver at all: "We couldn't reach this homeserver"
### Steps to reproduce

1. I download latest version from GitHub Releases
2. Opened the app, click continue
3. Why not matrix org, continue

## What did you expect?
Login and password prompt

## What happened instead?
We couldn't reach this homeserver. Please check that you have entered the homeserver USL correctly. If the URL is correct, contact your homeserver administrator for further help.

No matter what homeserver I try to login, it always gives this error

### Your phone model

POCO X5 Pro 5G

### Operating system version

Android 14

Element X 0.5.3

### Homeserver

I tried a lot of homeserver, none worked, even matrix.org

### Will you send logs?

Yes

No

ShadowRZ:
## Content

Ignore global proxy settings from `Settings.Global.getString(context.contentResolver, Settings.Global.HTTP_PROXY)` if the system says we have no proxy

## Motivation and context

Closes #3447

See https://github.com/element-hq/element-x-android/issues/3447#issuecomment-2841110041 for technical details.

## Tests

- Step 1: In the provisioned device, Run `adb shell settings put global http_proxy :0`
- Step 2: Check if Element X Android can select a identity provider without saying `We couldn't reach the homeserver`

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Android 15

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/proxy/DefaultProxyProvider.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/proxy/DefaultProxyProvider.kt
@@ -8,7 +8,9 @@
 package io.element.android.libraries.matrix.impl.proxy
 
 import android.content.Context
+import android.net.ConnectivityManager
 import android.provider.Settings
+import androidx.core.content.getSystemService
 import com.squareup.anvil.annotations.ContributesBinding
 import io.element.android.libraries.di.AppScope
 import io.element.android.libraries.di.ApplicationContext
@@ -32,6 +34,13 @@ class DefaultProxyProvider @Inject constructor(
     private val context: Context
 ) : ProxyProvider {
     override fun provides(): String? {
+        val defaultProxy = context.getSystemService<ConnectivityManager>()?.defaultProxy
+        if (defaultProxy == null) {
+            // Note: can be tested by running:
+            // adb shell settings put global http_proxy :0
+            Timber.d("No default proxy")
+            return null
+        }
         return Settings.Global.getString(context.contentResolver, Settings.Global.HTTP_PROXY)
             ?.also {
                 Timber.d("Using global proxy")


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/proxy/DefaultProxyProvider.kt
```diff
@@ -32,9 +34,11 @@ class DefaultProxyProvider @Inject constructor(
...
-        return Settings.Global.getString(context.contentResolver, Settings.Global.HTTP_PROXY)
-            ?.also {
-                Timber.d("Using global proxy")
-            }
+        return context.getSystemService<ConnectivityManager>()?.defaultProxy?.let {
+            Settings.Global.getString(context.contentResolver, Settings.Global.HTTP_PROXY)
+                ?.also {
+                    Timber.d("Using global proxy")
+                }
+        }
```
bmarty, line 42: Maybe rewrite to:
```kotlin
    override fun provides(): String? {
        val defaultProxy = context.getSystemService<ConnectivityManager>()?.defaultProxy
        if (defaultProxy == null) {
            // Note can be tested by runnin:
            // adb shell settings put global http_proxy :0
            Timber.d("No default proxy")
            return null
        }
        return Settings.Global.getString(context.contentResolver, Settings.Global.HTTP_PROXY)
            ?.also {
                Timber.d("Using global proxy")
            }
    }
```

for clarity?

ShadowRZ:
I did some further testing on the emulator:

* `adb shell settings put global http_proxy :1`: Direct connection
* `adb shell settings put global http_proxy :65535`: Direct connection
* `adb shell settings put global http_proxy 1:65536`: Direct connection
* `adb shell settings put global http_proxy 1:65535`: A proxy defined at `1:65535`

So I think that an invalid `http_proxy` value in Android means no proxy (direct), and to workaround this in the Rust SDK side we can just introduce a "lossy" proxy configration function where it'll throw away invalid proxy configurations.

jmartinesp:
I gave this a quick try and it seems like the current implementation works fine, the `defaultProxy` value is null if the proxy is not valid.

> So I think that an invalid http_proxy value in Android means no proxy (direct), and to workaround this in the Rust SDK side we can just introduce a "lossy" proxy configration function where it'll throw away invalid proxy configurations.

Isn't this what you already did in the PR? Unless I missed some corner case I think we can merge it as is, it's better than the current behaviour.

bmarty (approved):
Thanks for the update @ShadowRZ , and thanks for the extra test @jmartinesp !

PR #4758, opened 2025-05-21
Add `ActiveRoomHolder` to manage the active room for a session

jmartinesp:
## Content

What the title says

## Motivation and context

This allows us to reuse these rooms for several actions that would otherwise create a new instance of the room, which, if the room is already available, would cause delays of several seconds since the live timeline can take a while to restore if it contains many events.

## Tests

These actions should be noticeably faster:

- Starting or joining a call.
- Sharing media from outside the app.
- Syncing when a notification is received.
- Replying to a notification.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


appnav/src/main/kotlin/io/element/android/appnav/room/joined/JoinedRoomLoadedFlowNode.kt
```diff
@@ -95,6 +98,7 @@ class JoinedRoomLoadedFlowNode @AssistedInject constructor(
             },
             onDestroy = {
                 Timber.v("OnDestroy")
+                activeRoomHolder.removeRoom(inputs.room.sessionId)
```
bmarty, line 101: This does not work well with permalink. If a room A opens a room B, when going back to room A, the holder will be empty.

bmarty (commented):
Some remarks

appnav/src/main/kotlin/io/element/android/appnav/room/joined/JoinedRoomLoadedFlowNode.kt
bmarty, line 101: I still believe that this would be useful to provide the `roomId` to the `removeRoom` api, to double check that we are removing the correct room, and not the last added one.

Also one question, what will happen if there is an active call in the room, but the user closes the timeline? I guess the only issue is that there will be no cache, so not a real problem.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/tasks/ClearCacheUseCase.kt
```diff
@@ -37,8 +38,11 @@ class DefaultClearCacheUseCase @Inject constructor(
     private val ftueService: FtueService,
     private val pushService: PushService,
     private val seenInvitesStore: SeenInvitesStore,
+    private val activeRoomHolder: ActiveRoomHolder,
 ) : ClearCacheUseCase {
     override suspend fun invoke() = withContext(coroutineDispatchers.io) {
+        // Active rooms should be disposed of before clearing the cache
+        activeRoomHolder.removeRoom(matrixClient.sessionId)?.destroy()
```
bmarty, line 45: Now that `activeRoomHolder` stores a list of rooms, I think this code needs to be updated to destroy all the rooms.

services/appnavstate/api/src/main/kotlin/io/element/android/services/appnavstate/api/ActiveRoomHolder.kt
```diff
@@ -0,0 +1,59 @@
...
+import java.util.concurrent.ConcurrentHashMap
+import javax.inject.Inject
+
+/**
+ * Holds the active rooms for a given session so they can be reused instead of instantiating new ones.
+ *
+ * This works as a FILO (First In Last Out) stack, meaning that the last room added for a session will be the first one to be removed.
+ */
+@SingleIn(AppScope::class)
+class ActiveRoomHolder @Inject constructor() {
```
bmarty, line 24: Maybe rename to `ActiveRoomsHolder`?

bmarty (commented):
Thanks for the update. A few more comments.

appnav/src/main/kotlin/io/element/android/appnav/room/joined/JoinedRoomLoadedFlowNode.kt
jmartinesp: > I still believe that this would be useful to provide the roomId to the removeRoom api, to double check that we are removing the correct room, and not the last added one.

Makes sense.

> Also one question, what will happen if there is an active call in the room, but the user closes the timeline? I guess the only issue is that there will be no cache, so not a real problem.

I'd say nothing really. A call needs the sync to work and an active timeline to *start* the call, but if it's already running there should be no issues, at least with the current features.

bmarty (approved):
Thanks for the update!
2 more non-blocking remarks.

PR #4769, opened 2025-05-23
On boarding flow: add a screen to select account provider among a fixed list

bmarty:
## Content

When `EnterpriseService` provides a fixed list of account providers, i.e. at least 2 account providers and no "*", the login flow follows a new path so that when the user clicks on "Sign in manually", a screen with the list of available account providers is displayed.

## Motivation and context

Handle specific case where a fixed list of account providers is configured in the application.

## Screenshots / GIFs

See recorded screenshot. Note that the UI does not exactly match what was provided in Figma, but match with the existing code. I may iterate the UI in a future dedicated PR.

## Tests

- Need to hack the code. Let `DefaultEnterpriseService.defaultHomeserverList()` return a list of homeserver like `listOf("matrix.org", "element.io")`
- Run the app and start the login flow
- Click on "Sign in manually
- See the new screen
- Click on "matrix.org" this should open mas.matrix.org
- Go back
- Click on "element.io" this should open mas.element.io

The login flow can be finished normally

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


bmarty:
Good catch, let me handle this case a bit better, even if IRL it should barely occurs.

Generally the list can have the values:
- listOf("server.tld") -> one server, no way to use another one
- listOf("server1.tld", "server2.tld") -> fixed list of servers
- listOf("server.tld", "*") -> one sever, allow to use any other

The code should maybe just ensure that "*" is moved to the end of the list

jmartinesp (approved):
LGTM, thanks for fixing the issue. Having to select a provider then click on 'Continue' is a bit weird, although I guess this will be improved in the following PRs.

bmarty:
> LGTM, thanks for fixing the issue. Having to select a provider then click on 'Continue' is a bit weird, although I guess this will be improved in the following PRs.

This is the design but it's true that it's for iOS, so we may take some shortcut for Android...

image

PR #4771, opened 2025-05-23
Stronger lambda error

bmarty:
## Content

Ensure that `lambdaError()` really breaks the test in all circumstances.
It can happen that the throw Error is caught by the code, and this is not something expected.

Also there is a commit to harmonize how we create presenters to test them. I could have created a separate PR but this does not worth it I think.

## Motivation and context

More accurate tests
Same code pattern for creating presenter under test.

## Screenshots / GIFs

NA

## Tests

- Run all tests, they should not fail

## Tested devices

- NA

## Checklist

### DIFF
--- a/tests/testutils/src/main/kotlin/io/element/android/tests/testutils/lambda/Error.kt
+++ b/tests/testutils/src/main/kotlin/io/element/android/tests/testutils/lambda/Error.kt
@@ -7,8 +7,19 @@
 
 package io.element.android.tests.testutils.lambda
 
+import kotlin.system.exitProcess
+
 fun lambdaError(
     message: String = "This lambda should never be called."
 ): Nothing {
-    throw AssertionError(message)
+    // Throwing an exception here is not enough, it can be caught.
+    // Instead exit the process to make sure the test fails.
+    // The error will be:
+    // "Could not stop all services."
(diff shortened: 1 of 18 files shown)


bmarty:
@jmartinesp you're 💯 right and this is probably something we will do at some point. Can you create a new issue (task) with that? In the mean time, the idea of this "quick fix" is to avoid losing time when writing tests, which happened to me a few days ago.

tests/testutils/src/main/kotlin/io/element/android/tests/testutils/lambda/Error.kt
```diff
@@ -7,8 +7,19 @@
...
+
 fun lambdaError(
     message: String = "This lambda should never be called."
 ): Nothing {
-    throw AssertionError(message)
+    // Throwing an exception here is not enough, it can be caught.
+    // Instead exit the process to make sure the test fails.
+    // The error will be:
+    // "Could not stop all services."
+    // In this case, but a breakpoint here and run the test in debug mode to identify which lambda is failing.
```
ganfra, line 19: typo : 
> put a breakpoint

ganfra (approved):

Issue #4778, opened 2025-05-26
[Task] Replace usages of `runCatching/mapCatching` with our own versions


jmartinesp:
We're extensively using `runCatching` in the codebase because it's very useful, but it's also a code smell (see https://github.com/Kotlin/kotlinx.coroutines/issues/1814). This can catch and ignore:

- `Error` types that should be thrown (see https://github.com/element-hq/element-x-android/pull/4771).
- Coroutine cancellations that are needed to make sure the async code works as expected.

We should replace those usages with our own versions, maybe using some of the suggestions in the issue from the Kotlin repo.

PR #4797, opened 2025-05-29
Add `catchingExceptions` method to replace `runCatching`

jmartinesp:
## Content

- Add `catchingExceptions(block())`: this method creates a `Result` by catching only `Exception` throwables, also re-throwing `CancellationException` so coroutines can work as expected.
- Add also `Result.mapCatchingException(block())` to replace `Result.mapCatching`, with the same behaviour of catching only exceptions.
- Make `tryOrNull` also catch only exceptions and re-throw cancellation exceptions.
- Replace usages of these methods in the project and thrown error/throwables so the tests pass again.
- Create a custom detekt rules module with an specific rule to prevent `runCatching` and `mapCatching` from being used.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/4778

## Tests

The app builds, the tests should pass and we have Maestro for runtime.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


libraries/core/src/main/kotlin/io/element/android/libraries/core/extensions/Result.kt
```diff
@@ -7,6 +7,61 @@
...
+import kotlin.coroutines.cancellation.CancellationException
+
+/**
+ * Can be used to catch exceptions in a block of code and return a [Result].
+ * If the block throws a [CancellationException], it will be rethrown.
+ * If it throws any other exception, it will be wrapped in a [Result.failure].
+ *
+ * [Error]s are not caught by this function, as they are not meant to be caught in normal application flow.
+ */
+inline fun <T> catchingExceptions(
```
bmarty: To be closer to `runCatching`, maybe rename to `runCatchingExceptions`?

Also I am wondering if we could still have a safe mode (for production?) where we still catch `Error`s. Here we take the risk to have crashes on production. Maybe we could have only this on debug build, so that we have time to fix existing problems?

libraries/core/src/main/kotlin/io/element/android/libraries/core/extensions/Result.kt
```diff
@@ -7,6 +7,61 @@
...
+    }
+}
+
+/**
+ * Can be used to transform a [Result] into another [Result] by applying a [block] to the value if it is successful.
+ * If the original [Result] is a failure, the exception will be wrapped in a new [Result.failure].
+ *
+ * This is a safer version of [Result.mapCatching].
+ */
+inline fun <R, T> Result<T>.mapCatchingException(
```
bmarty: Maybe rename to `mapCatchingExceptions`? It seems that this is what you wanted to do according to the description of the detekt rule you wrote :)

libraries/core/src/main/kotlin/io/element/android/libraries/core/extensions/Result.kt
```diff
@@ -7,6 +7,61 @@
...
+    block: (T) -> R,
+): Result<R> {
+    return when {
+        isSuccess -> catchingExceptions { block(getOrNull()!!) }
+        else -> Result.failure(exceptionOrNull()!!)
+    }
+}
```
bmarty, line 63: Maybe rewrite to
```kotlin
    return fold(
        onSuccess = { value -> catchingExceptions { block(value) } },
        onFailure = { exception -> Result.failure(exception) }
    )
```

to avoid usage of `!!`.

libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/fixtures/fakes/FakeSessionVerificationController.kt
```diff
@@ -0,0 +1,16 @@
...
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.matrix.impl.fixtures.fakes
+
+import org.matrix.rustcomponents.sdk.NoPointer
+import org.matrix.rustcomponents.sdk.SessionVerificationController
+import org.matrix.rustcomponents.sdk.SessionVerificationControllerDelegate
+
+class FakeSessionVerificationController : SessionVerificationController(NoPointer) {
```
bmarty: The (not-written) convention is to name this class `FakeRustSessionVerificationController` so that we see that it's not a regular `Fake` class that we use for Unit test.
It seems that I did not follow the convention for `FakeQrCodeData`, my bad...

tests/testutils/src/main/kotlin/io/element/android/tests/testutils/lambda/Error.kt
```diff
@@ -17,9 +15,5 @@ fun lambdaError(
     // The error will be:
     // "Could not stop all services."
     // In this case, put a breakpoint here and run the test in debug mode to identify which lambda is failing.
-    System.err.println(message)
-    Thread.currentThread().stackTrace.forEach {
-        System.err.println(it)
-    }
-    exitProcess(1)
+    throw AssertionError(message)
```
bmarty, line 13: The comment above can be removed I think!

bmarty (approved):
LGTM, some small remark.

On Element Android we had an opt-in fail fast option in the developer settings, useful to avoid crash at runtime (when disabled) as much as possible. This is maybe something we want to introduce here?

libraries/core/src/main/kotlin/io/element/android/libraries/core/extensions/Result.kt
jmartinesp: I don't think we can, not easily 🫤 . This is in a java library module so we don't have the concept of `BuildConfig` to check if it's a debug or release version and we can't inject the `BuildMeta` here either.

PR #4801, opened 2025-05-30
Use Localazy's `langAliases` for Indonesian language

jmartinesp:
## Content

Instead of manually moving all the `values-id` to `values-in` in some post-processing, this instructs the CLI to already download them in the right folders.

## Motivation and context

This way we can remove the 'move all translations' step in the download strings script. I was looking at something else when I noticed this option for the Localazy CLI tool.

## Tests

I tested this locally with `./tools/localazy/downloadStrings.sh --all` and it worked as expected.

## Checklist


bmarty (approved):
Thanks, this is removing some fairly long steps!

Issue #4811, opened 2025-06-03
Attachment should send files without mime so no video/audio conversion happens


escix:
### Steps to reproduce

1. Where are you starting? What can you see?
Trying to send a video as attachment as currently there is a bug in videos increasing size when uploaded.

#### What did you expect?

Attachments sent as just files.

#### What happened instead?

Attachment uploaded as video and the video conversion kicks in with increased size.

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

PR #4815, opened 2025-06-04
fix: When sending media as files use the `octet-stream` type

jmartinesp:
## Content

What the title says.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/4811.

## Tests

- Send an image/video as a file.
- You should see no preview, the upload should be immediate.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
```diff
@@ -165,13 +165,13 @@ class MessageComposerPresenter @AssistedInject constructor(
...
-            handlePickedMedia(uri)
+            handlePickedMedia(uri, MimeTypes.OctetStream)
         }
         val cameraPhotoPicker = mediaPickerProvider.registerCameraPhotoPicker { uri ->
-            handlePickedMedia(uri, MimeTypes.IMAGE_JPEG)
+            handlePickedMedia(uri, MimeTypes.Jpeg)
         }
         val cameraVideoPicker = mediaPickerProvider.registerCameraVideoPicker { uri ->
-            handlePickedMedia(uri, MimeTypes.VIDEO_MP4)
+            handlePickedMedia(uri, MimeTypes.Mp4)
```
bmarty, line 175: I am wondering if we should not do the opposite and remove our `MimeTypes` object.

bmarty (approved):
LGTM, hence I am pretty sure that we will have people complaining that media is rendered as file in the timeline (if I understand correctly the consequences)

jmartinesp:
> LGTM, hence I am pretty sure that we will have people complaining that media is rendered as file in the timeline (if I understand correctly the consequences)

Probably, but this is the behaviour we wanted to implement from the beginning AFAICT.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
jmartinesp: Maybe we could think about it in another PR. It's weird though, the 'official' `MimeTypes` class has no `application/octet-stream` constant.

PR #4756, opened 2025-05-20
Stop sending call notifications

robintown:
## Content

In an upcoming version of Element Call, the Element Call widget will assume responsibility for sending notification events when you start a call. So we no longer need to have the client application send them.

## Motivation and context

https://github.com/element-hq/element-call/pull/3248

## Screenshots / GIFs

N/A

## Tests

TODO

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [ ] Pull request title will be used in the release note, it clearly define what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #4819, opened 2025-06-04
Add extra logs the 'send call notification' flow

jmartinesp:
## Content

Add some extra logs for the 'send call notification' flow when starting a call.

## Motivation and context

It's been observed the notify call event is not always sent. This should help debug why.

## Checklist

### DIFF
--- a/features/roomcall/impl/src/main/kotlin/io/element/android/features/roomcall/impl/RoomCallStatePresenter.kt
+++ b/features/roomcall/impl/src/main/kotlin/io/element/android/features/roomcall/impl/RoomCallStatePresenter.kt
@@ -46,14 +46,18 @@ class RoomCallStatePresenter @Inject constructor(
                 (currentCall as? CurrentCall.RoomCall)?.roomId == room.roomId
             }
         }
-        val callState = when {
-            isAvailable.not() -> RoomCallState.Unavailable
-            roomInfo.hasRoomCall -> RoomCallState.OnGoing(
-                canJoinCall = canJoinCall,
-                isUserInTheCall = isUserInTheCall,
-                isUserLocallyInTheCall = isUserLocallyInTheCall,
-            )
-            else -> RoomCallState.StandBy(canStartCall = canJoinCall)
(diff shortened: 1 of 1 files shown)


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/ActiveCallManager.kt
```diff
@@ -181,6 +181,9 @@ class DefaultActiveCallManager @Inject constructor(
             Timber.tag(tag).w("Call type $callType does not match the active call type, ignoring")
             return
         }
+
+        Timber.tag(tag).d("Hung up call: $callType")
```
bmarty, line 185: Is there some sensitive information in `callType`? Maybe safer to shrink some data first?

jmartinesp: I don't think there's anything there that won't be logged at some other point (URL in webview or session id + room id for room calls).

jmartinesp:
> Thanks. I am wondering if the send may be blocked because there is a unverified device in the room. Could it be possible?

This can happen, but according to @manuroe in his case it's intermittent, so I don't think it's related. Either that or some user in that room was adding and verifying/signing out sessions non-stop 😅 .

bmarty:
Also we have #4756, so 🤷‍♂️

jmartinesp:
> Also we have #4756, so 🤷‍♂️

I think that may be a bit stale. Most of the logs in this PR would still be valuable even if we remove the code to send the call notification (knowing when the call has started, when we receive an incoming call, when we hang up, the type of call, etc.).

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/ActiveCallManager.kt
jmartinesp: I added some changes to prevent the URL from being logged.

PR #4830, opened 2025-06-05
Fix room and user avatar downloaded with a `.bin` extension.

### LINKED ISSUE #4803: Downloading pfp/room picture corrupts it
When I attempt to download it, I do not get the original file that I expect. Instead I get a corrupted version that lacks a usable file extension (.bin)

I usually resort to manually finding out what file type it is and then adding the correct extension, although this is still ridiculous.

Image

Image

Image

bmarty:
## Content

Let the SDK set a correct extension to downloaded file by providing a full mimetype when a base mimetype is available.

## Motivation and context

Fixes #4803

## Tests

- Open a room avatar and downloaded it to the system.
- Observe that the extension of the downloaded file is no `jpeg`, it was `bin` previously.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/media/RustMediaLoader.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/media/RustMediaLoader.kt
@@ -72,7 +72,15 @@ class RustMediaLoader(
                     val mediaFile = innerClient.getMediaFile(
                         mediaSource = mediaSource,
                         filename = filename,
-                        mimeType = mimeType?.takeIf { MimeTypes.hasSubtype(it) } ?: MimeTypes.OctetStream,
+                        mimeType = when {
+                            mimeType == null -> MimeTypes.OctetStream
+                            MimeTypes.hasSubtype(mimeType) -> mimeType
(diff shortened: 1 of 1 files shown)


jmartinesp:
I think this is doing what iOS does, assuming images are jpgs. At least, when I downloaded this image it appeared as `.jpeg` and when you open it with a hex editor you can see it's a PNG.

tmpe4nNwf

It's better than what we have, but I was thinking about something more like [this option](https://stackoverflow.com/a/75305893).

bmarty:
> It's better than what we have, but I was thinking about something more like [this option](https://stackoverflow.com/a/75305893).

Yes, I was thinking about that too, but here, it appears that it's the Rust SDK which downloads the file and [decides](https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk/src/media.rs#L329) of its extension based on the provided mime type. So I'd argue that this problem could be fixed SDK side.

In the meantime I think that this fix is enough as a first step. Do you agree?

jmartinesp:
Yes, it's probably fine. Also, it seems like there are some very handy mime type libraries on Rust that could solve this problem instead of us re-implementing them ourselves. It's probably worth a shot doing it there.

ganfra (approved):

PR #4835, opened 2025-06-06
Log the push resolving failure reason if available

jmartinesp:
## Content

What the title says.

## Motivation and context

When processing the push events in bulk we lost the ability to get the reason behind a failure. I believe I forgot to add a TODO to fix the placeholder message, sorry 🫤 .

## Tests

- When a push fails to resolve, check the underlying reason is logged, or at least the associated exception type.

## Checklist

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/DefaultPushHandler.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/DefaultPushHandler.kt
@@ -75,22 +75,35 @@ class DefaultPushHandler @Inject constructor(
                 for (request in requests) {
                     // Log the result of the push notification event
                     val result = resolvedEvents[request]
-                    if (result?.isSuccess == true) {
-                        pushHistoryService.onSuccess(
-                            providerInfo = request.providerInfo,
-                            eventId = request.eventId,
-                            roomId = request.roomId,
-                            sessionId = request.sessionId,
-                            comment = "Push handled successfully",
-                        )
-                    } else {
-                        pushHistoryService.onUnableToResolveEvent(
-                            providerInfo = request.providerInfo,
-                            eventId = request.eventId,
-                            roomId = request.roomId,
-                            sessionId = request.sessionId,
-                            reason = "Push not handled",
-                        )
+                    when {
+                        result == null -> {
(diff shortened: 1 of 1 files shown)


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/DefaultPushHandler.kt
```diff
@@ -75,22 +75,35 @@ class DefaultPushHandler @Inject constructor(
...
+                            pushHistoryService.onSuccess(
+                                providerInfo = request.providerInfo,
+                                eventId = request.eventId,
+                                roomId = request.roomId,
+                                sessionId = request.sessionId,
+                                comment = "Push handled successfully",
+                            )
+                        }
+                        result.isFailure -> {
+                            val exception = result.exceptionOrNull()!!
```
ganfra: Are we sure this can't crash?

bmarty: yes, but it's better to use `fold()` I guess

jmartinesp: Ok, changing this to a fold call.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/DefaultPushHandler.kt
```diff
@@ -75,21 +75,34 @@ class DefaultPushHandler @Inject constructor(
...
+                                    comment = "Push handled successfully",
+                                )
+                            },
+                            onFailure = { exception ->
+                                pushHistoryService.onUnableToResolveEvent(
+                                    providerInfo = request.providerInfo,
+                                    eventId = request.eventId,
+                                    roomId = request.roomId,
+                                    sessionId = request.sessionId,
+                                    reason = exception.message ?: exception.javaClass.simpleName,
```
bmarty, line 103: This is matching what we had [before](https://github.com/element-hq/element-x-android/pull/4722/files#diff-2d8c2336c93c4152ea5c38b36b8aae57056727b09313bc1e12f525635d71c0feL158), thanks!

PR #4845, opened 2025-06-09
Add a banner to ask the user to disable battery optimization when Event cannot be resolved from Push

### LINKED ISSUE #4611: Add check/info to disable battery optimization to F-Droid-build's notification settings/troubleshooter

Have a reminder or test if battery optimization is set to "unrestricted" in Element X F-Droid build's notification settings/troubleshooter.

Recently, I installed Element X to 2 de-Googled phones, set up unified push using Nextcloud/NextPush and then ntfy, but in both cases notifications worked fine when the phones were unlocked (awake) but kept failing after they were locked for a few minutes.  
(Failing in a sense that they never arrived, were simply skipped/lost).

Initially I kept checking all the battery optimizations and tweaks of NextPush and ntfy, without success.  
Then I realized that maybe also Element needs to be unrestricted.  
And this fixed the issue.

It would be handy to have hint/check about the battery optimizations also for Element in the notification settings/troubleshooter, as all the green checkmarks kinda create a false sense of "here's everything alright" even tho it's not 😁 

**Thanks for a great app** 😉 ❤ 

Image

Image

_No response_

### Additional context

_No response_

No

bmarty:
## Content

When an Event cannot be resolved from Push, let the application render a banner in the room list to propose the user to disable battery optimization.

When battery optimization is disabled, it should help to resolve more events when a Push is received. The cost is a higher battery consumption. User always has the ability to revert the change by navigating to the system settings.

Draft as we're waiting for product decision and the wording of the banner may be updated (and must be added to Localazy).

## Motivation and context

Improve Notification resolution and limit the risk of not shown notification because not resolve. In a separate PR we will ensure that a notification is always displayed when a not resolved Push is happening.

Closes #4611

## Screenshots / GIFs

https://github.com/user-attachments/assets/5bed12ad-fdce-41db-9942-cf50d435a5a8

https://github.com/user-attachments/assets/3ddf267b-3ef2-4061-8aa2-90b2f4e71251

## Tests

- Have an unresolved event
- Open the app
- See the banner
- Option 1: Close the banner: it should not be displayed again. From the application, clear the cache to be able to test it again
- Option 2: click on "Yes, disable"
- A system dialog is displayed (can depend on OS / device)
- Option 1: click on "Deny"
- The system dialog is gone and the banner is gone until the application is started again
- Option 2: click on "Accept"
- The system dialog is gone and the banner is gone until the battery optimization setting remains unchanged.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/push/api/src/main/kotlin/io/element/android/libraries/push/api/battery/BatteryOptimizationEvents.kt
+++ b/libraries/push/api/src/main/kotlin/io/element/android/libraries/push/api/battery/BatteryOptimizationEvents.kt
@@ -9,5 +9,5 @@ package io.element.android.libraries.push.api.battery
 
 sealed interface BatteryOptimizationEvents {
     data object Dismiss : BatteryOptimizationEvents
-    data object DoAction : BatteryOptimizationEvents
+    data object RequestDisableOptimizations : BatteryOptimizationEvents
 }
--- a/libraries/push/api/src/main/kotlin/io/element/android/libraries/push/api/battery/BatteryOptimizationEvents.kt
+++ b/libraries/push/api/src/main/kotlin/io/element/android/libraries/push/api/battery/BatteryOptimizationEvents.kt
@@ -7,7 +7,7 @@
 
 package io.element.android.libraries.push.api.battery
 
-data class BatteryOptimizationState(
-    val shouldDisplayBanner: Boolean,
-    val eventSink: (BatteryOptimizationEvents) -> Unit,
-)
+sealed interface BatteryOptimizationEvents {
+    data object Dismiss : BatteryOptimizationEvents
+    data object DoAction : BatteryOptimizationEvents
+}
--- a/libraries/push/api/src/main/kotlin/io/element/android/libraries/push/api/battery/BatteryOptimizationState.kt
+++ b/libraries/push/api/src/main/kotlin/io/element/android/libraries/push/api/battery/BatteryOptimizationState.kt
@@ -1,0 +1,16 @@
+/*
+ * Copyright 2025 New Vector Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.push.api.battery
+
+import android.app.Activity
+
+data class BatteryOptimizationState(
+    val shouldDisplayBanner: Boolean,
+    val dismiss: () -> Unit,
+    val openSettings: (Activity?) -> Unit,
+)
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/battery/BatteryOptimization.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/battery/BatteryOptimization.kt
@@ -57,14 +57,28 @@ class AndroidBatteryOptimization @Inject constructor(
 
     @SuppressLint("BatteryLife")
     override fun requestDisablingBatteryOptimization(): Boolean {
+        val ignoreBatteryOptimizationsResult = launchAction(Settings.ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS, withData = true)
+        if (ignoreBatteryOptimizationsResult) {
+            return true
+        }
+        // Open settings as a fallback if the first attempt fails
+        return launchAction(Settings.ACTION_IGNORE_BATTERY_OPTIMIZATION_SETTINGS, withData = false)
+    }
+
+    private fun launchAction(
+        action: String,
+        withData: Boolean,
+    ): Boolean {
         val intent = Intent()
-        intent.action = Settings.ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS
-        intent.data = ("package:" + context.packageName).toUri()
+        intent.action = action
+        if (withData) {
+            intent.data = ("package:" + context.packageName).toUri()
+        }
         return try {
             externalIntentLauncher.launch(intent)
             true
         } catch (exception: ActivityNotFoundException) {
-            Timber.w(exception, "Cannot request ignoring battery optimizations.")
+            Timber.w(exception, "Cannot launch intent with action $action.")
             false
         }
     }
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/battery/BatteryOptimization.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/battery/BatteryOptimization.kt
@@ -8,7 +8,6 @@
 package io.element.android.libraries.push.impl.battery
 
 import android.annotation.SuppressLint
-import android.app.Activity
 import android.content.ActivityNotFoundException
 import android.content.Context
 import android.content.Intent
@@ -19,6 +18,7 @@ import androidx.core.net.toUri
 import com.squareup.anvil.annotations.ContributesBinding
 import io.element.android.libraries.di.AppScope
 import io.element.android.libraries.di.ApplicationContext
+import io.element.android.services.toolbox.api.intent.ExternalIntentLauncher
 import timber.log.Timber
 import javax.inject.Inject
 
@@ -39,16 +39,16 @@ interface BatteryOptimization {
      * This will open the system settings where the user can disable battery optimizations.
      * See https://developer.android.com/training/monitoring-device-state/doze-standby#exemption-cases
      *
-     * @param activity The activity from which to start the settings intent.
      * @return true if the intent was successfully started, false if the activity was not found
      */
-    fun requestDisablingBatteryOptimization(activity: Activity?): Boolean
+    fun requestDisablingBatteryOptimization(): Boolean
 }
 
 @ContributesBinding(AppScope::class)
 class AndroidBatteryOptimization @Inject constructor(
     @ApplicationContext
     private val context: Context,
+    private val externalIntentLauncher: ExternalIntentLauncher,
 ) : BatteryOptimization {
     override fun isIgnoringBatteryOptimizations(): Boolean {
         return context.getSystemService<PowerManager>()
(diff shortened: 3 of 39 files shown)


libraries/push/impl/src/test/kotlin/io/element/android/libraries/push/impl/push/DefaultPushHandlerTest.kt
```diff
@@ -268,11 +268,35 @@ class DefaultPushHandlerTest {
...
+    }
+
+    @Test
+    fun `when classical PushData is received, but not able to resolve the event, the banner to disable battery optimization will be displayed`() {
+        `test notification resolver failure`(
+            notificationResolveResult = { requests: List<NotificationEventRequest> ->
+                Result.success(
+                    requests.associateWith { Result.failure(ResolvingException("Unable to resolve event")) }
+                )
+            },
```
bmarty, line 287: The existing test was not really testing the notification resolver failure (the type is a `Result` of a Map of `Result`, so it's a bit confusing), but this change fixes it.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/battery/BatteryOptimization.kt
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Tells if the application ignores battery optimizations.
+     *
+     * Ignoring them allows the app to run in background to make background sync with the homeserver.
+     * This user option appears on Android M but Android O enforces its usage and kills apps not
+     * authorised by the user to run in background.
+     *
+     * @return true if battery optimisations are ignored
+     */
+    fun isIgnoringBatteryOptimizations(): Boolean
```
jmartinesp, line 35: Nit, so feel free to ignore, but this looks more like something that could be a property instead of a function.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/store/DefaultPushDataStore.kt
```diff
@@ -84,4 +110,10 @@ class DefaultPushDataStore @Inject constructor(
             it.clear()
         }
     }
+
+    companion object {
+        const val BATTERY_OPTIMIZATION_BANNER_STATE_INIT = 0
+        const val BATTERY_OPTIMIZATION_BANNER_STATE_SHOW = 1
+        const val BATTERY_OPTIMIZATION_BANNER_STATE_DISMISSED = 2
+    }
```
jmartinesp, line 118: Can't we move this to the interface declaration instead and use these values in `MutableBatteryOptimizationStore` instead of the raw ints?

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/MutableBatteryOptimizationStore.kt
```diff
@@ -0,0 +1,31 @@
...
+class DefaultMutableBatteryOptimizationStore @Inject constructor(
+    private val defaultPushDataStore: DefaultPushDataStore,
+) : MutableBatteryOptimizationStore {
+    override suspend fun showBatteryOptimizationBanner() {
+        defaultPushDataStore.setBatteryOptimizationBannerState(1)
+    }
+
+    override suspend fun onOptimizationBannerDismissed() {
+        defaultPushDataStore.setBatteryOptimizationBannerState(2)
+    }
```
jmartinesp: Seeing this only relays the action to an actual `Store` implementation, maybe we should remove this component? Or at least rename it so we don't mistake it for some store with its own data source.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/battery/BatteryOptimizationPresenter.kt
```diff
@@ -0,0 +1,66 @@
...
+        LifecycleResumeEffect(Unit) {
+            isSystemIgnoringBatteryOptimizations = batteryOptimization.isIgnoringBatteryOptimizations()
+            if (isRequestSent) {
+                localShouldDisplayBanner = false
+            }
+            onPauseOrDispose {}
+        }
+
+        return BatteryOptimizationState(
+            shouldDisplayBanner = localShouldDisplayBanner && storeShouldDisplayBanner && !isSystemIgnoringBatteryOptimizations,
```
jmartinesp: Just double checking, but this means if the user denies the battery optimisation change in the dialog we'd just keep displaying the banner, right?

jmartinesp (commented):
Thanks for the changes! I added a few quick review comments.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/store/DefaultPushDataStore.kt
bmarty: `PushDataStore` is more like the external interface (read only store, except for the `reset()` method).
For the raw ints, it's been fixed in 9956ed60e6db1067de7fe83cb5f622faff86a5a5

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/battery/BatteryOptimizationPresenter.kt
bmarty: Yes, the banner will be dismissed for the lifetime of the Presenter. I think that this is OK regarding all the various possible UX flows.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/MutableBatteryOptimizationStore.kt
bmarty: Having a higher level interface helps for clarity and allow to inject a different implementation for the tests. Do you have a better name to suggest?

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/battery/BatteryOptimization.kt
bmarty: I am not a big fan of `val` backed up by a method which may change their result any time. If we change it to a `val`, it will look weird (to me at least) that we request the value each time the app is resumed in `BatteryOptimizationPresenter`

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/MutableBatteryOptimizationStore.kt
jmartinesp: > Do you have a better name to suggest?

Not really no... I was thinking this looks like an use case, but it has 2 methods so it doesn't really match. I think it's ok to keep it like this if it helps with the tests.

libraries/push/api/src/main/kotlin/io/element/android/libraries/push/api/battery/BatteryOptimizationState.kt
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+package io.element.android.libraries.push.api.battery
+
+import android.app.Activity
+
+data class BatteryOptimizationState(
+    val shouldDisplayBanner: Boolean,
+    val dismiss: () -> Unit,
+    val openSettings: (Activity?) -> Unit,
```
ganfra: Use EventSink instead please?
Also we don't want Activity reference in our `State`, the activity should be grabbed in the dependency if possible. By the way, appContext is not enough here?

bmarty: You're right, I have improved the code, also on the banner for full screen intent.

mxandreas:
A couple of suggestions on the wording:

> Notification tip

Notifications not arriving?

> To be sure to receive all the notifications, it can help to disable the battery optimization for this application.

Disable battery optimization for this app, to make sure all notifications are received.

> Yes, disable

Disable Optimization

Am I right that we do not directly check if the battery optimization is enabled, we derive this from the fact that there was an unresolved event? So in theory it could be that the battery optimization actually is already disabled and there was another reason why the event was unresolved?

bmarty:
Hello, @mxandreas , thanks for the feedback.

> Am I right that we do not directly check if the battery optimization is enabled, we derive this from the fact that there was an unresolved event? So in theory it could be that the battery optimization actually is already disabled and there was another reason why the event was unresolved?

No, Element X also checks if the battery optimisation is already disabled. If battery optimization is already disabled, we do not show the banner. ([code](https://github.com/element-hq/element-x-android/pull/4845/files#diff-4ce09f9c1e874b7ed1add2b63a83b5772622b884d9e66d573ac422ecba9cb895R67))

libraries/push/api/src/main/kotlin/io/element/android/libraries/push/api/battery/BatteryOptimizationEvents.kt
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.push.api.battery
+
+sealed interface BatteryOptimizationEvents {
+    data object Dismiss : BatteryOptimizationEvents
+    data object DoAction : BatteryOptimizationEvents
```
jmartinesp: `DoAction` sounds way too generic IMO 😅 . Why not just `DisableOptimizations`? Even if we end up opening the app settings as a fallback for this I think it's fine.

bmarty: Ah yes, renamed to `RequestDisableOptimizations`. I wanted to use a common Event interface for the other banner, but I changed my mind and forgot to rename the data object.

Issue #4844, opened 2025-06-09
App crashes when trying to start a call.


YamatoRyou:
### Steps to reproduce

1. Update to 25.6.1;
2. Launch and enter an EC enabled room;
3. Try to start a call.

#### What did you expect?
n/a

#### What happened instead?
I tested the app on 3 phones, 2 of which crashed when starting a call:
- HMD TA-1054 (Android 8.1) ❌️
- Mi 9T (MIUI 12.5 by xiaomi.eu, Android 11) ❌️
- OnePlus MT2111 (Oxygen OS 15, Android 15) ✅️

Logs (From Mi 9T):
```
ERROR elementx: FATAL EXCEPTION ElementX Version : 1.0Phone : Redmi K20 (V12.5.2.0.RFJCNXM 11 REL)
Memory statuses 
usedSize   16 MB
freeSize   2 MB
totalSize   18 MB
Thread: main, Exception: java.lang.NoClassDefFoundError: Failed resolution of: Lio/element/android/features/call/impl/utils/WebViewAudioManager$$ExternalSyntheticLambda5;
	at io.element.android.features.call.impl.utils.WebViewAudioManager.<init>(SourceFile:103)
	at com.freeletics.flowredux.dsl.BaseBuilderBlock$$ExternalSyntheticLambda1.invoke(SourceFile:921)
	at com.bumble.appyx.core.node.Node$$ExternalSyntheticLambda3.invoke(SourceFile:1718)
	at androidx.compose.ui.viewinterop.ViewFactoryHolder.<init>(SourceFile:1)
	at androidx.compose.ui.viewinterop.AndroidView_androidKt$createAndroidViewNodeFactory$1$1.invoke(SourceFile:23)
	at androidx.compose.runtime.changelist.Operation$UpdateValue.execute(SourceFile:145)
	at androidx.compose.runtime.changelist.Operations.executeAndFlushAllPendingOperations(SourceFile:22)
	at androidx.compose.runtime.changelist.Operation$InsertSlotsWithFixups.execute(SourceFile:41)
	at androidx.compose.runtime.changelist.Operations.executeAndFlushAllPendingOperations(SourceFile:22)
	at androidx.compose.runtime.changelist.ChangeList.executeAndFlushAllPendingChanges(SourceFile:3)
	at androidx.compose.runtime.CompositionImpl.applyChangesInLocked(SourceFile:48)
	at androidx.compose.runtime.CompositionImpl.applyChanges(SourceFile:6)
[...]

PR #4847, opened 2025-06-09
Fix crash when using Element Call on API <= 30

jmartinesp:
## Content

Make `commsDeviceChangedListener` a lazy property to avoid crashes related to API versions (thanks @bmarty for the idea!).

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/4844

`commsDeviceChangedListener` is added as a non-null property but the `AudioManager.OnCommunicationDeviceChangedListener` class it implements is only available on API 31+. Although it's not used in APIs <= 30 the property definition is still there and the app will crash when trying to run the code.

## Tests

On a device with Android 11 or lower, start or join a call.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 11

## Checklist


bmarty (approved):
Thanks!

PR #4854, opened 2025-06-10
Element Call: add delay before selecting the default audio device

jmartinesp:
## Content

What the title says.

## Motivation and context

Even with the new audio callback, there are still a few times when the audio device is selected too soon and the right audio stream isn't used. This delay *should* fix it.

## Tests

Join several calls, make sure the in-call audio stream is always used.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


bmarty (approved):
OK, thanks!

PR #4876, opened 2025-06-13
Disable BT audio devices for Element Call on Android < 12

jmartinesp:
## Content

- Disable audio devices on Android devices with an OS version < 12 - using some legacy audio APIs.
- When this happens, mute the audio from the call, display a dialog and unmute the call once a different audio device is selected.
- Also add a workaround for an issue where some devices weren't able to easily leave the call after hanging up.

## Motivation and context

After trying several strategies to make BT audio devices play nice with the embedded Element Call instance on Android versions using the legacy audio APIs - Android 11 and lower - and it failing in several ways - wrong audio device being used when switching, audio just dying until the app is force closed or the app crashing with a native OS error and no stacktrace - we had to make a difficult decision of preventing using these devices at all.

This change makes it so when selecting a BT audio device or when it's selected by default by the OS, we notify the user about the device not being usable, and mute the audio to prevent it from being streamed through some other audio device. Once a new device is selected, the audio is enabled again and that device can be used normally.

## Screenshots / GIFs

image

## Tests

With an Android device using Android 11 or lower:

- Open a call with a BT audio device connected. You should see the error dialog and be able to switch to another device and have the audio working properly with those.
- When selecting the BT audio device again you should see the same result.
- Everything else related to audio in calls should still work as before.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 10

## Checklist


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenView.kt
```diff
@@ -157,6 +170,22 @@ internal fun CallScreenView(
...
+@Composable
+private fun InvalidAudioDeviceDialog(
+    invalidAudioDeviceReason: InvalidAudioDeviceReason,
+    onDismiss: () -> Unit,
+) {
+    ErrorDialog(
+        content = when (invalidAudioDeviceReason) {
+            InvalidAudioDeviceReason.BT_AUDIO_DEVICE_DISABLED -> {
+                // TODO: use a string resource when available
+                "Element Call does not support using Bluetooth audio devices in this Android version. Please select a different audio device."
```
jmartinesp: This is replaced by the actual string in a following commit.

PR #4884, opened 2025-06-16
Simplify message composer layout

jmartinesp:
## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


jmartinesp:
> On my device, I cannot have a RichTextEditor fullscreen as before, i.e. switch to RTE, open the bottom sheet at maximum and close the keyboard. I get this:

This is expected, since we now set a max height for the bottom sheet. This is how it works on iOS, so I hope it's good enough, but the code could probably be tweaked to get actual full screen though.

> Also in this state (bottom sheet expanded, and keyboard closed, I cannot open the keyboard. I need to reduce the bottom sheet first. Strange, but not really new apparently.

Oh, I hadn't noticed this. I'll see if I can reproduce it.

---

Now I see the issue: the problem is the actual backing `EditText` only reaches the same height as the text / existing lines, so tapping on the 'editor box' below where the text reaches does nothing - and it needs to be kept like this to make the intrinsic size measuring work. Maybe I can find a way to map this click event to requesting focus for the `EditText`.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/ExpandableBottomSheetLayout.kt
```diff
@@ -0,0 +1,279 @@
...
+// This is just for preview purposes
+@SuppressLint("AppCompatCustomView")
+private class PreviewEditText(context: Context) : EditText(context) {
+    override fun onAttachedToWindow() {
+        super.onAttachedToWindow()
+
+        parent?.requestDisallowInterceptTouchEvent(true)
+    }
+
+    override fun onTouchEvent(event: MotionEvent?): Boolean {
```
ElementBot, line 270: - :warning: Custom view `PreviewEditText` overrides `onTouchEvent` but not `performClick`

ganfra (approved):
Thanks, LGTM!

libraries/textcomposer/impl/src/main/kotlin/io/element/android/libraries/textcomposer/TextComposer.kt
```diff
@@ -493,20 +499,21 @@ private fun TextInputBox(
...
     Column(
-        modifier = Modifier
+        modifier = modifier
             .clip(roundedCorners)
             .border(0.5.dp, borderColor, roundedCorners)
             .background(color = bgColor)
             .requiredHeightIn(min = 42.dp)
-            .fillMaxSize(),
+            .fillMaxSize()
+            .then(modifier),
```
bmarty, line 516: @jmartinesp It's weird that `modifier` is used twice here

jmartinesp: Oh damn, that first usage should have been a Modifier instead.

jmartinesp: https://github.com/element-hq/element-x-android/pull/4928

PR #4889, opened 2025-06-17
Show generic notification when Event cannot be resolved

bmarty:
## Content

Ensure that when a Push is received and the application is unable to resolve the Event (network error, etc.) the application still shows a notification to the user.

This PR also contain a few more commit around Push, can be reviewed commit per commit

## Motivation and context

Less missed messages due to system blocking network request when the application is in background.

## Screenshots / GIFs

https://github.com/user-attachments/assets/b59f6697-a0ed-449c-96e1-8d903f04a0d4

## Tests

- Hard to repro the issue manually without patching the code.
- Apply the patch : [notifcationFallback.patch](https://github.com/user-attachments/files/20777880/notifcationFallback.patch)
- Run the application
- Receive a message
- Observe the fallback notification
- The battery optimization banner may also get displayed.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


libraries/push/impl/src/main/res/values/localazy.xml
```diff
@@ -13,7 +13,7 @@
     <item quantity="one">"%d notification"</item>
     <item quantity="other">"%d notifications"</item>
   </plurals>
-  <string name="notification_fallback_content">"Notification"</string>
+  <string name="notification_fallback_content">"You have new message(s)."</string>
```
jmartinesp: This sounds a bit odd to me, maybe a native speaker could suggest some alternative?

frebib: Seems fine to me

frebib: iirc WhatsApp says `You may have new messages` which is a less commited variant of the same thing

bmarty:
> I don't think we should display the battery optimisation banner in those cases

With the fallback notification, we will never show the banner then?

libraries/push/impl/src/main/res/values/localazy.xml
bmarty: @mxandreas / @amshakal any suggestion for this wording? I agree that `You may have new messages` is probably a better option.

ganfra:
> > I don't think we should display the battery optimisation banner in those cases
> 
> With the fallback notification, we will never show the banner then?

 Is it displayed only once? If the user dismiss it, this won't be shown again on the next fallback notification? This is what I failed to see in the PR, but maybe it's already the case?

bmarty:
Ah OK, yes, it's shown only once. If the user dismiss it (or disable battery optimization) it will not be shown again. If the user clear the cache, it can be shown again (see this [commit](https://github.com/element-hq/element-x-android/pull/4889/commits/017664f6728281e5c2653f1e6281f882f4dad1d2))

PR #4902, opened 2025-06-20
a11y: improve accessibility on grouped state events header.

bmarty:
## Content

a11y: improve accessibility on grouped state events header. Remove the custom a11y text we had and let the header be "toggleable", which will let the screen reader speak the new state out when it changes. The wording will be "checked" (for the expanded text) and "not checked" for the collapsed state. It seems that there is no way to tweak this wording. And there is no better `Role` for this use case.

## Motivation and context

Fix internal issue https://element-io.atlassian.net/browse/PSB-959

## Tests

- Enable screen reader
- Open a room with collapsed state event (Matrix-HQ is a good candidate)
- Observe how the application behaves when the grouped event header is focused and double tapped.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemGroupedEventsRow.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemGroupedEventsRow.kt
@@ -32,6 +32,7 @@ import io.element.android.libraries.designsystem.preview.ElementPreview
 import io.element.android.libraries.designsystem.preview.PreviewsDayNight
 import io.element.android.libraries.matrix.api.core.EventId
 import io.element.android.libraries.matrix.api.user.MatrixUser
+import io.element.android.libraries.ui.utils.time.isTalkbackActive
 import io.element.android.wysiwyg.link.Link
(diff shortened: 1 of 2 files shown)


jmartinesp (approved):
LGTM, thanks. 

I think I made everything a single item because some user said it was easier for an impaired user to just start listening to the whole group as a single item and skip to the next/previous one as soon as they're no longer interested in the contents instead of hearing the title, toggling, then moving through the sub-items one by one, etc. 

But if a client has explicitly asked us to revert it then it's fine, I guess they also did their own UX research.

bmarty:
> LGTM, thanks.
> 
> I think I made everything a single item because some user said it was easier for an impaired user to just start listening to the whole group as a single item and skip to the next/previous one as soon as they're no longer interested in the contents instead of hearing the title, toggling, then moving through the sub-items one by one, etc.
> 
> But if a client has explicitly asked us to revert it then it's fine, I guess they also did their own UX research.

I agree that the previous behavior was OK too. But also I guess it can be a bit annoying for impaired users to listen to a list (which can be pretty long) of state event changes. And with this change, the screen reader acts more like the "UI".

PR #4903, opened 2025-06-20
a11y: Make isTalkbackActive() live.

bmarty:
## Content

We made the assumption that the user needs to navigate to the setting to enable or disable the talkback, but there is a way to add a talkback switch on the bottom navigation bar. So the talkback can be enabled/disabled when the application is resumed. Since the UI may render differently depending on the talkback state, we need to make the composable `isTalkbackActive()` backed on a mutable state.

## Motivation and context

Ensure the UI is responsive regarading the talkback state.

## Screenshots / GIFs

https://github.com/user-attachments/assets/e51a2861-39cd-4f1e-8dd3-41bfaddc7eae

## Tests

- From the talkback (screen reader) setting, enable the shortcut. On my device, this is like this:
image
- open EXA
- open a timeline
- toggle the accessibility shortcut
- observe that the timeline is now getting reverse when the screen reader is enabled.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


ganfra (approved):
Good catch!

PR #4906, opened 2025-06-20
a11y: Add scripts to enable and disable the talkback service

jmartinesp:
## Content

What the title says.

## Motivation and context

It's better than going back through the command history or setting it manually on the test device 😅 .

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/tools/adb/disable_talkback.sh
+++ b/tools/adb/disable_talkback.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+# Copyright 2025 New Vector Ltd.
+#
+# SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+# Please see LICENSE files in the repository root for full details.
+
+adb shell settings put secure enabled_accessibility_services null
+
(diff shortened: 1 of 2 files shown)


tools/adb/disable_talkback.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+# Copyright 2025 New Vector Ltd.
+#
+# SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+# Please see LICENSE files in the repository root for full details.
+
+adb shell settings put secure enabled_accessibility_services com.android.talkback/com.google.android.marvin.talkback.TalkBackService
```
bmarty: This is like a trick to put a service that does not exist, right?
Maybe change the value to something more explicit like

`adb shell settings put secure enabled_accessibility_services null`

bmarty (approved):
Helpful, thanks!

tools/adb/disable_talkback.sh
jmartinesp: Good call, I thought I had tested that before and it hadn't worked, but I retried now and removing the setting works as expected.

PR #4908, opened 2025-06-20
Restore `MarkdownEditText.focusSearch` override

jmartinesp:
## Content

What the title says, we had this before and we removed it because we though the issue it prevented was gone.

It *might* not be needed if https://github.com/element-hq/element-x-android/pull/4884 is merged, but it's hard to say as it's not crashing consistently after the Compose 1.8 upgrade, as it used to do in Compose 1.7.x.

## Motivation and context

This prevents a crash when embedding the view in a Compose layout.

## Tests

I don't really know how to test this, since the issue can't be reproduced consistently anymore.

## Checklist


bmarty (approved):
Thanks!

PR #4919, opened 2025-06-23
Display error dialog if Element Call can't be joined

jmartinesp:
## Content

When on room call type, we listen for the 'join' action for 10s, and if we don't receive one we display an error dialog which will close the call screen.

## Motivation and context

This will allow us to remove the top app bar in a following PR.

## Tests

It's kind of difficult to test unless you add an artificial delay before receiving the join action when joining a call, but maybe it's possible by using an unstable network connection. In my case I just added the delay.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
@@ -44,13 +44,11 @@ import io.element.android.services.appnavstate.api.ActiveRoomsHolder
 import io.element.android.services.appnavstate.api.AppForegroundStateService
 import io.element.android.services.toolbox.api.systemclock.SystemClock
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
-import kotlinx.serialization.json.contentOrNull
-import kotlinx.serialization.json.jsonObject
-import kotlinx.serialization.json.jsonPrimitive
 import timber.log.Timber
 import java.util.UUID
 import kotlin.time.Duration.Companion.seconds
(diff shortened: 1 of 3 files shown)


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
```diff
@@ -145,12 +154,10 @@ class CallScreenPresenter @AssistedInject constructor(
                         if (parsedMessage?.direction == WidgetMessage.Direction.FromWidget) {
                             if (parsedMessage.action == WidgetMessage.Action.Close) {
                                 close(callWidgetDriver.value, navigator)
-                            } else if (parsedMessage.action == WidgetMessage.Action.SendEvent) {
-                                // This event is received when a member joins the call, the first one will be the current one
-                                val type = parsedMessage.data?.jsonObject?.get("type")?.jsonPrimitive?.contentOrNull
-                                if (type == "org.matrix.msc3401.call.member") {
-                                    isJoinedCall = true
-                                }
```
jmartinesp, line 153: According to the VOIP team, this should work the same. I didn't see any changes in behaviour when I tested it.

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
```diff
@@ -121,6 +121,15 @@ class CallScreenPresenter @AssistedInject constructor(
 
         callWidgetDriver.value?.let { driver ->
             LaunchedEffect(Unit) {
+                loadCallTimeoutJob = launch {
```
ganfra: Maybe we could just use : 
```
withTimeout(10){
driver.incomingMessages.firstOrNull {parsedMessage.action == WidgetMessage.Action.Join }
}
```
This would be a bit less convoluted

jmartinesp: I think I achieved something similar in https://github.com/element-hq/element-x-android/pull/4919/commits/2b504f295f678741d9608b4da92b3eb875737abe, just checking if the call is joined after the delay.

PR #4942, opened 2025-06-26
Try a workaround for the unstable call state in Maestro

jmartinesp:
## Content

Attempt to work around the issues we have with Maestro at the moment, so we can always tap on the start/join call button.

This seems to work, but then it seems like there is some issue with hanging up?

## Motivation and context

Fix the broken Maestro tests.

## Tests

Maestro should pass, or at least advance more.

## Checklist


jmartinesp:
Weird, from what I can see in the recordings the hang up button is clicked in the webview, but then nothing happens.

jmartinesp:
Closing this as it's not working.

Issue #4867, opened 2025-06-12
Message bubbles can be rendered almost empty until scrolled off-screen and back


rom4nik:
### Steps to reproduce

1. Upgrade app to v26.05.1 (that's when these issues started to appear, AFAIR)
2. Read messages in rooms until issue (randomly?) appears

#### What did you expect?
Messages to be shown correctly the first time.
#### What happened instead?

Event ID for affected message in video below: `$5p8DnhyR7teqonR2uQrlIYYt7b9GVoVvG0XGEIjsKhM`

https://github.com/user-attachments/assets/8f66f298-90e8-4160-be98-ee203afbebeb

And in video below: `$_MJ8W1XY0aXuFeqdLmcgOImyfy-qTNOMp3rjkeqUjvU`

https://github.com/user-attachments/assets/ca24619b-d477-4d3c-9fb5-bdd8f7a9595c

### Your phone model

Google Pixel Fold

### Operating system version

Android 15

### Application version and app store

v26.05.2

### Homeserver

Synapse v1.131.0

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

Destinyg133:
Can confirm, happend to me once, forgot to record it. Only very first part of message gets rendered in box rest is fixed once you re-scroll

jmartinesp:
It's probably a regression after upgrading Compose to v1.8.0. The size of the bubble is measured correctly but it seems like the contents are clipped to the min visible size 🤔 .

Destinyg133:
@jmartinesp still not fixed as of 25.07.0
Now half of the bubble was rendered and other half was missing
Can we please reopen this

rom4nik:
This fix isn't part of v25.07.0, it was landed just 3 days ago, meanwhile that version was released on July 4th.

rom4nik:
Unfortunately I'm still having this issue on v25.07.1, which contains fix from #4999.

jmartinesp:
There is another possible fix for this issue in https://github.com/element-hq/element-x-android/pull/5056, which should be included in tomorrow's nightly version.

rom4nik:
I haven't noticed this issue anymore on last few versions, thank you!

PR #4999, opened 2025-07-08
Make `ContentAvoidingLayoutData` an immutable data class

jmartinesp:
## Content

This *may* fix the issue causing the message contents to sometimes be clipped to a very tiny size while the message bubble has the right size. This was part of some optimisation we tried doing a long time ago and later removed as it didn't work as expected, and this data class wasn't reverted to the non-mutable version.

## Motivation and context

Try to fix https://github.com/element-hq/element-x-android/issues/4867, but it's difficult to check since I've never reproduced it.

## Tests

If you can reproduce the issue easily, scroll through the messages and check no messages have their contents clipped. Otherwise ¯\\_(ツ)\_/¯.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


bmarty (approved):
🤞

PR #5000, opened 2025-07-08
Remove unused composable and leanup colors

bmarty:
## Content

Remove unused BigCheckmark
Add preview for mention inside message bubble / composer. 

## Motivation and context

Cleanup codebase, less custom colors, more usefull screenshot

## Screenshots / GIFs

See recorded one

## Tests

- Should not have any impact.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/ColorAliases.kt
```diff
@@ -22,14 +22,14 @@ import kotlinx.collections.immutable.persistentMapOf
...
 
-@Composable
-fun ElementTheme.roomListRoomMessage() = colors.textSecondary
+val SemanticColors.roomListRoomMessage
+    get() = textSecondary
 
-@Composable
-fun ElementTheme.roomListRoomMessageDate() = colors.textSecondary
+val SemanticColors.roomListRoomMessageDate
+    get() = textSecondary
```
pixlwave, line 32: Random drive-by question out of curiosity… How come you're creating custom tokens that reference `textSecondary`, `iconAccentTeriary` and `bgSubtleSecondary`? Do you not just use the compound tokens directly?

bmarty: Yes, we could, here the idea of this file (ColorAliases) was to ensure that all the places where we have a room name in a room list use the same color. Now the added value is low, but it will be useful when implementing spaces maybe? I also created this file to ease customization of colors, but were not using it for that so far.

bmarty: (and maybe if one day we have no custom colors anymore I'll fully remove this file)

PR #5003, opened 2025-07-09
Home screen iteration

bmarty:
## Content

Apply blur effect on the top bar (only if Spaces flag is enabled).
Fix various layout issue.
Iterate on the top gradient.

## Motivation and context

Follow design files.

## Tests

- No behavior change expected

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


jmartinesp (approved):
Thanks for the fixes for the padding issues. 

I'm approving the PR, although if we want to enable the blur for these components we need to fix the issue that causes the blur to randomly disappear after a screen navigation and it leaves a view with no background instead before we enable this feature by default...

Issue #5002, opened 2025-07-09
Videos not respecting orientation


escix:
### Steps to reproduce

1. Where are you starting? What can you see?
Take a video or upload a video

#### What did you expect?
Videos to respect original video orientation

#### What happened instead?
All videos are uploaded in landscape orientation

### Your phone model

S23 Ultra

### Homeserver

selfhosted v1.132.0

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

bmarty:
Hello @escix , can you post in the issue a video with the problem so we can investigate? I have no problem uploading video with portrait orientation here. Thanks!

escix:
@bmarty 

This one:

https://github.com/user-attachments/assets/ac071517-42a9-4ff9-99d9-20e7fa9cb1ee

Interestingly this one (converted with video converted sometime back) uploaded in the correct orientation:

https://github.com/user-attachments/assets/1ec03c55-c9d8-4eba-85ce-d8192c3087b2

![Image](https://github.com/user-attachments/assets/19a165cb-27a1-4db1-92ef-0d6a01680b3a)

jmartinesp:
It's a metadata issue I think, the original video has a resolution of `1280x720 px`, which should render it in landscape, but then it also has a `Rotation: 90º` metadata field which the players will use to display it in portrait orientation instead. It seems like when we resize/transcode the video this metadata field is silently removed... instead we should either keep it or apply the rotation while transcoding, but I don't know how easy that would be given the library we use for that.

jmartinesp:
Actually, it seems like it would be quite simple to apply the rotation to the transcoded video. @bmarty do you want me to take this one over?

bmarty:
Yes, please, I was suspecting an issue regarding orientation metadata

PR #5008, opened 2025-07-10
Keep video rotation metadata when transcoding

jmartinesp:
## Content

Don't use the `PassthroughStrategy` when the video file contains a rotation metadata value.

## Motivation and context

The `PassthroughStrategy` used for video files to avoid doing any transcoding at all removes all the metadata for the video, removing also the tag used for describing the rotation of the video.

Fixes https://github.com/element-hq/element-x-android/issues/5002.

## Tests

Try sending the video mentioned in https://github.com/element-hq/element-x-android/issues/5002.

If the video is displayed in portrait orientation, it works.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/VideoCompressor.kt
+++ b/libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/VideoCompressor.kt
@@ -90,7 +90,7 @@ class VideoCompressor @Inject constructor(
                 val height = it.extractMetadata(MediaMetadataRetriever.METADATA_KEY_VIDEO_HEIGHT)?.toIntOrNull() ?: -1
                 val bitrate = it.extractMetadata(MediaMetadataRetriever.METADATA_KEY_BITRATE)?.toLongOrNull() ?: -1
                 val framerate = it.extractMetadata(MediaMetadataRetriever.METADATA_KEY_CAPTURE_FRAMERATE)?.toIntOrNull() ?: -1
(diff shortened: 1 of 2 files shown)


libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/VideoCompressor.kt
```diff
@@ -90,6 +90,7 @@ class VideoCompressor @Inject constructor(
                 val height = it.extractMetadata(MediaMetadataRetriever.METADATA_KEY_VIDEO_HEIGHT)?.toIntOrNull() ?: -1
                 val bitrate = it.extractMetadata(MediaMetadataRetriever.METADATA_KEY_BITRATE)?.toLongOrNull() ?: -1
                 val framerate = it.extractMetadata(MediaMetadataRetriever.METADATA_KEY_CAPTURE_FRAMERATE)?.toIntOrNull() ?: -1
+                val rotation = it.extractMetadata(MediaMetadataRetriever.METADATA_KEY_VIDEO_ROTATION)?.toIntOrNull()
```
bmarty: Should maybe default to `0` here? Else I think the later test (line 155 below) `rotation == 0` will fail.

Maybe all the fields of `VideoFileMetadata` could be non-nullable?

jmartinesp: Fixed in https://github.com/element-hq/element-x-android/pull/5008/commits/1ea6f5ba39cadedf803b0be2e10b93d02a7358c8

PR #5010, opened 2025-07-10
Add a feature flag to reuse the last `pos` value for initial syncs

jmartinesp:
## Content

What the title says. Instead of enabling it by default, we're making it configurable and disabled by default until we consider it's stable enough.

This should speed up initial syncs, but can have side effects so it's disabled by default.

The Rust SDK was upgraded to `v25.7.10` to be able to use the new option.

## Motivation and context

Enabling the 'share pos' feature by default has resulted in unstable sync for some users, so we're disabling it by default and allowing users to enable it manually for testing purposes.

## Tests

It's kind of difficult to check it's working without logs or a mitm proxy to check the `pos` value used in the initial sync.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
```diff
@@ -175,4 +175,13 @@ enum class FeatureFlags(
         // False so it's displayed in the developer options screen
         isFinished = false,
     ),
+    SharePos(
+        key = "feature.share_pos",
+        title = "Share pos",
```
ganfra: Maybe the title should include "Sliding sync" or "Sync"?

ganfra (approved):

libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/5010/commits/28e90d6fd547ed6caec4261ca4aa5fb5ed832e1a

PR #5039, opened 2025-07-17
Make sure we clean up the pre-processed and uploaded media

jmartinesp:
## Content

- Add `MediaSender.cleanUp()` and `MediaPreProcessor.cleanUp()` to remove temporary media files - results of the media processing - that have already been uploaded or whose upload was canceled.
- Fix back handler not being triggered on the attachments preview screen while pre-processing - it should only be disabled while uploading.

## Motivation and context

Fix an issue that caused temporary media to not be cleaned up after being used.

## Tests

A few unit tests were included, but if you can access your `/data/data/...` folder you can also check temp media files are deleted after being uploaded or when the upload is cancelled.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewView.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewView.kt
@@ -66,7 +66,7 @@ fun AttachmentsPreviewView(
         state.eventSink(AttachmentsPreviewEvents.CancelAndClearSendState)
     }
 
-    BackHandler(enabled = state.sendActionState !is SendActionState.Sending) {
+    BackHandler(enabled = state.sendActionState !is SendActionState.Sending.Uploading) {
         postCancel()
     }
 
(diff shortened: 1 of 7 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewView.kt
```diff
@@ -66,7 +66,7 @@ fun AttachmentsPreviewView(
         state.eventSink(AttachmentsPreviewEvents.CancelAndClearSendState)
     }
 
-    BackHandler(enabled = state.sendActionState !is SendActionState.Sending) {
+    BackHandler(enabled = state.sendActionState !is SendActionState.Sending.Uploading) {
```
ganfra, line 69: good catch!

libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/AndroidMediaPreProcessor.kt
```diff
@@ -100,6 +104,28 @@ class AndroidMediaPreProcessor @Inject constructor(
         }
     }.mapFailure { MediaPreProcessor.Failure(it) }
 
+    override fun cleanUp() {
+        // Clear temporary files created in older versions of the app
```
ganfra, line 108: Do we really want to handle this?

ganfra (approved):

libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/AndroidMediaPreProcessor.kt
jmartinesp: There might be a few hundred MBs there if the user has uploaded videos before 😬 .

Issue #5035, opened 2025-07-16
v25.07.1 Pre-release Video playback is overlapped with previous/next video


escix:
### Steps to reproduce

Video playback is overlapped with previous videos in the chat.

### Outcome

Upload a couple of videos and try to playback one.
The videos are overlapped.

![Image](https://github.com/user-attachments/assets/95da56a7-9a14-451b-af8e-3e8e49661c4c)

### Application version and app store

25.07.1

### Homeserver

selfhosted

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

bmarty:
This seems to be a regression due to https://github.com/element-hq/element-x-android/pull/4916. If I comment out the `zoomable` modifier [here](https://github.com/element-hq/element-x-android/blob/main/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/local/video/MediaVideoView.kt#L237), I do not have the issue anymore.

It seems that the video on the next page is zoomed in too much when it's loaded, and then is rendered in the current page.

rom4nik:
As reported by https://github.com/element-hq/element-x-android/pull/4916#issuecomment-3045937924 current video might start zoomed in as well, and afaict this happens (mostly? only?) when video has different aspect ratio than the screen.

In a situation where three 9:16 videos were sent one after another, opening video in the middle causes both preceding and following videos to cover it, in addition to all three being zoomed in. This is on Pixel Fold's cover screen which has 1080x2092 resolution, so a 9:17.4 aspect ratio.

vteraz:
I have the same issue on 25.07.1. Samsung Galaxy A54

jmartinesp:
This should be fixed by https://github.com/element-hq/element-x-android/pull/5055 and will be included in the next release. Please reopen the issue if it's not the case.

PR #5055, opened 2025-07-21
Remove the scaling added in `Player.Listener.onVideoSizeChanged`

jmartinesp:
## Content

Disable the video scaling, as it's probably conflicting with the zoomable state. This worsens the zoom-in experience a bit, but AFAICT fixes the issue with over-zoomed videos and 2 videos appearing at the same time in the media gallery.

## Motivation and context

Tentative fix for https://github.com/element-hq/element-x-android/issues/5035.

## Tests

In a room with several media, ideally with at least 2 videos uploaded after each other:

- Open the media gallery or one of the videos.
- Check the video is not larger than the free space (previously some videos, especially portrait ones, were over-scaled).
- When swiping left and right in the media gallery, there is no overlap between videos.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/local/video/MediaVideoView.kt
```diff
@@ -158,12 +157,7 @@ private fun ExoPlayerMediaVideoView(
             override fun onVideoSizeChanged(videoSize: VideoSize) {
                 // Ensure that the user cannot zoom/move outside of the video bounds
                 localMediaViewState.zoomableState.setContentLocation(
-                    ZoomableContentLocation.scaledInsideAndCenterAligned(
-                        Size(
-                            videoSize.width.toFloat(),
-                            videoSize.height.toFloat(),
-                        )
-                    )
+                    ZoomableContentLocation.SameAsLayoutBounds
```
ganfra: I think we can remove entirely the code here, as it's the default state already

PR #5056, opened 2025-07-21
Use lower level APIs to draw the message bubbles

jmartinesp:
## Content

Replace `Surface + clip & background modifiers` with drawing to the canvas. This should improve performance and simplify how the clipping is applied to the layout.

## Motivation and context

Replacing the modifier and making the clipping happen at the canvas level instead *could* be a fix for https://github.com/element-hq/element-x-android/issues/4867. I think the issue is somehow related to the cut-out part on the top-left corner (or top-right if you're on a RTL language) of the message bubble, as the clipping was applied to both the bubble and its contents.

## Screenshots / GIFs

There should be no screenshot test changes 🤞 .

## Tests

Since reproducing the original issue is quite difficult, I guess we should just check the 'normal' cases work as expected.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


jmartinesp:
Well, it seems like there are some changes in screenshots:

1. The rendering of the corners is probably not 100% the same, although I couldn't find any difference myself.
2. The text color for messages received from other users changes in the previews once we remove the `Surface` to the right one (darker, not light gray). It didn't seem to have any effect on the actual components when rendered in the app.

---

Changes in the rendering of the corners:

<details>
<summary>Screenshot</summary>
image</details>

tests/uitests/src/test/snapshots/images/features.messages.impl.pinned.list_PinnedMessagesListView_Day_3_en.png
jmartinesp, line 1: An example of the change in text color mentioned above.

tests/uitests/src/test/snapshots/images/features.messages.impl.pinned.list_PinnedMessagesListView_Night_3_en.png
jmartinesp, line 1: ... which does not happen on the dark version 😬

ganfra (approved):
LGTM thanks, lets see how it goes!

PR #5068, opened 2025-07-23
Calculate video output size taking into account portrait mode

jmartinesp:
## Content

Calculate video output size taking into account if the video is in portrait/landscape mode.

## Motivation and context

This is the right configuration, as not following this botches the compression for portrait videos, as well as their thumbnails. I don't know why it seemed to work well before, maybe we had this code but was lost in one of the many tests/rebases, or it interacted with some other code that fixed this behaviour but was later removed.

## Tests

Send a portrait video, both the thumbnail and the actual video should keep the portrait orientation.

Previously, they'd be placed into a landscape orientation with black borders on both sides.

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): 14, 16

## Checklist

### DIFF
--- a/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/bitmap/Bitmap.kt
+++ b/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/bitmap/Bitmap.kt
@@ -64,7 +64,7 @@ fun BitmapFactory.Options.calculateInSampleSize(desiredWidth: Int, desiredHeight
  * Decodes the [inputStream] into a [Bitmap] and applies the needed rotation based on [orientation].
  * This orientation value must be one of `ExifInterface.ORIENTATION_*` constants.
  */
-fun Bitmap.rotateToMetadataOrientation(orientation: Int): Bitmap {
+fun Bitmap.rotateToExifMetadataOrientation(orientation: Int): Bitmap {
     val matrix = Matrix()
     when (orientation) {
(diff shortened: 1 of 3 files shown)


libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/bitmap/Bitmap.kt
```diff
@@ -64,7 +64,7 @@ fun BitmapFactory.Options.calculateInSampleSize(desiredWidth: Int, desiredHeight
  * Decodes the [inputStream] into a [Bitmap] and applies the needed rotation based on [orientation].
  * This orientation value must be one of `ExifInterface.ORIENTATION_*` constants.
  */
-fun Bitmap.rotateToMetadataOrientation(orientation: Int): Bitmap {
+fun Bitmap.rotateToExifMetadataOrientation(orientation: Int): Bitmap {
```
jmartinesp, line 67: While I was experimenting with this, I realised this method takes an `Int` value here that could be mistaken for rotation degrees coming from `MediaMetadataRetriever`, so I renamed it to mention the `ExifMetadata` values which should be used instead.

ganfra (approved):

ganfra:
maestro failed :/

jmartinesp:
> curl: (28) Failed to connect to get.maestro.mobile.dev port 443 after 300579 ms: Timeout was reached

It couldn't install the CLI, so it failed. However, it didn't fail at that step 🤔 .

PR #5137, opened 2025-08-08
When mapping an invalid notification event, only drop that one

jmartinesp:
## Content

What the title says. I added a regression test too.

## Motivation and context

Previously, this meant the code processing the whole notification batch result failed and other notifications would be lost too.

## Tests

If the new unit test passes, it should be fine.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/fixtures/fakes/FakeFfiTimelineEvent.kt
+++ b/libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/fixtures/fakes/FakeFfiTimelineEvent.kt
@@ -14,7 +14,7 @@ import org.matrix.rustcomponents.sdk.NoPointer
 import org.matrix.rustcomponents.sdk.TimelineEvent
 import org.matrix.rustcomponents.sdk.TimelineEventType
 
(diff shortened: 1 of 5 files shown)


libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/fixtures/fakes/FakeFfiTimelineEvent.kt
```diff
@@ -14,7 +14,7 @@ import org.matrix.rustcomponents.sdk.NoPointer
 import org.matrix.rustcomponents.sdk.TimelineEvent
 import org.matrix.rustcomponents.sdk.TimelineEventType
 
-class FakeFfiTimelineEvent(
+open class FakeFfiTimelineEvent(
```
bmarty, line 17: An other option would have to replace 
`val timelineEventType: TimelineEventType = aRustTimelineEventTypeMessageLike()`
by something like 
`val timelineEventTypeResult: () -> TimelineEventType =  { aRustTimelineEventTypeMessageLike() }`

But OK for this shortcut.

bmarty (approved):
LGTM, thanks!

PR #5134, opened 2025-08-07
Hide the recovery key while we are entering it

andybalaam:
This is the Element X Android part of element-hq/element-meta#2888

 
## Content

As described in https://github.com/element-hq/element-meta/issues/2888, make the recovery key hidden when the user is typing it, and allow the user to show it if needed.

## Motivation and context

The recovery key is sensitive, so should be hidden by default.

## Screenshots / GIFs

|Before|After|
|
image
|
image
image
|
|||
 -->

## Tests

- Start fresh
- Log in 
- Choose Enter Recovery Key
- Note whether the recovery key is hidden and whether you can toggle it hidden/visible

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):
    - API 35

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [x] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly define what will change for the user
- [x] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #5147, opened 2025-08-11
Hide the recovery key while we are entering it

jmartinesp:
## Content

- Adds a toggle to hide/display the recovery key in the `RecoveryKeyView`, done by @andybalaam .
- Move the variable controlling the visibility to the state so we can modify it in the presenter, and use it to test the changes and add some new screenshots for UI validation.

## Motivation and context

Takes over after #5134 . This is the Element X Android part of element-hq/element-meta#2888.

## Screenshots / GIFs

|Before|After|
|-|-|
|||
|||

## Tests

Log in with a new account, choose to enter you recovery key. The toggle should be visible and the contents of the textfield hidden.

If you then access this screen from some other context, like creating a new recovery key, it should display the textfield contents.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


features/securebackup/impl/src/main/kotlin/io/element/android/features/securebackup/impl/tools/RecoveryKeyHiddenVisualTransformation.kt
```diff
@@ -0,0 +1,21 @@
...
+import androidx.compose.ui.text.input.PasswordVisualTransformation
+import androidx.compose.ui.text.input.TransformedText
+import androidx.compose.ui.text.input.VisualTransformation
+
+class RecoveryKeyHiddenVisualTransformation : VisualTransformation {
+    private var rc = RecoveryKeyVisualTransformation()
+    private var pw = PasswordVisualTransformation()
+
+    override fun filter(text: AnnotatedString): TransformedText =
+        rc.filter(pw.filter(text).text)
```
bmarty: Maybe not a big deal, but I think that this will always include some space between groups of 4 dots, even when the recovery key has space and so is a passphrase. Maybe this could be improved a bit to match the rendering when the chars are visible?

jmartinesp: The default behaviour was similar to this, but if the thing we're typing is hidden, should we really make this distinction? It would leak the text has one or more whitespaces.

bmarty: In this case I would just use a `PasswordVisualTransformation` without any spaces. Spaces was added to group random chars to let them a bit more readable by human (like a credit card number). For bullets this is not necessary I think.

jmartinesp: But the default state of this screen will be with the characters hidden, so unless the user toggles it they'll never see the recovery key grouped with an useful format. Since using a recovery key is more likely than using a passphrsae (I think) I added the key formatting by default when hidden.

jmartinesp: In the end and after discussing it with the iOS team, we'll use your suggestion or using a `PasswordVisualTransformation` . See https://github.com/element-hq/element-x-android/pull/5147/commits/ecabdcac77e7b68d2e51a9dc60d2ce410ec3244a.

Issue #5152, opened 2025-08-11
Joining a replacement room from a tombstone fails with generic "there was an error" due to no via= being sent


frebib:
To reproduce:
1. Tombstone a room and replace it with a v12 room
2. Try to join the replacement room in EX by clicking the "the conversation continues here" button.
3. Observe the generic error dialog in the client and this error in the homeserver logs:
`2025-08-11T21:20:40.017Z [INFO] synapse.http.server:133 - GET-21157 - <XForwardedForRequest at 0x5646912c6cf0 method='GET' uri='/_matrix/client/unstable/im.nheko.summary/rooms/!mauT12AzsoqxV7Abvy_ApA-HNPK1LcT4GbP70_AOPyQ/summary' clientproto='HTTP/1.0' site='unix'> SynapseError: 400 - Missing via to query remote room`

jmartinesp:
It seems like the problem happens when upgrading an `org.matrix.hydra.11` version room to version `12`. The SDK would previously fetch the domain name from the room id and use it for the `via` parameter as a fallback - which would usually work - but now it won't work anymore since there is no domain name in the room id.

frebib:
This presumably would also be an issue going from a v12 room to any other room version. I guess also it would be a problem if the room that created the server had gone away too

jmartinesp:
@frebib just to confirm: 

1. Were you using 2 different accounts for the one that upgraded the room and the one that tried to open the successor one?
2. If you were, was the room public or this 2nd account invited to the new room?

frebib:
> 1. Were you using 2 different accounts for the one that upgraded the room and the one that tried to open the successor one?

In this case, I didn't upgrade the room. It was #test:maunium.net

> 2. If you were, was the room public or this 2nd account invited to the new room?

I don't believe I was invited:
```
{
  "content": {
    "avatar_url": "mxc:\/\/nerdhouse.io\/ilZklVAylseRtariMNFFlbfx",
    "displayname": "frebib",
    "membership": "join"
  },
  "event_id": "$XKK3cshD5gdbQgpfWaSTInj5cSZIfCWNiKHvVqBAHHs",
  "origin_server_ts": 1754947428366,
  "sender": "@frebib:nerdhouse.io",
  "state_key": "@frebib:nerdhouse.io",
  "type": "m.room.member",
  "unsigned": {
    "age": 3,
    "membership": "join"
  }
}
```

jmartinesp:
[...]

PR #5155, opened 2025-08-12
Provide calculated server names when opening a room from another

jmartinesp:
## Content

When opening a room from an already displayed one, calculate the server names used for asking the HS for its info using the domain names from the user ids from the users in that room.

## Motivation and context

Fixes or workarounds https://github.com/element-hq/element-x-android/issues/5152. This is needed because the homeserver can no longer use the domain part of the room id as the fallback value for this since it won't be available in room versions >= `12`.

This helps when following permalinks or navigating to the successor room of a tombstoned one. Previously, the `/summary` endpoint was failing because no server names were used in the `via` parameters.

## Tests

- Create a room with version `org.matrix.hydra.11` or `12`.
- Have some other user of yours **in another HS** join it.
- Upgrade the room to `12` again.
- Navigate to the successor room from the user who joined the room.

If it works as expected, displaying the invite and the preview info, it's fixed.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/core/UserId.kt
```diff
@@ -29,4 +29,7 @@ value class UserId(val value: String) : Serializable {
         get() = value
             .removePrefix("@")
             .substringBefore(":")
+
+    val domainName: String?
```
bmarty, line 33:
```suggestion
    val domainName: String
```

libraries/core/src/main/kotlin/io/element/android/libraries/core/data/TakeAtMost.kt
```diff
@@ -0,0 +1,22 @@
...
+ * Returns a collection containing at most [maxSize] elements from the original collection.
+ * If the original collection has fewer than [maxSize] elements, it returns all of them.
+ * If [maxSize] is less than or equal to 0, it returns an empty collection.
+ */
+fun <T> Collection<T>.takeAtMost(maxSize: Int): Collection<T> {
+    val maxLength = min(maxSize, this.size)
+    if (maxLength <= 0) return emptyList()
+
+    return take(maxLength)
+}
```
bmarty: I do not see the added value compare to just using `take()`. If this is returning an empty list instaed of throwing an exception, this is not used in your usage `.takeAtMost(1_000)`. What am I missing?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
```diff
@@ -179,7 +180,14 @@ class TimelinePresenter @AssistedInject constructor(
                     resolveVerifiedUserSendFailureState.eventSink(ResolveVerifiedUserSendFailureEvents.ComputeForMessage(event.event))
                 }
                 is TimelineEvents.NavigateToRoom -> {
-                    navigator.onNavigateToRoom(event.roomId)
+                    // Workaround for not having the server names available, get possible server names from the user ids of the room members
```
bmarty: I was confused because I thought that `TimelineEvents.NavigateToRoom` was use to navigate to any roomId, but it is actually only used to navigate to the predecessor and to the sucessor room. Maybe rename to `TimelineEvents.NavigateToPredecessorOrSuccessorRoom`? (I cannot think of a common name to reflect the idea of Predecessor and Successor)

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
```diff
@@ -179,7 +180,14 @@ class TimelinePresenter @AssistedInject constructor(
...
+                    val serverNames = room.membersStateFlow.value.roomMembers()
+                        .orEmpty()
+                        .takeAtMost(1_000)
+                        .mapNotNull { it.userId.domainName }
+                        .distinct()
+
+                    navigator.onNavigateToRoom(event.roomId, serverNames)
```
bmarty, line 184: The algorithm to get the `via` parameter should take care of the number of user per domain.
This algorithm could be a bit better:

```kotlin
                    val serverNames = room.membersStateFlow.value.roomMembers()
                        .orEmpty()
                        .map { it.userId.domainName }
                        .groupingBy { it }
                        .eachCount()
                        .toMutableMap()
                        .apply { 
                            // Ensure the current user's server name is included
                            this[room.sessionId.domainName] = Int.MAX_VALUE
                        }
                        .let { map ->
                            map.keys.sortedByDescending { map[it] }
                        }
                        .take(3)
```

See the kotlin SDK impl: https://github.com/element-hq/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/permalinks/ViaParameterFinder.kt#L53

bmarty (approved):
LGTM, thanks for the fix!
A few suggestions though.

libraries/core/src/main/kotlin/io/element/android/libraries/core/data/TakeAtMost.kt
jmartinesp: Why is it not used? If the member list has 10 users, `members.take(1_000)` would crash, while `members.takeAtMost(1_000)` would return those 10 users. I added it to limit the case where we have rooms with hundreds of thousands of users (aka Matrix HQ).

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/core/UserId.kt
jmartinesp: I think I'd rather leave it optional and add the missing `.takeIf { it.isNotEmpty() }`.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
jmartinesp: Maybe `NavigateToRelatedRoom` with some doc comment? Or I can use the `NavigateToPredecessorOrSuccessorRoom` suggestion.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
jmartinesp: Do you think it's worth it to remove the limit on how may members to use here? It feels like for huge public rooms it may affect performance a bit, but maybe I'm overthinking it.

libraries/core/src/main/kotlin/io/element/android/libraries/core/data/TakeAtMost.kt
bmarty: > members.take(1_000) would crash

no, it will just return the 10 members.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
bmarty: Maybe `NavigateToLinkedRoom`?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
bmarty: I have tested on large public room and `room.membersStateFlow.value.roomMembers()` returned only 13 members anyway :).

I do not think we should bother about performance here TBF.

libraries/core/src/main/kotlin/io/element/android/libraries/core/data/TakeAtMost.kt
jmartinesp: Oh, you're right. As I said, slow brain day. Sigh.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
jmartinesp: That could be a room linked using a permalink 😅 . Maybe `NavigateToPredecessorOrSuccessorRoom` is the best option even if it's super verbose.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
jmartinesp: I think it depends on whether those users were previously loaded or not. When the timeline is loaded, we start fetching the members, which in very large rooms can take several seconds, even 10-20s. When that is finished, `roomMembers()` should return all of them if I'm not mistaken.

jmartinesp: In example, in `Matrix HQ` after loading 133k members, getting the server names took `1.337s` with no limit.

```
elementx: ## Navigate to predecessor or successor room, serverNames: [matrix.org, libera.chat, linuxdelta.com], elapsed: 1337 ms | TimelinePresenter.kt:378
```

With just 1000 results I got similar server names and it took `37ms` instead, but I guess the proposed server names aren't as accurate:

```
elementx: ## Navigate to predecessor or successor room, serverNames: [matrix.org, libera.chat, half-shot.uk], elapsed: 37 ms | TimelinePresenter.kt:378
```

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
bmarty: `NavigateToAffiliatedRoom` ?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
bmarty: ok, let's keep the limit then!

jmartinesp: After some changes I got it to ~350ms which is way better, so I think I'll remove the limit until someone complains in the future.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
```diff
@@ -353,3 +355,19 @@ private fun FocusRequestState.onFocusEventRender(): FocusRequestState {
         else -> this
     }
 }
+
+// Workaround for not having the server names available, get possible server names from the user ids of the room members
+private fun calculateServerNamesForRoom(room: JoinedRoom): List<String> {
+    // If we have no room members, return right ahead
+    val serverNames = room.membersStateFlow.value.roomMembers() ?: return emptyList()
```
bmarty, line 362: The variable name is not really correct

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/core/UserId.kt
bmarty: It *should* never be empty, but OK.

PR #5157, opened 2025-08-12
Do not automatically initialize `DefaultVideoMetadataExtractor`'s data source

jmartinesp:
## Content

Make sure we don't automatically set the data source for `DefaultVideoMetadataExtractor`'s `MediaMetadataRetriever`: this will cause a crash for image attachments.

## Motivation and context

Fix the crash.

## Tests

On any room, select an image to send. If the preview screen is displayed, it's fixed.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/video/VideoMetadataExtractor.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/video/VideoMetadataExtractor.kt
@@ -29,7 +29,7 @@ interface VideoMetadataExtractor : AutoCloseable {
 
 @ContributesBinding(AppScope::class)
 class DefaultVideoMetadataExtractor @AssistedInject constructor(
-    @ApplicationContext context: Context,
+    @ApplicationContext private val context: Context,
     @Assisted private val uri: Uri,
 ) : VideoMetadataExtractor {
     @ContributesBinding(AppScope::class)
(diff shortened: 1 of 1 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/video/VideoMetadataExtractor.kt
```diff
@@ -38,13 +39,18 @@ class DefaultVideoMetadataExtractor @AssistedInject constructor(
         override fun create(uri: Uri): DefaultVideoMetadataExtractor
     }
 
+    private val isDataSourceReady = AtomicBoolean(false)
+
     private val mediaMetadataRetriever = MediaMetadataRetriever()
 
-    init {
-        mediaMetadataRetriever.setDataSource(context, uri)
+    private fun ensureDataSourceIsReady() {
```
bmarty: Maybe another simpler fix would be to lazy create the MediaMetaDataRetriever?
```kotlin
    private val mediaMetadataRetriever by lazy {
        MediaMetadataRetriever().also {
            it.setDataSource(context, uri)
        }
    }
```
Or the crash do not occur anymore since the exception is caught by `runCatchingExceptions`? In this case I am not sure that my suggestion will work the same way.

bmarty (approved):
Thanks!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/video/VideoMetadataExtractor.kt
jmartinesp: That should work too. In fact, I wanted to make it lazy but I don't know why I discarded the idea 😅 . 

I guess I'm in a 'slow brain' day. I'll try your suggestion and update the code if it works fine.

jmartinesp: Well, it works in the sense that it doesn't crash anymore, but if we called one of those methods by mistake, the exception wouldn't be caught by `runCatchingExceptions`.

jmartinesp: Ok, if we don't use `by lazy` but `= lazy` instead it does work as expected, trying to re-initialize the value and throwing the exception in the calling code.

PR #5160, opened 2025-08-12
Remove old feature flags

### LINKED ISSUE #5154: Clean up feature flags
When we are actively developing a feature, we start by creating a feature flag so that users do not have access to the WIP. The list of feature flags has increased a lot and we have a way to make the flag not visible in the developer setting screen, which is used. But now we want to go further and fully remove the flags.

Here is the list of flags, with the current state and the action to do on it.

## Flags to remove:

- `LocationSharing`
- `Polls`
- `NotificationSettings`
- `VoiceMessages`
- `PinUnlock`
- `MarkAsUnread`
- `QrCodeLogin`
- `IncomingShare`
- `PinnedEvents`
- `SyncOnPush`
- `MediaUploadOnSendQueue`
- `MediaCaptionCreation` 
- `MediaCaptionWarning`
- `MediaGallery`

## Flags to keep

- `RoomDirectorySearch`
- `ShowBlockedUsersDetails`
- `OnlySignedDeviceIsolationMode`
- `EnableKeyShareOnInvite`
- `Space` -> Keep (WIP)
- `PrintLogsToLogcat`
- `SelectableMediaQuality` -> it has been added recently.
- `SharePos` -> it's been enabled recently

## Others

- `Knock` -> Disabled by default. Enable and remove?

bmarty:
## Content

Remove a bunch of feature flags.
Can be reviewed commit per commit.

<del>Draft as I spot a regression in the composer preview.</del> There is actually no regression, the current behavior is the same. When adding a caption, the send voice message button is displayed. We should fix that separately. (I think there is an existing issue about it, but GH is broken now, so I cannot find it).

## Motivation and context

Closes #5154

## Tests

- There should not be any change in the default application behavior.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


app/src/main/AndroidManifest.xml
```diff
@@ -147,14 +147,6 @@
...
                 <data android:host="room" />
             </intent-filter>
-        </activity>
-
-        <!-- Using an activity-alias for incoming share intent, in order
-        to be able to disable the feature programmatically -->
-        <activity-alias
-            android:name=".ShareActivity"
-            android:exported="true"
-            android:targetActivity=".MainActivity">
```
jmartinesp, line 157: Are we sure we want to remove this one? I think we wanted to keep it for a certain customer, but I'm not 100% sure.

libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
```diff
@@ -34,13 +34,6 @@ enum class FeatureFlags(
         defaultValue = { false },
         isFinished = false,
     ),
-    SyncOnPush(
-        key = "feature.syncOnPush",
-        title = "Sync on push",
-        description = "Subscribe to room sync when a push is received",
-        defaultValue = { true },
-        isFinished = false,
-    ),
```
jmartinesp: We might want to keep this one for testing, it's a bit difficult to track what's retrieved by the notification client if we perform another sync at the same time.

app/src/main/AndroidManifest.xml
bmarty: I removed the flag since I think that disabling incoming share does not need to be a feature flag. If a customer maintaining a fork wants to disable the feature, they just have to remove the `intent-filter`s on their fork.

libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
bmarty: OK, reverted on f99b6f30c4c8a582287ac2b22a645695074c3d43

PR #5179, opened 2025-08-18
Update Compose BOM to version 2025.12.00.

bmarty:
## Content

Upgrade to latest compose version

## Motivation and context

Be up to date

## Tests

- No expected change, let see the screenshot test result.
- Smoke test <del>in progress</del> OK

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/pinned/list/PinnedMessagesListNode.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/pinned/list/PinnedMessagesListNode.kt
@@ -15,6 +15,7 @@ import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalView
+import androidx.compose.ui.res.stringResource
 import com.bumble.appyx.core.modality.BuildContext
 import com.bumble.appyx.core.node.Node
 import com.bumble.appyx.core.plugin.Plugin
@@ -100,6 +101,7 @@ class PinnedMessagesListNode(
             LocalTimelineItemPresenterFactories provides timelineItemPresenterFactories,
         ) {
             val context = LocalContext.current
+            val toastMessage = stringResource(CommonStrings.common_copied_to_clipboard)
             val view = LocalView.current
             val state = presenter.present()
             PinnedMessagesListView(
@@ -113,8 +115,8 @@ class PinnedMessagesListNode(
                         HapticFeedbackConstants.LONG_PRESS
                     )
                     context.copyToClipboard(
-                        it.url,
-                        context.getString(CommonStrings.common_copied_to_clipboard)
+                        text = it.url,
+                        toastMessage = toastMessage,
                     )
                 },
                 modifier = modifier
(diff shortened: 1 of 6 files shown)


jmartinesp:
There seem to be some diffs in the screenshots, should we generate new ones?

jmartinesp:
It seems like this version broke <del>all</del> the dialog previews in screens using `Scaffold` 🫤 .

jmartinesp:
I opened an issue in Paparazzi's repo: https://github.com/cashapp/paparazzi/issues/2071, maybe it's Compose, maybe it's a combination of that and the alpha version.

bmarty:
It also seems to affect the preview in Android Studio, see for instance this hack, where the dialog is rendered only on the first preview:

image

jmartinesp:
> It also seems to affect the preview in Android Studio, see for instance this hack, where the dialog is rendered only on the first preview:

Could it be an IDE cache issue? I tried using the same preview and the dialog isn't displayed in any preview for me:

<details>
<summary>Screenshot</summary>
image
</details>

jmartinesp:
> It also seems to affect the preview in Android Studio, see for instance this hack, where the dialog is rendered only on the first preview:

I can reproduce it.

Maybe it's because of [this commit](https://cs.android.com/androidx/platform/frameworks/support/+/8eeb97904f2691906a4a3b7cb679478922ff2c1e:compose/ui/ui/src/androidMain/kotlin/androidx/compose/ui/window/AndroidDialog.android.kt;dlc=30ef14cc8a8d60c925eb7326e1abc04dee1714fb)? It's the latest one added to `AndroidDialog.android.kt`, and it affects the measured size of the dialog.

---

It started happening on [1.9.0-alpha01](https://developer.android.com/jetpack/androidx/releases/compose-ui#1.9.0-alpha01), so any of the changes added in that version could have caused it.

jmartinesp:
Just FYI, the Compose BOM `2025.08.01` still has this issue. I created [an issue](https://issuetracker.google.com/issues/439560299) in the Google Issue tracker, but they moved it to a different category and is no longer visible for public users (not even for me!).

---

I was hoping [this commit](https://android.googlesource.com/platform/frameworks/support/+/c6ff220e4785a6b03d703b4b037d0422b34475a1) in the latest alpha version would fix it, assuming it restored the previous behaviour, but it doesn't.

---

Oh, maybe it's [this one instead](https://android.googlesource.com/platform/frameworks/support/+/4628d30a7ea04c2c8788a642385ce3232eb58de9%5E%21/#F0), in which case there's not much we can do 🫤 .

jmartinesp:
> Oh, maybe it's [this one instead](https://android.googlesource.com/platform/frameworks/support/+/4628d30a7ea04c2c8788a642385ce3232eb58de9%5E%21/#F0), in which case there's not much we can do 🫤 .

I can confirm that's it, sadly. I created a copy of `Dialog` with the previous `DisposableEffect(dialog) { dialog.show() }` instead of `LaunchedEffect(dialog) { dialog.show() }` and the previews in the IDE are working again.

So there's not much we can do about it, that's how the Compose component will behave from now on. They could probably fix it on their end by changing which effect to use based on `LocalInspectionMode.current`, but if they don't we're stuck with the new behaviour.

bmarty:
Regression tracked by https://issuetracker.google.com/issues/442273708

jmartinesp:
The dialog issues seem to be fixed on the AS Narwhal RC versions <del>and the current Paparazzi 2.0.0-SNAPSHOT</del>. I guess we'll need to wait until the Paparazzi releases a new RC, by that time we should also have an stable AS version with the fix for previews.

PR #5116, opened 2025-08-04
Report shortcut usage for outgoing messages

networkException:
## Content

This pull request adds support for creating and pushing dynamic long-lived shortcuts for outgoing messages. This together with an existing reference to the roomId used by the shortcuts as an identifer allows conversations to be prioritized.

See https://developer.android.com/training/sharing/direct-share-targets#report-usage-outgoing

## Motivation and context

See #1547 and #1586

## Screenshots / GIFs

image

image

## Tests

- Go to a room
- Send a message
- Leave the app and long press on the launcher icon
- See a shortcut for the recent message

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [x] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly define what will change for the user
- [x] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #5180, opened 2025-08-18
Add shortcut suggestions for rooms, remove then when leaving

jmartinesp:
## Content

Continuation of https://github.com/element-hq/element-x-android/pull/5116.

Added the following changes:

- `InitialsAvatarBitmapGenerator` to create the initial avatars icon for those rooms that don't have an actual avatar - previously it just used a blank icon.
- Remove shortcuts when the room is left locally or from some other session. Note: this means removing the shortcut from the list of suggested shortcuts displayed when long pressing in the app's icon, not from the home screen of your device.

## Motivation and context

Iterate on https://github.com/element-hq/element-x-android/pull/5116.

## Screenshots / GIFs

image

## Tests

- Send a message in a room.
- Go to your home screen, long press on the app's icon. The room should appear as an available shortcut.
- If you create a shortcut for it in your home screen, it'll open the right room.
- Now leave the room either from the current session or a different one - if you leave from a different one, bring the app to foreground once, then go back to the home screen.
- The shortcut should still be present in the home screen, but the suggestion for it should now be gone from the long press context menu on the app's icon.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -215,6 +217,12 @@ class LoggedInFlowNode @AssistedInject constructor(
                     }
                     .launchIn(lifecycleScope)
             },
+            onResume = {
+                lifecycleScope.launch {
+                    val availableRoomIds = matrixClient.getJoinedRoomIds().getOrNull() ?: return@launch
+                    notificationConversationService.onAvailableRoomsChanged(sessionId = matrixClient.sessionId, roomIds = availableRoomIds)
+                }
+            },
```
jmartinesp, line 216: I'm not really convinced of adding it here, but the alternative would be to trigger it when the room list updates which is way too often IMO... maybe there's a better middle ground.

features/leaveroom/impl/src/main/kotlin/io/element/android/features/leaveroom/impl/LeaveRoomPresenter.kt
```diff
@@ -78,6 +80,7 @@ class LeaveRoomPresenter @Inject constructor(
             client.getRoom(roomId)!!.use { room ->
                 room
                     .leave()
+                    .onSuccess { notificationConversationService.onLeftRoom(client.sessionId, roomId) }
```
bmarty, line 83: Why not let `NotificationConversationService` implementation use `RoomMembershipObserver`?

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/conversations/DefaultNotificationConversationService.kt
```diff
@@ -0,0 +1,100 @@
...
+        val categories = setOfNotNull(
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) ShortcutInfo.SHORTCUT_CATEGORY_CONVERSATION else null
+        )
+
+        val client = matrixClientProvider.getOrRestore(sessionId).getOrNull() ?: return
+        val imageLoader = imageLoaderHolder.get(client)
+
+        val defaultShortcutIconSize = ShortcutManagerCompat.getIconMaxWidth(context)
+        val useDarkTheme = context.resources.configuration.uiMode and Configuration.UI_MODE_NIGHT_MASK == Configuration.UI_MODE_NIGHT_YES
+        val icon = bitmapLoader.getRoomBitmap(roomAvatarUrl, imageLoader)?.let(IconCompat::createWithBitmap)
```
bmarty: On my device, I got an extra padding on the generated image:

image

Using `createWithAdaptiveBitmap` instead of `createWithBitmap` does not fully solve the issue as the icon is now too big:

image

I also try to use `getExtraInsetFraction`, but the result is the same

```kotlin
        val extraInsetFraction = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
            AdaptiveIconDrawable.getExtraInsetFraction()
        } else {
            0f
        }
        val defaultShortcutIconSize = (ShortcutManagerCompat.getIconMaxWidth(context) * (1 + 2 * extraInsetFraction)).toInt()
```

libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/media/InitialsAvatarBitmapGenerator.kt
```diff
@@ -0,0 +1,131 @@
...
+        val bitmap = remember(isLightTheme) {
+            val generator = InitialsAvatarBitmapGenerator(useDarkTheme = !isLightTheme)
+            generator.generateBitmap(512, avatarData)?.asImageBitmap()
+        }
+
+        bitmap?.let {
+            Image(bitmap = it, contentDescription = "Initials Avatar", modifier = Modifier.size(48.dp))
+        } ?: Text("No avatar generated")
+    }
+}
```
bmarty, line 137: Maybe improve the Preview by rendering all the possible colors. Something like:

```kotlin
@Composable
@PreviewsDayNight
internal fun InitialsAvatarBitmapGeneratorPreview() = ElementPreview {
    Column(
        verticalArrangement = Arrangement.spacedBy(4.dp)
    ) {
        repeat(6) { index ->
            val avatarData = remember { AvatarData(id = index.toString(), name = Char('0'.code + index).toString(), size = AvatarSize.IncomingCall) }
            val isLightTheme = ElementTheme.isLightTheme
            val bitmap = remember(isLightTheme) {
                val generator = InitialsAvatarBitmapGenerator(useDarkTheme = !isLightTheme)
                generator.generateBitmap(512, avatarData)?.asImageBitmap()
            }

            bitmap?.let {
                Image(bitmap = it, contentDescription = null, modifier = Modifier.size(48.dp))
            } ?: Text("No avatar generated")
        }
    }
}
```

This will update the preview to:

image

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/conversations/DefaultNotificationConversationService.kt
```diff
@@ -0,0 +1,100 @@
...
+            ?: InitialsAvatarBitmapGenerator(useDarkTheme = useDarkTheme)
+                .generateBitmap(defaultShortcutIconSize, AvatarData(id = roomId.value, name = roomName, size = AvatarSize.RoomHeader))
+                ?.let(IconCompat::createWithAdaptiveBitmap)
+
+        val shortcutInfo = ShortcutInfoCompat.Builder(context, "$sessionId-$roomId")
+            .setShortLabel(roomName)
+            .setIcon(icon)
+            .setIntent(intentProvider.getViewRoomIntent(sessionId, roomId, threadId = null))
```
bmarty, line 111: Use `threadId` here? Or maybe just remove it from the API?

```suggestion
            .setIntent(intentProvider.getViewRoomIntent(sessionId, roomId, threadId = threadId))
```

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/conversations/DefaultNotificationConversationService.kt
```diff
@@ -0,0 +1,100 @@
...
+            .setLongLived(true)
+            .let {
+                when (roomIsDirect) {
+                    true -> it.addCapabilityBinding("actions.intent.SEND_MESSAGE")
+                    false -> it.addCapabilityBinding("actions.intent.SEND_MESSAGE", "message.recipient.@type", listOf("Audience"))
+                }
+            }
+            .build()
+
+        ShortcutManagerCompat.pushDynamicShortcut(context, shortcutInfo)
```
bmarty: Maybe safer to protect all methods from `ShortcutManaerCompat` with `runCatchingExceptions`, as we had some issue on Element Classic regarding this.

bmarty (changes requested):
The current Element Classic [implementation](https://github.com/element-hq/element-android/blob/main/vector/src/main/java/im/vector/app/features/home/ShortcutsHandler.kt#L46) does not set any shortcut if a pin code is set up, for privacy reason. Maybe we should mimic the same behavior on EX. Also maybe a good idea to check the Element Classic implementation regarding `isRequestPinShortcutSupported` and `maxShortcutCountPerActivity`, and see if it could aplpy in EX.

Last point, when the user signs out, the shortcut of the session should be removed.

features/leaveroom/impl/src/main/kotlin/io/element/android/features/leaveroom/impl/LeaveRoomPresenter.kt
jmartinesp: That only takes into account leaving the room locally if I'm not mistaken.

jmartinesp:
Is the usage of `maxShortcutCountPerActivity` really necessary? The current implementation is simpler than the one in the legacy app, instead of publishing the most active rooms we're just using the most recently used ones, since we still don't have a way to track which rooms are used the most.

With the current approach just pushing the latest one removes the oldest entry from the list, which I'd say it's fine.

libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/media/InitialsAvatarBitmapGenerator.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/5180/commits/3ed00166b7bc726dd89de649da10be935bd2e2c1

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/conversations/DefaultNotificationConversationService.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/5180/commits/37172f097b63d025c11cc6ff366abb0f78d81463

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/conversations/DefaultNotificationConversationService.kt
jmartinesp: I hope https://github.com/element-hq/element-x-android/pull/5180/commits/e62f6693c616ed053f5f6d57e879fc8c67a77ea3 fixes it, but I'm not sure since I can't really reproduce it 🫤 .

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/conversations/DefaultNotificationConversationService.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/5180/commits/462031ab0fac813e0e3f3960333952c591263a53

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/conversations/DefaultNotificationConversationService.kt
jmartinesp: With the latest changes I see this in my device:

image

However, since different resolution and screen density could affect this I'm not sure it's fixed.

bmarty:
> With the current approach just pushing the latest one removes the oldest entry from the list, which I'd say it's fine.

IIRC it can depend on device implementation. But we can try this and see if issues are reported.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/conversations/DefaultNotificationConversationService.kt
```diff
@@ -0,0 +1,195 @@
...
+    private val matrixClientProvider: MatrixClientProvider,
+    private val imageLoaderHolder: ImageLoaderHolder,
+    private val lockScreenService: LockScreenService,
+    sessionObserver: SessionObserver,
+    @AppCoroutineScope private val coroutineScope: CoroutineScope,
+) : NotificationConversationService {
+    private val isRequestPinShortcutSupported = ShortcutManagerCompat.isRequestPinShortcutSupported(context)
+
+    init {
+        sessionObserver.addListener(object : SessionListener {
```
bmarty, line 60: With this, I think that `DefaultNotificationConversationService` should have a `@SingleIn` annotation.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/conversations/DefaultNotificationConversationService.kt
bmarty: This is not better:

image

I am using a classic API 36 emulator if it helps. Can also be improved later, it's not really blocking.

bmarty (approved):
Just one last issue (about adding `@SingleIn`), then I think that we can merge it!

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/conversations/DefaultNotificationConversationService.kt
jmartinesp: Good news is I can reproduce it on an emulator, as you suggested!
Bad news is I have no idea how to fix it yet 😅 .

frebib: If it makes you feel any better, the Signal app also has this same issue on my phone, as well as EX

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/conversations/DefaultNotificationConversationService.kt
```diff
@@ -49,15 +50,17 @@ class DefaultNotificationConversationService @Inject constructor(
...
-            ?: InitialsAvatarBitmapGenerator()
-                .generateBitmap(512, AvatarData(id = roomId.value, name = roomName, size = AvatarSize.RoomHeader))?.let(
-                IconCompat::createWithAdaptiveBitmap
-            )
+            ?: InitialsAvatarBitmapGenerator(useDarkTheme = useDarkTheme)
+                .generateBitmap(defaultShortcutIconSize, AvatarData(id = roomId.value, name = roomName, size = AvatarSize.RoomHeader))
+                ?.let(IconCompat::createWithAdaptiveBitmap)
 
-        val shortcutInfo = ShortcutInfoCompat.Builder(context, roomId.value)
+        val shortcutInfo = ShortcutInfoCompat.Builder(context, "$sessionId-$roomId")
```
networkException, line 108: Unfortunately I only noticed this change now, this breaks the prioritization of notifications as the ids haven't been adjusted in the notifications itself. See

https://github.com/element-hq/element-x-android/blob/2fbfd25fd169c322d8d4cf376233f588bde23661/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt#L139

networkException: This can be tested using `adb shell dumpsys notification --noredact` and looking for `shortcut=` where in `<id> found valid? <boolean>` `id` should match a registered shortcut in `adb shell cmd shortcut get-shortcuts io.element.android.x` and `boolean` should be true.

frebib: Does https://github.com/element-hq/element-x-android/pull/5192 fix that?

networkException: Yes

PR #5131, opened 2025-08-07
Add media file limit size warning and media quality selection

jmartinesp:
## Content

- Load the home server's max file upload size, and use it to calculate which files can or can't be uploaded, when possible.
- Add `VideoCompressionPreset` and use it instead of the boolean `compressMedia`.
- Add feature flag to enable selecting the media quality per media upload or keep the current behaviour.
- Allow selecting the default video upload quality to use in advanced settings.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/5072, implements the Android part of https://github.com/element-hq/element-meta/issues/2916, https://github.com/element-hq/element-meta/issues/2917 and https://github.com/element-hq/element-meta/issues/2915. 

## Screenshots / GIFs

There are few new screenshots in the PR.

## Tests

- Try sending a large file. An error dialog should be displayed almost instantly and close the attachments preview screen.
- Go into developer settings and enable 'select media quality per upload'.
- Go into advanced settings and verify you can modify both the image upload quality and the video one.
- Then try sending an image: you should see a toggle you can enable/disable to override the default value. The default value should still be taken into account for the initial value displayed.
- Now try sending a video: the initial value will be the default one *if* the estimated video size for it is lower than the max upload size of the home server, otherwise a lower one will be selected or you'll see the 'file too large' error. Modifying the video quality should translate into sending the video with higher/lower quality as well.
- Uploading avatars or sharing media from outside the app should still work as it used to.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly define what will change for the user
- [x] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #5183, opened 2025-08-18
Fix bitrate value used for video transcoding

jmartinesp:
## Content

Fix the estimated video transcoding sizes.

## Motivation and context

It should be 1000 times what it is now. The video size estimation was wrong since the retrieved duration value was in milliseconds, not seconds. It's a regression we didn't catch in https://github.com/element-hq/element-x-android/pull/5131.

## Tests

If you send a video before this change and the same one after it, the video sizes should be dramatically different, as well as the video quality, which should now be way better.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/video/DefaultMediaOptimizationSelectorPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/video/DefaultMediaOptimizationSelectorPresenter.kt
@@ -84,25 +84,25 @@ class DefaultMediaOptimizationSelectorPresenter @AssistedInject constructor(
                 return@produceState
             }
 
-            val (videoDimensions, durationMs) = mediaExtractor.use {
+            val (videoDimensions, duration) = mediaExtractor.use {
                 val size = it.getSize()
                     .getOrElse { exception ->
                         value = AsyncData.Failure(exception)
                         return@produceState
                     }
 
-                val durationMs = it.getDuration()
+                val duration = it.getDuration()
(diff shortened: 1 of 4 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/video/DefaultMediaOptimizationSelectorPresenter.kt
```diff
@@ -84,25 +84,26 @@ class DefaultMediaOptimizationSelectorPresenter @AssistedInject constructor(
...
-                size to duration
+                size to durationMs
             }
 
             val sizeEstimations = VideoCompressionPreset.entries
                 .map { preset ->
-                    val bitRate = preset.compressorHelper().calculateOptimalBitrate(videoDimensions, 30)
-                    val calculatedSize = (bitRate * duration / 8 * 1.1).roundToLong() // Adding 10% overhead for safety
+                    val bitRateAsBytes = preset.compressorHelper().calculateOptimalBitrate(videoDimensions, 30) / 8f
+                    val durationInSeconds = durationMs.toFloat() / 1_000
```
bmarty: Maybe `VideoMetadataExtractor.getDuration()` could return `Result<Duration>` instead of `Result<Long>` to improve the code clarity?

jmartinesp: Good idea, thanks!

PR #5192, opened 2025-08-20
Set shortcut ID on received notifications to make them appear as a Conversation

### LINKED ISSUE #1547: Notifications aren't set as Conversation
### Steps to reproduce

1. Receive a notification
2. See it as random notification instead of Conversation

Here's a screenshot displaying difference notification between old Element and Element X : 

#### What did you expect?

To have enriched notifications labeled and grouped in Android 13 as a conversation notification and have quick options as "Mark as read" or "Quick Reply"

#### What happened instead?

Normal blank notification (see screenshot above)

#### Possible cause

I didn't see any issue about this, so I'm guessing it can be because of my phone configuration. I'm using a de-googled smartphone running MicroG instead of Google Play Service :shrug: 

### Your phone model

Poco F3

### Operating system version

Android 13

Element X 0.2.3

### Homeserver

matrix-org/synapse:v1.93.0

### Will you send logs?

No

No

frebib:
This is the missing piece for Android to treat a notification as a "conversation". Note that this requires shortcuts to have already been created prior to receiving the notification.

 
## Content

Notifications should now show as "conversations" (provided a shortcut for that conversation has already been created by sending a message with #5180)

## Motivation and context

Partially resolves #1547 

## Screenshots / GIFs

Here the user `Steve` doesn't have a shortcut in the left image, but they do the right (as I sent a message to them). User `frebib ⚡` does in both for contrast. The main difference seems to be the size of the avatar and the positioning of the notification within the tray

|Before|After|
|-|-|
|Screenshot_20250820-220930~2|Screenshot_20250820-221007|

Screenshot_20250820-213700Screenshot_20250820-213711

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 16

## Checklist

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
@@ -136,7 +136,10 @@ class DefaultNotificationCreator @Inject constructor(
                 // that can be displayed in not disturb mode if white listed (the later will need compat28.x)
                 .setCategory(NotificationCompat.CATEGORY_MESSAGE)
                 // ID of the corresponding shortcut, for conversation features under API 30+
-                .setShortcutId(roomInfo.roomId.value)
+                // Must match those created in the ShortcutInfoCompat.Builder()
+                // for the notification to appear as a "Conversation":
+                // https://developer.android.com/develop/ui/views/notifications/conversations
+                .setShortcutId("${roomInfo.sessionId.value}-${roomInfo.roomId.value}")
                 // Auto-bundling is enabled for 4 or more notifications on API 24+ (N+)
                 // devices and all Wear devices. But we want a custom grouping, so we specify the groupID
                 .setGroup(roomInfo.sessionId.value)


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
```diff
@@ -174,6 +174,10 @@ class DefaultNotificationCreator @Inject constructor(
             .setSmallIcon(smallIcon)
             // Set primary color (important for Wear 2.0 Notifications).
             .setColor(accentColor)
+            // Set shortcut id. Must match those created in the ShortcutInfoCompat.Builder()
+            // for the notification to appear as a "Conversation":
+            // https://developer.android.com/develop/ui/views/notifications/conversations
+            .setShortcutId("${roomInfo.sessionId}-${roomInfo.roomId}")
```
frebib: Matches https://github.com/element-hq/element-x-android/pull/5180/files#diff-cf5cbcc175237da121694fbac0f7feb8e045248c0320e3f6efc5d9afcd6011a2R108

frebib:
As per the code example in https://developer.android.com/develop/ui/views/notifications/bubbles#bubble-api I think it would make sense to also invoke the `ShortcutInfoCompat.Builder()` here to create the shortcut (if it doesn't exist) so notifications always appear as conversations. Would be nice to not copy-paste the code to make that happen though. I'm not sure I understand Kotlin/the design patterns used in EX enough to make that happen though.

networkException:
For some reason GitHub doesn't back-reference this, but see also https://github.com/element-hq/element-x-android/pull/5180/files#r2301381374

networkException:
> I think it would make sense to also invoke the ShortcutInfoCompat.Builder() here to create the shortcut (if it doesn't exist) so notifications always appear as conversations.

The [API docs](https://developer.android.com/training/sharing/direct-share-targets#report-usage-incoming) aren't exactly clear on when and where to report these for incoming messages, I could imagine issues if we don't take care to not do this in a throttled notification handler and with general prioritization / ranking of shortcuts in face of the limits mentioned in the docs throughout.

I'd be in favor of merging this pretty simple fix (technically a regression fix since this worked in #5116) and iterate on making shortcuts more discoverable / usable down the road

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
```diff
@@ -136,7 +136,10 @@ class DefaultNotificationCreator @Inject constructor(
                 // that can be displayed in not disturb mode if white listed (the later will need compat28.x)
                 .setCategory(NotificationCompat.CATEGORY_MESSAGE)
                 // ID of the corresponding shortcut, for conversation features under API 30+
-                .setShortcutId(roomInfo.roomId.value)
+                // Must match those created in the ShortcutInfoCompat.Builder()
```
bmarty, line 139: Maybe create a shared fun to enforce this requirement? Something like:

```kotlin
internal fun createShortcutId(sessionId: SessionId, roomId: RoomId) = "$sessionId-$roomId"
```

bmarty (approved):
Thanks for the fix! I think it will also fix an issue that may occurs in multi accounts, in case both accounts are member of the same room and a notification need to be displayed.

Issue #4857, opened 2025-06-11
Videos won't embed


babybackribsgivemeback:
### Steps to reproduce

Sometimes when I send a video file, the video does not get embedded. 

Here is a file where that issue happens with me. 

https://github.com/user-attachments/assets/571abacb-47d9-4965-bb4f-c806da2f3b4d

If i turn on "optimise media quality" then videos from my phone get sent fine and embed, but it would be nice to be able to send them without having to compress the video.

But some stuff downloaded from the internet still seems to be an issue.
### Outcome

for the video to be embedded, or atleast easy viewable.

### Your phone model
Pixel 8 pro

### Operating system version
2025060200

### Application version and app store

25.05.4

### Homeserver

default matrix homeserver

### Will you send logs?

No

### Are you willing to provide a PR?

No

Kidswiss:
> If i turn on "optimise media quality" then videos from my phone get sent fine and embed, but it would be nice to be able to send them without having to compress the video.

For me, it doesn't work if this is enabled or disabled. It doesn't work with any videos at all.

watermelon0339:
I am using ElementX App version 25.06.02 and encountering the same issue: when uploading videos via the EXA + button, others cannot play them, and the file format is displayed as application/octet-stream.

Even when I download the file to my phone, it cannot be played.

This issue significantly affects daily usage, and I hope it can be resolved as soon as possible.

profitrollgame:
This happens quite often to me as well (Android 14, EXA 25.06.2).

Many videos cannot be encoded and are being sent as "application/octet-stream" which confuses pretty much all media players.

The only option to view the video is to download it and open it from the file manager, which is extremely annoying.

KeenMaron:
Issue still persists. This feature works fine in the old Element App...

ODeadlock:
I am also experiencing this issue, but seemingly only with mp4 files. The optimize media option makes no difference. Webm and gif files upload without issue, but mp4 files upload with application/octet-stream.  This is with a synapse homeserver that is making full use of workers for handling media. Everything works without issue from Element Desktop.

Downgrading to 25.05.0 solves the issue.

reivilibre:
One of my family reproduced this yesterday, I noticed the msgtype was set to `m.file`.
If I edit and resend the event in Element Web, by setting msgtype to `m.video` and the MIME type to `video/mp4`, the video plays normally.
[...]

PR #5223, opened 2025-08-26
Use variable bitrate mode when transcoding to ensure compatibility with old devices

jmartinesp:
## Content

This should be compatible with more old devices that may lack the needed codecs to properly encode using constant bitrate mode (CBR).

Also, rework the output size computation again. It seems to work fine now with both landscape and portrait videos, unless there's some unknown corner case I haven't found.

## Motivation and context

Fix https://github.com/element-hq/element-x-android/issues/4857.

## Tests

I had to use an old phone (OnePlus 8 I think) with Android 10 to reproduce the issue and test the fix.

- Using the stable version of the app, try sending a normal H264 video. If it gets sent as a file, the transcoding failed because of this issue.
- Then build and install this branch, send the same video. It should now be transcoded and sent as a video, not a file.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 10

## Checklist


bmarty (approved):
LGTM, thanks!

PR #5236, opened 2025-08-27
[a11y] Add content descriptions to room list item indicators

jmartinesp:
## Content

What the title says. These can now be read aloud as 'ongoing call', 'new messages', 'new mentions'.

## Motivation and context

Closes https://github.com/element-hq/customer-success/issues/573.

## Tests

Enable a TTS solution like Talkback, then tap on different room list items/summaries.

- Tap on one without indicators: it should read just the room, timestamp and last message.
- Tap on one with an ongoing call: it should say also 'ongoing call'.
- Tap on one with mentions: it should say 'mentions', possible also 'new messages'.
- Tap on one with new messages: it should say 'new messages'.
- Tap on a muted one: it should say 'muted'.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/atoms/UnreadIndicatorAtom.kt
+++ b/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/atoms/UnreadIndicatorAtom.kt
@@ -15,12 +15,16 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.semantics.contentDescription
+import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import io.element.android.compound.theme.ElementTheme
 import io.element.android.libraries.designsystem.preview.ElementPreview
 import io.element.android.libraries.designsystem.preview.PreviewsDayNight
 import io.element.android.libraries.designsystem.theme.unreadIndicator
+import io.element.android.libraries.ui.strings.CommonStrings
 
 @Composable
 fun UnreadIndicatorAtom(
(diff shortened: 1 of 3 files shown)


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/atoms/UnreadIndicatorAtom.kt
```diff
@@ -29,8 +33,10 @@ fun UnreadIndicatorAtom(
     color: Color = ElementTheme.colors.unreadIndicator,
     isVisible: Boolean = true,
 ) {
+    val contentDescription = stringResource(CommonStrings.a11y_notifications_new_messages)
```
bmarty: `UnreadIndicatorAtom` is also used for [invitation](https://github.com/element-hq/element-x-android/blob/develop/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/components/RoomSummaryRow.kt#L335), so maybe let `contentDescription` be a parameter and add a string `a11y_notifications_new_invitation`?

tests/uitests/src/test/snapshots/images/features.home.impl_HomeViewA11y_en.png
bmarty, line 1: image

Maybe the indicator should be read first, and then the last message? It will require some change in the layout, but I believe that it will allow a faster screen reader navigation between rooms. We may handle this later though.

Note: happy to see that the a11y screenshot test is useful!

bmarty (approved):
Thanks! 2 suggestions that may be handled later.

tests/uitests/src/test/snapshots/images/features.home.impl_HomeViewA11y_en.png
jmartinesp: Maybe I can play with the `traversalIndex` here, but something tells me it'll be harder than that 😅

jmartinesp: In the end I had to use `zIndex` again. It seems like `traversalIndex` won't work for nodes that merge their descendants.

Issue #5242, opened 2025-08-28
Infinite loading after timeout when forwarding a message


jbrenorv:
### Steps to reproduce

1. Create a context where forwarding a message results in a timeout.

#### What did you expect?
The state to be updated to error.

#### What happened instead?
The state remains in loading forever.

### Will you send logs?

No

### Are you willing to provide a PR?

Yes

bmarty:
> Create a context where forwarding a message results in a timeout.

Can you be more specific please?

PR #5243, opened 2025-08-28
Update state in runUpdatingState when CancellationException occurs

jbrenorv:
## Content

`AsyncAction` > `runUpdatingState` wasn't updating the state when `CancellationException` occurs. This kind of exception happens, for example, when a timeout occurs. The state was remaining stuck in loading forever.

## Motivation and context

Issue
https://github.com/element-hq/element-x-android/issues/5242

## Tests

- A unit test was added to cover this scenario.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android API 33 (Samsung A32)

## Checklist

### DIFF
--- a/libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/AsyncAction.kt
+++ b/libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/AsyncAction.kt
@@ -10,6 +10,7 @@ package io.element.android.libraries.architecture
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.Stable
 import io.element.android.libraries.core.extensions.runCatchingExceptions
+import kotlinx.coroutines.TimeoutCancellationException
 import kotlin.contracts.ExperimentalContracts
 import kotlin.contracts.InvocationKind
 import kotlin.contracts.contract
@@ -159,16 +160,19 @@ suspend inline fun <T> runUpdatingState(
         callsInPlace(resultBlock, InvocationKind.EXACTLY_ONCE)
     }
     state.value = AsyncAction.Loading
-    return resultBlock().fold(
+    return try {
+        resultBlock()
+    } catch (e: TimeoutCancellationException) {
+        state.value = AsyncAction.Failure(errorTransform(e))
+        throw e
+    }.fold(
         onSuccess = {
             state.value = AsyncAction.Success(it)
             Result.success(it)
         },
         onFailure = {
             val error = errorTransform(it)
-            state.value = AsyncAction.Failure(
-                error = error,
-            )
+            state.value = AsyncAction.Failure(error)
             Result.failure(error)
         }
     )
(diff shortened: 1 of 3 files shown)


libraries/architecture/src/test/kotlin/io/element/android/libraries/architecture/AsyncActionTest.kt
```diff
@@ -0,0 +1,41 @@
...
+        val operationTimeMillis = 1000L
+
+        try {
+            runUpdatingState(state = state) {
+                withTimeout(timeoutMillis.milliseconds) {
+                    delay(operationTimeMillis)
+                }
+                Result.success(0)
+            }
+            assertTrue("Expected CancellationException, but nothing was thrown", false)
```
bmarty: You can simply use `fail`:

```suggestion
            fail("Expected CancellationException, but nothing was thrown")
```

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/AsyncAction.kt
```diff
@@ -159,17 +160,25 @@ suspend inline fun <T> runUpdatingState(
...
+            },
+            onFailure = {
+                val error = errorTransform(it)
+                state.value = AsyncAction.Failure(
+                    error = error,
+                )
+                Result.failure(error)
+            }
+        )
+    } catch (e: CancellationException) {
```
jmartinesp: I wonder if it wouldn't be a better solution to instead add a specific `catch (e: TimeoutCancellationException)` entry. My reasoning is:

1. We want to catch those exceptions most of the time, the only reason we don't is because they extend `CancellationException` which we want to rethrow so the coroutines work as expected.
2. We probably don't want to catch most `CancellationException`s, since these could happen i.e. when you exit a screen if some long task is running and could display error messages/dialogs while this screen transition is happening.

jbrenorv: You're right! That's exactly what I want to do.

bmarty: Also, the `try` block should probably only wrap `resultBlock()` and not the code in the `fold` lambdas, even if in practice the code behavior should be quite similar.

jbrenorv: How about this?

```kotlin
return try {
    resultBlock()
} catch (e: TimeoutCancellationException) {
    state.value = AsyncAction.Failure(errorTransform(e))
    throw e
}.fold(
    onSuccess = {
        state.value = AsyncAction.Success(it)
        Result.success(it)
    },
    onFailure = {
        val error = errorTransform(it)
        state.value = AsyncAction.Failure(error)
        Result.failure(error)
    }
)
```

Issue #5248, opened 2025-08-31
Send attachment as reply to selected message


AnXPyDev:
#### What would you like to do?
When selecting a message to reply to and then selecting an attachment to send (image or whatever else), reply to the selected message with the attachment/s.

#### Why would you like to do it?
Because the current behavior is frustrating and nonsensical to anyone who has used any other messaging app before.

#### How would you like to achieve it?
...

### Have you considered any alternatives?

no

### Additional context

no

### Are you willing to provide a PR?

No

frebib:
This is a duplicate of #1603 so #5261 will also close that too

bmarty:
Done in https://github.com/element-hq/element-x-android/pull/5261

PR #5261, opened 2025-09-02
Allow replying to a message with an attachment

jmartinesp:
## Content

When on `MessageComposerMode.Reply`, adding an attachment will send it as a reply to the message you were replying to.

## Motivation and context

Implements https://github.com/element-hq/element-x-android/issues/5248. I think not having this was an oversight on our side, since the Rust SDK API for it exists.

## Tests

In a room, start replying to a message. Then open the add attachment menu and send either some media or a plain file. It should be sent as a reply to the message.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


frebib:
Works great. One maybe-oddity is that the "reply to" context doesn't appear in the composer for media captions. To be clear, the correct message gets sent, it's just not displayed to the user in that screen

bmarty (approved):
LGTM, thanks!

PR #5268, opened 2025-09-03
Reload member list after moderation actions

jmartinesp:
## Content

- When a moderation action is done (kicking, banning, unbanning), reload the member list instead of trying to modify the member list Flow.
- Ensure the `Ready` state is set to the flow, even if the `LaunchedEffect/collectLatest` scope is cancelled.

## Motivation and context

The previous `runActionAndWaitForMembershipChange` logic wasn't really doing anything, as the modified flow was never used. This means kicking, banning or unbanning a user didn't reload the member list consistently.

## Tests

In a room you're an admin/moderator:

- Invite a couple of users. The room member list is updated.
- Kick one of them. The list should update.
- Then ban and unban the other. In both cases, the updates should be visible.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/roommembermoderation/impl/src/main/kotlin/io/element/android/features/roommembermoderation/impl/RoomMemberModerationPresenter.kt
+++ b/features/roommembermoderation/impl/src/main/kotlin/io/element/android/features/roommembermoderation/impl/RoomMemberModerationPresenter.kt
@@ -84,6 +84,9 @@ class RoomMemberModerationPresenter @Inject constructor(
                     )
                 }
                 is RoomMemberModerationEvents.ProcessAction -> {
+                    // First, hide any list of existing actions that could be displayed
+                    moderationActions.value = persistentListOf()
+
                     when (event.action) {
                         is ModerationAction.DisplayProfile -> Unit
                         is ModerationAction.KickUser -> {
@@ -99,7 +102,6 @@ class RoomMemberModerationPresenter @Inject constructor(
                             unbanUserAsyncAction.value = AsyncAction.ConfirmingNoParams
                         }
                     }
-                    moderationActions.value = persistentListOf()
                 }
                 is InternalRoomMemberModerationEvents.DoKickUser -> {
                     selectedUser?.let {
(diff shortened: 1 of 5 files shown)


features/roommembermoderation/impl/src/main/kotlin/io/element/android/features/roommembermoderation/impl/RoomMemberModerationPresenter.kt
```diff
@@ -97,6 +99,7 @@ class RoomMemberModerationPresenter @Inject constructor(
                             unbanUserAsyncAction.value = AsyncAction.ConfirmingNoParams
                         }
                     }
+                    moderationActions.value = persistentListOf()
```
bmarty: I am not sure to understand the effect of this. It is supposed to close the bottom sheet as soon as an action is selected, to avoid that a bottom sheet is displayed at the same time as a confirmation dialog, but this is already the current behavior. Is there some other rationale?

bmarty (approved):
LGTM, thanks

features/roommembermoderation/impl/src/main/kotlin/io/element/android/features/roommembermoderation/impl/RoomMemberModerationPresenter.kt
jmartinesp: > It is supposed to close the bottom sheet as soon as an action is selected, to avoid that a bottom sheet is displayed at the same time as a confirmation dialog, but this is already the current behavior.

It wasn't after I made these changes, I had to manually set it to an empty list or both the action list and the confirmation dialog would be displayed at the same time.

jmartinesp: Oh, it's because the actions were computed for the user I was banning, then I tried to unban that same user, and it was displaying the previously calculated actions, which no longer apply.

bmarty: ok, thanks!

PR #5292, opened 2025-09-08
Hide the home navigation bar if the user is not a member of any Space.

bmarty:
## Content

Hide the navigation bar if the user is not a member of any space.

## Motivation and context

Part of https://github.com/element-hq/element-meta/issues/2906:
`The tab bar with the option to view joined spaces is only shown when the user has at least one space that they have joined (because otherwise they have no clue what to do in here).`

## Tests

- Login using an account which is not in any space
- Enable the Space FF
- Observe that the navigation bar on the home screen is still not rendered
- From a client capable of doing this, join a space
- The navigation bar should become visible.
- Navigate to the space tab
- From a client capable of doing this, leave the space
- Observe that the room list is displayed again, and the navigation bar is not displayed anymore

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/home/impl/src/test/kotlin/io/element/android/features/home/impl/HomePresenterTest.kt
+++ b/features/home/impl/src/test/kotlin/io/element/android/features/home/impl/HomePresenterTest.kt
@@ -12,9 +12,11 @@ import app.cash.molecule.moleculeFlow
 import app.cash.turbine.test
 import com.google.common.truth.Truth.assertThat
 import io.element.android.features.home.impl.roomlist.aRoomListState
+import io.element.android.features.home.impl.spaces.HomeSpacesState
 import io.element.android.features.home.impl.spaces.aHomeSpacesState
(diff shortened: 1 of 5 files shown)


features/home/impl/src/test/kotlin/io/element/android/features/home/impl/HomePresenterTest.kt
```diff
@@ -145,21 +148,50 @@ class HomePresenterTest {
         }
     }
 
-    private fun TestScope.createHomePresenter(
+    @Test
+    fun `present - NavigationBar is hidden when the last space is left`() = runTest {
+        val homeSpacesPresenter = MutablePresenter(aHomeSpacesState())
```
bmarty, line 153: I first reinvented `MutablePresenter` and see that it was already a thing, but not used (yet!)

Issue #5321, opened 2025-09-10
Align max file size text


Destinyg133:
### Steps to reproduce

![Image](https://github.com/user-attachments/assets/9f306640-ae62-4cbd-a1a9-c8466cf78016)

Text should align, looks funny

#### What did you expect?
To be aligned

#### What happened instead?
Was not aligned

### Application version and app store

25.09.1

### Will you send logs?

No

### Are you willing to provide a PR?

No

PR #5332, opened 2025-09-11
Do not center the dialog title text for dialogs with no icon

jmartinesp:
## Content

What the title says.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/5321. Matches [Figma](https://www.figma.com/design/G1xy0HDZKJf5TCRFmKb5d5/Compound-Android-Components?node-id=911-343492), AFAICT. Centering the title when it overflows seems like a mistake.

## Screenshots / GIFs

There are lots of screenshots with dialogs that move ~1px 😬 .

## Checklist


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/AlertDialogContent.kt
```diff
@@ -142,7 +142,6 @@ internal fun SimpleAlertDialogContent(
                 Text(
                     text = titleText,
                     style = ElementTheme.typography.fontHeadingSmMedium,
-                    textAlign = TextAlign.Center,
```
bmarty, line 145: This does not match the PR description `Do not center the dialog title text for dialogs with no icon`. Maybe keep centering the text when there is an icon?

Else we may have this:

Image

See also the change in QrCodeIntroView Preview

(But this is maybe OKish?)

tests/uitests/src/test/snapshots/images/features.login.impl.screens.qrcode.intro_QrCodeIntroView_Day_1_en.png
bmarty, line 1: Not sure we want this change.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/theme/components/AlertDialogContent.kt
jmartinesp: Ah true, I guess we should change how this is centered: instead  of centering the `Text` component when it has an icon, we should make if fill the max width and use `TextAlign.Center` instead in that case.

jmartinesp: Done in https://github.com/element-hq/element-x-android/pull/5332/commits/60518986bc5a6c340dc85e1b182731dd4c663258, I'm creating new screenshots.

jmartinesp:
Ok, I think both the implementation and screenshots should now be better. The titles for dialogs with no icon should be start-aligned and the ones for dialogs with icons center-aligned, per the [M3 guidelines](https://m3.material.io/components/dialogs/specs#6771d107-624e-47cc-b6d8-2b7b620ba2f1).

PR #5355, opened 2025-09-15
Add thread decoration with latest event details

jmartinesp:
## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
```diff
@@ -288,6 +300,79 @@ fun TimelineItemEventRow(
...
 
+@Composable
+private fun ThreadSummaryView(
+    threadSummary: ThreadSummary,
+    latestEventText: String?,
+    isOutgoing: Boolean,
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    val bubbleShape = MessageEventBubbleDefaults.shape(false, TimelineItemGroupPosition.None, isOutgoing)
```
bmarty: It is complex to use `MessageEventBubbleDefaults.shape(...)` with 3 parameters here since the result will always be a simple `RoundedCornerShape(BUBBLE_RADIUS)`.

Maybe add a val in `MessageEventBubbleDefaults` like `val threadShape = RoundedCornerShape(BUBBLE_RADIUS)`?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt
```diff
@@ -288,6 +300,79 @@ fun TimelineItemEventRow(
...
+                )
+                Avatar(
+                    avatarData = avatarData,
+                    avatarType = AvatarType.User,
+                )
+
+                Spacer(modifier = Modifier.width(8.dp))
+
+                Text(
+                    text = latestEvent.senderProfile.getDisplayName() ?: latestEvent.senderId.value,
```
bmarty: We should use the disambiguated displayname here:


```suggestion
                    text = latestEvent.senderProfile.getDisambiguatedDisplayName(latestEvent.senderId),
```

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemEventFactory.kt
```diff
@@ -69,6 +71,27 @@ class TimelineItemEventFactory(
...
+            summaryFormatter.format(threadSummaryContent)
+        }
+        val mappedThreadInfo = threadInfo?.let {
+            TimelineItemThreadInfo(
+                threadRootId = it.threadRootId,
+                latestEventText = threadSummaryText,
+                threadSummary = it.threadSummary,
+            )
+        }
+
```
bmarty, line 98: I rewrite the code the "Kotlin" way:

```
        val mappedThreadInfo = currentTimelineItem.event.threadInfo()?.let { threadInfo ->
            TimelineItemThreadInfo(
                threadRootId = threadInfo.threadRootId,
                latestEventText = threadInfo.threadSummary.latestEvent.dataOrNull()?.let {
                    contentFactory.create(
                        itemContent = it.content,
                        eventId = it.eventOrTransactionId.eventId,
                        isEditable = false,
                        sender = it.senderId,
                        senderProfile = it.senderProfile,
                    )
                }?.let(summaryFormatter::format),
                threadSummary = threadInfo.threadSummary,
            )
        }
```

I'll let you decide whether it's better or not.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemEventFactory.kt
```diff
@@ -87,7 +110,7 @@ class TimelineItemEventFactory(
             readReceiptState = currentTimelineItem.computeReadReceiptState(roomMembers),
             localSendState = currentTimelineItem.event.localSendState,
             inReplyTo = currentTimelineItem.event.inReplyTo()?.map(permalinkParser = permalinkParser),
-            threadInfo = currentTimelineItem.event.threadInfo() ?: EventThreadInfo(threadRootId = null, threadSummary = null),
+            threadInfo = mappedThreadInfo ?: TimelineItemThreadInfo(threadRootId = null, threadSummary = null, latestEventText = null),
```
bmarty: Event.threadInfo` should be nullable instead of having a data class full of null val.

```suggestion
            threadInfo = mappedThreadInfo,
```

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/TimelineItem.kt
```diff
@@ -130,3 +131,9 @@ sealed interface TimelineItem {
         val aggregatedReadReceipts: ImmutableList<ReadReceiptData>,
     ) : TimelineItem
 }
+
+data class TimelineItemThreadInfo(
+    val threadRootId: ThreadId?,
+    val latestEventText: String?,
+    val threadSummary: ThreadSummary?,
+)
```
bmarty: I would make `threadRootId` and `threadSummary` not nullable, but `Event.threadInfo` nullable

`MessageContent.threadInfo` should be nullable too.

jmartinesp: I don't think we can make them non-nullable, since the root of a thread will have a non-null thread summary but null threadRootId (since it is already the root id) and any other item in the thread will have the opposite info.

jmartinesp: I'm double checking this in case I'm wrong and it is possible, since that would be a way better API.

jmartinesp: Yeah, when we get a `ThreadSummary` we don't get the root id. We could assume it's the thread root and manually set the value to the event's own id though. But the opposite won't work (given any other event in a thread but the root, get its summary).

jmartinesp: In the end I've modeled this as a `sealed interface` where you can either have a `ThreadRoot(ThreadSummary)` or a `ThreadResponse(ThreadId)`, or just `null` if it's not any of those 2 cases.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
```diff
@@ -328,7 +329,10 @@ class MessagesPresenter(
                 val displayThreads = featureFlagService.isFeatureEnabled(FeatureFlags.Threads)
                 if (displayThreads) {
                     // Get either the thread id this event is in, or the event id if it's not in a thread so we can start one
-                    val threadId = targetEvent.threadInfo.threadRootId ?: targetEvent.eventId!!.toThreadId()
+                    val threadId = when (targetEvent.threadInfo) {
+                        is TimelineItemThreadInfo.ThreadResponse -> targetEvent.threadInfo.threadRootId
+                        is TimelineItemThreadInfo.ThreadRoot, null -> targetEvent.eventId?.toThreadId()
```
bmarty, line 334: Not a big deal but since we are not using the smart cast on `TimelineItemThreadInfo.ThreadRoot` (I was confused about that at the beginning), maybe using `else` is cleaner.


```suggestion
                        else -> targetEvent.eventId?.toThreadId()
```

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/TimelineItem.kt
bmarty: Clever solution, thanks!

jmartinesp:
> Thanks for the update. There is just this [remark](https://github.com/element-hq/element-x-android/pull/5355#discussion_r2371643842) missing. Also maybe bubbleShape ([here](https://github.com/element-hq/element-x-android/blob/236c1b04906b934fe6362837b5179e6b04e0186d/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/TimelineItemEventRow.kt#L308)) could be inlined.

Ah damn, I changed it by mistake in the `AvatarData` instead 🤦 .

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
jmartinesp: I wanted to explicitly list all the possible cases so if we end up adding one in the future (unlikely, but...) we'll have to come back here to manage it.

Issue #5031, opened 2025-07-15
Duplicate room on romo list


ara4n:
### Steps to reproduce

1. @kongo09 reported duplicate room in list (not around midnight, so it's not #4182, plus the room preview is the same):

<img width="1048" height="504" alt="Image" src="https://github.com/user-attachments/assets/33706c14-068c-45d5-ae21-ddcb8d08bb82" />

#### What did you expect?

No dup rooms

#### What happened instead?

Dup room with dup room preview

### Application version and app store

Nightly 2025-07-14

### Homeserver

element.io

### Will you send logs?

No

### Are you willing to provide a PR?

No

bustikiller:
Also happening in version 25.06.2 installed from F-Droid.

![Image](https://github.com/user-attachments/assets/56f7a03a-6b07-43f9-9f57-0ccde2979edb)

jmartinesp:
> Also happening in version 25.06.2 installed from F-Droid.

So it's definitely not new 🫤 . We don't have an easy way to debug this, sadly... It could come either from the SDK, emitting the wrong data, or something in the app that added the same room twice / displayed the wrong data for the items.

jmartinesp:
Now that I look at it:

> Nightly 2025-07-14

That's also a quite outdated version, could you upgrade to the latest nightly @kongo09 and report back if it happens again?

PR #5364, opened 2025-09-18
Add logs to detect duplicates in the room list

jmartinesp:
## Content

Adds some logs to detect duplicates in the room list and include relevant info about what triggered the change and what the change implied.

## Motivation and context

Try to diagnose https://github.com/element-hq/element-x-android/issues/5031.

## Tests

We didn't change any logic, so everything should work just like before. Room list updates may be slightly slower though, since we now compute duplicates.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/datasource/RoomListDataSource.kt
+++ b/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/datasource/RoomListDataSource.kt
@@ -30,6 +30,7 @@ import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
+import timber.log.Timber
 import kotlin.time.Duration.Companion.seconds
 
 @Inject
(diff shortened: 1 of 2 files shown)


features/home/impl/src/main/kotlin/io/element/android/features/home/impl/datasource/RoomListDataSource.kt
```diff
@@ -102,13 +103,39 @@ class RoomListDataSource(
...
+                    }
+                    buildAndCacheItem(roomSummaries, index)
+                }
             } else {
+                roomSummaries.getOrNull(index)?.roomId?.let {
+                    val pairs = cachingResults.getOrDefault(it, mutableListOf())
+                    pairs.add(index to false)
+                    cachingResults[it] = pairs
+                }
                 buildAndCacheItem(roomSummaries, index)
```
ganfra: The code is a bit hard to follow here, what is happing exactly?

jmartinesp: Some comments would have definitely helped, true 🫤 . What I tried to do there is, for each item that's computed, associated with its room id, add the index where it is and whether it came from the cache or not. If for a room id we find 2 pairs of index:cached, then something went wrong and we can log what change caused the issue.

features/home/impl/src/main/kotlin/io/element/android/features/home/impl/datasource/RoomListDataSource.kt
```diff
@@ -103,27 +103,30 @@ class RoomListDataSource(
...
                     }
                 } ?: run {
                     roomSummaries.getOrNull(index)?.roomId?.let {
                         val pairs = cachingResults.getOrDefault(it, mutableListOf())
-                        pairs.add(index to false)
+                        pairs.add(CacheResult(index, fromCache = false))
                         cachingResults[it] = pairs
                     }
                     buildAndCacheItem(roomSummaries, index)
                 }
```
ganfra: Maybe it'd be simpler to understand without the `also ?: run `, the `it` are not obvious here I think

ganfra (approved):
Thanks for the change, one final remark, but otherwise LGTM!

PR #5399, opened 2025-09-23
When joining a call, wait for the `content_loaded` action

jmartinesp:
## Content

We now wait for `content_loaded` action to assume the call hasn't timed out instead of `io.element.join`.

## Motivation and context

This ensures the widget won't be disposed when the user joins a call and loads the lobby screen for room calls on EC `0.16.x`.

## Tests

Join a room call and stay in the lobby > 10s. If it doesn't display an error, it's working fine.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
@@ -152,7 +152,7 @@ class CallScreenPresenter(
                 delay(10.seconds)
 
                 if (!isWidgetLoaded) {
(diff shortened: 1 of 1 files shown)


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
```diff
@@ -151,7 +151,7 @@ class CallScreenPresenter(
                 // Wait for the call to be joined, if it takes too long, we display an error
                 delay(10.seconds)
 
-                if (!isJoinedCall) {
+                if (!isWidgetLoaded) {
```
bmarty, line 154: Maybe update the log below?
s/joined/loaded/

PR #5255, opened 2025-09-02
Add emoji search to the reaction emoji picker

jmartinesp:
## Content

Add search functionality to the emoji reaction selector.

Thanks @ChelseaDH for his initial work on https://github.com/element-hq/element-x-android/pull/2892, which served as inspiration for this one.

<del>Note the bottom sheet being collapsed when the keyboard opens remains an issue, but since the only alternative is implementing our own bottom sheet, there's not much we can do about it.</del> Actually this issue was fixed on the latest Material 3 Compose library, which we can't include yet.

## Motivation and context

Manually searching for an emoji is slow and inefficient.

## Screenshots / GIFs

In the PR.

## Tests

In a room, long press to display the timeline item menu and tap on the 'emoji +' button to open the whole list of emoji reactions.

You can use it to search for different emojis based on its tags and shortcodes - I could also have added the label, but it's usually a combination of the tags.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly define what will change for the user
- [x] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #5402, opened 2025-09-24
Use shared recent emoji reactions from account data

jmartinesp:
## Content

- Add `AddRecentEmoji` and `GetRecentEmojis` use cases to avoid injecting the whole `MatrixClient` for just one of these operations.
- Update the UI and logic of the emoji picker and message context menu to include the recent emojis.
- Add `CoroutineDispatchers.Default` with the defaults coroutines to use in the app for ease of use.

## Motivation and context

Iterate on https://github.com/element-hq/element-x-android/pull/5255, continue integrating the work done for the hackathon around temojis.

## Screenshots / GIFs

image

image

Then on Element Web for the same account:

image

## Tests

1. Long press on a message to open the message context menu.
2. If you already used some emoji reactions in Element Web with that account, you should see the same emojis being displayed there. Otherwise, the default ones will be displayed.
3. If you tap in the button to open the emoji picker and you have recent emoji reactions, you should now see a 'recents' tab with the items added in the same order.
4. Adding a new emoji reaction should update the list. Removing one won't.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/customreaction/picker/EmojiPicker.kt
```diff
@@ -53,105 +55,83 @@ fun EmojiPicker(
     modifier: Modifier = Modifier,
 ) {
     val coroutineScope = rememberCoroutineScope()
-    val categories = state.categories
-    val pagerState = rememberPagerState(pageCount = { EmojibaseCategory.entries.size })
-
+    val pagerState = rememberPagerState(pageCount = { state.categories.size })
     Column(modifier) {
-        SearchBar(
```
frebib, line 60: Why are you removing the search bar? 😞

jmartinesp: Oh, it must be a rebase issue, thanks for spotting it!

jmartinesp: This should be fixed now. Sorry for the misunderstanding!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/actionlist/ActionListView.kt
```diff
@@ -338,40 +347,65 @@ private val emojiRippleRadius = 24.dp
...
-                onClick = onEmojiReactionClick
-            )
+        val backgroundColor = ElementTheme.colors.bgCanvasDefault
+
+        val emojis = remember(recentEmojis.isEmpty()) {
+            if (recentEmojis.isEmpty()) {
+                persistentListOf("👍️", "👎️", "🔥", "❤️", "👏")
+            } else {
+                recentEmojis.take(50)
+            }
```
bmarty: Do we have a product decision on this?
I think it can be improved, because before the user sends a very first reaction, they will see the 5 default emojis, and after they send a first one, it will be the only one in the list which is weird since it can be any random one.
Maybe we should always display the 5 default emojis after the list of the recent ones?
```
recentEmojis.take(50) + listOf("👍️", "👎️", "🔥", "❤️", "👏").filter { it !in recentEmojis }
```

bmarty (commented):
Not tested, but one first comment

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/actionlist/ActionListView.kt
jmartinesp: > Do we have a product decision on this?

Not really, I just kept the previous suggested emoji reactions for the default values.

> Maybe we should always display the 5 default emojis after the list of the recent ones?

That's one option, but then it breaks the 1:1 match with EW this PR wanted to provide.

@mxandreas any opinion on this?

jmartinesp:
There were some issues with recording screenshots, let's see if the latest changes fix them 🤞 .

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/actionlist/ActionListView.kt
mxandreas: I am not sure I fully understand everything here (if it easy, perhaps re-state the question). But generally consistency with EW is key, unless EW's behavior is known to be problematic and should be improved.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/customreaction/picker/EmojiPicker.kt
frebib: Yay, thanks!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/actionlist/ActionListView.kt
bmarty: You can't compare with Element Web, which has everything:

image

bmarty: (this message was for Jorge)

jmartinesp: > I am not sure I fully understand everything here (if it easy, perhaps re-state the question). But generally consistency with EW is key, unless EW's behavior is known to be problematic and should be improved.

When we don't have any recently used emojis, we suggest using any of these:

```
👍️    👎️   🔥    ❤️    👏
```

But then, once an emoji has been used (i.e. 🔹), the user will see just this for the suggested emoji reactions: 

```
🔹
```

Which is a bit odd, but then they'd also see just that single suggested emoji in EW if I'm not mistaken. @bmarty proposes having this after the 1st emoji is used:

```
🔹    👍️    👎️   🔥    ❤️    👏
```

But then it wouldn't match EW.

bmarty: For @mxandreas if I understand correctly the PR, when long clicking on any message, on an account that has not sent any reaction, user will see this:

image

and once a reaction 💩 will be sent, they will see this:

image

I propose that they always see the default reaction:

image

jmartinesp: It's true we can't really match EW 1:1 since we're using their 'quick actions' as the suggested ones, and then we'd want to replace them / add the recent ones to those, which on EW are 2 different sections. Maybe @bmarty 's suggestion is the best option. WDYT @mxandreas ? - sorry for pinging you so much on this topic, btw 😅.

spaetz: Please, no :-). Imagine I used a few popular emojis, then I would see:
👍️    🔥    👍️    👎️   🔥    ❤️    👏
and other weird duplications, which looks incredibly clumsy.

Proposal: If getRecentEmojis is empty, prime the list once with setRecentEmojis.addEmoji('👍️') (or whatever the exact API is) and do this for 3-4 emojis. One could do this once on startup and be done with it or each time the emoji list is shown.
Will shut up now

jmartinesp: In the end we're doing a mix: apparently, iOS was already using a concatenation of the suggested emojis + the recent emojis provided by the OS, removing duplicates. We'll do the same here, since there were some users who complained about the quick reactions changing too frequently with the recent emojis and they tapped the wrong emojis by muscle memory.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/actionlist/ActionListPresenter.kt
```diff
@@ -153,14 +155,15 @@ class DefaultActionListPresenter(
                 ),
                 displayEmojiReactions = displayEmojiReactions,
                 verifiedUserSendFailure = verifiedUserSendFailure,
-                actions = actions.toImmutableList()
+                actions = actions.toImmutableList(),
+                recentEmojis = getRecentEmojis().getOrNull()?.toImmutableList() ?: persistentListOf()
```
bmarty, line 159: Maybe simpler with:
```suggestion
                recentEmojis = getRecentEmojis().getOrNull().orEmpty().toImmutableList()
```

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/customreaction/CustomReactionPresenter.kt
```diff
@@ -9,29 +9,39 @@ package io.element.android.features.messages.impl.timeline.components.customreac
...
+
         val target: MutableState<CustomReactionState.Target> = remember {
             mutableStateOf(CustomReactionState.Target.None)
         }
 
-        val localCoroutineScope = rememberCoroutineScope()
         fun handleShowCustomReactionSheet(event: TimelineItem.Event) {
             target.value = CustomReactionState.Target.Loading(event)
             localCoroutineScope.launch {
+                recentEmojis = getRecentEmojis().getOrNull()?.toImmutableList() ?: persistentListOf()
```
bmarty:
```suggestion
                recentEmojis = getRecentEmojis().getOrNull().orEmpty().toImmutableList()
```

bmarty (approved):
LGTM, thanks!

PR #5407, opened 2025-09-24
Feature/bma/check element call support

### LINKED ISSUE #5406: When Element Call is disabled incoming call notification are still displayed
SessionEnterpriseService.isElementCallAvailable() should be checked before displaying incoming call notification

bmarty:
## Content

Do not show incoming call notification if Element Call is disabled in Element Pro.

Please review the diff here too: https://github.com/element-hq/element-android-enterprise/compare/95789d40119499eba8a79284df9dd2306405b099...311c7d282a2b0b00cad23da84146235228a17942

## Motivation and context

Closes #5406 

## Tests

- FOSS is not impacted
- Using Element Pro on a server with Element Call disable, let the user receive an incoming call
- Before: a notification was displayed and the call could be ansered
- Now: nothing happens.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


features/enterprise/impl-foss/src/main/kotlin/io/element/android/features/enterprise/impl/DefaultEnterpriseService.kt
```diff
@@ -35,5 +36,7 @@ class DefaultEnterpriseService : EnterpriseService {
     override fun firebasePushGateway(): String? = null
     override fun unifiedPushDefaultPushGateway(): String? = null
 
+    override suspend fun isElementCallAvailable(matrixClient: MatrixClient) = true
```
jmartinesp: Having to pass a `MatrixClient` as a parameter seems a bit weird, to be honest. Can't we make this work with just the session id, then have the enterprise service get or restore the client internally?

bmarty: Very true. I will update.

I was also wondering if this is up to the PushHandler to check Element Call availability or rather `DefaultElementCallEntryPoint`. I tend to think that it is better to let `DefaultElementCallEntryPoint` do the check. What is your POV @jmartinesp ?

jmartinesp: Maybe it's better to do it inside the entrypoint, since it seems more related to it than the push handler.

bmarty: I have updated the PR and force pushed the branches to reduce noise in the git history.

jmartinesp (approved):
Thanks, this looks way better!

bmarty:
We decided to not update the code regarding incoming call notification when Element Call is not supported. Element Call support should only be checked when initiating a call.

PR #5423, opened 2025-09-27
Feature: Add minimized video service (YouTube-style PiP

### LINKED ISSUE #5421: Feature Request: System-wide floating video player (overlay service)
### Your use case

What would you like to do?

I would like to add support for a floating video player overlay that allows users to keep watching a video while navigating outside the Element X app. The floating window should be draggable, resizable, and persistent on top of other apps (similar to YouTube’s floating player).

Why would you like to do it?

This would improve multitasking. Users often want to continue watching a call recording, conference, or media file while browsing messages, switching rooms, or even leaving the app. A floating player ensures media playback is not interrupted when leaving the Element app.

How would you like to achieve it?

My idea is to implement a  service with display a draggable video window. The service would:

Render the video in a floating view using VideoView or ExoPlayer.

Provide basic controls (play/pause, close, maximize).

Manage overlay permissions (Settings.canDrawOverlays).

Optionally integrate with Picture-in-Picture mode for devices that support it.

This approach would be separate from the in-app video viewer, ensuring it works system-wide, even if the user switches apps.

Yes, I’ve considered the following:

Android Picture-in-Picture (PiP) mode:
Android already supports PiP for videos, but it is limited in customization (only certain aspect ratios, no custom UI beyond basic controls) and behaves differently across OEMs. My proposal offers more flexibility (draggable window, custom controls, consistent UI).

In-app mini-player:
Keeping the video minimized within the Element app UI is simpler, but it does not solve the use case of watching video while switching to other apps.

Background audio playback:
Another alternative is allowing video audio to keep playing when the app is in the background, but this does not cover users who want to see the video while multitasking.

I believe a floating overlay provides the most complete experience for users who want video to remain visible while using other apps.

### Additional context

Many modern messaging apps (e.g., Telegram, WhatsApp) support floating or picture-in-picture video playback. This sets a strong user expectation for multitasking during video calls or media playback.

Because Element X is built with a Compose-first architecture, implementing this feature in the exact same way as other messengers may not be straightforward. However, by using an overlay service approach (similar to YouTube’s floating player), it is possible to provide a similar experience where the video stays visible and draggable even outside the app.

This would bring Element X closer to feature parity with other messaging apps while also offering a unique implementation adapted to Element’s architecture.

No

aliIsazadeh:
## What’s Changed
- Introduced a minimized video service, similar to YouTube’s PiP experience.
- Allows users to keep watching videos while navigating the app.
- Designed within Element X’s architecture constraints, since implementing it like Telegram/WhatsApp is not straightforward.
- Ensures smooth transitions between fullscreen and minimized modes.

## Why This Matters
Many modern messengers (e.g. Telegram, WhatsApp) already support minimized video.  
Element X’s structure makes this challenging, but this approach provides a practical way to implement it while staying close to the app’s design principles.

## Demo
[https://github.com/user-attachments/assets/c5480d10-9b42-4ed2-81fd-e3080a506366]

Closes #5421 
---

## Notes
This was really fun to implement, and I believe it can improve the user experience a lot. 🚀  
Also — I’m open to remote Android opportunities. If anyone is interested in collaborating or has a project/position where I can contribute, feel free to reach out: **[iszadhali@gmail.com]**


libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerNode.kt
```diff
@@ -127,15 +134,27 @@ class MediaViewerNode(
 
     @Composable
     override fun View(modifier: Modifier) {
+        val context = LocalContext.current
+        val (isMinimized, setMinimized) = remember { mutableStateOf(false) }
```
bmarty, line 142: TIL we can assigned remembered mutable state like this. Handy!

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerView.kt
```diff
@@ -483,6 +490,20 @@ private fun MediaViewerTopBar(
...
+                IconButton(
+                    onClick = {
+                        setMinimize(true)
+                        onBackClick()
+                    },
+                    modifier = Modifier
+                ) {
+                    Icon(
+                        imageVector = Icons.Default.FullscreenExit,
+                        contentDescription = "Minimize"
```
bmarty: a11y

app/src/main/AndroidManifest.xml
```diff
@@ -26,7 +26,10 @@
         android:supportsRtl="true"
         android:theme="@style/Theme.ElementX"
         tools:targetApi="33">
-
+        <service
```
bmarty: This declaration should be located in the AndroidManifest of the module where `FloatingVideoService` is defined.

app/src/main/AndroidManifest.xml
```diff
@@ -9,7 +9,7 @@
 
     <!-- To be able to install APK from the application -->
     <uses-permission android:name="android.permission.REQUEST_INSTALL_PACKAGES" />
-
+    <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW" />
```
bmarty: This declaration should be located in the AndroidManifest of the module where `FloatingVideoService` is defined.

libraries/mediaviewer/impl/src/main/res/drawable/ic_close.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="960"
+    android:viewportHeight="960">
+  <path
+      android:pathData="m256,760 l-56,-56 224,-224 -224,-224 56,-56 224,224 224,-224 56,56 -224,224 224,224 -56,56 -224,-224 -224,224Z"
+      android:fillColor="#e3e3e3"/>
+</vector>
```
bmarty: You should not add new icons but use icons provided by Compound. For this one, it is `io.element.android.compound.R.drawable.ic_compound_close`.

libraries/mediaviewer/impl/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?><!--
+  ~ Copyright 2025 New Vector Ltd.
+  ~
+  ~ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+  ~ Please see LICENSE files in the repository root for full details.
+  -->
+
+<resources>
+    <string name="video_playback_error">Video playback error</string>
```
bmarty: We do not add string manually in the project. For now you can just remove it and use `CommonStrings.error_unknown`.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,727 @@
...
+            }
+
+            override fun onStopTrackingTouch(seekBar: SeekBar?) {
+                startProgressUpdater()
+                showControls()
+            }
+        })
+
+        container.addView(overlayContainer)
+        return container
```
bmarty: It would be awesome to use Compose UI to design the View. It will simplify the code a lot (I think), and allow to have previews. Maybe https://gist.github.com/kishan-vadoliya/9fbd1e3c1590de1e4a1a830c5d4edb3f can inspire?

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerView.kt
aliIsazadeh: sure

app/src/main/AndroidManifest.xml
aliIsazadeh: oh of course

libraries/mediaviewer/impl/src/main/res/drawable/ic_close.xml
aliIsazadeh: cool 
what about minimize and maximize icons it seems Compound dose not have them
i will use something else but i will comment in the code that it need to change

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
aliIsazadeh: I will try that and then I will tell you what happend
 thanks

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerView.kt
bmarty: Oh sorry, my comment seems a bit rude, it was not supposed to be sent but was a note for me to not forget. But I forgot to edit it.
I have create a new String `CommonStrings.action_minimize` with this value. It will be available on `develop` next time we sync the string (probably in the coming days, but for sure on next Monday).

libraries/mediaviewer/impl/src/main/res/drawable/ic_close.xml
bmarty: The available compound icons are visible here: https://github.com/element-hq/compound-android/blob/main/compound/screenshots/Compound%20Icons%20-%20Light.png. You may want to use `expand` and `collapse` for your use case?

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerView.kt
aliIsazadeh: 😂 Nah I undrstanded what you mean you were talking about accessibility and hard coded string.

aliIsazadeh:
such a cool checks

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerView.kt
bmarty: oh yes, and the comment should have been i18n of course!

aliIsazadeh: i changed string to unknown and also i add a comment about it needs to be replaced with CommonStrings.action_minimize

app/src/main/AndroidManifest.xml
aliIsazadeh: done

app/src/main/AndroidManifest.xml
aliIsazadeh: done

libraries/mediaviewer/impl/src/main/res/drawable/ic_close.xml
aliIsazadeh: done

libraries/mediaviewer/impl/src/main/res/values/strings.xml
aliIsazadeh: done

app/src/main/AndroidManifest.xml
bmarty: and remove the declaration from here?

libraries/mediaviewer/impl/src/main/res/values/strings.xml
bmarty: not removed?

app/src/main/AndroidManifest.xml
aliIsazadeh: oh shoot sorry i forgot about that 
yeah now i removed it

libraries/mediaviewer/impl/src/main/res/values/strings.xml
aliIsazadeh: now it's removed.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
aliIsazadeh: Quick update on the floating video overlay refactor. I migrated it from XML/ViewBinding to Compose, but ran into an issue: Compose recompositions don’t work reliably in a Service context, even with proper lifecycle setup. State updates happen, but the UI doesn’t recompose.

Because of that, I switched to a hybrid approach: Compose is used for layout, but window size/aspect ratio updates are applied directly via WindowManager and callbacks instead of relying on recomposition.

Main changes:

- > Dynamic aspect ratio for portrait/landscape
- > Removed controller bar (state-driven updates weren’t working) → replaced with tap-to-play/pause gesture
- > Fixed window resizing when toggling minimized/maximized
- > Improved drag/tap gesture handling

So the overlay is Compose-based now, but with direct view updates for stable behavior in Service.
@bmarty

.idea/codeStyles/Project.xml
```diff
@@ -1,5 +1,40 @@
...
+          <emptyLine />
+          <package name="java" withSubpackages="true" static="false" />
+          <emptyLine />
+          <package name="javax" withSubpackages="true" static="false" />
+          <emptyLine />
+          <package name="" withSubpackages="true" static="false" />
+          <emptyLine />
+        </value>
+      </option>
+    </JavaCodeStyleSettings>
```
bmarty: You must revert this change.
I have the same change locally, this is a pain, I have not find a way to get ride of it. The problem is that other developers do not have it so we cannot touch this file on each PR...

gradle.properties
```diff
@@ -54,3 +54,7 @@ ksp.allow.all.target.configuration=false
 
 # Used to prevent detekt from reusing invalid cached rules
 detekt.use.worker.api=true
+org.gradle.java.home=/usr/lib/jvm/java-17-openjdk-amd64
+org.gradle.jvmargs=-Xmx4g -Dfile.encoding=UTF-8
+kotlin.compiler.execution.strategy=in-process
+kotlin.daemon.jvm.options=-Xmx2g
```
bmarty: Please revert all the changes in this file. You can place your local configuration in the file `local.properties` instead, which is ignore by VCS.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,520 @@
...
+import androidx.savedstate.SavedStateRegistryController
+import androidx.savedstate.SavedStateRegistryOwner
+import androidx.savedstate.setViewTreeSavedStateRegistryOwner
+import io.element.android.libraries.matrix.api.media.MediaSource
+import io.element.android.libraries.designsystem.theme.components.Icon
+import io.element.android.libraries.ui.strings.CommonStrings
+import java.io.File
+import androidx.core.net.toUri
+
+class FloatingVideoService : Service(), LifecycleOwner, ViewModelStoreOwner, SavedStateRegistryOwner {
```
bmarty: We try to avoid long file. Please split the code into multiple file. The UI can be extracted, it will force the code to be a bit more decoupled.

Also it will allow to add preview of the UI, for instance:

```kotlin
@PreviewsDayNight
@Composable
internal fun FloatingVideoOverlayPreview() = ElementPreview {
    FloatingVideoOverlay(
        // params
    )
}
```

The method `getVideoUriFromMediaSource` could be in a separate file too.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,520 @@
...
+        } else {
+            val displayMetrics = DisplayMetrics()
+            @Suppress("DEPRECATION") windowManager?.defaultDisplay?.getMetrics(displayMetrics)
+            displayMetrics.heightPixels
+        }
+    }
+}
+
+@OptIn(UnstableApi::class)
+fun getVideoUriFromMediaSource(mediaSource: MediaSource): Uri {
```
bmarty: It would be nicer to use an extension, which also allow to simplify the name:
`fun MediaSource.getVideoUri(): Uri {`
Using a block like `with(url) {}` could also simplify the code, but I am not a big fan of having several `this` in a block.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,520 @@
...
+    return try {
+        val url = mediaSource.url
```
bmarty: then you can just remove this line.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,520 @@
...
+        Log.d("VideoPlayer", "MediaSource URL: $url")
```
bmarty: We use Timber on this project, please do not use `Log`.
```
Timber.tag("VideoPlayer").d("MediaSource URL: $url")
```

Also logging the url should be avoided for privacy reason. The log can be sent by the user to our company, so we avoid logging private data (the url could contains sensitive information)

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,520 @@
...
+                        }
+                    }
+            )
+
+            Row(
+                modifier = Modifier
+                    .align(Alignment.TopCenter)
+                    .fillMaxWidth()
+                    .background(
+                        brush = androidx.compose.ui.graphics.Brush.verticalGradient(
```
bmarty: please use import (there are several place to fix in this file)

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,520 @@
...
+            videoView = null
+        }
+    }
+
+    @Composable
+    fun FloatingVideoOverlay(
+        onClose: () -> Unit,
+        onToggleFullScreen: () -> Unit
+    ) {
+        var currentAspectRatio by remember { mutableStateOf(16f / 9f) }
```
bmarty: Please use `mutableFloatStateOf`

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,520 @@
...
+import androidx.compose.foundation.gestures.detectDragGestures
+import androidx.compose.foundation.gestures.detectTapGestures
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.material.IconButton
```
bmarty: You must use `io.element.android.libraries.designsystem.theme.components.IconButton` instead.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,520 @@
...
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Close
+import androidx.compose.material.icons.filled.Fullscreen
```
bmarty: You should use CompoundIcon instead of Icon from material.

bmarty (changes requested):
Thanks for the update and the migration to Compose!
I have added more remarks.

I have also tested the feature and it seems to work pretty well. I would maybe add some padding, here the minimized video was displayed at the bottom start of the screen. Also maybe having a way to Play/Pause the video on the minimize video would be awesome. Last point, the View should maybe automatically close at the end of the Playback?

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/VideoDataRepository.kt
```diff
@@ -0,0 +1,42 @@
...
+    companion object {
+        @Volatile
+        private var INSTANCE: VideoDataRepository? = null
+
+        fun getInstance(): VideoDataRepository {
+            return INSTANCE ?: synchronized(this) {
+                INSTANCE ?: VideoDataRepository().also { INSTANCE = it }
+            }
+        }
+    }
```
bmarty: We used DI on this project, so you should remove this companion and let the DI handle the singleton:
```
@Inject
@SingleIn(AppScope::class)
class VideoDataRepository {
```

Then `startFloating` could have a `VideoDataRepository` parameter and `VideoDataRepository` could be injected in `FloatingVideoService`. You can see how members are injected in the Activites for instance.

.idea/codeStyles/Project.xml
aliIsazadeh: oh sure i put  this file in gitignore from now on.

gradle.properties
aliIsazadeh: My apologies — I’m not sure when these changes were made and committed. It wasn’t intentional.

aliIsazadeh:
for play and pause just tap on the video

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/VideoDataRepository.kt
aliIsazadeh: done

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
aliIsazadeh: done

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
aliIsazadeh: done

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
aliIsazadeh: done

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
aliIsazadeh: Done

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
aliIsazadeh: Done.
and you are right  it was there for test.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
aliIsazadeh: Done.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/util/VideoUriFromMediaSource.kt
```diff
@@ -0,0 +1,45 @@
...
+
+import android.net.Uri
+import androidx.core.net.toUri
+import io.element.android.libraries.matrix.api.media.MediaSource
+import timber.log.Timber
+import java.io.File
+
+fun MediaSource.getVideoUriFromMediaSource () : Uri{
+    return try {
+        val url = this.url
```
bmarty: This line can just be removed.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/util/VideoUriFromMediaSource.kt
```diff
@@ -0,0 +1,45 @@
...
+
+package io.element.android.libraries.mediaviewer.impl.floatingvideo.util
+
+import android.net.Uri
+import androidx.core.net.toUri
+import io.element.android.libraries.matrix.api.media.MediaSource
+import timber.log.Timber
+import java.io.File
+
+fun MediaSource.getVideoUriFromMediaSource () : Uri{
```
bmarty: Rename to `getUri`? No need to repeat MediaSource as it's the receiver and also no need to precise `Video`, it can stay generic.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/util/VideoUriFromMediaSource.kt
```diff
@@ -0,0 +1,45 @@
...
+                // Remote URL
+                url.toUri()
+            }
+            url.startsWith("file://") -> {
+                // Already a file URI
+                url.toUri()
+            }
+            url.startsWith("/") -> {
+                // Local file path, convert to file URI
+                Uri.fromFile(File(url))
```
bmarty: Maybe change to

```suggestion
                File(url).toUri()
```

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/ui/FloatingVideoOverlay.kt
```diff
@@ -0,0 +1,214 @@
...
+
+@Composable
+fun FloatingVideoOverlay(
+    onClose: () -> Unit,
+    onToggleFullScreen: (Float) -> Unit,
+    currentVideoData : MediaViewerPageData.MediaViewerData?,
+    isMaximized : Boolean,
+    windowManager : WindowManager?,
+    windowLayoutParams : WindowManager.LayoutParams,
+    floatingView : View?,
```
bmarty: Instead of providing `windowManager` and `floatingView`, `isMaximized`, I'll provide some lambda parameter and handle all the computation in `FloatingVideoService`.

At the end, the signature should be something like:

```kotlin
fun FloatingVideoOverlay(
    uri: Uri?,
    onClose: () -> Unit,
    onToggleFullScreen: () -> Unit,
    updateAspectRatio: (Float) -> Unit,
    movePosition: (Int, Int) -> Unit,
)
```

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/ui/FloatingVideoOverlay.kt
```diff
@@ -0,0 +1,214 @@
...
+    onCompleted : () -> Unit
+) {
+    var currentAspectRatio by remember { mutableFloatStateOf(16f / 9f) }
+    val videoViewRef = remember { mutableStateOf<VideoView?>(null) }
+
+
+
+    var resolvedUri: Uri = Uri.EMPTY
```
bmarty: This should be remembered

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,305 @@
...
+    private var currentVideoData: MediaViewerPageData.MediaViewerData? = null
+    private var currentPosition: Long = 0L
+    private var isMaximized = true
+
+    private lateinit var windowLayoutParams: WindowManager.LayoutParams
+
+    companion object {
+        const val ACTION_START_FLOATING = "START_FLOATING"
+        const val ACTION_STOP_FLOATING = "STOP_FLOATING"
+        const val ACTION_UPDATE_POSITION = "UPDATE_POSITION"
```
bmarty: It appears to be dead code, can you remove please?

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,305 @@
...
+                    addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                }
+
+
+                context.startActivity(intent)
+                return
+            }
+
+            // Generate unique ID for this video session
+            val videoId = "floating_video_${System.currentTimeMillis()}"
```
bmarty: Better to use `UUID.randomUUID().toString()` to generate an ID. The id could also be created by the videoDataRepository and returned by the method `storeVideoData`.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,305 @@
...
+
+            // Store the video data in repository via DI
+            context.bindings<FloatingVideoServiceBindings>().videoDataRepository().storeVideoData(videoId, videoData)
```
bmarty: Instead of doing this, `MediaViewerNode` could have the video data repository and store the data before invoking `startFloating`. The `videoId` would be a parameter of `startFloating`.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,305 @@
...
+import androidx.savedstate.SavedStateRegistryOwner
+import androidx.savedstate.setViewTreeSavedStateRegistryOwner
+import androidx.core.net.toUri
+import io.element.android.libraries.mediaviewer.impl.floatingvideo.ui.FloatingVideoOverlay
+import io.element.android.libraries.mediaviewer.impl.floatingvideo.util.getScreenHeight
+import io.element.android.libraries.mediaviewer.impl.floatingvideo.util.getScreenWidth
+import dev.zacsweers.metro.Inject
+import io.element.android.libraries.architecture.bindings
+
+class FloatingVideoService : Service(), LifecycleOwner, ViewModelStoreOwner, SavedStateRegistryOwner {
```
bmarty: Does this service need to implement `ViewModelStoreOwner`? Also `viewModelStore` is unused, just remove it?

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,305 @@
...
+
+        windowLayoutParams.width = width
+        windowLayoutParams.height = height
+        windowLayoutParams.flags = WindowManager.LayoutParams.FLAG_NOT_FOCUSABLE or
+            WindowManager.LayoutParams.FLAG_LAYOUT_NO_LIMITS
+        windowManager?.updateViewLayout(floatingView, windowLayoutParams)
+
+        Handler(Looper.getMainLooper()).post {
+            wm.updateViewLayout(view, layoutParams)
+        }
```
bmarty: Please replace by:
```
        lifecycleScope.launch(coroutineDispatchers.io) {
            wm.updateViewLayout(view, layoutParams)
        }
```

coroutineDispatchers can be injected.

Same thing for the Looper used in `FloatingVideoOverlay`. All the treatment done in `FloatingVideoOverlay` must be moved to the service.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,305 @@
...
+import androidx.savedstate.setViewTreeSavedStateRegistryOwner
+import androidx.core.net.toUri
+import io.element.android.libraries.mediaviewer.impl.floatingvideo.ui.FloatingVideoOverlay
+import io.element.android.libraries.mediaviewer.impl.floatingvideo.util.getScreenHeight
+import io.element.android.libraries.mediaviewer.impl.floatingvideo.util.getScreenWidth
+import dev.zacsweers.metro.Inject
+import io.element.android.libraries.architecture.bindings
+
+class FloatingVideoService : Service(), LifecycleOwner, ViewModelStoreOwner, SavedStateRegistryOwner {
+    private var windowManager: WindowManager? = null
```
bmarty, line 47:
```suggestion
    private val windowManager = getSystemService<WindowManager>()!!
```

The code will be simplified by this change, and I believe that the system always provides the `WindowManager` service.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,305 @@
...
+    private var floatingView: View? = null
+    private var videoView: VideoView? = null
+    private var currentVideoData: MediaViewerPageData.MediaViewerData? = null
+    private var currentPosition: Long = 0L
+    private var isMaximized = true
+
+    private lateinit var windowLayoutParams: WindowManager.LayoutParams
```
bmarty, line 54: I think it's possible to remove lateinit, change to val and initialize the value here.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,305 @@
...
+                Toast.makeText(context, "To show the floating video, please allow 'Display over other apps' permission.", Toast.LENGTH_LONG).show()
+
+                // Request overlay permission
+                val intent = Intent(Settings.ACTION_MANAGE_OVERLAY_PERMISSION).apply {
+                    data = "package:${context.packageName}".toUri()
+                    addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                }
+
+
+                context.startActivity(intent)
```
bmarty: This code should be moved to `SystemUtils.kt` and the exception for `ActivityNotFoundException` should be caught.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerNode.kt
```diff
@@ -132,20 +134,27 @@ class MediaViewerNode(
...
+            Box(modifier = modifier.fillMaxSize()) {
+                MediaViewerView(
+                    state = state,
+                    textFileViewer = textFileViewer,
+                    modifier = modifier,
+                    audioFocus = audioFocus,
+                    onBackClick = ::onDone,
+                    setMinimize = setMinimized
+                )
+                if (isMinimized && data?.mediaInfo?.mimeType.isMimeTypeVideo() && data != null) {
```
bmarty: Checking for `data != null` first will reduce the number of needed `?`

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerNode.kt
```diff
@@ -132,20 +134,27 @@ class MediaViewerNode(
...
+
+                    FloatingVideoService.startFloating(context, data, 0L)
```
bmarty: A LaunchEffect is missing here, else the service will be started on each recomposition. The final code should be something like:
```kotlin
    @Composable
    override fun View(modifier: Modifier) {
        val context = LocalContext.current
        val (isMinimized, setMinimized) = remember { mutableStateOf(false) }
        ForcedDarkElementTheme {
            val state = presenter.present()
            MediaViewerView(
                state = state,
                textFileViewer = textFileViewer,
                modifier = modifier,
                audioFocus = audioFocus,
                onBackClick = ::onDone,
                setMinimize = setMinimized
            )
            LaunchedEffect(isMinimized) {
                if (isMinimized) {
                    val data = state.listData
                        .getOrNull(state.currentIndex) as? MediaViewerPageData.MediaViewerData
                    if (data != null && data.mediaInfo.mimeType.isMimeTypeVideo()) {
                        // Store the video data in repository
                        val videoId = videoDataRepository.storeVideoData(data)
                        FloatingVideoService.startFloating(context, videoId, 0L)
                    }
                }
            }
        }
    }
```

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerNode.kt
```diff
@@ -132,20 +134,27 @@ class MediaViewerNode(
...
+        ForcedDarkElementTheme {
             val state = presenter.present()
-            MediaViewerView(
-                state = state,
-                textFileViewer = textFileViewer,
-                modifier = modifier,
-                audioFocus = audioFocus,
-                onBackClick = callback::onDone,
-            )
+            val data = state.listData
```
bmarty, line 154: This need to be moved below

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerNode.kt
```diff
@@ -132,20 +134,27 @@ class MediaViewerNode(
...
+                .getOrNull(state.currentIndex) as? MediaViewerPageData.MediaViewerData
+            Box(modifier = modifier.fillMaxSize()) {
```
bmarty, line 164: No need to add a Box here I think.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoServiceBindings.kt
```diff
@@ -0,0 +1,19 @@
...
+
+package io.element.android.libraries.mediaviewer.impl.floatingvideo
+
+import dev.zacsweers.metro.AppScope
+import dev.zacsweers.metro.ContributesTo
+
+@ContributesTo(AppScope::class)
+interface FloatingVideoServiceBindings {
+    fun inject(service: FloatingVideoService)
+    fun videoDataRepository(): VideoDataRepository
```
bmarty, line 16: This fun can be removed if you handle my other remarks.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,305 @@
...
+
+        // 2. Restore state (if any)
+        savedStateRegistryController.performRestore(null)
+
+        // 3. Now move lifecycle forward
+        lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_CREATE)
+    }
+
+    private var currentVideoId: String? = null
+    private var eventId: String? = null
```
bmarty, line 113: Not used, remove?

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
```diff
@@ -0,0 +1,305 @@
...
+            layoutParams.height = WindowManager.LayoutParams.WRAP_CONTENT
+            layoutParams.x = 0
+            layoutParams.y = 0
+        }
+
+        windowLayoutParams.width = width
+        windowLayoutParams.height = height
+        windowLayoutParams.flags = WindowManager.LayoutParams.FLAG_NOT_FOCUSABLE or
+            WindowManager.LayoutParams.FLAG_LAYOUT_NO_LIMITS
+        windowManager?.updateViewLayout(floatingView, windowLayoutParams)
```
bmarty: This seems to be done twice (here and below), can you check?

aliIsazadeh:
Thanks for the detailed review. I’ll apply these changes and I really appreciate the explanations.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
aliIsazadeh: thanks for your recomendation but it raised and error while creating the service 
                                                                                                    java.lang.RuntimeException: Unable to create service io.element.android.libraries.mediaviewer.impl.floatingvideo.FloatingVideoService: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String android.content.Context.getSystemServiceName(java.lang.Class)' on a null object reference

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/floatingvideo/FloatingVideoService.kt
aliIsazadeh: I dont see how to remove ACTION_START_FLOATING  
to me it seems to be essential to have it

aliIsazadeh:
I wanted to give you a quick update on the minimized-video branch. While trying to merge develop into minimized-video and handle conflicts, I accidentally merged the two branches. Now both branches are identical.

I also tried to fix the history via rebase, but it didn’t work as expected. I plan to attempt a proper rebase and cleanup later to restore a clean commit history.

I’ll update you once it’s resolved.

Thanks for your understanding.

PR #5424, opened 2025-09-29
Sync Strings

ElementBot:


bmarty:
Will trigger another sync when a change on a string will be merged.

Issue #4864, opened 2025-06-11
EXA: When no other devices are verified, don't offer "Use another device" for verification


mxandreas:
### Steps to reproduce

1. Log out everywhere.
2. Login with EW but do not verify your device.
3. Login with EXA, on the "Confirm your identity" screen click "Use another device" and follow the process.

#### What did you expect?

Not to have the option "Use another device" if there are no verified devices available.

#### What happened instead?

It was possible to use "Use another device" and the verification process starts as "business as usual", until it gets stuck on the emoji screen.

Note:  EXI has the same [issue](https://github.com/element-hq/element-x-ios/issues/4190), the difference is that it gets even further in the process. EW does not have this bug, e.g. it does not offer to use another device if there is no verified device.

### Will you send logs?

Yes

### Are you willing to provide a PR?

Yes

PR #5433, opened 2025-09-30
Only offer to verify if a cross-signed device is available

uhoreg:
Fixes https://github.com/element-hq/element-x-android/issues/4864

## Content

Currently, on logging in, it will offer to verify against another device if the user has any other devices, even if they are not cross-signed. This change uses the function introduced in https://github.com/matrix-org/matrix-rust-sdk/pull/5699 to check whether the user has any cross-signed devices, and will only offer to verify if one is available.

## Motivation and context

https://github.com/element-hq/element-x-android/issues/4864

## Screenshots / GIFs

image

No "Use another device" button

## Tests

- Signed in a test user with Element Web, but did not verify
- Ensured that the user had no other devices
- Signed in to Element X, and check that "Confirm your dentity" screen did not offer to "Use another device"
- Signed out of Element X
- Reset cross-signing in Element Web
- Signed in to Element X, and check that "Confirm your identity" screen now offers to "Use another device"

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Android 16 (API 36.0)

## Checklist

### DIFF
--- a/features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/choosemode/ChooseSelfVerificationModePresenter.kt
+++ b/features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/choosemode/ChooseSelfVerificationModePresenter.kt
@@ -26,7 +26,7 @@ class ChooseSelfVerificationModePresenter(
 ) : Presenter<ChooseSelfVerificationModeState> {
     @Composable
     override fun present(): ChooseSelfVerificationModeState {
-        val isLastDevice by encryptionService.isLastDevice.collectAsState()
+        val hasDevicesToVerifyAgainst by encryptionService.hasDevicesToVerifyAgainst.collectAsState()
         val recoveryState by encryptionService.recoveryStateStateFlow.collectAsState()
         val canEnterRecoveryKey by remember { derivedStateOf { recoveryState == RecoveryState.INCOMPLETE } }
 
@@ -39,7 +39,7 @@ class ChooseSelfVerificationModePresenter(
         }
 
         return ChooseSelfVerificationModeState(
-            isLastDevice = isLastDevice,
+            isLastDevice = !hasDevicesToVerifyAgainst,
             canEnterRecoveryKey = canEnterRecoveryKey,
             directLogoutState = directLogoutState,
             eventSink = ::eventHandler,
--- a/features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/choosemode/ChooseSelfVerificationModePresenter.kt
+++ b/features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/choosemode/ChooseSelfVerificationModePresenter.kt
@@ -39,7 +39,7 @@ class ChooseSelfVerificationModePresenter(
         }
 
         return ChooseSelfVerificationModeState(
-            isLastDevice = !hasDevicesToVerifyAgainst,
+            canUseAnotherDevice = hasDevicesToVerifyAgainst,
             canEnterRecoveryKey = canEnterRecoveryKey,
             directLogoutState = directLogoutState,
             eventSink = ::eventHandler,
--- a/features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/choosemode/ChooseSelfVerificationModeStateProvider.kt
+++ b/features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/choosemode/ChooseSelfVerificationModeStateProvider.kt
@@ -13,18 +13,18 @@ import io.element.android.features.logout.api.direct.aDirectLogoutState
 class ChooseSelfVerificationModeStateProvider :
     PreviewParameterProvider<ChooseSelfVerificationModeState> {
     override val values = sequenceOf(
-        aChooseSelfVerificationModeState(isLastDevice = true, canEnterRecoveryKey = true),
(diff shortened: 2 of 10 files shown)


uhoreg:
Trying to get Maestro running locally so I can look into the failure, because the CI log isn't very helpful.  The code coverage seems to be complaining about missing a test on the `mapRecoveryException` line.  Otherwise, I think this is pretty much ready for review.

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/choosemode/ChooseSelfVerificationModePresenter.kt
```diff
@@ -39,7 +39,7 @@ class ChooseSelfVerificationModePresenter(
         }
 
         return ChooseSelfVerificationModeState(
-            isLastDevice = isLastDevice,
+            isLastDevice = !hasDevicesToVerifyAgainst,
```
bmarty: For clarity / reduce tech debt, I think `ChooseSelfVerificationModeState.isLastDevice` could be renamed to `canUseAnotherDevice`. Do not forget to negate the boolean wherever this is necessary.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/encryption/RustEncryptionService.kt
```diff
@@ -96,6 +96,15 @@ internal class RustEncryptionService(
...
         .stateIn(sessionCoroutineScope, SharingStarted.Eagerly, false)
 
+    override val hasDevicesToVerifyAgainst: StateFlow<Boolean> = flow {
+        while (currentCoroutineContext().isActive) {
+            val result = hasDevicesToVerifyAgainst().getOrDefault(false)
+            emit(result)
+            delay(5_000)
+        }
+    }
+        .stateIn(sessionCoroutineScope, SharingStarted.Eagerly, false)
```
bmarty, line 112: Maybe add the same TODO than in the KDoc of `isLastDevice`. It is a bit annoying to request this value every 5s, when it's only used in the session verification screen.

bmarty (commented):
Thanks!

`isLastDevice` is also used in LogoutView, to warn the user if this is the last session. I am wondering if the new `hasDevicesToVerifyAgainst` could be used there too. What do you think?

Also, having this screen:

image

with actually no options to confirm seems a bit annoying for the user. We may want to iterate on the design in this case.
Can't we enter the recovery key?

uhoreg:
> Also, having this screen:
> image
> 
> with actually no options to confirm seems a bit annoying for the user. We may want to iterate on the design in this case. Can't we enter the recovery key?

Yes, if the user has a recovery key set up, they should have a button to use the recovery key.  In this case, my test user did not have a recovery key, so no button was shown, and the user's only option was "Can't confirm?", which resets the user's cross-signing keys (which is the only thing that can be done in this case).

uhoreg:
> `isLastDevice` is also used in LogoutView, to warn the user if this is the last session. I am wondering if the new `hasDevicesToVerifyAgainst` could be used there too. What do you think?

I'm not sure.  I just checked what Element Web did here, and it looks like it's only checking the backup status.  As far as I can tell, it doesn't seem to check at all whether there are any other devices.  So we might need to make a new issue to make the clients consistent, and figure out what criteria we should use to warn the user.  Anyways, I don't think `hasDevicesToVerifyAgainst` will be quite right, because it specifically excludes dehydrated devices, and if the user has a dehydrated device, they shouldn't be shown a warning on logout.

uhoreg:
Tried to reproduce the CI failures.  For [Test / Runs unit tests](https://github.com/element-hq/element-x-android/actions/runs/18177517049/job/51746618887?pr=5433) the failing task succeeds on my machine.  For [Maestro (local) / Maestro test suite](https://github.com/element-hq/element-x-android/actions/runs/18177517057/job/51747203781?pr=5433) I'm having trouble getting the Maestro suite to complete, but it gets past the login section, which is the only place my PR should affect.

features/ftue/impl/src/main/kotlin/io/element/android/features/ftue/impl/sessionverification/choosemode/ChooseSelfVerificationModeStateProvider.kt
```diff
@@ -13,18 +13,18 @@ import io.element.android.features.logout.api.direct.aDirectLogoutState
...
     PreviewParameterProvider<ChooseSelfVerificationModeState> {
     override val values = sequenceOf(
-        aChooseSelfVerificationModeState(isLastDevice = true, canEnterRecoveryKey = true),
-        aChooseSelfVerificationModeState(isLastDevice = true, canEnterRecoveryKey = false),
-        aChooseSelfVerificationModeState(isLastDevice = false, canEnterRecoveryKey = true),
-        aChooseSelfVerificationModeState(isLastDevice = false, canEnterRecoveryKey = false),
+        aChooseSelfVerificationModeState(canUseAnotherDevice = true, canEnterRecoveryKey = true),
+        aChooseSelfVerificationModeState(canUseAnotherDevice = true, canEnterRecoveryKey = false),
+        aChooseSelfVerificationModeState(canUseAnotherDevice = false, canEnterRecoveryKey = true),
+        aChooseSelfVerificationModeState(canUseAnotherDevice = false, canEnterRecoveryKey = false),
```
bmarty, line 17: The failing tests are related to the screenshot test, which are not run by default when running the test locally.
To avoid this problem, I think you need to use the opposite values for all boolean assigned to `canUseAnotherDevice`, and so the screenshot will be identical to the current version.

bmarty (approved):
Thanks!

PR #5440, opened 2025-10-01
Remove CurrentSessionIdHolder and inject SessionId instead.

bmarty:
## Content

We do not need CurrentSessionIdHolder anymore. The SessionId can be provided by SessionMatrixModule and injected in constructors directly.

Also improve and simplify a bit SignedOutPresenter.

## Motivation and context

Cleaner code.

## Tests

- No expected change

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/signedout/impl/src/test/kotlin/io/element/android/features/signedout/impl/DefaultSignedOutEntryPointTest.kt
+++ b/features/signedout/impl/src/test/kotlin/io/element/android/features/signedout/impl/DefaultSignedOutEntryPointTest.kt
@@ -28,7 +28,7 @@ class DefaultSignedOutEntryPointTest {
                 buildContext = buildContext,
                 plugins = plugins,
                 presenterFactory = { sessionId ->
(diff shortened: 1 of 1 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/di/SessionMatrixModule.kt
```diff
@@ -27,6 +28,11 @@ import kotlinx.coroutines.CoroutineScope
 @BindingContainer
 @ContributesTo(SessionScope::class)
 object SessionMatrixModule {
+    @Provides
+    fun providesSessionId(matrixClient: MatrixClient): SessionId {
```
bmarty, line 32: It could have `@SingleIn(SessionScope::class)`, but as all the other method in this object so not sure that it worth it.

ganfra (approved):

PR #5457, opened 2025-10-03
Internal compound

bmarty:
## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


libraries/compound/build.gradle.kts
```diff
@@ -12,13 +12,6 @@ plugins {
 android {
     namespace = "io.element.android.compound"
 
-    defaultConfig {
-        vectorDrawables {
-            useSupportLibrary = true
-            generatedDensities()
-        }
-    }
-
```
jmartinesp: I think not having this caused issues when trying to use vector icons in old API versions. It may be worth it to double check this: https://github.com/element-hq/compound-android/pull/1

Maybe now they're part of the same project and we bumped the min API version to 24 we won't have this issue.

build.gradle.kts
```diff
@@ -180,16 +181,22 @@ tasks.register("runQualityChecks") {
 // Make sure to delete old screenshots before recording new ones
 subprojects {
     val snapshotsDir = File("${project.projectDir}/src/test/snapshots")
+    val snapshotsDir2 = File("${project.projectDir}/screenshots")
```
jmartinesp: Nit: maybe `compoundSnapshotsDir`?

jmartinesp (approved):
LGTM, thanks. We should probably double check the icons work fine on API 24 after removing the code to disable PNG rasterization.

build.gradle.kts
bmarty: I have updated this code because it was bugged. Both sets of screenshots could not be deleted when we want to record one set 😅

bmarty:
> LGTM, thanks. We should probably double check the icons work fine on API 24 after removing the code to disable PNG rasterization.

Some icons on API 24 are bugged, with both the rasterization enabled or not. I will investigate the issue later.

libraries/compound/build.gradle.kts
bmarty: I checked on API 24 and there is no crash, but I will double check again.

bmarty: OK, so there is no crash on API 24, but we can see that some icons are broken. It's not new, so it's not blocking this PR.

image

PR #5460, opened 2025-10-06
Sync Strings

ElementBot:


bmarty (approved):
OK, but we should trigger a new sync before the release. I have spotted some issue in the French translation and we want more translation for the space announcement screen.

PR #5462, opened 2025-10-06
Add support for Spaces

bmarty:
## Content

Enable Spaces by default.

Please not that is the Spaces FF had been enabled then disabled, this PR will have no effect and Spaces will remain disabled.

## Motivation and context

Let Spaces be visible by deafult for the next release

## Tests

- Start the application
- See that Spaces are enabled by default.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
+++ b/libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
@@ -71,8 +71,7 @@ enum class FeatureFlags(
     Space(
         key = "feature.space",
         title = "Spaces",
-        description = "Spaces are under active development, only developers should enable this flag for now.",
-        defaultValue = { false },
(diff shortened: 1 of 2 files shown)


libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
```diff
@@ -71,8 +71,7 @@ enum class FeatureFlags(
     Space(
         key = "feature.space",
         title = "Spaces",
-        description = "Spaces are under active development, only developers should enable this flag for now.",
-        defaultValue = { false },
+        defaultValue = { true },
```
ganfra, line 74: Maybe we should keep a short description?

bmarty: that's the developer settings...

Issue #5430, opened 2025-09-30
[Task] Introduce a Labs section in EX


manuroe:
Please see and discuss the details in the meta issue.

PR #5465, opened 2025-10-06
Add Labs screen for beta testing of public features

jmartinesp:
## Content

- Make `Feature` have an `isInLabs` boolean to distinguish private feature flags from public ones.
- Have `FeatureFlagsService` provide the list of available flags.
- Display the labs item in the settings screen only if there are available public features.
- Remove public feature toggles from developer options.
- Implement the labs screen with the public features.

## Motivation and context

Implements https://github.com/element-hq/element-x-android/issues/5430.

## Screenshots / GIFs

image

image

## Tests

- Open Settings, the Labs item should be visible.
- Tap on it, you should see the 'Threads' feature. When toggled, it will enable/disable threads. Note: this needs an app restart, but the designs don't include that anywhere. I've asked the design team about this.
- If you go to developer settings, the 'Threads' item should be gone now.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/developer/DeveloperSettingsPresenter.kt
```diff
@@ -90,8 +90,8 @@ class DeveloperSettingsPresenter(
         }
 
         LaunchedEffect(Unit) {
-            FeatureFlags.entries
-                .filter { it.isFinished.not() }
+            featureFlagService.getAvailableFeatures()
```
bmarty, line 93: `getAvailableFeatures()` is already filtering with `!isFinished` so the filter below could be simplified.
Or maybe to reduce duplication of code, `getAvailableFeatures` could take a boolean parameter `isInLabs` to handle all the filtering?

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/labs/LabsEvents.kt
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright 2025 New Vector Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.features.preferences.impl.labs
+
+sealed interface LabsEvents {
+    data class ToggleFeature(val feature: String) : LabsEvents
```
bmarty: I will not use the `Feature.key` for the parameter to avoid mistake if a developer use `Feature.title`. I would change the API to 

```suggestion
    data class ToggleFeature(val feature: FeatureUiModel) : LabsEvents
```

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/labs/LabsPresenter.kt
```diff
@@ -0,0 +1,124 @@
...
+                val title = when (feature) {
+                    FeatureFlags.Threads -> stringProvider.getString(R.string.screen_labs_enable_threads)
+                    else -> feature.title
+                }
+                val description = when (feature) {
+                    FeatureFlags.Threads -> stringProvider.getString(R.string.screen_labs_enable_threads_description)
+                    else -> feature.description
+                }
+                val icon = when (feature) {
+                    FeatureFlags.Threads -> CompoundIcons.Threads()
```
bmarty, line 109: I would have let the View decide which string resource / Icon to use depending on the feature type, but this code works too, even if it looks a bit harder to read (to me!)

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/labs/LabsView.kt
```diff
@@ -0,0 +1,103 @@
...
+import io.element.android.libraries.designsystem.components.button.BackButton
+import io.element.android.libraries.designsystem.components.list.ListItemContent
+import io.element.android.libraries.designsystem.components.list.SwitchListItem
+import io.element.android.libraries.designsystem.preview.ElementPreview
+import io.element.android.libraries.designsystem.preview.PreviewsDayNight
+import io.element.android.libraries.designsystem.theme.components.TopAppBar
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun LabsView(
```
bmarty, line 44: Could be nice to have a Figma link in the KDoc (https://www.figma.com/design/V0dkfRAW6T3yCQKjahpzkX/ER-46-EX--Threads?node-id=2004-27319&t=yssy1yYYigsGON3s-0).

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/root/PreferencesRootPresenter.kt
```diff
@@ -112,6 +112,8 @@ class PreferencesRootPresenter(
                 .launchIn(this)
         }
 
+        val showLabsItem = remember { featureFlagService.getAvailableFeatures().any { it.isInLabs && !it.isFinished } }
```
bmarty, line 115: Kudos for checking if there are any labs feature to show its entry point!

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/labs/LabsView.kt
```diff
@@ -0,0 +1,103 @@
...
+            TopAppBar(
+                titleStr = stringResource(R.string.screen_labs_title),
+                navigationIcon = {
+                    BackButton(onClick = onBack, enabled = !state.isApplyingChanges)
+                }
+            )
+        },
+        header = {
+            IconTitleSubtitleMolecule(
+                modifier = Modifier.padding(horizontal = 24.dp),
```
bmarty: Need a top padding of `24.dp` here.

bmarty (approved):
Nice work, thanks! Really nice solution to use the clear cache use case in order to restart the app!
Only minor remarks, tested OK.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/developer/DeveloperSettingsPresenter.kt
jmartinesp: It's actually duplicated so we can check this has the right behaviour in unit tests (i.e. if `getAvailableFeatures` returned some unfinished features we'd still remove them).

> Or maybe to reduce duplication of code, `getAvailableFeatures` could take a boolean parameter isInLabs to handle all the filtering?

I thought about that, but this seems more related to how we want to handle some data in the UI than something that should be part of the logic of the service.

bmarty: Coming back here since I see that my remark was not handled :). Sorry I have missed your reply.

> It's actually duplicated so we can check this has the right behaviour in unit tests (i.e. if getAvailableFeatures returned some unfinished features we'd still remove them).

It feels wrong to duplicate the filtering in order to have to write extra test, no?

> I thought about that, but this seems more related to how we want to handle some data in the UI than something that should be part of the logic of the service.

I see the parameter `isInLabs` as an API with a higher level of service, so it's not a problem for me. We write the services to help the UI, so why not?

Since I am in this area for other reasons, I'll add some commits and we can discuss this in the coming PR.

PR #5469, opened 2025-10-07
Set a notification sound

bmarty:
## Content

This PR takes the sound from https://github.com/element-hq/element-web/pull/30804 and the spec and design from https://github.com/element-hq/element-x-ios/pull/4572

I had to update the [migration](https://github.com/element-hq/element-x-android/commit/d9d14f1abce077e43f07f9bc4ecb172972d74e03) codebase to be able to migrate differently in the case of app install of app upgrade.

I also had to set a new notification channel id since setting a new sound has no effect on already registered channels.

## Motivation and context

All Element clients have the same notification sound.

## Screenshots / GIFs

See recorded one. May include unrelated change due to the sync of strings.

## Tests

- Fresh install the app and login: the new notification sound banner is not displayed and will never be
- Install a previous version and login
- Install a build from this PR
- observe that a new notification sound banner is displayed and can be dismissed forever.
- ensure the device has notification sound volume is high enough
- receive a message which would trigger a noisy notification
- enjoy the new notification sound.

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s):

## Checklist


features/home/impl/src/main/kotlin/io/element/android/features/home/impl/roomlist/RoomListPresenter.kt
```diff
@@ -98,6 +98,7 @@ class RoomListPresenter(
         }
 
         var securityBannerDismissed by rememberSaveable { mutableStateOf(false) }
+        val showNewNotificationSoundBanner by appPreferencesStore.showNewNotificationSoundBanner().collectAsState(false)
```
ganfra, line 101: Would be great to use the `Announcement` feature no?

libraries/push/impl/src/main/res/raw/message.mp3
ganfra, line 1: Can we find a more explicit name than message.mp3 ?

features/home/impl/src/main/kotlin/io/element/android/features/home/impl/roomlist/RoomListPresenter.kt
bmarty: Yes, maybe, but here it's more embedded in the UI so it's probably easier like this.

libraries/push/impl/src/main/res/raw/message.mp3
bmarty: I kept the original name from ElementWeb for consistency: https://github.com/element-hq/element-web/tree/develop/res/media

PR #5486, opened 2025-10-08
Sync Strings

ElementBot:


features/login/impl/src/main/res/values-fr/translations.xml
```diff
@@ -39,7 +39,7 @@
   <string name="screen_login_title_with_homeserver">"Connectez-vous à %1$s"</string>
   <string name="screen_onboarding_app_version">"Version %1$s"</string>
   <string name="screen_onboarding_sign_in_manually">"Se connecter manuellement"</string>
-  <string name="screen_onboarding_sign_in_to">"Se connecter à %1$s"</string>
+  <string name="screen_onboarding_sign_in_to">"Connectez-vous à %1$s"</string>
```
bmarty, line 42: I do no like this change, and cannot understand where is it coming from. Probably some Localazy similar strings optimization...

PR #5503, opened 2025-10-09
Do some cleanup on our immutable annotation usage

bmarty:
## Content

Use compose compiler to detect state class that are unstable. Base on the result of this tool, fix all the existing issues on State classes, and remove unnecessary Immutable annotation.

## Motivation and context

Cleaner code, and better performance

## Tests

- No expected visible change

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/tools/compose/check_stability.sh
+++ b/tools/compose/check_stability.sh
@@ -18,7 +18,6 @@ find . -type f -name "*-classes.txt" | while read -r file; do
     # Check that there is no line containing "unstable class .*State {"
     if grep -E 'unstable class .*State \{' "$file"; then
         echo "❌ ERROR: Found unstable State class in $file"
-        # TEMPORARY CHANGE
-        # exit 1
+        exit 1
     fi
 done


features/home/impl/src/main/kotlin/io/element/android/features/home/impl/spaces/HomeSpacesPresenter.kt
```diff
@@ -39,7 +40,7 @@ class HomeSpacesPresenter(
 
         return HomeSpacesState(
             space = CurrentSpace.Root,
-            spaceRooms = spaceRooms,
+            spaceRooms = spaceRooms.toImmutableList(),
```
jmartinesp: I wonder if maybe we should be returning immutable lists from the stores, since otherwise we'd be creating a new immutable list from a `List` every time we want to produce a new state, even if it's not necessary.

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/configureroom/CreateRoomConfig.kt
```diff
@@ -8,10 +8,13 @@
 package io.element.android.features.createroom.impl.configureroom
 
 import android.net.Uri
+import androidx.compose.runtime.Immutable
 import io.element.android.libraries.matrix.api.user.MatrixUser
 import kotlinx.collections.immutable.ImmutableList
 import kotlinx.collections.immutable.persistentListOf
 
+// Annotate with @Immutable since `Uri` is unstable
+@Immutable
```
jmartinesp, line 17: The main reason we can't use Roborazzi instead of Paparazzi is because we use Android classes in the states (such as `Uri`). Maybe our long term plan should be to replace them with some `@Stable`/`@Immutable` version of ours, or just use a plain `String`.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/customreaction/picker/EmojiPickerState.kt
```diff
@@ -8,11 +8,14 @@
...
 
 import androidx.annotation.StringRes
+import androidx.compose.runtime.Immutable
 import io.element.android.emojibasebindings.Emoji
 import io.element.android.libraries.designsystem.theme.components.IconSource
 import io.element.android.libraries.designsystem.theme.components.SearchBarResultState
 import kotlinx.collections.immutable.ImmutableList
 
+// Emoji is unstable (because from an external library?), so we annotate with @Immutable
+@Immutable
```
jmartinesp, line 18: Huh, interesting. Maybe we should change the library and manually mark the data classes as `@Immutable`.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/AggregatedReactionSender.kt
```diff
@@ -7,10 +7,13 @@
 
 package io.element.android.features.messages.impl.timeline.model
 
+import androidx.compose.runtime.Immutable
 import io.element.android.libraries.matrix.api.core.UserId
 import io.element.android.libraries.matrix.api.user.MatrixUser
 import java.util.Date
 
+// Need to be marked as @Immutable for Date
+@Immutable
```
jmartinesp: Oh, nice catch. I could have sworn `java.util.Date` would be treated as immutable by the compose compiler.

jmartinesp:
There seem to be some issues still - caught by the check, which is nice 😅 .

bmarty:
Thanks for the review.

> There seem to be some issues still - caught by the check, which is nice 😅 .

Yes it's nice to see the CI failing (let's say it was on purpose!). What is less nice is that when I run the script locally the error is not reported :/. Investigating...

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/AggregatedReactionSender.kt
bmarty: I replaced the Data by a Long, since it's only use to do comparison.

features/createroom/impl/src/main/kotlin/io/element/android/features/createroom/impl/configureroom/CreateRoomConfig.kt
bmarty: Yes, we should maybe use `String` instead of `Uri`. Can be handled in a separate PR.

features/home/impl/src/main/kotlin/io/element/android/features/home/impl/spaces/HomeSpacesPresenter.kt
bmarty: Done in ea2307a39370249d890c8bcb757c75619ea76a60

Issue #2931, opened 2024-05-28
ability to playback speed of Voice notes


iamtalhaasghar:
#### What would you like to do?
Implement the feature to adjust the playback speed of voice notes.

#### Why would you like to do it?
This feature could enhance user experience by allowing them to listen to voice notes at a speed that suits their preference. Some users might prefer faster playback for quick listening, while others might prefer slower playback for better comprehension.

#### How would you like to achieve it?
There are a few approaches you could take to implement this feature:

1. In-app Controls: Add playback speed controls directly within the voice note playback interface of your app. Users could choose from preset speeds (e.g., 0.5x, 1x, 1.5x, 2x).

<img src="https://github.com/element-hq/element-x-android/assets/46928648/f8e9b803-9949-4c69-97b4-013e3e0e229e" height="200px">

2. Settings Menu: Include playback speed options in the app's settings menu, allowing users to set a default playback speed for all voice notes.

### Are you willing to provide a PR?

No

iamtalhaasghar:
@msusman1 bhai kindly check when free. This would be great contribution for whole matrix / element community.

kalix127:
@bmarty This feature has already been implemented in #5504. This issue can be closed

PR #5504, opened 2025-10-09
Add variable playback speed feature for voice messages

Medformatik:
## Content

Add playback speed control for voice messages with support for 0.5×, 1×, 1.5×, and 2× playback speeds. The speed button is displayed above the timestamp and cycles through the available speeds when tapped.

## Motivation and context

Allows users to adjust voice message playback speed for improved accessibility and user experience. This is a common feature in messaging apps that helps users consume voice content more efficiently.

## Screenshots / GIFs

|Before|After|
|-|-|
|Before|After|

## Tests

Manual testing:
- Step 1: Record or receive a voice message in a conversation
- Step 2: Tap the play button to start playback
- Step 3: Tap the playback speed button (displays "1×" by default) to cycle through speeds
- Step 4: Verify playback speed changes accordingly: 1× → 1.5× → 2× → 0.5× → 1×
- Step 5: Verify the speed setting is applied immediately to currently playing voice messages

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 16

## Checklist


ema987:
Hey! 👋 I was about to submit a PR for this same feature because I didn't notice this at first look, and then just saw yours: nice work!

We ended up making very similar changes. You can see my version in [this branch](https://github.com/ema987/element-x-android/tree/feat/add_voice_notes_playback_speed_control) and [this commit](https://github.com/ema987/element-x-android/commit/8324886eaa994e3fcdd3ec781e291019af9013e1).

In my implementation, I positioned the playback speed control at the *end of the row* (similar to WhatsApp) and used an `enum` to define the possible speeds.  

Screen Recording 2025-11-07 at 22 46 27

Most of the other changes are basically the same, which seems like a good sign that we’re on the right track!

Since there aren’t clear design guidelines for this yet, I wasn’t sure which layout the maintainers would prefer.  
It’d be great to get some direction so we can finalize the implementation and close the [related issue](https://github.com/element-hq/element-x-android/issues/2931). 
I would love to have this feature in production! 🎯

Medformatik:
Hey, thanks for your contribution to this. I'm glad that our implementations are similar. You're right about the design, the maintainers will have to comment on that. I have taken my inspiration from Signal. My design has the advantage over yours that no space is taken away from the waveform. On the other hand, the button is much smaller.

ema987:
Yeah, I agree with what you are saying. At this point, it's honestly a matter of preference and a design decision to follow current practices, because both solutions work.
We just need to wait for an input from the maintainers.

bmarty:
Thanks for the PR, technically correct!
Hello @amshakal , we need some design / UX input for the feature. Can you read the comments above and provide feedback please?
CC @mxandreas

mxandreas:
Thanks for the contributions!

I find the version with the _speed_ button on the right a bit more "accessible" - because the button is bigger and it is further away from the _play_ button. The small downside is that it shortens the waveform so scrolling is probably a bit less precise.

@amshakal wdyt?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemVoiceView.kt
```diff
@@ -172,6 +186,36 @@ private fun RetryButton(
...
+        else -> "${speed}×"
+    }
+    androidx.compose.foundation.layout.Box(
+        modifier = Modifier
+            .background(
+                color = ElementTheme.colors.bgCanvasDefault,
+                shape = RoundedCornerShape(12.dp)
+            )
+            .padding(horizontal = 8.dp, vertical = 4.dp)
+            .clickable(onClick = onClick),
```
bmarty: The clickable modifier should be above the padding, so that the clickable area is bigger.

bmarty (commented):
Hello @Medformatik,
Thanks for the proposal. We decided to accept your proposal and may iterate on design later.
I have a small remark on the clickable area, please see my comment.
Also do you think it would be useful to persist the playback speed in a data store?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemVoiceView.kt
Medformatik: Fixed in 2b81b18

Medformatik:
> Hello @Medformatik, Thanks for the proposal. We decided to accept your proposal and may iterate on design later.

Hello, I'm happy to hear that!

> I have a small remark on the clickable area, please see my comment.

Fixed!

> Also do you think it would be useful to persist the playback speed in a data store?

Good idea, but I find it hard to decide that. WhatsApp persists the playback speed, Signal doesn't. I guess I like it more when it's persistent, but I'm not sure. @ema987 do you have a preference on that? Or @bmarty, if the Element team wants to decide on that, let me know.

bmarty:
Thanks for the update. Let's decide later for the playback speed persistence.
I'll merge your PR in a branch I own to finalize the work and record the screenshots.

bmarty:
Follow up PR is at #5963. I implemented a store to persist the playback speed as it was actually solving rendering issues and synchronisation issue between the player speed and the UI.

PR #5514, opened 2025-10-12
Display 'join anyway' button on room preview when the state can't be loaded

ShadowRZ:
## Content

Remove unknown room display condition.

## Motivation and context

The current logic for when to display unknown room screen doesn't work:

> So to explain, Element X Android had a condition of showing the Join Anyway button: https://github.com/element-hq/element-x-android/blob/b6e2208ccb6abccd1d4037ef9e00dda37607d17d/features/joinroom/impl/src/main/kotlin/io/element/android/features/joinroom/impl/JoinRoomPresenter.kt#L144-L148
> (State `ContentState.UnknownRoom` is the state that will ask the view layer to present an unknown room content in `JoinRoomView`)
>
> EXA calls [`matrix_sdk_ffi::Client::get_room_preview_from_room_alias`](https://github.com/matrix-org/matrix-rust-sdk/blob/588d6046533a9657bcb2eb98d2e886437e7964e4/bindings/matrix-sdk-ffi/src/client.rs#L1435-L1450) (in the room alias case) which eventually calls the code in https://github.com/matrix-org/matrix-rust-sdk/blob/588d6046533a9657bcb2eb98d2e886437e7964e4/crates/matrix-sdk/src/room_preview.rs#L142-L175
>
> The only error that can be returned at this point is `matrix_sdk::Error::InsufficientData`, then converted to `matrix_sdk_ffi::ClientError::Generic`, and then sent back to EXA as Kotlin exception

## Tests

- Login to EXA with a server that doesn't have room summary enabled (should be optional)
- Find a room that can't preview and join before this PR, and try to join with this PR applied

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): Android 14 / 15

## Checklist

### DIFF
--- a/features/joinroom/impl/src/main/kotlin/io/element/android/features/joinroom/impl/JoinRoomPresenter.kt
+++ b/features/joinroom/impl/src/main/kotlin/io/element/android/features/joinroom/impl/JoinRoomPresenter.kt
@@ -39,8 +39,6 @@ import io.element.android.libraries.core.meta.BuildMeta
 import io.element.android.libraries.matrix.api.MatrixClient
 import io.element.android.libraries.matrix.api.core.RoomId
 import io.element.android.libraries.matrix.api.core.RoomIdOrAlias
-import io.element.android.libraries.matrix.api.exception.ClientException
-import io.element.android.libraries.matrix.api.exception.ErrorKind
 import io.element.android.libraries.matrix.api.room.CurrentUserMembership
 import io.element.android.libraries.matrix.api.room.RoomInfo
 import io.element.android.libraries.matrix.api.room.RoomMembershipDetails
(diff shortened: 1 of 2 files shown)


jmartinesp:
Thanks for your PR! In the beginning I was thinking that maybe we should try to detect and propagate the `InsufficientData` error from Rust to the FFI layer and into the app, but we discussed it with the iOS team and apparently they're just using the same strategy as you proposed, any unknown error would assume an unknown room preview state and let you try to join the room.

Could you:

1. Run `./gradlew ktlintFormat runQualityChecks` to autofix formatting and check no lint issues are present.
2. Fix the broken unit tests?

I can also merge this into a branch of mine and take it over if you prefer that.

ShadowRZ:
@jmartinesp Codestyle and quality checks should pass now, tests also get fixed.

`present - when room is not known RoomPreview is loaded with error - dismiss` test was removed since it didn't make sense now.

jmartinesp (approved):
Thanks!

Issue #5530, opened 2025-10-13
Notification issue


PH03N1X999:
While doing notification troubleshooting I'm getting an error.
I did this test because sometimes I'm not getting notifications. Also I tried disabling battery optimizations and everything to confirm but still same. Can some one help me what's going on.... Attaching screenshot for reference
Also to note im running ess community edition 

![Screenshot_2025-10-13-20-29-16-62_508fb70dcdbd51b3b84edd38d0498869.jpg](https://github.com/user-attachments/assets/31fff535-d971-4378-be33-079227a72889)

Shivox:
Same issue. Call notifications (with `element-call`) have been broken for me for a while now. It works for some family members, doesn't work for some others.

rajil:
Same here, 'Test Push loop back' fails for Google Play Store app on v25.10.0.

jmartinesp:
[This PR](https://github.com/element-hq/element-x-android/pull/5541) should fix the test failing, that was caused by a breaking change that didn't take this test into account.

As for why you may not be receiving notifications, that's unrelated to the changes above and we'd need more info, like the Android version, or if it's always the same rooms, or rooms vs DMs, etc. 

If it's happening on Android 15 or higher, it may be caused by an OS bug that forbids the app from getting network connectivity when trying to fetch the notification data: the app receives the notification on background, it's woken up, but then when it tries to fetch the associated event data to display it just receives an error because the Android OS sometimes blocks network access when the app is not active.

Shivox:
@jmartinesp Yes. That checks out. Since my phone was updated to Android 15 (and now 16), notifications stopped working. When I installed the nightly build, I could see a couple of notifications received when called, but both show `notification was filtered out`.

PR #5541, opened 2025-10-15
Fix 'test push loop back' notification check

jmartinesp:
## Content

What the title says: we now include the client secret as the HS would do, since it's mandatory. Otherwise, the received push notification would not have it and we'd discard it, leading to the check failing.

## Motivation and context

Fix one of the issues mentioned in https://github.com/element-hq/element-x-android/issues/5530.

## Tests

Go to settings -> notifications -> troubleshoot and run the tests.

The test push loop back step should now work again.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/pushgateway/PushGatewayNotifyRequest.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/pushgateway/PushGatewayNotifyRequest.kt
@@ -42,9 +42,12 @@ class DefaultPushGatewayNotifyRequest(
                     devices = listOf(
                         PushGatewayDevice(
                             params.appId,
-                            params.pushKey
+                            params.pushKey,
+                            PusherData(mapOf(
+                                "cs" to "A_FAKE_SECRET",
+                            ))
                         )
-                    )
+                    ),
                 )
             )
         )
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/pushgateway/PushGatewayDevice.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/pushgateway/PushGatewayDevice.kt
@@ -21,5 +21,14 @@ data class PushGatewayDevice(
      * Required. The pushkey given when the pusher was created.
      */
     @SerialName("pushkey")
-    val pushKey: String
+    val pushKey: String,
+    /** Optional. Additional pusher data. */
+    @SerialName("data")
+    val data: PusherData? = null,
+)
+
+@Serializable
+data class PusherData(
+    @SerialName("default_payload")
+    val defaultPayload: Map<String, String>,
 )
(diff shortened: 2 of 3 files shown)


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/pushgateway/PushGatewayNotifyRequest.kt
```diff
@@ -42,9 +47,12 @@ class DefaultPushGatewayNotifyRequest(
                     devices = listOf(
                         PushGatewayDevice(
                             params.appId,
-                            params.pushKey
+                            params.pushKey,
+                            PusherData(mapOf(
+                                "cs" to secret,
+                            ))
```
bmarty, line 48: Please use a FAKE_SECRET, it will be simpler (nearly all the change in this file can be reverted) and the secret will not be used anyway when the push will be received. It will also enforce the fact that there is no need to have a session to perform this test.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/test/TestPush.kt
```diff
@@ -20,7 +20,7 @@ interface TestPush {
     suspend fun execute(config: CurrentUserPushConfig)
 }
 
-@ContributesBinding(AppScope::class)
+@ContributesBinding(SessionScope::class)
```
bmarty: Revert?

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/troubleshoot/PushLoopbackTest.kt
```diff
@@ -28,7 +28,7 @@ import kotlinx.coroutines.withTimeout
 import timber.log.Timber
 import kotlin.time.Duration.Companion.seconds
 
-@ContributesIntoSet(AppScope::class)
+@ContributesIntoSet(SessionScope::class)
```
bmarty: Revert?

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/DefaultPushService.kt
```diff
@@ -29,8 +29,8 @@ import io.element.android.libraries.sessionstorage.api.observer.SessionObserver
 import kotlinx.coroutines.flow.Flow
 import timber.log.Timber
 
-@ContributesBinding(AppScope::class, binding = binding<PushService>())
-@SingleIn(AppScope::class)
+@ContributesBinding(SessionScope::class, binding = binding<PushService>())
+@SingleIn(SessionScope::class)
```
bmarty: Revert?

bmarty (commented):
Thanks, using a FAKE_SECRET would simplify a lot the needed changes. You may define it in `DefaultTestPush.companion`.

jmartinesp:
I force pushed a commit without all the session scope changes and having the fake secret value.

PR #5557, opened 2025-10-17
Sort feature flags

bmarty:
## Content

Improve the code around feature flag:
- ensure that the order in screen is the same that declared in the enum. It was painful to add a flag at the end of the list and  then have to look for it because the order was randomized.
- also apply the same fix on the lab screen. For now there is only one entry, but we would have seen the problem when other entries will be added
- add a ProgressDialog when clearing the cache in the developer settings (as it's done in the labs screen)
- improve api and test, especially use more fake data instead of real `FeatureFlags.Space` for instance

## Motivation and context

- More predictive behavior
- Improved tests.
- handle remarks in comment https://github.com/element-hq/element-x-android/pull/5465#discussion_r2409659913

## Screenshots / GIFs

See recorded one, the ProgressDialog should be displayed now (there was no need to update the State provider)

## Tests

- Open the developer screen and observe that the FF are ordered as in the code.
- Clear the cache and observe that a ProgressDialog is displayed.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/featureflag/impl/src/test/kotlin/io/element/android/libraries/featureflag/impl/DefaultFeaturesProviderTest.kt
+++ b/libraries/featureflag/impl/src/test/kotlin/io/element/android/libraries/featureflag/impl/DefaultFeaturesProviderTest.kt
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2025 New Vector Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.featureflag.impl
+
+import com.google.common.truth.Truth.assertThat
+import io.element.android.libraries.featureflag.api.FeatureFlags
+import org.junit.Test
+
+class DefaultFeaturesProviderTest {
+    @Test
+    fun `provide should return all features`() {
+        val provider = DefaultFeaturesProvider()
+        val features = provider.provide()
+        assertThat(features.size).isEqualTo(FeatureFlags.entries.size)
+        FeatureFlags.entries.forEach {
+            assertThat(features.contains(it)).isTrue()
+        }
+    }
+}


features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/model/EnabledFeature.kt
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.features.preferences.impl.model
+
+import io.element.android.libraries.featureflag.api.Feature
+
+data class EnabledFeature(
```
bmarty, line 12: `EnableableFeature` would be a better name, but the word `enableable` does not seem to exist. Suggestion welcome!

libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlagService.kt
```diff
@@ -35,7 +35,12 @@ interface FeatureFlagService {
...
 
     /**
-     * @return the list of available (not finished) features that can be toggled.
+     * @return the list of available features that can be toggled.
+     * @param includeFinishFeatures whether to include finished features, default is false
+     * @param isInLabs whether the user is in labs (to include lab features), default is false
      */
-    fun getAvailableFeatures(): List<Feature>
+    fun getAvailableFeatures(
+        includeFinishFeatures: Boolean = false,
```
jmartinesp: Typo, I think: `includeFinishedFeatures`?

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/model/EnabledFeature.kt
jmartinesp: `FeatureEnabledState`? I'm not really convinced though 😅 .

jmartinesp (approved):
Thanks, LGTM!

Issue #5505, opened 2025-10-10
Incoming call UI duration is too short (on certain devices).


YamatoRyou:
### Steps to reproduce

Lock your phone and wait for the call, but don't answer it.

#### What did you expect?
n/a

#### What happened instead?
On certain devices, even without delaying push notifications, the incoming call UI and ringtone duration is only 10-15 seconds, which is already the maximum on certain devices.

Demo:
https://www.youtube.com/watch?v=ctHyHpzWc1E
**As of the time of filming, the incoming call ringtone (including the UI) lasted less than 1 second, sometimes even just flashing and disappearing, making it unusable.**

My other three phones, even without delaying push notifications, can maintain this duration for less than 30 seconds:
Mi 9T (MIUI 12.5, Android 11)
HMD TA-1054 (Android 8.1)
iPhone SE 3 (iOS 18.3.1)

Logs:
[{510796C7-C682-D347-E48F-BF884D104421}.zip](https://github.com/user-attachments/files/22817523/510796C7-C682-D347-E48F-BF884D104421.zip)
Password: `{338A203D-3885-4F8F-E3DC-83877EBFFDF2}`
**Please download this file as soon as possible.**

----
On versions released in the last week or two (including EW and EXi), the default duration of the incoming call displayed and heard by the other party during a call is 30 seconds, which is still very short. I hope to increase this to around 60 seconds.

### Your phone model

OnePlus MT2111

### Operating system version

OxygenOS 15 (Android 15)

### Application version and app store

Element X Android 25.10.0

### Homeserver

Synapse 1.139.0

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

jmartinesp:
In the logs you uploaded I see:

```
2025-10-10T00:00:31.545747Z  INFO elementx: ## handleInternal() : Incoming call. | 
2025-10-10T00:00:31.549851Z DEBUG elementx: [DefaultActiveCallManager] Received incoming call for room id: !TxkGysRFDuDmqKPRpR:example.com, ringDuration(ms): 11539 | 
2025-10-10T00:00:31.549903Z DEBUG elementx: [SyncOrchestrator] isAppActive=true, isNetworkAvailable=true | 
2025-10-10T00:00:31.552561Z DEBUG elementx: [DefaultActiveCallManager] Acquiring partial wakelock | 
2025-10-10T00:00:31.560244Z DEBUG elementx: [DefaultActiveCallManager] Displaying ringing call notification | 
[...]

PR #5563, opened 2025-10-20
Make sure declining a call stops observing the ringing call state

jmartinesp:
## Content

- Centralise call cancellation logic in a single `removeCurrentCall` method.
- Make sure the existing iterations of `flatMapLatest` in `observeRingingCall` get cancelled when the active call is null or not ringing anymore by passing null values, then filtering them out. Previously these kept running even if the `activeCall` was no longer valid.

## Motivation and context

The logs in #5505 display this issue, although I'm not sure it's the source of the mentioned issues.

## Tests

I couldn't reproduce it in my device, but it should be possible to revert the latest commit in the list and run the regression test, which will fail with this issue.

## Checklist

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/ActiveCallManager.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/ActiveCallManager.kt
@@ -208,20 +208,16 @@ class DefaultActiveCallManager(
                 ?.declineCall(notificationData.eventId)
         }
 
-        timedOutCallJob?.cancel()
-
         removeCurrentCall()
     }
 
+    /**
+     * Removes the current active call and any associated UI, cancelling the timeouts and wakelocks.
+     */
     override suspend fun joinedCall(callType: CallType) = mutex.withLock {
         Timber.tag(tag).d("Joined call: $callType")
 
-        cancelIncomingCallNotification()
-        if (activeWakeLock?.isHeld == true) {
-            Timber.tag(tag).d("Releasing partial wakelock after joining call")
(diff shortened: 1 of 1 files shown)


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/ActiveCallManager.kt
```diff
@@ -281,73 +287,75 @@ class DefaultActiveCallManager(
...
-                    return@flatMapLatest flowOf()
-                }
+        val roomForActiveCallFlow: Flow<Pair<BaseRoom, EventId>?> = activeCall.mapLatest { activeCall ->
+            val callType = activeCall?.callType as? CallType.RoomCall ?: return@mapLatest null
+            val ringingInfo = activeCall.callState as? CallState.Ringing ?: return@mapLatest null
+            val client = matrixClientProvider.getOrRestore(callType.sessionId).getOrNull() ?: run {
+                Timber.tag(tag).d("Couldn't find session for incoming call: $activeCall")
+                return@mapLatest null
+            }
+            val room = client.getRoom(callType.roomId) ?: run {
```
ganfra, line 297: Room is never closed?

jmartinesp: With JNA we don't really need to, but I can make sure we close it manually.

jmartinesp: Or maybe I can't, since it's a value that's shared and doesn't have a clear lifetime 🤔 .

Issue #5580, opened 2025-10-22
A room timeline won't load while a previous one is working


jmartinesp:
### Steps to reproduce

1. Open a room.
2. Go to the media gallery in that room, let it paginate for a while.
3. Exit the room.
4. Go back to the room.

#### What did you expect?

The timeline to load as usual.

#### What happened instead?

The timeline loads after a *very long time* (10-20s). In the logs I can see the redecryption mechanism working non-stop, probably on items from the previous media timeline created for the gallery.

### Your phone model

Xiaomi 9T

### Operating system version

Android 14

### Application version and app store

Element X 25.10.1

### Homeserver

element.io

### Will you send logs?

No

### Are you willing to provide a PR?

Yes

jmartinesp:
This is probably caused by this line: https://github.com/element-hq/element-x-android/blob/dd395a52146712039969fdc87911906351069b19/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt#L224

[The associated function in the SDK](https://github.com/matrix-org/matrix-rust-sdk/blob/d36b68b7d12d19551c4eab5afcbcd114e54581d9/bindings/matrix-sdk-ffi/src/room/mod.rs#L671) side reinstantiates a temporary timeline, which will reload any existing cached items by the current one, which can be extremely slow for rooms you've paginated a lot on. And while this timeline is alive the cache will be kept, so opening the room again will also be extremely slow.

Some proposals to address this are:

- Call `room.liveTimeline.markAsRead` instead. This however is an async function and if we send it while we exit the room UI we'll be delaying the timeline cache clearing that makes everything fast again. It's better, but doesn't fix the issue.
- Send the fully read marker when scrolling to the bottom of the timeline instead of exiting the room: similar to the previous one, this won't fix the issue, but will make it even less likely.
- Create a new FFI function that bypasses the checks in the timeline so we can free it and we can just send the fully read marker. This will fix the issue with the loading times, but it means we'll always send fully read markers, and it's up to the homeserver to decide if they're valid or not.

PR #5582, opened 2025-10-22
Make sure we destroy rooms removed from the a `ActiveRoomsHolder`

jmartinesp:
## Content

What the title says.

## Motivation and context

This is needed to clear their timeline cache so their timeline creation is faster when they're re-opened. This would also happen after a GC, but that can take time to happen so if the room is re-opened shortly after closing it, this cache clear might not have happened.

It's probably what caused https://github.com/element-hq/element-x-android/issues/5580.

## Tests

- Open a room, and back paginate a lot. You can also open its gallery and let it back paginate on its own.
- Close the room.
- Open it again.

The opening time should be the same as when you first opened it. Previously, this could take 10-20s or more, depending on how much you had back paginated and how many timeline items the timeline had cached on the Rust side.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/services/appnavstate/api/src/main/kotlin/io/element/android/services/appnavstate/api/ActiveRoomsHolder.kt
+++ b/services/appnavstate/api/src/main/kotlin/io/element/android/services/appnavstate/api/ActiveRoomsHolder.kt
@@ -53,8 +53,11 @@ class ActiveRoomsHolder {
      */
     fun removeRoom(sessionId: SessionId, roomId: RoomId) {
         val roomsForSessionId = rooms[sessionId] ?: return
-        roomsForSessionId.find { it.roomId == roomId }?.destroy()
-        roomsForSessionId.removeIf { it.roomId == roomId }
+        roomsForSessionId.find { it.roomId == roomId }?.let { room ->
+            // Destroy the room to reset the live timeline
+            room.destroy()
+            roomsForSessionId.remove(room)
+        }
     }
 
     /**


services/appnavstate/api/src/main/kotlin/io/element/android/services/appnavstate/api/ActiveRoomsHolder.kt
```diff
@@ -53,6 +53,7 @@ class ActiveRoomsHolder {
      */
     fun removeRoom(sessionId: SessionId, roomId: RoomId) {
         val roomsForSessionId = rooms[sessionId] ?: return
+        roomsForSessionId.find { it.roomId == roomId }?.destroy()
```
bmarty: Maybe store the result of `roomsForSessionId.find { it.roomId == roomId }` to then destroy it and remove it (the line below). Here I think it is searched twice (which is not a big deal!).

jmartinesp: Oh, you're right, since I already have the instance I can just call `roomsForSessionId.remove(room)`. Good call!

jmartinesp:
Closing this, as I think my rooms loading faster wasn't related to these changes. As @ganfra mentioned in a chat, the live timeline is already destroyed when the message UI is dismissed. However, I can still see the decryption task running in background and something about read receipts. Maybe destroying the FFI reference for the timeline is not enough to cancel the tasks associated to it's Rust SDK counterpart.

Issue #5581, opened 2025-10-22
[Task] Threaded push notifications


manuroe:
Please see and discuss the details in the meta issue.

manuroe:
The associated PR contains all the work. Thanks @jmartinesp!

PR #5595, opened 2025-10-23
Split notifications for messages in threads

jmartinesp:
## Content

- Splits room notifications and thread notifications so they're grouped separately in the notification drawer if the associated feature flag is enabled.
- Add deep linking navigation to threads from their notifications.
- Following a notification (threaded or not) will focus in the event inside the room.
- Make sure redactions work for threads too.
- When opening a thread, this will also clear the notifications for that thread.

Known issues:
- Opening a thread from a notification will also clear the notifications for the parent room, since we very briefly visit that room's screen.
- Sometimes the notifications don't display the associated avatar. It's not always the same rooms/threads, it seems quite random. I'm not sure if it started with this PR.

## Motivation and context

Implements https://github.com/element-hq/element-x-android/issues/5581.

## Screenshots / GIFs

image

## Tests

- Enable the threads flag in labs.
- Receive a notification for a room/DM.
- Receive a notification for one or several threads in a room/DM.
- Tap on the room notification, check it takes you to the latest event in the notification.
- Then tap on the thread notification, which should now open the thread.
- Exit both the thread and the room.
- Receive a couple of new notifications for room and thread.
- Open the room from the room list. The notifications associated to it should be gone.
- Do the same for the thread, opening it from the timeline.
- Exit the thread.
- Receive a new notification for a thread.
- Redact that event after it's received.
- Check the notification is also redacted.

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): 14, 16

## Checklist

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationBroadcastReceiverHandler.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationBroadcastReceiverHandler.kt
@@ -125,6 +125,9 @@ class NotificationBroadcastReceiverHandler(
                     Timber.d("Marked room $roomId as read with receipt type $receiptType")
                 }
             }
+            ?.onFailure {
+                Timber.e(it, "Fails to mark as read with receipt type $receiptType")
+            }
         if (timeline?.mode != Timeline.Mode.Live) {
             timeline?.close()
         }


libraries/push/api/src/main/kotlin/io/element/android/libraries/push/api/notifications/NotificationIdProvider.kt
```diff
@@ -49,9 +49,8 @@ object NotificationIdProvider {
     private const val ROOM_MESSAGES_NOTIFICATION_ID = 1
     private const val ROOM_EVENT_NOTIFICATION_ID = 2
     private const val ROOM_INVITATION_NOTIFICATION_ID = 3
-    private const val ROOM_CALL_NOTIFICATION_ID = 3
-
     private const val FOREGROUND_SERVICE_NOTIFICATION_ID = 4
+    private const val ROOM_CALL_NOTIFICATION_ID = 5
```
jmartinesp: Since I spotted this issue I fixed it in this PR. I can also split it into a dedicated one.

bmarty: Not a big deal since it seems to be unused (?)

app/src/main/kotlin/io/element/android/x/intent/DefaultIntentProvider.kt
```diff
@@ -28,11 +29,12 @@ class DefaultIntentProvider(
     override fun getViewRoomIntent(
         sessionId: SessionId,
         roomId: RoomId?,
+        eventId: EventId?,
```
bmarty: Maybe move the new `eventId` parameter at the end of the list? The order will be consistent with what `deepLinkCreator.create()` takes.

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -517,7 +519,8 @@ class LoggedInFlowNode(
                 serverNames = serverNames,
                 trigger = trigger,
                 initialElement = RoomNavigationTarget.Messages(
-                    focusedEventId = eventId
+                    focusedEventId = eventId,
+                    inThreadId = threadId,
```
bmarty: I would expect to have a function like `attachThread`?

libraries/deeplink/impl/src/main/kotlin/io/element/android/libraries/deeplink/impl/DefaultDeepLinkCreator.kt
```diff
@@ -27,6 +28,15 @@ class DefaultDeepLinkCreator : DeepLinkCreator {
...
                 }
+
+                if (eventId != null) {
+                    if (threadId == null) {
+                        // Add extra separator to mark the thread slot as empty
+                        append("/")
+                    }
+                    append("/")
+                    append(eventId.value)
+                }
```
bmarty: It works, but maybe we could simplify a bit the complexity and have:
```kotlin
        return buildString {
            append("$SCHEME://$HOST/")
            append(sessionId.value)
            append("/")
            append(roomId?.value.orEmpty())
            append("/")
            append(threadId?.value.orEmpty())
            append("/")
            append(eventId?.value.orEmpty())
        }
```
at the price of having less nice links :)

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/conversations/DefaultNotificationConversationService.kt
```diff
@@ -88,9 +87,7 @@ class DefaultNotificationConversationService(
             return
         }
 
-        val categories = setOfNotNull(
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) ShortcutInfo.SHORTCUT_CATEGORY_CONVERSATION else null
-        )
+        val categories = setOfNotNull(ShortcutInfo.SHORTCUT_CATEGORY_CONVERSATION)
```
bmarty: Revert? I still see this warning: `Field requires API level 25 (current min is 24)`

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/PendingIntentFactory.kt
```diff
@@ -32,19 +32,19 @@ class PendingIntentFactory(
...
 
-    fun createOpenRoomPendingIntent(sessionId: SessionId, roomId: RoomId): PendingIntent? {
-        return createRoomPendingIntent(sessionId = sessionId, roomId = roomId, threadId = null)
+    fun createOpenRoomPendingIntent(sessionId: SessionId, roomId: RoomId, eventId: EventId?): PendingIntent? {
+        return createRoomPendingIntent(sessionId = sessionId, roomId = roomId, eventId = eventId, threadId = null)
     }
 
-    fun createOpenThreadPendingIntent(roomInfo: RoomEventGroupInfo, threadId: ThreadId?): PendingIntent? {
-        return createRoomPendingIntent(sessionId = roomInfo.sessionId, roomId = roomInfo.roomId, threadId = threadId)
+    fun createOpenThreadPendingIntent(roomInfo: RoomEventGroupInfo, eventId: EventId?, threadId: ThreadId): PendingIntent? {
```
bmarty: For parity this method should take `sessionId: SessionId, roomId: RoomId` instead of `roomInfo: RoomEventGroupInfo` (I now that this is not new though)

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
```diff
@@ -130,28 +145,40 @@ class DefaultNotificationCreator(
         val builder = if (existingNotification != null) {
             NotificationCompat.Builder(context, existingNotification)
         } else {
-            NotificationCompat.Builder(context, channelId)
+            NotificationCompat.Builder(context, channelId).apply {
```
bmarty: Can you revert this `apply` please?

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
```diff
@@ -130,28 +145,40 @@ class DefaultNotificationCreator(
...
+                    }
+                }
                 // Auto-bundling is enabled for 4 or more notifications on API 24+ (N+)
                 // devices and all Wear devices. But we want a custom grouping, so we specify the groupID
-                .setGroup(roomInfo.sessionId.value)
+                setGroup(roomInfo.sessionId.value)
+                setGroupSummary(false)
                 // In order to avoid notification making sound twice (due to the summary notification)
-                .setGroupAlertBehavior(NotificationCompat.GROUP_ALERT_ALL)
+                setGroupAlertBehavior(NotificationCompat.GROUP_ALERT_CHILDREN)
```
bmarty: Using `GROUP_ALERT_ALL` was wrong?

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
```diff
@@ -443,28 +457,33 @@ class DefaultNotificationCreator(
...
-                .setKey(sessionId.value)
+                .setKey(user.userId.value)
                 .build()
         ).also {
-            it.conversationTitle = roomName.takeIf { roomIsGroup }
-            it.isGroupConversation = roomIsGroup
+            it.conversationTitle = if (isThread) {
+                stringProvider.getString(CommonStrings.notification_thread_in_room, roomName)
+            } else {
+                roomName
```
bmarty, line 478: Previous code was `roomName.takeIf { roomIsGroup }`. Can you double check this change please?

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/ActiveNotificationsProvider.kt
```diff
@@ -44,9 +53,19 @@ class DefaultActiveNotificationsProvider(
         return getNotificationsForSession(sessionId).filter { it.id == notificationId }
     }
 
-    override fun getMessageNotificationsForRoom(sessionId: SessionId, roomId: RoomId): List<StatusBarNotification> {
+    override fun getMessageNotificationsForRoom(sessionId: SessionId, roomId: RoomId, threadId: ThreadId?): List<StatusBarNotification> {
         val notificationId = NotificationIdProvider.getRoomMessagesNotificationId(sessionId)
-        return getNotificationsForSession(sessionId).filter { it.id == notificationId && it.tag == roomId.value }
+        val expectedTag = if (threadId != null) {
+            "${roomId.value}|${threadId.value}"
```
bmarty: Maybe better to use `NotificationCreator.messageTag`:

```kotlin
val expectedTag = NotificationCreator.messageTag(roomId, threadId)`
```

libraries/push/api/src/main/kotlin/io/element/android/libraries/push/api/notifications/NotificationIdProvider.kt
bmarty: (Maybe add missing cases in `NotificationIdProviderTest`?)

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
jmartinesp: I don't think we need one because we'd just duplicate most of the logic, unless we want to open a thread *without* opening the room instead.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/conversations/DefaultNotificationConversationService.kt
jmartinesp: Agh, I had the codebase in Element X Pro mode, which has a way higher minSdk,  so that's why the IDE said it was an unnecessary check. Reverting!

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
jmartinesp: From what I could tell given their docs, it seems like `GROUP_ALERT_CHILDREN` makes more sense:

> Constant for NotificationCompat.Builder.setGroupAlertBehavior(int), meaning that the summary notification in a group should be silenced (no sound or vibration) even if they would otherwise make sound or vibrate. Use this constant to mute this notification if this notification is a group summary.
For example, you might want to use this constant if only the children notifications in your group have content and the summary is only used to visually group notifications rather than to alert the user that new information is available.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
jmartinesp: I think this would override the existing `Notification.Builder.setContentTitle` value. Since that's now gone, I think this is needed here, otherwise the notification wouldn't have any title.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/ActiveNotificationsProvider.kt
jmartinesp: Good catch! I suspected there was another usage around that I need to replace but I didn't find it at the time.

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
bmarty: Mhh, OK. I still believe that having a `attachThread` would be better to handle it with Appyx instead of having `LaunchedEffect(Unit)` in MessagesNode, but it's a bit of a mess because of our multi module project. Here is a patch with how I would implement it:

libraries/deeplink/impl/src/main/kotlin/io/element/android/libraries/deeplink/impl/DefaultDeepLinkCreator.kt
bmarty: Thanks for the rework. The links are still nice!

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
```diff
@@ -447,21 +459,26 @@ class DefaultNotificationCreator(
...
                 .setName(user.displayName?.annotateForDebug(50))
                 .setIcon(bitmapLoader.getUserIcon(user.avatarUrl, imageLoader))
-                .setKey(sessionId.value)
+                .setKey(user.userId.value)
                 .build()
         ).also {
-            it.conversationTitle = roomName.takeIf { roomIsGroup }
-            it.isGroupConversation = roomIsGroup
+            it.conversationTitle = if (isThread) {
+                stringProvider.getString(CommonStrings.notification_thread_in_room, roomName)
```
bmarty, line 476: For DM the title is a bit strange since the room name is the other user name:

Image

tests/testutils/src/main/kotlin/io/element/android/tests/testutils/lambda/Assertions.kt
```diff
@@ -51,7 +51,7 @@ class ParametersAssertions internal constructor(
         }
         parametersSequence.zip(matchersSequence).forEachIndexed { invocationIndex, (parameters, matchers) ->
             if (parameters.size != matchers.size) {
-                throw AssertionError("Expected ${matchers.size} parameters, but got ${parameters.size} parameters during invocation #$invocationIndex")
+                throw AssertionError("Expected ${parameters.size} parameters, but got ${matchers.size} parameters during invocation #$invocationIndex")
```
bmarty: Are you sure of this change? The current code look correct (after several readings).
Edit: I have made some test and I confirm that the current code is correct.

jmartinesp: That's odd, when I was testing this I received the message:

`Expected 5 parameters, but got 6` when the lambda was actually expecting 6 parameter and got 5.

And it took me a while to figure out what was wrong 🫤 . Let me see if I can reproduce it again.

appnav/src/main/kotlin/io/element/android/appnav/RootFlowNode.kt
```diff
@@ -388,26 +393,45 @@ class RootFlowNode(
...
             }
             is PermalinkData.UserLink -> {
                 attachUser(permalinkData.userId)
             }
         }
     }
 
+    private suspend fun RoomFlowNode.maybeAttachThread(threadId: ThreadId?, focusedEventId: EventId?) {
+        if (threadId != null) {
+            waitForNavTargetAttached { it is RoomFlowNode.NavTarget.JoinedRoom }
```
bmarty: I do not understand why this line is necessary (it was not in my patch).
`RoomFlowNode.attachThread` already contains `waitForChildAttached<JoinedRoomFlowNode>()`. This is the same thing, no? Or am I missing something? I am getting confused with all those Nodes :)

bmarty (commented):
One question.

tests/testutils/src/main/kotlin/io/element/android/tests/testutils/lambda/Assertions.kt
jmartinesp: Ah, never mind, in my mind this was the opposite:

`isCalledOnce().withSequence(any(), any(), any())` for me meant calling this function with 3 parameters, not 3 matchers 🤦 . Reverting it.

appnav/src/main/kotlin/io/element/android/appnav/RootFlowNode.kt
jmartinesp: From our EXA devs chat:

> there was another interesting bug
`LoggedInFlowNode.attachRoom` creates a new RoomFlowNode and waits until a `RoomFlowNode` is present
however, what happens if you do this while a `RoomFlowNode` was already present? the operation returns the existing one :D
and since that one is about to be disposed, anything pushed into it is ignored

You can test this by removing this code and:

1. Opening a room.
2. Receiving a notification for a thread while the room is visible (of that room or some other).
3. Opening the notification.

Instead of the thread UI being pushed, you'll see the messages screen for that room being pushed, and the thread screen is nowhere to be found. If you attach a debugger, you'll see the attach thread operation does happen... in the previous `RoomFlowNode`, not the one we just replaced it with.

bmarty: I will try to repro. But for me the problem you described was fixed by this code: https://github.com/element-hq/element-x-android/blob/cfcb821c8977ba3ac6844e311f9317849a693942/appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt#L521

bmarty: I have commented out the code and I do not have the issue you describe

jmartinesp: Ah sorry, I thought we were talking about that other one. This one may not be necessary. If it's not in your tests, we can just remove it, since I probably added it to check what was causing the weird navigation and forgot to remove it later.

bmarty:
Failing test is not failing locally...

bmarty (approved):
Thanks for the many updates and your patience!

We will treat the issue on the "Mark as read" action separately. Let's merge!

PR #5606, opened 2025-10-24
Fix issues with WorkManager on Android 12 and below

jmartinesp:
## Content

- Remove `setExpedited` value for Android 12 and lower, since it's not needed.
- Add `getForegroundInfo` in case some OS behaves unexpectedly and still forces us to display the foreground service notification to run the worker.

## Motivation and context

Fix an issue reported in the Element X Android public room, where the WorkManager crashed because the `setExpedited`  option used in the worker request would force the app to display a notification we weren't ready for.

## Tests

On Android <= 12, enabled the 'sync notifications using workmanager' feature and receive a couple of notifications.

Notifications should arrive normally.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 11, 12, 13.

## Checklist

### DIFF
--- a/libraries/push/api/src/main/kotlin/io/element/android/libraries/push/api/notifications/NotificationIdProvider.kt
+++ b/libraries/push/api/src/main/kotlin/io/element/android/libraries/push/api/notifications/NotificationIdProvider.kt
@@ -31,10 +31,6 @@ object NotificationIdProvider {
         return getOffset(sessionId) + FALLBACK_NOTIFICATION_ID
     }
 
-    fun getCallNotificationId(sessionId: SessionId): Int {
-        return getOffset(sessionId) + ROOM_CALL_NOTIFICATION_ID
-    }
-
     fun getForegroundServiceNotificationId(type: ForegroundServiceType): Int {
         return type.id * 10 + FOREGROUND_SERVICE_NOTIFICATION_ID
     }
@@ -49,7 +45,6 @@ object NotificationIdProvider {
     private const val ROOM_MESSAGES_NOTIFICATION_ID = 1
     private const val ROOM_EVENT_NOTIFICATION_ID = 2
     private const val ROOM_INVITATION_NOTIFICATION_ID = 3
-    private const val ROOM_CALL_NOTIFICATION_ID = 3
 
     private const val FOREGROUND_SERVICE_NOTIFICATION_ID = 4
 }
@@ -57,5 +52,4 @@ object NotificationIdProvider {
 enum class ForegroundServiceType(val id: Int) {
     INCOMING_CALL(1),
     ONGOING_CALL(2),
-    NOTIFICATION_FETCHING(3),
 }


jmartinesp:
@YamatoRyou since I can't reproduce the issue myself, can I ask you to test the debug APK that will be attached to this PR and see if it fixes your issue with workmanager? If it doesn't, please rageshake again and let us know.

YamatoRyou:
> @YamatoRyou since I can't reproduce the issue myself, can I ask you to test the debug APK that will be attached to this PR and see if it fixes your issue with workmanager? If it doesn't, please rageshake again and let us know.

I just installed the debug APK and tested it.
On this device, the previously mentioned issue has resolved.
{6335BF90-7F79-6A6E-87BB-D4D89109EBD5}

jmartinesp:
Sorry @YamatoRyou can I ask you to test again the newly built APK after some changes I made? This should make the 'Fetching notifications...' notification not necessary on your Android OS version, but I want to double check notifications still work as expected after removing it.

If you want to test it, you can just tap on the link/scan the QR code again.

YamatoRyou:
> Sorry @YamatoRyou can I ask you to test again the newly built APK after some changes I made? This should make the 'Fetching notifications...' notification not necessary on your Android OS version, but I want to double check notifications still work as expected after removing it.
> 
> If you want to test it, you can just tap on the link/scan the QR code again.

It looks normal. I also tested the incoming call UI and it also works fine.
{50095F8B-6CCD-0CDA-C33C-A1B2AE5ED87D}

----
App: `99cd75b5`

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/workmanager/SyncNotificationWorkManagerRequest.kt
```diff
@@ -46,7 +47,14 @@ class SyncNotificationWorkManagerRequest(
         return Result.success(
             OneTimeWorkRequestBuilder<FetchNotificationsWorker>()
                 .setInputData(workDataOf("requests" to json))
-                .setExpedited(OutOfQuotaPolicy.RUN_AS_NON_EXPEDITED_WORK_REQUEST)
+                .apply {
+                    // Expedited workers aren't needed Android 12 or lower:
```
bmarty:
```suggestion
                    // Expedited workers aren't needed on Android 12 or lower:
```

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/workmanager/FetchNotificationsWorker.kt
```diff
@@ -122,6 +132,22 @@ class FetchNotificationsWorker(
         }
     }
 
+    override suspend fun getForegroundInfo(): ForegroundInfo {
```
bmarty: Maybe document also here that this code should never be run?

libraries/push/impl/src/test/kotlin/io/element/android/libraries/push/impl/workmanager/SyncNotificationWorkManagerRequestTest.kt
```diff
@@ -48,8 +73,6 @@ class SyncNotificationWorkManagerRequestTest {
         val result = request.build()
         assertThat(result.isFailure).isTrue()
     }
-
-    // TODO add test for invalid serialization (how?)
```
bmarty: Already removed and replaced by a test on develop, careful during the merge!

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/workmanager/FetchNotificationsWorker.kt
```diff
@@ -122,6 +132,22 @@ class FetchNotificationsWorker(
...
+        val notificationChannelId = notificationChannels.getChannelIdForMessage(false)
+        val notification = NotificationCompat.Builder(context, notificationChannelId)
+            .setSmallIcon(CommonDrawables.ic_notification)
+            .setOngoing(true)
+            .setTicker(context.getString(CommonStrings.common_android_notification_sync_notifications_foreground_service_title))
+            .setContentTitle(context.getString(CommonStrings.common_android_notification_sync_notifications_foreground_service_title))
+            .build()
+        val notificationId = NotificationIdProvider.getForegroundServiceNotificationId(ForegroundServiceType.NOTIFICATION_FETCHING)
+        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
+            ForegroundInfo(notificationId, notification, ServiceInfo.FOREGROUND_SERVICE_TYPE_REMOTE_MESSAGING)
```
bmarty: It seems that using this foreground type requires the manifest permission `android.permission.FOREGROUND_SERVICE_REMOTE_MESSAGING` (see the doc). Since this is dead code maybe not a big deal, but we should check that if it is used it works.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/workmanager/FetchNotificationsWorker.kt
```diff
@@ -112,6 +125,25 @@ class FetchNotificationsWorker(
         }
     }
 
+    // This code technically *should* never run because the expedited scheduling is disabled in Android versions
+    // which would call this. However, some custom OS versions like MIUI have some unexpected behaviours, so we're
```
bmarty: Thinking again about it, since `getForegroundInfo()` is defined in a library that we embed, I do not really see why custom ROM could change the behavior of the library.

libraries/push/impl/src/main/AndroidManifest.xml
```diff
@@ -7,8 +7,8 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
-
     <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_REMOTE_MESSAGING" />
```
bmarty: I hope it will not make the PlayStore to require more work like providing a video demoing the usage of this permission. It is a bit of a shame for supposedly dead code. Can't we just take the risk to remove our override of `getForegroundInfo()`? If we got crash report, we can put it back but it should not happen I believe.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/workmanager/FetchNotificationsWorker.kt
jmartinesp: So should we remove it instead? I'm ok with that and adding it back if we ever find this issue in the wild.

bmarty: Yes, I'd just remove this code.

libraries/ui-strings/src/main/res/values/localazy.xml
```diff
@@ -176,6 +176,7 @@
   <string name="common_advanced_settings">"Advanced settings"</string>
   <string name="common_an_image">"an image"</string>
   <string name="common_analytics">"Analytics"</string>
+  <string name="common_android_notification_sync_notifications_foreground_service_title">"Fetching notifications…"</string>
```
bmarty, line 179: Note: I have removed this key from Localazy as the string is not used now.

bmarty (approved):
Thanks for the update. I have pushed a cleanup commit. Set to auto merge so we can have it in the next nightly.

PR #5612, opened 2025-10-27
Display only valid emojis in recent emoji list

jmartinesp:
## Content

- Create `:libraries:recentemojis` and move everything related to pure fetching/modification of recent emojis there.
- Make `GetRecentEmojis` remove duplicates and return only valid emojis (those that are in `Emojibase`).
- Move merging recent and suggested emojis to `ActionListPresenter` from `ActionListView`, so we can test it.

## Motivation and context

I was seeing some duplicate or empty emojis in my list of recent emojis. Some of them were emojis that looked the same visually but had a different representation (i.e. I had 2 green check mark ones, also in EW), but some others were being displayed blank (reactions for TWIM?).

## Tests

The unit tests should be enough, but you can try testing them these values by modifying the global account data.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/actionlist/ActionListView.kt
```diff
@@ -361,7 +361,7 @@ private fun EmojiReactionsRow(
         val backgroundColor = ElementTheme.colors.bgCanvasDefault
 
         val emojis = remember(recentEmojis) {
-            (suggestedEmojis + recentEmojis.filter { it !in suggestedEmojis })
+            (suggestedEmojis + recentEmojis).distinct()
```
bmarty: Nicer solution!

ganfra (approved):

Issue #5593, opened 2025-10-23
The incoming call UI appears much less frequently.


YamatoRyou:
### Steps to reproduce

When calling from another device, wait for the incoming call UI to appear.

#### What did you expect?
I wish this UI would appear for every incoming call.

#### What happened instead?
After updating to 25.10.1, I've noticed that this UI rarely appears.
I've made a video demonstrating this:
https://www.youtube.com/watch?v=iG54Pdpmn1g

The video shows three phones involved in the test, and the call process was repeated 11 times. This video is unedited.
As you can see from the video, the UI didn't appear 11 times on each device as expected (or at a satisfactory frequency). One device didn't even appear once from the beginning to the end.
This issue also occurred in previous versions, but not to the same degree as it is now.

**Battery, background restrictions, notifications, permissions, and other factors that affect notification appearance have been eliminated, and all devices use the same self-hosted ntfy instance as the push gateway.**

Logs:
[{BC758E23-B829-B71D-A702-783789107412}.zip](https://github.com/user-attachments/files/23093341/BC758E23-B829-B71D-A702-783789107412.zip)
Password: `***`
**Please download this file as soon as possible.**

### Your phone model

See Video.

### Operating system version

See Video.

### Application version and app store

Element X Android 25.10.1

### Homeserver

Synapse 1.140.0

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

jmartinesp:
I see lots of:

```
2025-10-23T07:58:54.067305Z DEBUG elementx: [DefaultActiveCallManager] Received timed-out incoming ringing call for room id: ..., cancel ringing | 
```

As if the received notification's expiration value was already past, so the call UI is not displayed.

And also a few of these:

```
2025-10-23T07:33:09.460542Z TRACE matrix_sdk_ui::notification_client: all notification events havenot  been found | crates/matrix-sdk-ui/src/notification_client.rs:596 | spans: try_sliding_sync
```

But I don't really know what causes them, there's not much info in the logs about it.

If you go back to the previous version, can you confirm this works well again? Maybe my changes in https://github.com/element-hq/element-x-android/pull/5563 somehow caused this regression.

Also, I'm on `25.10.1` and I can't reproduce it in my test device with a similar setup.

escix:
I also noticed this issue soon after upgrading to 25.10.1.
It doesn't even trigger the vibration - my phone is always on vibration mode - (neither on my phone nor on my smartwatch).

[...]

PR #5615, opened 2025-10-27
Revert "Make sure declining a call stops observing the ringing call state"

jmartinesp:
Reverts element-hq/element-x-android#5563

This PR apparently started causing issues for some users that stopped receiving notifications (see https://github.com/element-hq/element-x-android/issues/5593).


jmartinesp:
An affected user confirmed reverting the PR actually fixes their issues: https://github.com/element-hq/element-x-android/issues/5593#issuecomment-3451565940

jmartinesp:
Sadly, yes... The code inside isn't breaking anything in the app since everything is already cancelled when new events to cancel the call are received, but it messes up the logging quite a bit 🫤 .

PR #5628, opened 2025-10-29
Fix marking a room as read re-instantiates its timeline

jmartinesp:
## Content

- Stops using `Room.markAsRead` to mark room as read when exiting the room UI.
- `Timeline.getLatestEventId` + the `MarkAsFullyRead` use case will be used instead, or `Timeline.markAsRead`.
- Add a delay before displaying the loading indicator when loading a room.
- Move sending the fully read marker to a `MessagesPresenter` event handling case, which will do the operation as a fire and forget one and then navigate up.

## Motivation and context

Should fix https://github.com/element-hq/element-x-android/issues/5580.

## Tests

- Open an unread room.
- Exit the room.
- Check the unread marker is gone soon after/already gone when back at the room list.
- Open the room again.

This 2nd attempt to open the room should be as fast as the first one. If you take a look at the logs you shouldn't see any references to the timeline being instantiated again, as it happened before.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/utils/DelayedVisibility.kt
+++ b/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/utils/DelayedVisibility.kt
@@ -11,6 +11,7 @@ import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.movableContentOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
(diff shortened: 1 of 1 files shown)


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/utils/DelayedVisibility.kt
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Displays the content of [block] after a delay of [duration].
+ */
+@Composable
+fun DelayedVisibility(
+    duration: Duration = 300.milliseconds,
+    block: @Composable () -> Unit,
+) {
+    val isPreview = LocalInspectionMode.current
```
bmarty: Despite this, the Preview is impacted. Maybe just invoke `block()` when `isPreview` is `true`?

PR #5645, opened 2025-10-31
Improve rendering notification for multi account

bmarty:
## Content

Ensure that the userId is displayed in notification when several accounts are used.
Problem: the summary notification seems not to be displayed if there is only one MessagingStyle notification. So you need notification for at least 2 rooms for the account name to be visible. So better than before but not ideal yet.

## Motivation and context

Let the user know from which account the notification is coming in.

## Screenshots / GIFs

image

image

## Tests

- Login on 2 accounts
- Get notifications
- Observe that the notification may display the account name.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/push/impl/src/test/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotificationDrawerManagerTest.kt
+++ b/libraries/push/impl/src/test/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotificationDrawerManagerTest.kt
@@ -12,7 +12,6 @@ import androidx.compose.ui.graphics.Color
 import com.google.common.truth.Truth.assertThat
 import io.element.android.features.enterprise.api.EnterpriseService
 import io.element.android.features.enterprise.test.FakeEnterpriseService
-import io.element.android.libraries.matrix.test.AN_AVATAR_URL
 import io.element.android.libraries.matrix.test.AN_EVENT_ID
 import io.element.android.libraries.matrix.test.A_ROOM_ID
 import io.element.android.libraries.matrix.test.A_SESSION_ID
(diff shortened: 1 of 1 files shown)


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationRenderer.kt
```diff
@@ -41,24 +44,37 @@ class NotificationRenderer(
     ) {
         val color = enterpriseService.brandColorsFlow(currentUser.userId).first()?.toArgb()
             ?: NotificationConfig.NOTIFICATION_ACCENT_COLOR
+        val numberOfAccounts = sessionStore.getAllSessions().size
```
ganfra: We are getting all the session data each time we render a notification just to get the number of sessions? Not ideal

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationDataFactory.kt
```diff
@@ -29,41 +28,35 @@ import io.element.android.libraries.push.impl.notifications.model.SimpleNotifiab
 import io.element.android.services.toolbox.api.strings.StringProvider
 
 interface NotificationDataFactory {
-    suspend fun toNotifications(
-        messages: List<NotifiableMessageEvent>,
-        currentUser: MatrixUser,
+    suspend fun List<NotifiableMessageEvent>.toNotifications(
```
ganfra: This is a bit weird to use extension method on an interface like this

ganfra (commented):
Thanks for the changes.
Some remarks, otherwise LGTM

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationRenderer.kt
bmarty: True, I'll update this code when #5678 will be merged.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationDataFactory.kt
bmarty: OK, I'll revert this change

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationRenderer.kt
bmarty: Done in af12ff3c9aea3e9e3a997a0ed01760af80e7831d

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationDataFactory.kt
bmarty: be807f4

Issue #5608, opened 2025-10-24
"https://" required for the domain address of a custom server on setting up the app


Somebodyisnobody:
### Steps to reproduce

1. Start from fresh (new app installation or reset data)
2. Click on `Change account provider`
3. Click on `Other`
4. A message is shown below a field `Enter a search term or a domain address.`
5. Click on that field and enter a "domain address" (without the protocol as you should only enter "a search term or a domain address")
6. Click on "OK" on your keyboard to submit
7. Nothing happens

#### What did you expect?
Why the heck is "https://" required and not assumed?
Add here?:
https://github.com/element-hq/element-x-android/blob/f5e7a43f69f0adc83cb0191dc2248fdcfc184dd2/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/searchaccountprovider/SearchAccountProviderView.kt#L113-L116
```java
url = input;
if (!input.startsWith("https://")) {
  url = "https://" + input;
}
// call url
```

#### What happened instead?
Nohing. Just nothing. Really: Nothing. No Error message, no "you need to prefix with https://". Honestly: Nothing

### Your phone model

S25 EE

### Application version and app store

25.10.0

### Will you send logs?

No

### Are you willing to provide a PR?

No

davelowe1977:
I almost gave up on signing into my own homeserver because of this problem. Took me 20 minutes to figure it out. God help a new user.

Somebodyisnobody:
Yeah me too, I was really despaired and about to install the "Element Classic" app.
@jmartinesp I think this should get more attention. It is no big deal to
 - either clarify the message below the field,
 - show an appropriate error message
 - or assume the "https://" with a piece of code (preferred).

jmartinesp:
I'm not really sure what's happening, because the provided URL *should* eventually end up here:

https://github.com/element-hq/element-x-android/blob/97f3be3dc5352aea69d3568343f417fdbd59c8ef/libraries/core/src/main/kotlin/io/element/android/libraries/core/uri/UrlUtils.kt#L24

Which literally prepends `https://` to the provided value if it it doesn't start with `http`(we'd skip that for values already starting with `http` or `https`). So something is definitely not working as intended but I'm not sure what because I can log in just fine with `element.io`, `matrix.org`, `beta.matrix.org` and other homeservers without the need to add the `https://` prefix.

[...]

PR #5664, opened 2025-11-03
Use the SDK Client to check whether a homeserver is compatible

jmartinesp:
## Content

- Removes most classes associated with fetching the `.well-known/matrix/client` file and parsing it in the Android side.
- Instead, use `HomeServerLoginCompatibilityChecker` - name suggestions are welcome. Under the hood, this uses the `Client.homeserverLoginDetails()` function from the SDK to know if the homeserver has login types compatible with our client (OIDC, password).

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/5608

## Tests

Try to log in using a homeserver with password-only login, **without the https:// prefix**. If you don't know one, DM me and I can share one with you for the tests.

The server selection screen should list it as compatible. Also, you can try checking SSO-only homeservers aren't displayed, if you know some.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

### DIFF
--- a/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/searchaccountprovider/SearchAccountProviderStateProvider.kt
+++ b/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/searchaccountprovider/SearchAccountProviderStateProvider.kt
@@ -34,6 +34,7 @@ fun aSearchAccountProviderState(
 
 fun aHomeserverDataList(): List<HomeserverData> {
     return listOf(
+        aHomeserverData(homeserverUrl = AuthenticationConfig.MATRIX_ORG_URL),
         aHomeserverData(homeserverUrl = "https://no.sliding.sync"),
         aHomeserverData(homeserverUrl = "https://invalid"),
     )


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustHomeServerLoginCompatibilityChecker.kt
```diff
@@ -0,0 +1,32 @@
...
+@Inject
+class RustHomeServerLoginCompatibilityChecker(
+    private val clientBuilderProvider: ClientBuilderProvider,
+) : HomeServerLoginCompatibilityChecker {
+    override suspend fun check(url: String): Result<Unit> = runCatchingExceptions {
+        val client = clientBuilderProvider.provide().homeserverUrl(url).build()
+        client.use {
+            it.homeserverLoginDetails()
+        }.use {
+            assert(it.supportsOidcLogin() || it.supportsPasswordLogin()) {
```
bmarty: The usage of `assert` is a bit weird here, maybe the `check` method could return `Result<Boolean>` instead?

jmartinesp: I guess we can do that if we want to tell apart 'this is not a valid homeserver' and 'we couldn't check if it's a valid homeserver' results.

bmarty: Maybe create an interface to clarify?
```kotlin
sealed interface HomeServerLoginCompatibilityCheckerResult { // 
   object Invalid: HomeServerLoginCompatibilityCheckerResult
   data class Valid(
       val supportsOidcLogin: Boolean,
       val supportsPasswordLogin: Boolean,
   ) : HomeServerLoginCompatibilityCheckerResult
}
```
But maybe not necessary as you said

jmartinesp: I think using `Result.success` to indicate the server was reachable and compatible/not, then `Result.failure` for any actual errors is good enough.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/searchaccountprovider/SearchAccountProviderPresenter.kt
```diff
@@ -57,7 +57,7 @@ class SearchAccountProviderPresenter(
             userInput = userInput,
             userInputResult = data.value,
             changeServerState = changeServerState,
-            eventSink = ::handleEvents
+            eventSink = { handleEvents(it) },
```
bmarty: Was this change necessary?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustHomeServerLoginCompatibilityChecker.kt
```diff
@@ -0,0 +1,30 @@
...
+import io.element.android.libraries.matrix.api.auth.HomeServerLoginCompatibilityChecker
+import io.element.android.libraries.matrix.impl.ClientBuilderProvider
+
+@ContributesBinding(AppScope::class)
+@Inject
+class RustHomeServerLoginCompatibilityChecker(
+    private val clientBuilderProvider: ClientBuilderProvider,
+) : HomeServerLoginCompatibilityChecker {
+    override suspend fun check(url: String): Result<Boolean> = runCatchingExceptions {
+        val client = clientBuilderProvider.provide().homeserverUrl(url).build()
```
bmarty: This is probably the default behavior, but maybe make it explicit by calling `.inMemoryStore()`?

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/resolver/HomeserverResolver.kt
```diff
@@ -41,14 +38,13 @@ class HomeserverResolver(
...
-                        wellknownRetriever.getWellKnown(url)
-                    }
-                }
-                val isValid = wellKnown?.dataOrNull()?.isValid().orFalse()
+                val isValid = homeServerLoginCompatibilityChecker.check(url)
+                    .onFailure { Timber.w(it, "Failed to check compatibility with homeserver $url") }
+                    .getOrNull()
+                    ?: return@parallelMap
+
+                // Emit the list as soon as possible
```
bmarty, line 46: I believe that `HomeserverData.isWellknownValid` is not used anymore, maybe take the opportunity to do some cleanup? Can be done separately of course!

bmarty (commented):
When I try to change the account provider to `element` I get an infinite loading. With the previous code, `element.io` was found.

Also I think that something is missing to emit an empty list when there is no result (but this is maybe not a new issue?).
EDIT: Actually I think the code is [there](https://github.com/element-hq/element-x-android/blob/develop/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/searchaccountprovider/SearchAccountProviderPresenter.kt#L72), it is just taking ages to resolve the wellknown data using a Rust SDK.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/searchaccountprovider/SearchAccountProviderPresenter.kt
jmartinesp: Not really, I got struck with the weird reflection error and forgot it was associated to not consumed items 🙈 .

jmartinesp:
The new changes should fix the issue with `element.io`, I was using the wrong method to set the URL to discover.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/searchaccountprovider/SearchAccountProviderPresenter.kt
bmarty: We will improve that in the future 🤞

PR #5665, opened 2025-11-03
Disable forward of Event from the media viewer if coming from the Pinned timeline

bmarty:
## Content

Disable forward of Event from the media viewer if coming from the Pinned timeline.

## Motivation and context

Forwarding an event from the media viewer when opened from the pinned timeline can lead to error, as in this case the live timeline will be used, and the live timeline may not know the event from the pinned timeline. I did not find a quick solution for this issue so in the mean time, let's disable the feature. The use case should not happen often and it's still possible to forward the media from the pinned timeline.

## Screenshots / GIFs

No expected change

## Tests

- In a room pinned an image
- open the list of pinned events
- long click on the image: it's possible to forward the image
- click on the image: the image viewer is visible
- click on the "info" icon: the bottom sheet is visible and it's not possible to forward the image from here.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [ ] Pull request is based on the develop branch
- [ ] Pull request title will be used in the release note, it clearly define what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [ ] You've made a self review of your PR

PR #5669, opened 2025-11-03
Fix forward events from media viewer from pinned media timeline

bmarty:
## Content

Forwarding an event from the media viewer when opened from the pinned timeline can lead to error, as in this case the live timeline will be used, and the live timeline may not know the event from the pinned timeline.

Co-authored by @jmartinesp . Thanks for the fix!

## Motivation and context

Replacement for #5665

## Tests

- In a room pinned an old image (scroll far away in the past and pin an image)
- clear the cache to ensure that the live timeline does not know the just pinned event
- open the list of pinned events
- click on the image: the image viewer is visible
- click on the "info" icon and forward the image from here.
- it should work, previously an error occured

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/tests/uitests/src/test/snapshots/images/features.forward.impl_ForwardMessagesView_Day_3_en.png
+++ b/tests/uitests/src/test/snapshots/images/features.forward.impl_ForwardMessagesView_Day_3_en.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:5c6ab936964323971e9b4927684e3ae2da96884b340569aebe6e6a711bd6c241
-size 8446
+oid sha256:9bd3fe043d7fffab7b19b5af9de7ef58ce8d2d97018a058b5b3c94cb55b277b9
+size 11273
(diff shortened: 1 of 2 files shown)


libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerNode.kt
```diff
@@ -149,3 +145,11 @@ class MediaViewerNode(
...
     }
 }
+
+internal fun MediaViewerEntryPoint.MediaViewerMode.getTimelineMode(): Timeline.Mode? {
+    return when (this) {
+        is MediaViewerEntryPoint.MediaViewerMode.TimelineImagesAndVideos -> timelineMode
+        is MediaViewerEntryPoint.MediaViewerMode.TimelineFilesAndAudios -> timelineMode
+        else -> null
+    }
+}
```
jmartinesp, line 155: 👌

jmartinesp (approved):
Thanks!

PR #5672, opened 2025-11-03
Always use the handleEvent(s) function the same way.

bmarty:
Just a small cleanup on our codebase.

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
@@ -201,7 +201,7 @@ class CallScreenPresenter(
             userAgent = userAgent,
             isCallActive = isWidgetLoaded,
             isInWidgetMode = isInWidgetMode,
-            eventSink = { handleEvents(it) },
+            eventSink = ::handleEvents,
         )
     }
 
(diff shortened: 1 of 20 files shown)


jmartinesp:
This has the side effect of not displaying the actual error when something unexpected happens in the tests related to the state values, instead displaying an error saying it can't print the actual error because it doesn't know how to debug-print the state value because it doesn't have the kotlin reflection library.

We could maybe add [kotlin-reflect](https://mvnrepository.com/artifact/org.jetbrains.kotlin/kotlin-reflect) as a `testImplementation` dependency, see if that fixes the issue.

bmarty:
> This has the side effect of not displaying the actual error when something unexpected happens in the tests related to the state values, instead displaying an error saying it can't print the actual error because it doesn't know how to debug-print the state value because it doesn't have the kotlin reflection library.

When we have such issue in the test, we now know that it's because there are unconsumed state(s) (some `awaitItem()` are missing), so I think it's fine, do you agree?

jmartinesp:
> > This has the side effect of not displaying the actual error when something unexpected happens in the tests related to the state values, instead displaying an error saying it can't print the actual error because it doesn't know how to debug-print the state value because it doesn't have the kotlin reflection library.
> 
> When we have such issue in the test, we now know that it's because there are unconsumed state(s) (some `awaitItem()` are missing), so I think it's fine, do you agree?

*We* do, but it's not the first time - and won't be the last - I've had to tell some contributor why that happened (ask @richvdh ). If we can fix it with adding the library just for the unit tests it seems like a good trade-off.

jmartinesp (approved):
It seems like just adding the lib doesn't fix the mentioned issue. It's ok then, let's just merge this and we can later iterate on it.

PR #5679, opened 2025-11-04
Extract save change dialog

bmarty:
Extract the dialog to its own file.

This dialog will then be used in other places.


bmarty:
> Sorry for the late review. LGTM, but I think we need to fix some Konsist checks before merging.

Yes, I'll handle that. I have noticed that a previous dialog had no title. I think it's better now, but the wording is clearly not ideal in the context of creating a poll. This is not new and should not happen a lot, so...

PR #5684, opened 2025-11-05
Do not use the bestDescription but the caption for images, when available

### LINKED ISSUE #3945: empty notification
### Steps to reproduce

I dunno how to reproduce this

### Outcome

image

### Your phone model

op9p

### Operating system version

crdroid 9.7 on android 13

0.7.4

### Homeserver

_No response_

### Will you send logs?

No

No

bmarty:
## Content

Do not use the bestDescription but the caption for images because else the filename will be rendered in the notification and for image / gif / png we do not want that since the filename can be quite ugly and does not bring much.  Example:

image

Also fixes the issue when images is not rendered on some system and so they can be empty notification.

## Motivation and context

Avoid having random filename to be rendered in the notification below the images.
Also closes #3945, which was less visible since a text message was always rendered below the image, but with this change, it's not the case anymore

## Tests

- Receive images with caption: the images and the captions should be displayed in the notification drawer
- Receive images without caption: only the image should be displayed. If the image cannot be displayed, the filename will be displayed

## Tested devices

- [ ] Physical
- [x] Emulator, in particular Pixel 9A with API 36 which can repro #3945. It's now possible to have only images rendered in the notification.
- OS version(s):

## Checklist

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
@@ -8,7 +8,10 @@
 package io.element.android.libraries.push.impl.notifications
 
 import android.content.Context
+import android.graphics.ImageDecoder
 import android.net.Uri
+import android.os.Build
+import androidx.core.app.NotificationCompat
 import androidx.core.content.FileProvider
 import dev.zacsweers.metro.AppScope
 import dev.zacsweers.metro.ContributesBinding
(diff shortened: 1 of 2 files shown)


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
```diff
@@ -299,13 +308,18 @@ class DefaultNotifiableEventResolver(
...
+        hasImageUri: Boolean,
+    ): String? {
         return when (val messageType = content.messageType) {
             is AudioMessageType -> messageType.bestDescription
             is VoiceMessageType -> stringProvider.getString(CommonStrings.common_voice_message)
             is EmoteMessageType -> "* $senderDisambiguatedDisplayName ${messageType.body}"
             is FileMessageType -> messageType.bestDescription
-            is ImageMessageType -> messageType.bestDescription
+            is ImageMessageType -> if (hasImageUri) {
+                messageType.caption
```
jmartinesp, line 319: Should we add some fallback value if the caption is null?

bmarty: When there is no caption, only the image will be displayed, and this is what we expect.

PR #5709, opened 2025-11-10
Add alert to encrypted rooms with visible history (Android).

kaylendog:
## Content

- Adds a dismissable alert that is displayed whenever the user opens a room with `history_visibility` != `joined`. When cleared, this is recorded in the app's data store.
- When opening a room with `history_visibility` = `joined`, this flag is cleared.`

## Motivation and context

This is part of ongoing work to implement production-ready encrypted room history. 

Issue: https://github.com/element-hq/element-meta/issues/2875

## Screenshots / GIFs

image

## Tests

1. Create a new private (encrypted) room.  Since Element X defaults to `invited` history visibility, the new banner should be visible immediately upon opening the room.
3. Dismiss the banner, then re-open the room. It should not be visible.

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s):

## Checklist


appconfig/src/main/kotlin/io/element/android/appconfig/LearnMoreConfig.kt
```diff
@@ -13,4 +12,5 @@ object LearnMoreConfig {
     const val DEVICE_VERIFICATION_URL: String = "https://element.io/help#encryption-device-verification"
     const val SECURE_BACKUP_URL: String = "https://element.io/help#encryption5"
     const val IDENTITY_CHANGE_URL: String = "https://element.io/help#encryption18"
+    const val HISTORY_VISIBLE_URL: String = "https://example.com"
```
kaylendog: The proper URL is still pending.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/historyvisible/HistoryVisibleAcknowledgementRepository.kt
```diff
@@ -0,0 +1,42 @@
...
+import androidx.datastore.preferences.core.booleanPreferencesKey
+import androidx.datastore.preferences.core.edit
+import dev.zacsweers.metro.ContributesBinding
+import io.element.android.libraries.di.SessionScope
+import io.element.android.libraries.matrix.api.core.RoomId
+import io.element.android.libraries.preferences.api.store.PreferenceDataStoreFactory
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+
+interface HistoryVisibleAcknowledgementRepository {
```
kaylendog, line 21: I'm not sure if this is the correct naming scheme to use?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/historyvisible/HistoryVisibleState.kt
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+package io.element.android.features.messages.impl.crypto.historyvisible
+
+import io.element.android.libraries.matrix.api.room.history.RoomHistoryVisibility
+
+data class HistoryVisibleState(
+    val roomHistoryVisibility: RoomHistoryVisibility,
+    val roomIsEncrypted: Boolean,
+    val acknowledged: Boolean,
```
kaylendog: I could collapse this into a single `visible` flag that gets computed by the presenter. Unsure what is best practice.

bmarty: Yes, the best practice is to limit computations/logic in the View and do it in the presenter.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/historyvisible/HistoryVisibleStatePresenter.kt
```diff
@@ -0,0 +1,54 @@
...
+) : Presenter<HistoryVisibleState> {
+    @Composable
+    override fun present(): HistoryVisibleState {
+        val roomInfo by room.roomInfoFlow.collectAsState()
+        // Implicitly acknowledge the initial event to avoid flashes in UI.
+        val acknowledged by repository.hasAcknowledged(room.roomId).collectAsState(initial = true)
+
+        val coroutineScope = rememberCoroutineScope()
+
+        LaunchedEffect(roomInfo.historyVisibility) {
```
jmartinesp: Note this will run every time `roomInfo.historyVisibility` changes, which means at least once when the value is received. Do we always mark it as not acknowledged in the repository?

jmartinesp (commented):
Sorry for the long delay in reviewing this. I added a question, because I'm not sure I understand the logic behind one of the changes.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/historyvisible/HistoryVisibleStatePresenter.kt
kaylendog: This is done to make sure the UX is done per https://github.com/element-hq/element-meta/issues/2875:

> * Whenever the user opens an encrypted room with `history_visibility` == `joined`, clear the dismissed flag, if it is set. This ensures that if a room is changed public -> private -> public, we show the banner again when it is set back to public.

I believe this is the best way to implement this?

jmartinesp: Ok, now I understand the logic:

- If you join a room with `visibility != joined`, that means you can read previous messages so we display this banner unless it was acknowledged before.
- If the room `visibility == joined` there's no need to display the banner, and we remove the acknowledgement in the store so if the visibility becomes `!= joined` again the user will have to acknowledge again too.

jmartinesp:
Could you record new screenshots following the instructions [here](https://github.com/element-hq/element-x-android/pull/5709#issuecomment-3511420292)? I've tried doing it myself but I think it'll fail since the action won't have permissions to add code to your fork. Thanks!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/historyvisible/HistoryVisibleAcknowledgementRepository.kt
```diff
@@ -0,0 +1,42 @@
...
+    fun hasAcknowledged(roomId: RoomId): Flow<Boolean>
+    suspend fun setAcknowledged(roomId: RoomId, value: Boolean)
+}
+
+@ContributesBinding(SessionScope::class)
+class DefaultHistoryVisibleAcknowledgementRepository(
+    preferenceDataStoreFactory: PreferenceDataStoreFactory,
+) : HistoryVisibleAcknowledgementRepository {
+    val store = preferenceDataStoreFactory.create("elementx_historyvisible")
```
bmarty: To properly support multi-account, I believe that we would need to have a store per account. You can inject the `sessionId` in the constructor and use it to create a dedicated store.
Also when the session is deleted (log out), I guess we should do some cleanup? We can handle this part maybe since the code is not ideal so far regarding this.

bmarty: an other option is to have a single store, but include the `sessionId` in the preference keys.

kaylendog: Should be added in https://github.com/element-hq/element-x-android/pull/5709/commits/706f12f081f77702a712cdbdc3d25db4ced7246f

bmarty: (cross-posting my comment here for visibility)
Even if the spec is strict on `userId` grammar, we can see weird things in the wild, so it's better to compute a hash of it before trying to use it as filenames. You can do the same as [here](https://github.com/element-hq/element-x-android/blob/develop/features/invite/impl/src/main/kotlin/io/element/android/features/invite/impl/DefaultSeenInvitesStore.kt#L45) to compute the string.

kaylendog: Done! https://github.com/element-hq/element-x-android/pull/5709/commits/68f979a97f5c692c4a978722d3547cf022c0a346

appconfig/src/main/kotlin/io/element/android/appconfig/LearnMoreConfig.kt
```diff
@@ -13,4 +13,5 @@ object LearnMoreConfig {
     const val DEVICE_VERIFICATION_URL: String = "https://element.io/help#encryption-device-verification"
     const val SECURE_BACKUP_URL: String = "https://element.io/help#encryption5"
     const val IDENTITY_CHANGE_URL: String = "https://element.io/help#encryption18"
+    const val HISTORY_VISIBLE_URL: String = "https://element.io/en/help#e2ee-history-sharing"
```
bmarty, line 16: This HTML anchor is not live yet, right? I cannot see it.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
```diff
@@ -476,6 +477,10 @@ private fun MessagesViewComposerBottomSheetContents(
                         state = state.identityChangeState,
                         onLinkClick = onLinkClick,
                     )
+                    HistoryVisibleStateView(
+                        state = state.historyVisibleState,
+                        onLinkClick = onLinkClick,
+                    )
```
bmarty: We can have a double banner with the banner `IdentityChangeStateView` above. I guess we do not to have this situation?

Image

Maybe nest in a if block with condition `state.identityChangeState.roomMemberIdentityStateChanges.isEmpty()`?

features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl/crypto/historyvisible/HistoryVisibleStatePresenterTest.kt
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    private fun createHistoryVisibleStatePresenter(
+        room: JoinedRoom = FakeJoinedRoom(),
+        enabled: Boolean = true,
+        acknowledged: Boolean = false
+    ): HistoryVisibleStatePresenter {
+        return HistoryVisibleStatePresenter(
+            room = room,
+            featureFlagService = FakeFeatureFlagService(mapOf("feature.enableKeyShareOnInvite" to enabled)),
```
bmarty, line 123: Better to avoid hard coded strings. Can you replace by `FeatureFlags.EnableKeyShareOnInvite.key`?
(the same apply to all the occurrences of this string in this file)

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/historyvisible/HistoryVisibleStatePresenter.kt
```diff
@@ -0,0 +1,56 @@
...
+    @Composable
+    override fun present(): HistoryVisibleState {
+        val isFeatureEnabled by featureFlagService.isFeatureEnabledFlow(FeatureFlags.EnableKeyShareOnInvite).collectAsState(initial = false)
+        val roomInfo by room.roomInfoFlow.collectAsState()
+        // Implicitly assume the alert is initially acknowledged to avoid flashes in UI.
+        val acknowledged by repository.hasAcknowledged(room.roomId).collectAsState(initial = true)
+
+        val coroutineScope = rememberCoroutineScope()
+
+        LaunchedEffect(roomInfo.historyVisibility) {
```
bmarty: I think you need to use `acknowledged` as a key too, so that the launch effect is started again when the value change. It's more correct technically, even if IRL it should not cause any issue.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/historyvisible/HistoryVisibleStatePresenter.kt
```diff
@@ -0,0 +1,56 @@
...
+            if (isFeatureEnabled && roomInfo.historyVisibility == RoomHistoryVisibility.Joined && acknowledged) {
```
bmarty: I would not bother about testing the value of `isFeatureEnabled` here.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/historyvisible/HistoryVisibleStatePresenter.kt
```diff
@@ -0,0 +1,56 @@
...
+                repository.setAcknowledged(room.roomId, false)
+            }
+        }
+
+        return HistoryVisibleState(
+            showAlert = isFeatureEnabled && roomInfo.historyVisibility != RoomHistoryVisibility.Joined && roomInfo.isEncrypted == true && !acknowledged,
+            eventSink = { event ->
```
bmarty: Neat: we usually use a fun `handleEvent` in other presenters.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/historyvisible/MessagesViewWithHistoryVisiblePreview.kt
```diff
@@ -0,0 +1,45 @@
...
+import io.element.android.features.messages.impl.aMessagesState
+import io.element.android.features.messages.impl.messagecomposer.aMessageComposerState
+import io.element.android.libraries.designsystem.preview.ElementPreview
+import io.element.android.libraries.designsystem.preview.PreviewsDayNight
+import io.element.android.libraries.textcomposer.model.aTextEditorStateMarkdown
+
+@PreviewsDayNight
+@Composable
+internal fun MessagesViewWithHistoryVisiblePreview(
+    @PreviewParameter(HistoryVisibleStateProvider::class) historyVisibleState: HistoryVisibleState
```
bmarty: The added preview with the first state (banner not visible) does not bring any added value, so I'd remove the preview parameter `historyVisibleState` and replace it with `aHistoryVisibleState(showAlert = true)` at line 32.

bmarty (changes requested):
Thanks for the update, a few more remarks

appconfig/src/main/kotlin/io/element/android/appconfig/LearnMoreConfig.kt
kaylendog: It isn't, no - this is pending on @mxandreas's end, I believe.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
kaylendog: Addressed in https://github.com/element-hq/element-x-android/pull/5709/commits/7a3f5b3399c1a5086b20e0739ea78e1268059dd7

features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl/crypto/historyvisible/HistoryVisibleStatePresenterTest.kt
kaylendog: Addressed in 2b5169dc4f742be20022e1d5454d1732d3ddf411

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/historyvisible/HistoryVisibleStatePresenter.kt
kaylendog: Addressed in 2b5169dc4f742be20022e1d5454d1732d3ddf411

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/historyvisible/HistoryVisibleStatePresenter.kt
kaylendog: Addressed in 2b5169dc4f742be20022e1d5454d1732d3ddf411

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/crypto/historyvisible/MessagesViewWithHistoryVisiblePreview.kt
kaylendog: Addressed in 2b5169dc4f742be20022e1d5454d1732d3ddf411, it's probably better to be consistent about these things!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesStateProvider.kt
```diff
@@ -83,6 +87,19 @@ open class MessagesStateProvider : PreviewParameterProvider<MessagesState> {
...
+            ),
+            aMessagesState(
+                composerState = aMessageComposerState(textEditorState = aTextEditorStateMarkdown()),
+                historyVisibleState = aHistoryVisibleState(showAlert = true)
+            ),
+            aMessagesState(
+                composerState = aMessageComposerState(textEditorState = aTextEditorStateMarkdown()),
+                identityChangeState = anIdentityChangeState(listOf(aRoomMemberIdentityStateChange())),
+                historyVisibleState = aHistoryVisibleState(showAlert = true)
+            )
```
kaylendog, line 102: These may belong in `MessagesViewWithHistoryVisible` -  I thought it would be good to test what happens when these two alerts both meet their visibility conditions at the same time.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
```diff
@@ -472,10 +473,17 @@ private fun MessagesViewComposerBottomSheetContents(
...
+                        IdentityChangeStateView(
+                            state = state.identityChangeState,
+                            onLinkClick = onLinkClick,
+                        )
+                    } else {
+                        HistoryVisibleStateView(
+                            state = state.historyVisibleState,
+                            onLinkClick = onLinkClick,
+                        )
+                    }
```
bmarty, line 500: Thanks for the fix. It's OK like that, even if it can make reader feel that `HistoryVisibleStateView` has no condition to be displayed (I know the method can return early if nothing has to be displayed)

features/messages/impl/src/main/res/values/temporary.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
+    <string name="crypto_history_visible">Messages you send will be shared with new members invited to this room. %1$s</string>
```
bmarty: Do you need any help to add this string to Localazy? I understand that this is temporary, but It can't stay here, it won't be translated in other languages.

kaylendog:
I believe the failing Maestro CI is the same issue we've had before since I'm operating off a fork.

jmartinesp:
Yeah, that's expected. Maestro is not super stable at the moment, and it never worked fine in forks.

features/messages/impl/src/main/res/values/temporary.xml
kaylendog: If you could walk me through, that would be great for subsequent string additions!

appconfig/src/main/kotlin/io/element/android/appconfig/LearnMoreConfig.kt
mxandreas: Yes, it is not live - and I'd like a few copy/UI adjustments in the room settings https://github.com/element-hq/element-meta/issues/3029 before it goes live as with the current copy it is very hard to write the Learn more (FAQ). But I have the draft for the FAQ and have asked to review/prepare it for publishing.

features/messages/impl/src/main/res/values/temporary.xml
```diff
@@ -0,0 +1,4 @@
...
+</resources>
```
bmarty: Remove the file now?

kaylendog: Oh yes! Silly me. Addressed in f7006a559296839f49f03cc6043b08118697545e

PR #5711, opened 2025-11-10
Fix rendering notifications after receiving redundant push

SpiritCroc:
## Content

Deduplicate event IDs before iterating them, to ensure we only ever look up (and thus destroy) any item at most once.

## Motivation and context

Since merging v25.11.2 into SchildiChat Next, notifications have often been failing with "Unable to resolve event: Unknown error: TimelineEvent object has already been destroyed".

I have not 100% confirmed that this patch fixes it yet, doing the PR now anyway so I don't forget later. I will keep using my app with this patch for a while to observe.

I can artificially reproduce the error by forcing the code to iterate over the same event ID twice, so it may be good to have either way. Last time I ran into this and looked at the logs, I also found two "elementx: Queueing notification: NotificationEventRequest" log lines for the event that eventually failed, which may also support this theory. But again, to be further observed.

## Tests

Use the app for a while and observe if notifications fail to render via "Unable to resolve event: Unknown error: TimelineEvent object has already been destroyed".

## Tested devices

- [x] Physical: (currently testing on SchildiChat Next on v25.11.2 base, to be observed in the long run)
- [ ] Emulator
- OS version(s): Android 15

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/notification/RustNotificationService.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/notification/RustNotificationService.kt
@@ -45,7 +45,7 @@ class RustNotificationService(
             }
             val items = notificationClient.getNotifications(requests)
             buildMap {
-                val eventIds = requests.flatMap { it.eventIds }
+                val eventIds = requests.flatMap { it.eventIds }.distinct()
                 for (rawEventId in eventIds) {
                     val roomId = RoomId(requests.find { it.eventIds.contains(rawEventId) }?.roomId!!)
                     val eventId = EventId(rawEventId)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/notification/RustNotificationService.kt
```diff
@@ -45,7 +45,7 @@ class RustNotificationService(
             }
             val items = notificationClient.getNotifications(requests)
             buildMap {
-                val eventIds = requests.flatMap { it.eventIds }
+                val eventIds = requests.flatMap { it.eventIds }.distinct()
```
jmartinesp, line 48: Oh, very good catch! I tried to fix this before by making sure we didn't have several `.use { ... }` blocks for the same object, but it never occurred to me this could be another source for the issue. In any case, it seems like a sensible check, so LGTM.

Issue #5688, opened 2025-11-05
EXA 25.11.02 crash when opening after sharing a video


escix:
### Steps to reproduce

1. Where are you starting? What can you see?

Shared a video from filebrowser and when I tried to open EXA it shows me the bug report screen

#### What did you expect?
No crash/crash report

#### What happened instead?
Showing a Report bug screen

### Your phone model

S23 Ultra

### Application version and app store

EXA 25.11.2

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

escix:
I can't even send the bug report

![Image](https://github.com/user-attachments/assets/ce49942b-19ae-4019-8bbf-7dd725aeff6e)

jmartinesp:
It seems to be an stack overflow related to navigation:

```
at androidx.activity.compose.BackHandlerKt$BackHandler$backCallback$1$1.handleOnBackPressed(BackHandler.kt:89)
at androidx.activity.OnBackPressedDispatcher.onBackPressed(OnBackPressedDispatcher.kt:260)
at androidx.activity.ComponentActivity.onBackPressed(ComponentActivity.kt:588)
at com.bumble.appyx.core.integrationpoint.ActivityIntegrationPoint.handleUpNavigation(ActivityIntegrationPoint.kt:43)
at com.bumble.appyx.core.node.Node.navigateUp(Node.kt:221)
at io.element.android.appnav.LoggedInFlowNode$resolve$3.onDone(LoggedInFlowNode.kt:486)
at io.element.android.features.share.impl.ShareNode$resolve$callback$1.onCancel(ShareNode.kt:65)
at io.element.android.libraries.roomselect.impl.RoomSelectNode$View$1$1.invoke(RoomSelectNode.kt:45)
at io.element.android.libraries.roomselect.impl.RoomSelectNode$View$1$1.invoke(RoomSelectNode.kt:45)
at io.element.android.libraries.roomselect.impl.RoomSelectViewKt.RoomSelectView$onBackButton(RoomSelectView.kt:87)
at io.element.android.libraries.roomselect.impl.RoomSelectViewKt.RoomSelectView$lambda$1$0(RoomSelectView.kt:91)
at io.element.android.libraries.roomselect.impl.RoomSelectViewKt.$r8$lambda$ATzTdmFHg1O_IUUoA40_p1yKU68(Unknown Source:0)
[...]

PR #5724, opened 2025-11-13
Fix navigation stack overflow when sharing media

jmartinesp:
## Content

Makes sure the `BackHandler` can only navigate up once in the `RoomSelectView` component.

This screen is a good candidate for the `SafeBackHandler` we previously talked about @bmarty .

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/5688.

## Tests

- From another app, share some file or media.
- When cancelling or actually sending the file, you should navigate to the room and see the file being sent.

Previously, this would get in a stack overflow where `navigateUp` triggered a back event that got caught by `BackHandler`, which in turn navigated up, and ended up looping in that logic.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/libraries/roomselect/impl/src/main/kotlin/io/element/android/libraries/roomselect/impl/RoomSelectView.kt
+++ b/libraries/roomselect/impl/src/main/kotlin/io/element/android/libraries/roomselect/impl/RoomSelectView.kt
@@ -89,7 +89,7 @@ fun RoomSelectView(
     fun onBackButton(state: RoomSelectState) {
         if (state.isSearchActive) {
             state.eventSink(RoomSelectEvents.ToggleSearchActive)
-        } else {
+        } else if (canHandleBack) {
             canHandleBack = false
             onDismiss()
         }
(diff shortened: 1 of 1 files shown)


libraries/roomselect/impl/src/main/kotlin/io/element/android/libraries/roomselect/impl/RoomSelectView.kt
```diff
@@ -81,15 +85,20 @@ fun RoomSelectView(
         )
     }
 
+    var canHandleBack by remember { mutableStateOf(true) }
     fun onBackButton(state: RoomSelectState) {
         if (state.isSearchActive) {
             state.eventSink(RoomSelectEvents.ToggleSearchActive)
         } else {
+            canHandleBack = false
```
bmarty, line 93: I think this is OK, but we do not check the value of `canHandleBack` before setting it and invoking `onDismiss`? Asking because `navigationIcon` click also invokes the fun `onBackButton`.

jmartinesp:
> Thanks! Why not creating a `SafeBackHandler` in a shared module so that we can start using it?

I thought you wanted to do that, so I was just adding this as a workaround in the meantime 😅 .

libraries/roomselect/impl/src/main/kotlin/io/element/android/libraries/roomselect/impl/RoomSelectView.kt
jmartinesp: Hmm, calling `onBackButton` from the back navigation icon could cause a different issue (2 or more back paginations if called quickly?), but it wouldn't interfere with the `BackHandler` here AFAICT.

PR #5726, opened 2025-11-13
Notification robustness

bmarty:
## Content

Ensure pusher (and topic in case of usage of UnifiedPush) are unregistered if the user disables the notifications. This will ensure that no Push will be sent (and ignored) if the user disables the notifications.

If the user has configured UnifiedPush, let EX be robust against manual topic deletion by creating a new endpoint and registering it. This will be done only once in a delay of 1 minute.

## Motivation and context

Robust notifications

## Screenshots / GIFs

https://github.com/user-attachments/assets/766389b1-462c-46c5-a08f-c0ef42355336

EDIT: after 2 successive removals, the pusher is not registered again automatically and the notification is displayed.

https://github.com/user-attachments/assets/de708e38-2348-413b-9772-7640e3b219d4

## Tests

- From the notification settings screen, disable the notification for the session
- The pusher should be unregistered. If using ntfy, the topic should vanish from the list
- enable back the notification
- the pusher should be registered again and a new topic should appear.

From ntfy application
- delete the topic
- EX should detect that and register a new pusher with a new topic.
- A new topic is visible to the list
- In case of error, a notification will be displayed to alert the user that notification will not work anymore.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
@@ -360,14 +360,11 @@ class DefaultNotificationCreator(
         notificationAccountParams: NotificationAccountParams,
     ): Notification {
         val userId = notificationAccountParams.user.userId
-        val text = if (notificationAccountParams.showSessionId) {
-            // TODO i18n
-            "$userId will not receive notifications anymore."
-        } else {
-            // TODO i18n
-            "You will not receive notifications anymore."
-        }
+        val text = stringProvider.getString(R.string.notification_error_unified_push_unregistered_android)
         return NotificationCompat.Builder(context, notificationChannels.getChannelIdForTest())
+            .setSubText(userId.value)
+            // The text is long and can be truncated so use BigTextStyle.
+            .setStyle(NotificationCompat.BigTextStyle().bigText(text))
             .setContentTitle(stringProvider.getString(CommonStrings.dialog_title_warning))
             .setContentText(text)
             .configureWith(notificationAccountParams)
@@ -479,7 +476,7 @@ class DefaultNotificationCreator(
                 .build()
         ).also {
             it.conversationTitle = if (isThread) {
-                stringProvider.getString(CommonStrings.notification_thread_in_room, roomName)
+                stringProvider.getString(R.string.notification_thread_in_room, roomName)
             } else {
                 roomName
             }
--- a/libraries/push/impl/src/main/res/values-cs/translations.xml
+++ b/libraries/push/impl/src/main/res/values-cs/translations.xml
@@ -42,6 +42,7 @@
   <string name="notification_sender_me">"Já"</string>
   <string name="notification_sender_mention_reply">"%1$s zmínil(a) nebo odpověděl(a)"</string>
   <string name="notification_test_push_notification_content">"Prohlížíte si oznámení! Klikněte na mě!"</string>
+  <string name="notification_thread_in_room">"Vlákno v %1$s"</string>
   <string name="notification_ticker_text_dm">"%1$s: %2$s"</string>
   <string name="notification_ticker_text_group">"%1$s: %2$s %3$s"</string>
   <plurals name="notification_unread_notified_messages">
--- a/libraries/push/impl/src/main/res/values-da/translations.xml
+++ b/libraries/push/impl/src/main/res/values-da/translations.xml
@@ -38,6 +38,7 @@
   <string name="notification_sender_me">"Mig"</string>
   <string name="notification_sender_mention_reply">"%1$s nævnt eller besvaret"</string>
   <string name="notification_test_push_notification_content">"Du ser notifikationen! Klik på mig!"</string>
+  <string name="notification_thread_in_room">"Tråd i %1$s"</string>
   <string name="notification_ticker_text_dm">"%1$s: %2$s"</string>
   <string name="notification_ticker_text_group">"%1$s: %2$s %3$s"</string>
   <plurals name="notification_unread_notified_messages">
--- a/libraries/push/impl/src/main/res/values-de/translations.xml
+++ b/libraries/push/impl/src/main/res/values-de/translations.xml
@@ -38,6 +38,7 @@
   <string name="notification_sender_me">"Ich"</string>
   <string name="notification_sender_mention_reply">"%1$s hat Dich erwähnt oder geantwortet"</string>
   <string name="notification_test_push_notification_content">"Du siehst dir die Benachrichtigung an! Klicke hier!"</string>
+  <string name="notification_thread_in_room">"Thread in %1$s"</string>
   <string name="notification_ticker_text_dm">"%1$s: %2$s"</string>
   <string name="notification_ticker_text_group">"%1$s: %2$s %3$s"</string>
   <plurals name="notification_unread_notified_messages">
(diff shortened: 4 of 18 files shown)


appnav/src/main/kotlin/io/element/android/appnav/loggedin/LoggedInPresenter.kt
```diff
@@ -133,59 +143,6 @@ class LoggedInPresenter(
         currentSlidingSyncVersion == SlidingSyncVersion.Proxy
     }
 
-    private suspend fun ensurePusherIsRegistered(pusherRegistrationState: MutableState<AsyncData<Unit>>) {
```
bmarty, line 136: Code moved to `DefaultPushService` so that can be reused by the notification settings screen

libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushRemovedGatewayHandler.kt
```diff
@@ -0,0 +1,53 @@
...
+import timber.log.Timber
+
+private val loggerTag = LoggerTag("UnifiedPushRemovedGatewayHandler", LoggerTag.PushLoggerTag)
+
+/**
+ * Handle new endpoint received from UnifiedPush. Will update the session matching the client secret.
+ */
+interface UnifiedPushRemovedGatewayHandler {
+    suspend fun handle(clientSecret: String): Result<Unit>
+}
```
jmartinesp, line 32: On one hand, I'm super happy you added this because it'll be useful for a draft PR of mine. On the other, I see quite a few conflicts in my near future 😅 .

libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnregisterUnifiedPushUseCase.kt
```diff
@@ -21,12 +21,19 @@ interface UnregisterUnifiedPushUseCase {
     /**
      * Unregister the app from the homeserver, then from UnifiedPush.
```
jmartinesp: Maybe update the doc comments?

libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushRemovedGatewayHandler.kt
```diff
@@ -29,25 +30,53 @@ class DefaultUnifiedPushRemovedGatewayHandler(
...
+                                    Timber.tag(loggerTag.value).w(it, "Unable to register with current data")
+                                }
+                        } else {
+                            Result.failure(IllegalStateException("Unable to register again"))
+                        }
+                    }
+                    .onFailure {
+                        // Let the user know
+                        pushService.onServiceUnregistered(userId)
+                    }
```
jmartinesp: Could we extract this to its own function? Having those nested flatMaps and onFailure calls makes it a bit difficult to follow.

libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushRemovedGatewayHandler.kt
```diff
@@ -58,14 +80,21 @@ class DefaultUnifiedPushRemovedGatewayHandler(
...
-                            pushService.registerWith(
-                                client,
-                                pushProvider,
-                                distributor,
-                            )
-                                .onFailure {
-                                    Timber.tag(loggerTag.value).w(it, "Unable to register with current data")
-                                }
+                            if (unifiedPushRemovedGatewayThrottler.canRegisterAgain()) {
+                                pushService.registerWith(
```
jmartinesp: Maybe add some comments for the reason this is needed?

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
```diff
@@ -382,6 +356,28 @@ class DefaultNotificationCreator(
...
+        notificationAccountParams: NotificationAccountParams,
+    ): Notification {
+        val userId = notificationAccountParams.user.userId
+        val text = if (notificationAccountParams.showSessionId) {
+            // TODO i18n
+            "$userId will not receive notifications anymore."
+        } else {
+            // TODO i18n
+            "You will not receive notifications anymore."
+        }
```
jmartinesp: Just a reminder to wait until we have the actual strings here before merging. Also, the text should explain why the notifications won't be received anymore.

bmarty: Yes, there is room for improvement. I did not want to add to Localazy until we have a final version.
I am not sure how to phrase this TBH. Maybe something like `You will not receive notifications anymore because some changes happened in the UnifiedPush application.`. We may be able to include the push distributor name.

Do you have a better suggestion? I think it can be a bit technical.

libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushRemovedGatewayHandler.kt
bmarty: I agree, I will try to improve. The thing is that for some error I want the user to be notified (`pushService.onServiceUnregistered(userId)`) and for other no. `onFailure` are also used for logs, but yes, it's brain consuming. UT will help to rework this without breaking anything.

libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushRemovedGatewayHandler.kt
bmarty: Sure, you mean explain why `canRegisterAgain()` is needed?

jmartinesp: Yeah, a description of the flow: we try to register, fail, then try to register again in < 60s, which instead of retrying displays an error to the user (if I understood the code correctly).

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
jmartinesp: What about "The UnifiedPush notification distributor couldn't be registered, so you will not receive notifications anymore. Please check the notifications settings of the app and the status of the push distributor." ?

libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushRemovedGatewayHandler.kt
bmarty: Ah that's not really that. We do not automatically retry in case of failure.
See my second video in the description. We avoid registering again it the user persist to delete the topic in th ntfy application.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
bmarty: Thanks. Added to Localazy and here is a preview:

image

PR #5731, opened 2025-11-14
Send read receipts using the current timeline, not the live timeline

jmartinesp:
## Content

This fixes sending read receipts in threads.

cc @frebib 

## Motivation and context

If you opened a thread, previously the app would try to send read receipts for it using the live timeline. As those events aren't really in the live timeline, this would be rejected by the homeserver.

## Tests

You'll need 2 clients (EX/EW) with the same account and one with a different account to test it. With threads enabled:

- From another account, send a message in a thread that this client will receive and take as an unread.
- With this client, open such thread.
- From the other client with the same account, check the read receipt is now gone.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
@@ -389,13 +389,17 @@ class TimelinePresenter(
     ) = launch(dispatchers.computation) {
         // If we are at the bottom of timeline, we mark the room as read.
         if (firstVisibleIndex == 0) {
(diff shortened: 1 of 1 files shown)


jmartinesp:
Note there is a similar issue, causing the live timeline to send an incorrect read receipt when the latest event of the room is in a thread even if there were unread events in the live timeline: similarly, it will try sending a read receipt for a threaded event from the live timeline, but as this event is not in this timeline, the HS will reject it. That will be handled in https://github.com/matrix-org/matrix-rust-sdk/pull/5864 instead.

jmartinesp:
Should we do it in this PR or move it to a different one?

bmarty:
> Should we do it in this PR or move it to a different one?

I guess it can be a follow up PR :)

Issue #5723, opened 2025-11-12
Unified Push test fails with 406


mcepl:
### Steps to reproduce

1. Using Element X from F-Droid (version 25.11.0) and SunUp 1.1.0 (also from F-Droid)
2. Application seems to work reasonably well (I haven’t tested it long enough to be certain about it)
3. When I go to (back-translating from Czech) Settings/Annoucements/Fixing problems/Run tests, most tests pass except for few:

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/08a3cb25-ae1c-446d-8cb2-5466df4900b2" />

The text of the error is:

> Test push gateway Fail to check the gateway https://updates.push.services.mozilla .com/ matrix/push/v1/notify: HTTP 406 
> Otestovat push zpětnou smyčku Chyba: HTTP 406 

(BTW, the Czech translation here is horrendous, I am not sure even what it is supposed to mean)

### Your phone model

Motorola Edge 30 neo (XT2245-1)

### Operating system version

Android 14 UISSMS34.31-64-10-9

### Application version and app store
Element X from F-Droid (version 25.11.0) and SunUp 1.1.0 (also from F-Droid)

### Homeserver

Element One as @mcepl:one.ems.host

### Will you send logs?

Not sure what exactly you want

### Are you willing to provide a PR?

No idea how

jmartinesp:
When I ran some tests related to this yesterday using Sunup with the default server, this worked fine. So it seems likely something changed or broke in the push server, since the SunUp app hasn't been updated in 7 months on F-Droid.

mcepl:
Just to note that the lookup test still fails with SinUp 1.2.2 and Element X 25.12.0.

p1gp1g:
I've just tested and it works. You may have been in an inconsistent state with a previous version. Try to disable UnifiedPush and re-enable again

<img width="540" height="2400" alt="Screenshot of the troubleshoot notifications page from Element-X, using Sunup, all tests pass" src="https://github.com/user-attachments/assets/2c175010-cb3d-47f3-821e-980cd0087e37" />

bmarty:
> (BTW, the Czech translation here is horrendous, I am not sure even what it is supposed to mean)

Side note: translations into Czech are provided by the community, please feel free to improve them from here: https://localazy.com/p/element/phrases/60. Thanks!

bmarty:
Is the issue fixed on your side @mcepl ?

mcepl:
[...]

PR #5741, opened 2025-11-15
Fix push gateway with some push provider (Sunup/autopush)

p1gp1g:
## Content

This PR adds a few HTTP codes that imply the push server doesn't have any matrix gateway: it returns NoMatrixGateway instead of Error.
 
Fixes: https://github.com/element-hq/element-x-android/issues/5723

## Motivation and context

Autopush (Sunup push server) returns a 406 to the matrix gateway discovery request (GET to /_matrix/push/v1/notify), which is currently considered as a network error and therefore fallback to the stored push gateway if it exists or to the custom gateway \*.

\* It should fallback to the default gateway - but I'm opening another PR for this, in case there is a need I am not aware of

## Tests

- Install Sunup
- Enable UnifiedPush and selection Sunup
- Go to notification troubleshooting list
- Observe the nonexistent gateway being used

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/pushproviders/unifiedpush/src/test/kotlin/io/element/android/libraries/pushproviders/unifiedpush/DefaultUnifiedPushGatewayResolverTest.kt
+++ b/libraries/pushproviders/unifiedpush/src/test/kotlin/io/element/android/libraries/pushproviders/unifiedpush/DefaultUnifiedPushGatewayResolverTest.kt
@@ -118,10 +118,40 @@ class DefaultUnifiedPushGatewayResolverTest {
     }
 
     @Test
-    fun `when a custom url is forbidden (403), Error is returned`() = runTest {
+    fun `when a custom url is forbidden (403), NoMatrixGateway is returned`() = runTest {
         val unifiedPushApiFactory = FakeUnifiedPushApiFactory(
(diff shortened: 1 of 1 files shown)


libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushGatewayResolver.kt
jmartinesp, line 80: Super nitpick: since this isn't a constant, it should start with lower case. We can change this later though.

jmartinesp (approved):
Thanks, this fixed my issues with using Sunup! @bmarty since you're the most knowledgeable around unified push, would you like to take a look too?

jmartinesp:
It seems like there are some unit tests to fix: https://github.com/element-hq/element-x-android/actions/runs/19391981774/job/55563077771?pr=5741

p1gp1g:
Oh yeah, I forgot to patch it

PR #5742, opened 2025-11-15
Fallback to the default push gateway on error

p1gp1g:
## Content

Based on #5741.

Fallback to the default gateway if the discovery request returns an error and no gateway was recorded for the push server yet.

## Motivation and context

We have the issue #5723 because the gateway resolver uses the custom gateway by default, when the request fails and the push server doesn't have any gateway yet. This exact issue is fixed with #5741 because 406 is included in valid responses showing no gateway is available. But we wouldn't have that issue if the fallback was to the default gateway.

I have split the 2 patchs in 2 PR, because I may not be aware of a need that let you choose to fallback to the custom gateway instead of the default, and I don't want to block the other patch.

## Tests

- Install Sunup
- Enable UnifiedPush and selection Sunup
- Go to notification troubleshooting list
- Observe the nonexistent gateway being used

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushGatewayUrlResolver.kt
+++ b/libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushGatewayUrlResolver.kt
@@ -28,8 +28,9 @@ class DefaultUnifiedPushGatewayUrlResolver(
     ): String {
         return when (gatewayResult) {
             is UnifiedPushGatewayResolverResult.Error -> {
-                // Use previous gateway if any, or the provided one
-                unifiedPushStore.getPushGateway(instance) ?: gatewayResult.gateway
+                // Use previous gateway if any, or the default one
+                unifiedPushStore.getPushGateway(instance)
+                    ?: defaultPushGatewayHttpUrlProvider.provide()
             }
             UnifiedPushGatewayResolverResult.ErrorInvalidUrl,
             UnifiedPushGatewayResolverResult.NoMatrixGateway -> {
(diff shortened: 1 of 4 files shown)


libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushGatewayUrlResolver.kt
bmarty, line 33: Maybe instead of patching the code here, `gatewayResult.gateway` could contains the value of `defaultPushGatewayHttpUrlProvider.provide()`?

I am confused since we decided to provide the `customUrl` to the [Error](https://github.com/element-hq/element-x-android/blob/main/libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushGatewayResolver.kt#L71) class in the past. Have we changed our mind?

bmarty (commented):
Thanks, just a question, see the other comment

libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/UnifiedPushGatewayUrlResolver.kt
p1gp1g: It used to return the default gateway: https://github.com/element-hq/element-x-android/commit/20370656cd6f523b1ab4ca5f1f0d513c40859195#diff-54db23386c84f03ffe631a6af4461af92d353f352f1b23025f14875d1175a0b4R49

jmartinesp: I think this changed in https://github.com/element-hq/element-x-android/pull/3388, there is even [a comment](https://github.com/element-hq/element-x-android/pull/3388#issuecomment-2451406904) from @p1gp1g in it 😅 .

IMO, using the default/fallback gateway when the custom one isn't present or working makes sense. Although maybe we should display some toast/notification in that case so the user is aware of that?

p1gp1g: I don't know if the toast is a good idea, some users are already confused that the gateway isn't their push server, and it may add more confusion

There are some action regarding the webpush MSC, I hope this gateway will be an old story soon :)

bmarty:
Let's merge this PR, sorry for the delay

PR #5749, opened 2025-11-17
Add media retention policy

jmartinesp:
## Content

- Add a custom media retention policy, as the one that exists on iOS. The SDK already has a default one, but these new values might be better than the default ones.
- Add `ByteSize` class to help with conversions between byte units.

## Motivation and context

Parity with iOS, also try to keep fewer values in the media store, which may speed up the app a tiny bit.

## Tests

I don't think this can be tested.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/libraries/core/src/main/kotlin/io/element/android/libraries/core/data/ByteSize.kt
+++ b/libraries/core/src/main/kotlin/io/element/android/libraries/core/data/ByteSize.kt
@@ -7,25 +7,21 @@
 
 package io.element.android.libraries.core.data
 
-private const val kilo = 1024L
-private const val mega = kilo * kilo
-private const val giga = mega * kilo
-
-enum class ByteUnit(val multiplier: Long) {
-    BYTES(1L),
-    KB(kilo),
-    MB(mega),
-    GB(giga)
+enum class ByteUnit(val bitShift: Int) {
+    BYTES(0),
+    KB(10),
+    MB(20),
+    GB(30)
 }
 
 class ByteSize internal constructor(val value: Long, val unit: ByteUnit) {
     fun to(dest: ByteUnit): Long {
         if (unit == dest) return value
-        return (value * unit.multiplier) / dest.multiplier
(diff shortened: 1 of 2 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -70,6 +75,19 @@ class RustMatrixClientFactory(
             .username(sessionData.userId)
             .use { it.build() }
 
+        client.setMediaRetentionPolicy(
+            MediaRetentionPolicy(
+                // Make this 500MB instead of 400MB
+                maxCacheSize = 500.megaBytes.to(ByteUnit.BYTES).toULong(),
```
bmarty, line 81: Maybe for a future improvement, make the value depend on the system available resource?

libraries/core/src/main/kotlin/io/element/android/libraries/core/data/ByteSize.kt
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.core.data
+
+private const val kilo = 1024L
+private const val mega = kilo * kilo
+private const val giga = mega * kilo
```
bmarty: Maybe just inline those val.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
jmartinesp: Sounds like a good idea 👍

Issue #5739, opened 2025-11-15
Reporting a problem from the app doesn't work


Rustacian:
### Steps to reproduce

1. Open the menu "Report a problem" via Settings
2. Type some text, check "Contact Me" (let "Allow logs enabled", "Send notification settings" disabled)
3. Press "Send"
4. See Error:"
> ### Error
> Could not save report

#### What did you expect?

report is sent including configured details

#### What happened instead?

above error

### Your phone model

Fairphone 3+

### Operating system version

Android 13 (6.A.039.4)

### Application version and app store

Google Play, 25.11.2 (202511020) nY1MLWpBJj

### Homeserver

matrix.org

### Will you send logs?

No (not possible)

### Are you willing to provide a PR?

No

jmartinesp:
I just reproduced it with this error:

```
elementx: ## compressFile() failed
    java.io.FileNotFoundException: /data/user/0/io.element.android.x.debug/cache/logs/element.io: open failed: EISDIR (Is a directory)
```

No idea why it's trying to upload the directory too though, we need to investigate it. May be related to multi-account, can you confirm if you've enabled the multi-account feature flag, @Rustacian ?

That said, the bug report *was uploaded* after the error dialog was displayed.

Rustacian:
No, no multi-account feature enabled. I would not even know how to enable it. Labs only offers thread replies.

jmartinesp:
Then it's probably a different issue than mine. 

I'm sorry, but since it may be quite difficult for us to debug the issue without logs (I tested this right now on my personal device and it worked fine), can I ask you to go into the bug report screen, try to send a bug report, then once it fails go into 'view logs' and send both the first visible file (should be named `logcat.log`) and the last one (will be named something like `logs.<year>-<month>-<day>-<hour>.log`, matching your current time in UTC) to either the email `android@element.io` or to me at my user `@jorgem:element.io` so they're only shared with the team? You can tap on those entries and then send them from there with the share icon, or download them (they'll be saved in your Downloads folder) and then share them from there.

jmartinesp:
Actually, the error you're seeing (`Could not save report`) comes from [the bug report server](https://github.com/matrix-org/rageshake/blob/21d52a003ba25f883263c0cf823b983cae6718f1/submit.go#L270), that somehow failed to save the data. If you try to report a bug again is it working fine now? It's working for me, at least.

Rustacian:
1) still not working (same error)
[...]

PR #5750, opened 2025-11-17
Enable logging OkHttp traffic based on the current log level

jmartinesp:
## Content

What the title says. Previously it was enabled only in debuggable builds.

## Motivation and context

It would have helped when debugging https://github.com/element-hq/element-x-android/issues/5739, since there is no way to turn on Retrofit HTTP logging when the app is in release mode.

## Tests

In a nightly/release build, set the log level to either debug or trace. The next logs should include HTTP logging coming from OkHttp. You can check this by running the notification troubleshoot, i.e.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/features/rageshake/api/src/main/kotlin/io/element/android/features/rageshake/api/reporter/BugReporter.kt
+++ b/features/rageshake/api/src/main/kotlin/io/element/android/features/rageshake/api/reporter/BugReporter.kt
@@ -45,5 +45,5 @@ interface BugReporter {
     /**
      * Save the logcat.
      */
-    fun saveLogCat()
+    fun saveLogCat(): File?
 }
--- a/features/rageshake/impl/src/main/kotlin/io/element/android/features/rageshake/impl/reporter/DefaultBugReporter.kt
+++ b/features/rageshake/impl/src/main/kotlin/io/element/android/features/rageshake/impl/reporter/DefaultBugReporter.kt
@@ -95,8 +95,6 @@ class DefaultBugReporter(
     private val logcatCommandDebug = arrayOf("logcat", "-d", "-v", "threadtime", "*:*")
     private var currentTracingLogLevel: String? = null
 
-    private val logCatErrFile: File
-        get() = File(logDirectory(), LOG_CAT_FILENAME)
     private val baseLogDirectory = File(context.cacheDir, LOG_DIRECTORY_NAME)
     private var currentLogDirectory: File = baseLogDirectory
 
@@ -160,10 +158,14 @@ class DefaultBugReporter(
                 }
                 if (withCrashLogs || withDevicesLogs) {
                     saveLogCat()
-                    val gzippedLogcat = compressFile(logCatErrFile)
-                    if (gzippedLogcat != null) {
-                        gzippedFiles.add(0, gzippedLogcat)
-                    }
+                        ?.let { logCatFile ->
+                            compressFile(logCatFile).also {
+                                logCatFile.safeDelete()
+                            }
+                        }
+                        ?.let { gzippedLogcat ->
+                            gzippedFiles.add(0, gzippedLogcat)
+                        }
                 }
                 val sessionData = sessionStore.getLatestSession()
                 val numberOfAccounts = sessionStore.numberOfSessions()
@@ -387,7 +389,8 @@ class DefaultBugReporter(
             onException = { Timber.e(it, "## getLogFiles() failed") }
         ) {
             val logDirectory = logDirectory()
-            logDirectory.listFiles()?.toList()
+            logDirectory.listFiles()
+                ?.filter { it.isFile && !it.name.endsWith(LOG_CAT_FILENAME) }
         }.orEmpty()
     }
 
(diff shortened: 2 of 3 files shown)


libraries/network/src/main/kotlin/io/element/android/libraries/network/NetworkModule.kt
```diff
@@ -26,14 +28,15 @@ object NetworkModule {
...
-        buildMeta: BuildMeta,
         userAgentInterceptor: UserAgentInterceptor,
+        featureFlagService: FeatureFlagService,
     ): OkHttpClient = OkHttpClient.Builder().apply {
         connectTimeout(30, TimeUnit.SECONDS)
         readTimeout(60, TimeUnit.SECONDS)
         writeTimeout(60, TimeUnit.SECONDS)
         addInterceptor(userAgentInterceptor)
-        if (buildMeta.isDebuggable) addInterceptor(providesHttpLoggingInterceptor())
+        val addLogInterceptor = runBlocking { featureFlagService.isFeatureEnabled(FeatureFlags.PrintLogsToLogcat) }
```
bmarty: Is it possible to avoid this runBlocking? I had a quick look, but not sure how to do it simply.

jmartinesp: I actually changed this to use the `LogLevel` and re-did the PR, sorry. I believe that makes more sense. Also, the new way it's done makes sure the code that calls this `runBlocking` runs in a different thread, so it shouldn't matter if it's blocked for a few ms.

libraries/network/src/main/kotlin/io/element/android/libraries/network/NetworkModule.kt
```diff
@@ -26,21 +27,34 @@ object NetworkModule {
...
+        val loggingLevel = HttpLoggingInterceptor.Level.BODY
+        val logger = FormattedJsonHttpLogger(loggingLevel)
+        val interceptor = HttpLoggingInterceptor(logger)
+        interceptor.level = loggingLevel
+        return interceptor
+    }
+
+    @Provides
+    @SingleIn(AppScope::class)
+    fun provideDynamicHttpLoggingInterceptor(
```
bmarty: `@Provides` should be used only for classes that we do not own. In the case of `DynamicHttpLoggingInterceptor`, the constructor can just be annoted with `@Inject` (and `@SingleIn(AppScope::class)`)

jmartinesp:
> Thanks for the update. It is working fine. But I wonder if the solution is a bit over engineered. We could simply observe the flow `appPreferencesStore.getTracingLogLevelFlow()` and set the value of `HttpLoggingInterceptor.level` accordingly (`Level.NONE` or `Level.BODY`). WDYT?

I tried doing something similar before uploading these changes, but that is created asynchronously and observed asynchronously. If what you suggest and what I did match, using something like a `StateFlow` to observe the value, the first time that's checked is almost guaranteed to have the default value and not print anything, so we'd consistently miss the logs for the first request.

libraries/network/src/main/kotlin/io/element/android/libraries/network/interceptors/DynamicHttpLoggingInterceptor.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    override fun intercept(chain: Interceptor.Chain): Response {
+        // This is called in a separate thread, so calling `runBlocking` here should be fine, it should be also instant after the value is cached
+        val intercept = runBlocking { shouldDisplayLogs(appPreferencesStore.getTracingLogLevelFlow().first()) }
+        return if (intercept) {
+            loggingInterceptor.intercept(chain)
+        } else {
+            chain.proceed(chain.request())
+        }
+    }
```
bmarty, line 36: Regarding my comment (and your response), maybe keep this solution, but simplify it like this:
```kotlin
val logLevel = runBlocking { appPreferencesStore.getTracingLogLevelFlow().first() }
loggingInterceptor.level = if (logLevel >= LogLevel.DEBUG) Level.BODY else Level.NONE
return loggingInterceptor.intercept(chain)
```

Note that I am trying this, and I do not see the logs anymore. Maybe it's not working?

libraries/network/src/main/kotlin/io/element/android/libraries/network/NetworkModule.kt
```diff
@@ -26,21 +27,34 @@ object NetworkModule {
...
-    val logger = FormattedJsonHttpLogger(loggingLevel)
-    val interceptor = HttpLoggingInterceptor(logger)
-    interceptor.level = loggingLevel
-    return interceptor
+    @Provides
+    fun providesHttpLoggingInterceptor(): HttpLoggingInterceptor {
+        val loggingLevel = HttpLoggingInterceptor.Level.BODY
+        val logger = FormattedJsonHttpLogger(loggingLevel)
+        val interceptor = HttpLoggingInterceptor(logger)
+        interceptor.level = loggingLevel
```
bmarty: This line could be removed with my proposal I think.
```kotlin
    @Provides
    fun providesHttpLoggingInterceptor(): HttpLoggingInterceptor {
        val logger = FormattedJsonHttpLogger(HttpLoggingInterceptor.Level.BODY)
        return HttpLoggingInterceptor(logger)
    }
```

bmarty:
@jmartinesp I think we can merge this.
I used to repro the issue with the error 500 but not anymore.
We would need to have more info by looking at the server log (from [here](https://github.com/matrix-org/rageshake/blob/21d52a003ba25f883263c0cf823b983cae6718f1/submit.go#L269))

libraries/network/src/main/kotlin/io/element/android/libraries/network/interceptors/DynamicHttpLoggingInterceptor.kt
bmarty: The log was probably filtered out because of the usage of `Timber.v`

PR #5753, opened 2025-11-18
Stop overriding the homeserver when restoring a `Client`

jmartinesp:
## Content

Removes the `setHomeserverUrl` call when restoring a `Client` from its `SessionData`.

## Motivation and context

This call isn't necessary and overrides the existing data previously saved by the SDK, resulting in losing data such as the `Client::server` (the discovery server URL).

In turn, this caused the app to be unable to refresh the server info in some homeservers. Sadly, this means users that are already logged in will be affected by this issue since their configured URLs would have been overridden after the first `Client` restoration, so they won't be able to refresh their server info until they log out and in again.

## Tests

- Freshly log into `element.io` or some other HS whose server URL doesn't match their HS URL (the host for the `/.well-known/matrix/client` doesn't match the actual HS).
- Then clear the cache from the developer settings. You should not see this messages in the logs anymore:

```
matrix_sdk::client: Failed to fetch client well-known: the server returned an error: [404] <non-json bytes> | crates/matrix-sdk/src/client/mod.rs:1994
```

Also, if right after restoring the `Client` you add a log containing the `Client.server()` value, it should display the server URL instead of null.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
@@ -66,7 +66,6 @@ class RustMatrixClientFactory(
             passphrase = sessionData.passphrase,
             slidingSyncType = ClientBuilderSlidingSync.Restored,
         )
-            .homeserverUrl(sessionData.homeserverUrl)
             .username(sessionData.userId)
             .use { it.build() }
 


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -66,7 +66,6 @@ class RustMatrixClientFactory(
             passphrase = sessionData.passphrase,
             slidingSyncType = ClientBuilderSlidingSync.Restored,
         )
-            .homeserverUrl(sessionData.homeserverUrl)
```
bmarty, line 69: Thanks for the fix!
So it seems that the application is storing the actual homeserver Url (CS baseUrl) and this is not really expected. I was looking at the other usage of `homeserverUrl` (to see if it could be just removed) and I see that the only other usage of `homeserverUrl` is [here](https://github.com/element-hq/element-x-android/blob/develop/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt#L179), and `toSession` is invoked at line 73 below.

So it is actually OK that the CS baseUrl is stored?

bmarty: The other usage is actually not used by the SDK. Maybe some cleanup can be done SDK side later.

jmartinesp: > The other usage is actually not used by the SDK. Maybe some cleanup can be done SDK side later.

This was implemented in https://github.com/matrix-org/matrix-rust-sdk/pull/5874.

PR #5758, opened 2025-11-18
Fix null pointer exception on room notification settings. 

bmarty:
## Content

Ensure the comparator does not compare null values. This is not supported by `Collator`.
Cleanup the codebase by removing duplication of class, and limit instance creations of `PowerLevelRoomMemberComparator`.
Add unit tests to cover the room sorting, also in the case of null name.

## Motivation and context

Closes https://github.com/element-hq/element-x-android-rageshakes/issues/8517

Thanks @progserega for reporting!

## Screenshots / GIFs

No expected change

## Tests

- With rooms with no name, navigate to settings, notifications, notify me for Group chat.
- If the app does not crash, the bug is fixed.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/edit/EditDefaultNotificationSettingPresenter.kt
+++ b/features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/edit/EditDefaultNotificationSettingPresenter.kt
@@ -23,6 +23,7 @@ import io.element.android.libraries.architecture.AsyncAction
 import io.element.android.libraries.architecture.Presenter
 import io.element.android.libraries.architecture.runUpdatingStateNoSuccess
 import io.element.android.libraries.designsystem.components.avatar.AvatarSize
+import io.element.android.libraries.matrix.api.core.RoomId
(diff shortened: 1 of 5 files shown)


features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/edit/EditDefaultNotificationSettingPresenter.kt
```diff
@@ -49,6 +49,10 @@ class EditDefaultNotificationSettingPresenter(
         fun create(oneToOne: Boolean): EditDefaultNotificationSettingPresenter
     }
 
+    private val collator = Collator.getInstance().apply {
```
ganfra, line 53: If the Locale changes, this won't be updated

bmarty: it will be updated the next time this screen is open, so I think that this is acceptable, do you agree?

PR #5760, opened 2025-11-19
Add some performance metrics for Sentry

jmartinesp:
## Content

- Add `AnalyticsService.startTransaction(...)` to start a logging transaction that can be uploaded to Sentry if the user enabled the analytics upload.
- Add `AnalyticsTransaction` wrapper to abstract the Sentry ones.
- Added several helper methods to improve the UX around these transactions.
- Then measure:
  - Time until the first sync, and how it ended.
  - Time until the first rooms are displayed.
  - Time to load a room or a preview.
  - Time to load a timeline.

## Motivation and context

Add some performance measurements to the app so we can track when something becomes slow.

## Tests

There's not much to test, other than running this code and checking your transactions have been uploaded to Sentry.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
@@ -19,7 +19,7 @@ import io.element.android.libraries.core.coroutine.childScope
 import io.element.android.libraries.matrix.api.sync.SyncService
 import io.element.android.libraries.matrix.api.sync.SyncState
 import io.element.android.services.analytics.api.AnalyticsService
-import io.element.android.services.analytics.api.recordAsyncTransaction
+import io.element.android.services.analytics.api.recordTransaction
 import io.element.android.services.appnavstate.api.AppForegroundStateService
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.FlowPreview
@@ -72,12 +72,12 @@ class SyncOrchestrator(
             // Perform an initial sync if the sync service is not running, to check whether the homeserver is accessible
             // Otherwise, if the device is offline the sync service will never start and the SyncState will be Idle, not Offline
             Timber.tag(tag).d("performing initial sync attempt")
-            analyticsService.recordAsyncTransaction("First sync", "syncService.startSync()") {
+            analyticsService.recordTransaction("First sync", "syncService.startSync()") { transaction ->
                 syncService.startSync()
 
                 // Wait until the sync service is not idle, either it will be running or in error/offline state
                 val firstState = syncService.syncState.first { it != SyncState.Idle }
-                setData("first_sync_state", firstState.name)
+                transaction.setData("first_sync_state", firstState.name)
             }
 
             observeStates()
--- a/services/analytics/api/src/main/kotlin/io/element/android/services/analytics/api/AnalyticsLongRunningTransaction.kt
+++ b/services/analytics/api/src/main/kotlin/io/element/android/services/analytics/api/AnalyticsLongRunningTransaction.kt
@@ -7,20 +7,9 @@
 
 package io.element.android.services.analytics.api
 
-import java.util.Objects
-
 sealed class AnalyticsLongRunningTransaction(
     val name: String,
     val operation: String?,
 ) {
-    override fun equals(other: Any?): Boolean {
-        if (other !is AnalyticsLongRunningTransaction) return false
-        return name == other.name && operation == other.operation
-    }
-
-    override fun hashCode(): Int {
-        return Objects.hash(name, operation)
-    }
-
     data object FirstRoomsDisplayed : AnalyticsLongRunningTransaction("First rooms displayed after login or restoration", null)
 }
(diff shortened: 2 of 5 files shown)


services/analytics/api/src/main/kotlin/io/element/android/services/analytics/api/AnalyticsService.kt
```diff
@@ -48,4 +49,33 @@ interface AnalyticsService : AnalyticsTracker, ErrorTracker {
...
+    val transaction = startTransaction(name, operation)
+    try {
+        val result = transaction.block()
+        return result
+    } finally {
+        transaction.finish()
+    }
+}
+
+suspend fun <T> AnalyticsService.recordAsyncTransaction(name: String, operation: String, block: suspend AnalyticsTransaction.() -> T): T {
```
bmarty: Not a big deal, but the method itself does not need to be `suspend`.

Also the code (at call site) will be a bit easier to understand to me if `AnalyticsTransaction` is a parameter and not a receiver of the block:


```suggestion
suspend fun <T> AnalyticsService.recordAsyncTransaction(name: String, operation: String, block: suspend (AnalyticsTransaction) -> T): T {
```

I let you decide if you want to change or not.

services/analytics/api/src/main/kotlin/io/element/android/services/analytics/api/AnalyticsLongRunningTransaction.kt
```diff
@@ -0,0 +1,26 @@
...
+    val operation: String?,
+) {
+    override fun equals(other: Any?): Boolean {
+        if (other !is AnalyticsLongRunningTransaction) return false
+        return name == other.name && operation == other.operation
+    }
+
+    override fun hashCode(): Int {
+        return Objects.hash(name, operation)
+    }
```
bmarty: Why do you implement `equals` and `hashCode` here?

appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
```diff
@@ -69,10 +72,13 @@ class SyncOrchestrator(
             // Perform an initial sync if the sync service is not running, to check whether the homeserver is accessible
             // Otherwise, if the device is offline the sync service will never start and the SyncState will be Idle, not Offline
             Timber.tag(tag).d("performing initial sync attempt")
-            syncService.startSync()
+            analyticsService.recordAsyncTransaction("First sync", "syncService.startSync()") {
```
bmarty: Maybe create some enum instead of hard coded string?

services/analytics/api/src/main/kotlin/io/element/android/services/analytics/api/AnalyticsLongRunningTransaction.kt
jmartinesp: `AnalyticsLongRunningTransaction` will be used as a `Map` key in `DefaultAnalyticsService`. Leaving those with the default implementation would create a different hash code per instance, right?

appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
jmartinesp: I actually wanted to keep these as strings so we could quickly add new transactions to track. I added the enums for the `AnalyticsLongRunningTransaction` so we can use them to fetch existing transactions, but it's not needed for other use cases and not having them allows more flexibility.

services/analytics/api/src/main/kotlin/io/element/android/services/analytics/api/AnalyticsLongRunningTransaction.kt
bmarty: I do not think so, since `FirstRoomsDisplayed` is an object, there will be only one instance. But I have not checked what happened at runtime. If you add this code to fix an issue it's fine, but if it was "just in case" I guess you should just remove it and ensure that the code still works as expected.

appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
bmarty: I agree that it makes more sense for `AnalyticsLongRunningTransaction` since the enum value can be used at any part of the codebase, and for the `recordAsyncTransaction` parameters, they are used only once. OK then!

services/analytics/api/src/main/kotlin/io/element/android/services/analytics/api/AnalyticsLongRunningTransaction.kt
jmartinesp: Ah, you're right. When I first implemented this, `FirstRoomsDisplayed` wasn't an object so this code was needed, but the way it works now we don't need it 👍 .

austom180-svg:
gh pr checkout 5760

austom180-svg:
gh pr checkout 5760

PR #5767, opened 2025-11-19
Add more performance checks

jmartinesp:
## Content

- Ensure long running transaction methods return an actual transaction we can modify. Also add `AnalyticsService.getLongRunningTransaction` and change `AnalyticsService.stopLongRunningTransaction` to `removeLongRunningTransaction`.
- Then add `AnalyticsRoomListStateWatcher` to measure how long it takes the app to receive the up to date room list after coming back from a background state.
- Add also `AnalyticsColdStartWatcher` to measure how long it takes the app to display the cached room list from a cold start.
- Add `NotificationTapOpensTimeline` to check how long it takes for a notification tap to open a room.
- Add `OpenRoom` to check how long it takes to open a room and display the initial timeline events.
- Add transaction grouping so we can have trace trees like:

image

## Motivation and context

Add some more metrics.

## Tests

Testing any of the cases added above should add those events to Sentry.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
@@ -206,6 +206,9 @@ class TimelinePresenter(
                     focusOnEvent(event.eventId, focusRequestState)
                 }.start()
                 is TimelineEvents.OnFocusEventRender -> {
+                    // If there was a pending 'notification tap opens timeline' transaction, finish it now we're focused in the required event
+                    analyticsService.finishLongRunningTransaction(NotificationTapOpensTimeline)
+
                     focusRequestState.value = focusRequestState.value.onFocusEventRender()
                 }
                 is TimelineEvents.ClearFocusRequestState -> {
@@ -242,7 +245,6 @@ class TimelinePresenter(
                     analyticsService.run {
                         finishLongRunningTransaction(DisplayFirstTimelineItems)
                         finishLongRunningTransaction(OpenRoom)
-                        finishLongRunningTransaction(NotificationTapOpensTimeline)
                     }
                 }
                 .launchIn(this)


app/src/main/kotlin/io/element/android/x/intent/DefaultIntentProvider.kt
```diff
@@ -36,6 +36,7 @@ class DefaultIntentProvider(
         return Intent(context, MainActivity::class.java).apply {
             action = Intent.ACTION_VIEW
             data = deepLinkCreator.create(sessionId, roomId, threadId, eventId).toUri()
+            putExtra("from_notification", true)
```
bmarty: Maybe create a const for the String?
Also this code is also invoked by `DefaultNotificationConversationService` so the entry point will not be a notification. But I guess we do not want to distinguish this case for now.

appnav/src/main/kotlin/io/element/android/appnav/NotLoggedInFlowNode.kt
```diff
@@ -64,6 +66,7 @@ class NotLoggedInFlowNode(
 
     override fun onBuilt() {
         super.onBuilt()
+        analyticsColdStartWatcher.whenLoggingIn()
```
bmarty, line 70: Note: This will be also invoked when adding an account.

features/home/impl/build.gradle.kts
```diff
@@ -28,6 +28,7 @@ setupDependencyInjection()
 
 dependencies {
     implementation(projects.appconfig)
+    implementation(projects.appnav)
```
bmarty: I'd move `AnalyticsColdStartWatcher` to the analytics api module to avoid adding this dependency.

Also `FakeAnalyticsColdStartWatcher` can be moved to `:analytics:test`

services/analytics/impl/src/main/kotlin/io/element/android/services/analytics/impl/watchers/DefaultAnalyticsColdStartWatcher.kt
```diff
@@ -0,0 +1,70 @@
...
+                }
+            }
+            .catch { Timber.w(it.message) }
+            .launchIn(appCoroutineScope)
+    }
+
+    override fun whenLoggingIn() {
+        if (isColdStart.getAndSet(false)) {
+            analyticsService.removeLongRunningTransaction(ColdStartUntilCachedRoomList)
+            Timber.d("Canceled cold start check: user is logging in")
```
bmarty, line 53: Neat: it seems that this is the only case where the returned value of `removeLongRunningTransaction` is not `finished`. Maybe add a parameter to `removeLongRunningTransaction`like `andFinish: Boolean = true`? This parameter will be `false` here.

services/analytics/impl/src/main/kotlin/io/element/android/services/analytics/impl/watchers/DefaultAnalyticsColdStartWatcher.kt
```diff
@@ -0,0 +1,70 @@
...
+                if (hasConsent) {
+                    if (isColdStart.get()) {
+                        Timber.d("Starting cold start check")
+                        analyticsService.startLongRunningTransaction(ColdStartUntilCachedRoomList)
+                    } else {
+                        error("The app is no longer in a cold start state")
+                    }
+                }
+            }
+            .catch { Timber.w(it.message) }
```
bmarty, line 46: There is a warning here, maybe change to:

```suggestion
            .catch { Timber.w(it, "Error") }
```

Maybe the `error` above was only to debug?

appnav/src/main/kotlin/io/element/android/appnav/NotLoggedInFlowNode.kt
jmartinesp: That's fine, this just wants to cancel the cold start time check. By the time you try to add an account, the check should already have finished or been canceled.

services/analytics/impl/src/main/kotlin/io/element/android/services/analytics/impl/watchers/DefaultAnalyticsColdStartWatcher.kt
jmartinesp: I'd rather just return the transaction so the developers can do what they need to them, including extra parameters if needed. If we add this, we'd be returning a transaction that's already finished, so it makes no sense to return it at all since you won't be able to alter what's sent to Sentry.

services/analytics/impl/src/main/kotlin/io/element/android/services/analytics/impl/watchers/DefaultAnalyticsColdStartWatcher.kt
jmartinesp: That's actually what I wanted, since I don't expect any actual error. I just emit the error above to cancel the flow, then use the log to just notify we've done that. I can't remember if there's a better way to cancel a hot flow from inside the flow, to be honest.

services/analytics/impl/src/main/kotlin/io/element/android/services/analytics/impl/watchers/DefaultAnalyticsColdStartWatcher.kt
bmarty: You make a point!
But maybe for clarity the API could be (or we could have extension on `AnalyticsService` like):
```kotlin
// Will just remove the transaction
analyticsService.cancelLongRunningTransaction(...)
// Will remove the transaction and finish it
analyticsService.finishLongRunningTransaction(...)
```

services/analytics/impl/src/main/kotlin/io/element/android/services/analytics/impl/watchers/DefaultAnalyticsColdStartWatcher.kt
bmarty: OK, the error will happen when the user gives their consent later from the settings, no?
I am running the code to check this.

**EDIT**: so yes, it's the case. An error will occur (and will be caught) when the user opt-in for analytics in the settings.

jmartinesp:
I was busy with some Rust these past couple of days, but I'll continue working on this tomorrow. @ganfra having your review too before I make any changes would be nice, so I don't have to change the code twice.

appnav/src/main/kotlin/io/element/android/appnav/analytics/AnalyticsRoomListStateWatcher.kt
```diff
@@ -0,0 +1,79 @@
...
+    private var currentCoroutineScope: CoroutineScope? = null
+    private val isWarmState = AtomicBoolean(false)
+
+    fun start() {
+        if (currentCoroutineScope != null) {
+            Timber.w("Can't start RoomListStateWatcher, it's already running.")
+            return
+        }
+
+        val coroutineScope = CoroutineScope(sessionCoroutineScope.coroutineContext + dispatchers.computation)
```
ganfra: Use `sessionCoroutineScope.childScope`?
Also I'd prefer we create the child scope in the constructor and use an atomic boolean to know if it's already started.

jmartinesp: So we'd store the jobs for the launched flows in another variable(s) so we can cancel them on `stop()`? I used the nullable `CoroutineScope` because cancelling it will also cancel all their children.

features/home/impl/build.gradle.kts
jmartinesp: Done in https://github.com/element-hq/element-x-android/pull/5767/commits/bfbe01e93fa73d9c9346ab15dc30d39cbb84f248.

app/src/main/kotlin/io/element/android/x/intent/DefaultIntentProvider.kt
jmartinesp: Handled in https://github.com/element-hq/element-x-android/pull/5767/commits/5bd4c4c019b9e746f16b95ab7182c8680af9b3c9. This should also ensure the value is only included for notifications.

services/analytics/impl/src/main/kotlin/io/element/android/services/analytics/impl/watchers/DefaultAnalyticsColdStartWatcher.kt
jmartinesp: Added these helper methods in https://github.com/element-hq/element-x-android/pull/5767/commits/1a79bb39bf5e869b771959dfc16ef37400f9656a.

appnav/src/main/kotlin/io/element/android/appnav/analytics/AnalyticsRoomListStateWatcher.kt
jmartinesp: Used `childScope` in https://github.com/element-hq/element-x-android/pull/5767/commits/bfbe01e93fa73d9c9346ab15dc30d39cbb84f248.

jmartinesp:
Some build/quality checks are failing but those *are* working locally. Maybe it's a cache issue?

---

Rebasing in case it helps with the cache issue 🤞 .

PR #5784, opened 2025-11-20
Use regex to check forbidden terms

jmartinesp:
## Content

What the title says.

## Motivation and context

We were returning and `Elementul` as a match for the forbidden term `Element`. It now checks for the full word.

## Tests

With the latest strings, run the script. It shouldn't return any matches. If you change it back to not using `\b` for the word boundary, the `Elementul` match will be found.

## Checklist


bmarty (approved):
Tested OK, thanks!

Issue #5781, opened 2025-11-20
The app does not accept shares from certain apps.


YamatoRyou:
### Steps to reproduce

1. Install MacroDroid and create a macro within it;
2. Save and share the macro after creation;
3. Select EXA as the first target for sending the macro, then select a room as the second target, and click "Send".

#### What did you expect?
n/a

#### What happened instead?
EXA error message: Failed to handle incoming share intent.
However, other apps can process it normally.
<img width="405" height="577" alt="Image" src="https://github.com/user-attachments/assets/0569f9b2-8906-444c-9cf1-8e1ce28fbdbe" />
### If you click the "Back" button (green arrow point to) at this point, EXA will crash.

----
Logs:
[{E4539EE7-BF93-AF7E-DBD7-5B560292D620}.zip](https://github.com/user-attachments/files/23655357/E4539EE7-BF93-AF7E-DBD7-5B560292D620.zip)
Password: `{24EB9E19-DE22-10C5-1AEF-69C05BD8E4E7}`
**Please download this file as soon as possible.**

### Your phone model

OnePlus MT2111

### Operating system version

Oxygen OS 15 (Android 15)

### Application version and app store

Element X Android 25.11.2

### Homeserver

Synapse 1.142.0

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

jmartinesp:
I believe this was just fixed by https://github.com/element-hq/element-x-android/pull/5773.

YamatoRyou:
> I believe this was just fixed by [#5773](https://github.com/element-hq/element-x-android/pull/5773).

In the recent nightly build (ece47a42), the crash issue has indeed been fixed, but sharing from MacroDroid is still not accepted.

jmartinesp:
@YamatoRyou please try the APK in https://github.com/element-hq/element-x-android/pull/5785 once it's uploaded. It fixed the issue for me, but I'd like to double check the fix.

YamatoRyou:
> [@YamatoRyou](https://github.com/YamatoRyou) please try the APK in [#5785](https://github.com/element-hq/element-x-android/pull/5785) once it's uploaded. It fixed the issue for me, but I'd like to double check the fix.

Tested and fixed.
App: `01fc964c`

Issue #5623, opened 2025-10-28
VC 202510010 fails to consume an `intent:#Intent;action=android.intent.action.SEND;type=message/rfc822` containing a file *and* `text/plain`.


beedell-roke:
#### How To Reproduce It

Attempt to export: [^1]

[^1]: [`tasks/tasks/issues/3952#issue-3562709063`](https://github.com/tasks/tasks/issues/3952#issue-3562709063)

<blockquote>

~~~YAML
cmd activity start --user 0 -a android.intent.action.SEND -t message/rfc822 --es android.intent.extra.SUBJECT Tasks logs --es android.intent.extra.TEXT ----------
Tasks: 14.8.3 (generic build 140818)
Android: 15 (FP5.VT2E.C.059.20250908)
Locale: en_GB_#u-fw-mon-mu-celsius
Model: Fairphone FP5
Product: FP5 (FP5)
Kernel: 5.4.274-qgki-g67a8c76d3b27 (VT2E)
----------
notifications: true
reminders: true
background location: false
foreground location: true
calendar: true
----------
dont keep activities: null
---------- --eu android.intent.extra.STREAM content://org.tasks.file_provider/internal_cache/logs.zip --esa android.intent.extra.EMAIL support@tasks.org -f 50331649
~~~

</blockquote>

…from:

<blockquote>

~~~ini
sourcePackage=org.tasks
versionCode=140818 minSdk=26 targetSdk=35
versionName=14.8.3
~~~

</blockquote>

…to:

<blockquote>

~~~ini
sourcePackage=io.element.android.x
versionCode=202510010
~~~

</blockquote>

#### The Outcome

-	#### What I Expected To Occur

	Being listed as an available consumer:

	<blockquote>

	~~~YAML
[...]

PR #5785, opened 2025-11-20
Uris should take precedence in plain text intents

jmartinesp:
## Content

When handling a plain text share intent, check if it contains Uris before handling its content.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/5781.

## Tests

- Create a macro using MacroDroid.
- Share it in the app.

If the file is shared, it's fixed.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/features/share/impl/src/main/kotlin/io/element/android/features/share/impl/ShareIntentHandler.kt
+++ b/features/share/impl/src/main/kotlin/io/element/android/features/share/impl/ShareIntentHandler.kt
@@ -58,8 +58,9 @@ class DefaultShareIntentHandler(
         onPlainText: suspend (String) -> Boolean,
     ): Boolean {
         val type = intent.resolveType(context) ?: return false
+        val uris = getIncomingUris(intent, type)
         return when {
-            type == MimeTypes.PlainText -> handlePlainText(intent, onPlainText)
+            uris.isEmpty() && type == MimeTypes.PlainText -> handlePlainText(intent, onPlainText)
             type.isMimeTypeImage() ||
                 type.isMimeTypeVideo() ||
                 type.isMimeTypeAudio() ||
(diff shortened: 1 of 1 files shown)


features/share/impl/src/main/kotlin/io/element/android/features/share/impl/ShareIntentHandler.kt
```diff
@@ -58,16 +58,16 @@ class DefaultShareIntentHandler(
         onPlainText: suspend (String) -> Boolean,
     ): Boolean {
         val type = intent.resolveType(context) ?: return false
+        val uris = getIncomingUris(intent, type)
         return when {
-            type == MimeTypes.PlainText -> handlePlainText(intent, onPlainText)
+            uris.isEmpty() && type == MimeTypes.PlainText -> handlePlainText(intent, onPlainText)
```
bmarty, line 63: Thanks. Now I am wondering why we are doing so much filtering on the types. This is maybe not necessary?
the code could be simplified by
```kotlin
        return if (uris.isEmpty() && type == MimeTypes.PlainText) {
            handlePlainText(intent, onPlainText)
        } else {
            val result = onUris(uris)
            revokeUriPermissions(uris.map { it.uri })
            result
        }
```
Not a blocker and not related to the issue that this PR is fixing, I am just thinking out loud.

jmartinesp: I'm not sure either, I just kept it like that because I couldn't remember the reasoning 😅 .

bmarty: Let's merge this fix and iterate later if issue are reported.

beedell-roke:
Does this at all relate to [`issues/5623`](https://github.com/element-hq/element-x-android/issues/5623#issue-3562986558), or might that be due to the file, instead?

PR #5786, opened 2025-11-20
Ensure confirmation dialog is displayed when an admin add other admin to a room

bmarty:
## Content

Ensure confirmation dialog is displayed when an admin add other admin to a room.
An issue in the code made admin of room owners with higher power level and in this case confirmation dialog does not have to be displayed.

Take the opportunity to cleanup the code, and add the room version rendering in the room details when developer mode is on.

## Motivation and context

Closes https://github.com/element-hq/customer-success/issues/647

## Tests

- In a room before version 12 where you are an admin, go to the screen to edit admin
- Add an admin
- Click save
- observe that the confirmation dialog is displayed as expected

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/rolesandpermissions/impl/src/test/kotlin/io/element/android/features/rolesandpermissions/impl/roles/ChangeRolesViewTest.kt
+++ b/features/rolesandpermissions/impl/src/test/kotlin/io/element/android/features/rolesandpermissions/impl/roles/ChangeRolesViewTest.kt
@@ -50,7 +50,6 @@ class ChangeRolesViewTest {
                 ),
             )
         }.exceptionOrNull()
-
         assertThat(exception).isNotNull()
     }
 
@@ -63,9 +62,7 @@ class ChangeRolesViewTest {
                 eventSink = eventsRecorder,
             ),
         )
-
         rule.pressBackKey()
-
         eventsRecorder.assertSingle(ChangeRolesEvent.ToggleSearchActive)
     }
 
@@ -78,9 +75,7 @@ class ChangeRolesViewTest {
                 eventSink = eventsRecorder,
             ),
         )
-
         rule.pressBackKey()
-
         eventsRecorder.assertList(listOf(ChangeRolesEvent.QueryChanged(""), ChangeRolesEvent.Exit))
     }
 
@@ -93,9 +88,7 @@ class ChangeRolesViewTest {
                 eventSink = eventsRecorder,
             ),
         )
-
         rule.pressBack()
-
         eventsRecorder.assertList(listOf(ChangeRolesEvent.QueryChanged(""), ChangeRolesEvent.Exit))
     }
 
@@ -108,9 +101,7 @@ class ChangeRolesViewTest {
                 eventSink = eventsRecorder,
             ),
         )
-
         rule.clickOn(CommonStrings.action_save)
-
         eventsRecorder.assertList(listOf(ChangeRolesEvent.QueryChanged(""), ChangeRolesEvent.Save))
     }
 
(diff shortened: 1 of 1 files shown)


features/rolesandpermissions/impl/src/main/kotlin/io/element/android/features/rolesandpermissions/impl/roles/ChangeRolesPresenter.kt
```diff
@@ -104,7 +104,7 @@ class ChangeRolesPresenter(
             }
         }
 
-        val hasPendingChanges = usersWithRole.value != selectedUsers.value
+        val hasPendingChanges = usersWithRole.value.toSet() != selectedUsers.value.toSet()
```
jmartinesp: If this will be executed in every recomposition maybe we should have it inside a `remember { derivedStateOf { ... } }` so we don't create two new sets every time.

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/RoomDetailsView.kt
```diff
@@ -737,6 +741,19 @@ private fun DebugInfoSection(roomId: RoomId) {
                 )
             },
         )
+        ListItem(
+            headlineContent = {
+                Text("Room version")
+            },
+            supportingContent = {
+                Text(
+                    text = roomVersion ?: "Unknown",
```
jmartinesp, line 750: Should we have some localazy strings for these?

libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/model/RoomInfoExtension.kt
```diff
@@ -23,12 +23,12 @@ fun RoomInfo.getAvatarData(size: AvatarSize) = AvatarData(
...
 /**
  * Returns the role of the user in the room.
- * If the user is a creator, returns [RoomMember.Role.Owner].
+ * If the user is a creator and [RoomInfo.privilegedCreatorRole] is true, returns [RoomMember.Role.Owner].
  * Otherwise, checks the power levels and returns the corresponding role.
  * If no specific power level is set for the user, defaults to [RoomMember.Role.User].
  */
 fun RoomInfo.roleOf(userId: UserId): RoomMember.Role {
-    return if (creators.contains(userId)) {
+    return if (privilegedCreatorRole && creators.contains(userId)) {
```
jmartinesp, line 31: Oh, nice catch!

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/RoomDetailsView.kt
bmarty: This should be handled as a global task to add all the strings from the developer settings to Localazy. For now I think we should not care.

jmartinesp: I figured that was the case, but since this is a developer UI even though it's not in the developer settings I wanted to double check first 👍 .

bmarty: Maybe one day lint will complain about hard-coded string like it used to do in Android Views, but for now let's be lazy!

jmartinesp (approved):

PR #5808, opened 2025-11-25
Enable Sentry in the SDK and allow bridging spans

jmartinesp:
## Content

Enable distributed tracing:
- Add a new `SentryConfig.SDK_DSN` value which will be passed to the SDK to setup its Sentry instance. This allows not only to log spans but also to upload panics automatically.
- Add `AnalyticsService.enterSdkSpan` function which will create an `AnalyticsSdkSpan`, which allows us to start a Sentry span in the SDK so we can have metrics of what's happening on the SDK side while the span is active.
- Add one of this bridge spans to the `Room.timelineWithConfiguration` client-side transaction.

## Motivation and context

Improve the details of the metrics.

## Screenshots / GIFs

image

## Tests

Open a timeline in the app, check it's hopefully added to Sentry. However, SDK spans won't be sent if not all processes spawned while the spans was active have finished, so sometimes the SDK span may be missing.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist


app/src/main/kotlin/io/element/android/x/di/AppBindings.kt
```diff
@@ -48,4 +49,7 @@ interface AppBindings {
     fun featureFlagService(): FeatureFlagService
 
     fun buildMeta(): BuildMeta
+
+    @SentrySdkDsn
+    fun sentrySdkDsn(): String?
```
ganfra: Maybe using a value class would be better?

ganfra (approved):
One small remark otherwise LGTM

app/src/main/kotlin/io/element/android/x/di/AppBindings.kt
jmartinesp: It shouldn't really matter, as you'd need the annotation to get this injected anywhere, but it's true it'd be easier to tell apart from our own app's DSN value.

jmartinesp: Done in https://github.com/element-hq/element-x-android/pull/5808/commits/beb419008cefc50b2457f3ecc459aa50f4b0a806

jmartinesp:
@ganfra I refactored the code a bit in https://github.com/element-hq/element-x-android/pull/5808/commits/64b2742a69f40e57003b9f70e0d4003344626ea3, let me know if you want to review it again.

ganfra:
@jmartinesp ok for me, you can merge!

PR #5809, opened 2025-11-25
misc : use newLatestEvent api from sdk

ganfra:
## Content
This PR makes sure to use the newLatestEvent api from the sdk : 
- expose the same api than the sdk in the matrix module, in case we want to do more than formatting
- rename a bunch of things to use _LatestEvent_  instead of LastMessage, to better match what is happening.

## Motivation and context
Use latest api so we have a better room list!

## Screenshots / GIFs
No ui change.

## Tests

- Play with the room list by sending messages in several rooms while being offline, using edition etc etc
- Check the room list is accurate

## Tested devices

- [X] Physical
- [ ] Emulator
- OS version(s):

## Checklist


features/space/impl/src/main/kotlin/io/element/android/features/space/impl/SpaceFlowNode.kt
```diff
@@ -120,7 +120,7 @@ class SpaceFlowNode(
                         backstack.pop()
                     }
 
-                    override fun navigateToSpaceInfo() {
+                    override fun navigateToEditDetails() {
```
jmartinesp: Unrelated to this PR? Let me know if you still want to merge it.

jmartinesp (approved):
Everything seems fine AFAICT, thanks!

features/space/impl/src/main/kotlin/io/element/android/features/space/impl/SpaceFlowNode.kt
ganfra: Oooooops, good catch!

PR #5817, opened 2025-11-26
Load `JoinedRoom` in home screen, pass it to the room flow

jmartinesp:
## Content

When clicking on an item from the room list, instead  of opening the room flows and loading the room data there:

- Asynchronously load the joined room, with a delayed loading dialog if this takes too long. If loading the joined room fails, just use the previous flow and pass the room id.
- Once the room is loaded, pass it through the `NavTarget` instances by adding an optional `JoinedRoom` parameter. Usually these should be only `Parcelable`  values, but we can use `@IgnoredOnParcel` to avoid serializing and deserializing these values.
- Tweak the room loading flows so we skip the room loading steps, which are no longer necessary when we do have a `JoinedRoom`.
- The flow restoration should still work.

## Motivation and context

Improve the UX when opening a room.

## Screenshots / GIFs

https://github.com/user-attachments/assets/3c561e32-e444-44a8-988d-c10aa54739f8

## Tests

Open a few rooms, check it works smoothly (it's better to check in a nightly/low overhead debug build).

When trying to open several rooms at a time, we should only navigate to the first one.

If you kill the app and restore it, it should work as before. Same for opening the room from other sources, like notifications.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/HomeFlowNode.kt
+++ b/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/HomeFlowNode.kt
@@ -62,6 +62,7 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.isActive
+import kotlinx.coroutines.job
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
@@ -193,18 +194,18 @@ class HomeFlowNode(
                         onSuccess = { joinedRoom ->
                             if (isActive) {
                                 callback.navigateToRoom(roomId, joinedRoom)
-                                loadingJoinedRoomJob.value = AsyncData.Success(loadingJoinedRoomJob.value.dataOrNull()!!)
+                                loadingJoinedRoomJob.value = AsyncData.Success(coroutineContext.job)
                                 // Wait a bit before resetting the state to avoid allowing to open several rooms
                                 delay(200.milliseconds)
                                 loadingJoinedRoomJob.value = AsyncData.Uninitialized
                             }
                         },
                         onFailure = {
-                            // If the operation wasn't cancelled, navigate without the room
+                            // If the operation wasn't cancelled, navigate without the room, using the room id
                             if (it !is CancellationException) {
                                 callback.navigateToRoom(roomId, null)
                             }
-                            loadingJoinedRoomJob.value = AsyncData.Failure(it, loadingJoinedRoomJob.value.dataOrNull()!!)
+                            loadingJoinedRoomJob.value = AsyncData.Failure(error = it, prevData = coroutineContext.job)
                             // Wait a bit before resetting the state to avoid allowing to open several rooms
                             delay(200.milliseconds)
                             loadingJoinedRoomJob.value = AsyncData.Uninitialized


frebib:
I've been running this for a day or so and it certainly improves the UX. Very occasionally the odd room will take several hundred ms to open but it's still less jarring than the old behaviour.
This feels much closer to opening rooms in WhatsApp as the room is fully populated as it animates in

features/home/impl/src/main/kotlin/io/element/android/features/home/impl/HomeFlowNode.kt
```diff
@@ -150,9 +165,58 @@ class HomeFlowNode(
...
+                }
+
+                val job = sessionCoroutineScope.launch {
+                    runCatchingExceptions {
+                        matrixClient.getJoinedRoom(roomId)
+                    }.fold(
+                        onSuccess = { joinedRoom ->
+                            if (isActive) {
+                                callback.navigateToRoom(roomId, joinedRoom)
+                                loadingJoinedRoomJob.value = AsyncData.Success(loadingJoinedRoomJob.value.dataOrNull()!!)
```
bmarty: This line is a bit hard to understand (it feels a bit hacky), but now that I have seen line 214, I get it. But does it add any value?
I guess it can just be removed. Not using `!!` is also a good thing.
Same for line 207, maybe just remove it?

appnav/src/main/kotlin/io/element/android/appnav/room/RoomFlowNode.kt
```diff
@@ -60,10 +60,13 @@ import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.shareIn
 import kotlinx.coroutines.launch
+import kotlinx.parcelize.IgnoredOnParcel
 import kotlinx.parcelize.Parcelize
 import timber.log.Timber
 import java.util.Optional
 import kotlin.jvm.optionals.getOrNull
+import im.vector.app.features.analytics.plan.JoinedRoom as JoinedRoomAnalyticsEvent
+import io.element.android.libraries.matrix.api.room.JoinedRoom as JoinedRoomInstance
```
bmarty, line 69: Too many (3?) `JoinedRoom` types...

bmarty (approved):
Thanks! Tested OK and no blocking remarks

features/home/impl/src/main/kotlin/io/element/android/features/home/impl/HomeFlowNode.kt
jmartinesp: I need to double check, but maybe we could move it to `Uninitialized`.

bmarty: you mean also remove the `delay` at the line below?

jmartinesp: Actually we can't remove it. It is needed to make sure we don't display the progress dialog while animating, the moment the state changes from `Loading` to something else. Then we keep it for a few ms so it acts as a debounce and makes sure we can't open several rooms at the same time.

I'll upload a commit improving the `!!` usage.

bmarty: OK

jmartinesp: That said, I don't like this implementation, so if we come up with a better one I'd be thrilled to remove it 😅 .

PR #5824, opened 2025-11-27
Add room list indicators about last message

bmarty:
## Content

Add room list indicators about last message:
- a clock when the message is being sent
- an error icon + critical text when the message cannot be sent

There is a known limitation that will be handled SDK side, see details in https://github.com/matrix-org/matrix-rust-sdk/issues/5904 

## Motivation and context

More info on the room list and stories:
- https://github.com/element-hq/element-meta/issues/2945
- https://github.com/element-hq/element-meta/issues/2946

## Screenshots / GIFs

See recorded ones.

## Tests

- Enter airplane mode
- send a message
- navigate back to the room list
- you should see the new icon

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


mxandreas:
@bmarty I tried the following scenario:

1. Turned off network.
2. Sent a message.

The timeline now has the pending message, however, in the room list - it says "Message failed to send". I think that's not correct - it should be a pending message like on the timeline.

features/home/impl/src/main/kotlin/io/element/android/features/home/impl/datasource/RoomListRoomSummaryFactory.kt
```diff
@@ -71,4 +73,28 @@ class RoomListRoomSummaryFactory(
...
     }
+
+    private fun computeLatestEvent(latestEvent: LatestEventValue, dm: Boolean): LatestEvent {
+        return when (latestEvent) {
+            is LatestEventValue.None -> {
+                LatestEvent.None
+            }
+            is LatestEventValue.Local -> {
+                if (latestEvent.isSending) {
+                    val content = roomLatestEventFormatter.format(latestEvent, dm).orEmpty()
```
ganfra, line 84: As we are already in a when, maybe the formatter could expose individual method to avoid re-checking the type?

features/home/impl/src/main/kotlin/io/element/android/features/home/impl/model/LatestEvent.kt
```diff
@@ -0,0 +1,34 @@
...
+
+package io.element.android.features.home.impl.model
+
+import androidx.compose.runtime.Immutable
+
+@Immutable
+sealed interface LatestEvent {
+    data object None : LatestEvent
+
+    data class Regular(
```
ganfra: Maybe `Received` or `Synced` would be a better name?

mxandreas:
> The timeline now has the pending message, however, in the room list - it says "Message failed to send". I think that's not correct - it should be a pending message like on the timeline.

And FWIW - this seems to be working correctly on iOS.

features/home/impl/src/main/kotlin/io/element/android/features/home/impl/datasource/RoomListRoomSummaryFactory.kt
bmarty: Done

features/home/impl/src/main/kotlin/io/element/android/features/home/impl/model/LatestEvent.kt
bmarty: Done

bmarty:
Thanks for checking @mxandreas, I also see what you are describing. @pixlwave confirmed that the mapping of the state that the SDK is sending is matching what iOS is doing, so it's weird that rendering is different. We may have a diff in the SDK version though, maybe worth checking again on the next RC.

PR #5826, opened 2025-11-27
change(room member): make sure we never display name/avatar when member is banned

ganfra:
## Content
What the title says.
Use [same logic ](https://github.com/element-hq/element-x-ios/blob/cf63a4dcc0c60000a21b2bf18874327cec3d2947/ElementX/Sources/Services/Users/UserProfileProxy.swift#L25-L25)than iOS.

## Motivation and context
Address https://github.com/element-hq/element-meta/issues/3003#issuecomment-3585178906 

## Screenshots / GIFs
See recorded screenshot for banned section.

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [X] Physical
- [ ] Emulator
- OS version(s):

## Checklist


ganfra:
> OK, but the filtering could be handled at the presentation layer maybe, it's a bit weird to do it in an extension.

I prefer to do it in the extension as it's already used from several places, this way we have less chance to forget, even if it's not perfect.

> Also we could reuse `FeatureFlags.ShowBlockedUsersDetails` flag.

I wasn't aware of the feature flag, but I don't think this will change

PR #5832, opened 2025-12-01
fix: usersWithRole(Owner) returns creators only if privilegedCreatorRole is true

ganfra:
## Content
What the title says.

## Motivation and context
Fix an issue on room version < 12 where creators shouldn't have a privileged creator role.

## Screenshots / GIFs
No ui change.

## Tests

- Open an old room where you have you and the user who created the room as admins.
- See he's displayed only once in the edit admin role.
- 

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/powerlevels/RoomMembersWithRole.kt
+++ b/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/powerlevels/RoomMembersWithRole.kt
@@ -36,7 +36,7 @@ fun BaseRoom.usersWithRole(predicate: (RoomMember.Role) -> Boolean): Flow<Immuta
     return readyMembersFlow.map { membersState ->
         membersState
             .activeRoomMembers()
-            .filter({ predicate(it.role) })
+            .filter { predicate(it.role) }
             .toImmutableList()
     }.distinctUntilChanged()
 }


libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/RoomInfo.kt
```diff
@@ -85,7 +85,7 @@ data class RoomInfo(
      */
     fun usersWithRole(role: RoomMember.Role): List<UserId> {
         return if (role is RoomMember.Role.Owner && role.isCreator) {
-            this.creators
+            if (privilegedCreatorRole) this.creators else emptyList()
```
bmarty: Maybe add a unit test to check that an empty list is returned when the conditions are met?

ganfra:
@bmarty I've actually rework that a bit more : 
I think this is a bit error prone to rely on the RoomInfo and PowerLevel for such things, so now we are only using the member list to compute those info. 
Also this is parity with iOS.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/powerlevels/RoomMembersWithRole.kt
```diff
@@ -34,12 +33,17 @@ fun BaseRoom.usersWithRole(role: RoomMember.Role): Flow<ImmutableList<RoomMember
...
-        .combine(readyMembersFlow) { powerLevels, membersState ->
-            membersState.activeRoomMembers()
-                .filter { powerLevels.contains(it.userId) }
-                .toImmutableList()
-        }
-        .distinctUntilChanged()
+    return readyMembersFlow.map { membersState ->
+        membersState
+            .activeRoomMembers()
+            .filter({ predicate(it.role) })
```
bmarty:
```suggestion
            .filter { predicate(it.role) }
```

bmarty (approved):
Thanks, the code looks a bit clearer.

Issue #3851, opened 2024-11-10
Message composer consumes entire window and covers messages bahind when crafting long reply


90vgt:
### Steps to reproduce

1. Where are you starting? What can you see?
From any room / DM, start typing out a long message in Element X. The composer box will grow as the message gets longer, eventually covering the entire message / room window behind it. This prevents you from being able to read back a longer message line by line and reply to it. 

#### What did you expect?
Expected behaviour would be for the composer to grow until a certain point when typing a longer message, but stopping to allow some of the room behind to remain visible to allow you to read back parts of longer messages while forming your reply. 

#### What happened instead?
Composer consumes the entire window, meaning you would have to either try and memorise a very long message while relpying, or close your keyboard frequently to try and read content in the background. This behaviour does not occur on the older Element Android, and every other messaging app (Signal, Whatsapp, SMS etc) also allows you to see some of message behind the composer. We frequently send and receive long messages, so this behaviour makes Element X difficult to use.

In the first example image, most of the room behind has been obscured.
![Screenshot_20241031_212756_Element X-cropped](https://github.com/user-attachments/assets/5f741101-c299-4052-9e25-50b174c1ae0e)

Continuing to type eventually obscures the entire room history
![Screenshot_20241031_212824_Element X-cropped](https://github.com/user-attachments/assets/63e74b17-6227-44fc-8653-b1b4b4e2748d)

The same message in Element Android (legacy) looks like this. Note that I'm using the rich text composer in the screenshot, but the behaviour is the same for the plain tect composer. 
![Screenshot_20241031_212944_Element-cropped](https://github.com/user-attachments/assets/135f7579-82f3-4528-a293-98cfceead265)

### Your phone model

Samsung Note20 Ultra 5G

### Operating system version

Android 13

### Application version and app store

Element X 0.7.2

### Homeserver

Self hosted, Synapse 118.0

### Will you send logs?

No

### Are you willing to provide a PR?

No

90vgt:
- I've tested this on another device, running Android 11, and the same issue occurs. 
- The issue is still present in EX 0.7.6 (latest at time of making this comment). 
- Checked the iOS version of the app, and found that the composer window stops expanding after a certain point and leaves a bit of a gap to allow reading of the message in the background still, which is the expected behaviour.

PR #5835, opened 2025-12-02
Limit composer height dynamically

bmarty:
## Content

Let the composer takes at max half of the available height. The value will be different if the soft keyboard is displayed or not.

First commit is a cleanup commit, second commit is the improvement

## Motivation and context

Ensure that last messages in the room are always visible, even if the content of the composer has a big height.

Closes https://github.com/element-hq/element-x-android/issues/3851

## Tests

- Open a room
- Type in the composer a message on several lines
- If the keyboard is closed, the composer should occupy 50% of the screen height
- If the keyboard is opened, the composer should occupy 50% of the remaining height, so that last messages of the room are always visible (and scrollable)

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/ExpandableBottomSheetLayoutState.kt
+++ b/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/components/ExpandableBottomSheetLayoutState.kt
@@ -11,7 +11,6 @@ package io.element.android.libraries.designsystem.components
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Stable
 import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableFloatStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue


jmartinesp (commented):
I *believe* this breaks the composer scrolling when maximized (might also happen without these changes): open the composer, switch to rich text editor, then write a lot of new lines.

If you try to scroll in those conditions one of these 2 things will happen:

1. Either the edit text won't scroll.
2. Or the composer will be resized to > 50% of the screen somehow.

At least that was my experience when testing this. I think the RTE layout uses the max size to detect whether it's maximized or not and these changes may break that logic.

bmarty:
> I _believe_ this breaks the composer scrolling when maximized (might also happen without these changes): open the composer, switch to rich text editor, then write a lot of new lines.
> 
> If you try to scroll in those conditions one of these 2 things will happen:
> 
>     1. Either the edit text won't scroll.
> 
>     2. Or the composer will be resized to > 50% of the screen somehow.
> 
> 
> At least that was my experience when testing this. I think the RTE layout uses the max size to detect whether it's maximized or not and these changes may break that logic.

I also observe 1, but I have the same behavior on develop, so it's not new. To scroll when using the RTE, I have to use a double fingers scroll for some reason. So I do not see any regression. I have not observed your point 2.

PR #5838, opened 2025-12-02
Upgrade the Rust SDK to `v25.12.2`

jmartinesp:
## Content

What the title says. I don't know why renovate didn't add the PR automatically.

Tracking read receipts now takes an enum value instead of a boolean.

## Motivation and context

Upgrading the SDK before the RC.

## Tests

(Once posthog has been downgraded again in https://github.com/element-hq/element-x-android/pull/5836:) use the app in general, check the latest room events in the room list, etc.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/fixtures/factories/NotificationItem.kt
+++ b/libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/fixtures/factories/NotificationItem.kt
@@ -64,6 +64,7 @@ fun aRustNotificationRoomInfo(
     isEncrypted: Boolean? = true,
     isDirect: Boolean = false,
     joinRule: JoinRule? = null,
+    isSpace: Boolean = false,
 ) = NotificationRoomInfo(
     displayName = displayName,
     avatarUrl = avatarUrl,
@@ -73,6 +74,7 @@ fun aRustNotificationRoomInfo(
     isEncrypted = isEncrypted,
     isDirect = isDirect,
     joinRule = joinRule,
+    isSpace = isSpace,
 )
 
 fun aRustNotificationEventTimeline(


jmartinesp:
Oh, now it opens the PR https://github.com/element-hq/element-x-android/pull/5837!

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/member/RoomMemberListFetcher.kt
```diff
@@ -118,7 +118,7 @@ internal class RoomMemberListFetcher(
                 while (true) {
                     // Loading the whole membersIterator as a stop-gap measure.
                     // We should probably implement some sort of paging in the future.
-                    coroutineContext.ensureActive()
+                    currentCoroutineContext().ensureActive()
```
ganfra: why this change here?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/JoinedRustRoom.kt
```diff
@@ -237,7 +238,7 @@ class JoinedRustRoom(
                     filter = filter,
                     internalIdPrefix = internalIdPrefix,
                     dateDividerMode = dateDividerMode,
-                    trackReadReceipts = trackReadReceipts,
+                    trackReadReceipts = if (trackReadReceipts) TimelineReadReceiptTracking.ALL_EVENTS else TimelineReadReceiptTracking.DISABLED,
```
ganfra, line 241: Are we sure about the usage here? What `MESSAGE_LIKE_EVENTS` is about?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/member/RoomMemberListFetcher.kt
jmartinesp: Not really needed, but there was a warning about `currentCoroutineContext()` being safer than calling `coroutineContext`, since that could reference a different context. I can revert this.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/JoinedRustRoom.kt
jmartinesp: It's about having the iOS behaviour where read receipts are only added to message like events, instead of any event in the timeline. It's also supposedly way slower than our current behaviour, so I believe we should keep it like this and maybe test the new behaviour in a nightly version instead.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/member/RoomMemberListFetcher.kt
jmartinesp: image

ganfra: Would be worth to be in a different PR yeah!

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/JoinedRustRoom.kt
pixlwave: Don't worry about the speed, those performance regressions were 100% programmer error 🙈. The final PR had no impact on performance at all :D

jmartinesp: We can still change this behaviour if needed after the RC though.

PR #5849, opened 2025-12-04
Fix work requests for inaccessible sessions being re-scheduled indefinitely

jmartinesp:
## Content

- Create `SessionRestorationException` with a couple of common cases for session restoration failures.
- Add `cause` value to `ClientException` so we can forward the root cause (i.e. the exception created above).
- When the session restoration fails when trying to fetch notifications, assume it won't work later either and avoid re-scheduling the work.

## Motivation and context

Sometimes when the user is logged out in a non-standard way, there can be pending operations to fetch notifications for sessions that no longer exist or can't be refreshed. If we don't stop this process, the work will be re-scheduled indefinitely.

## Tests

No easy way to test I'd say. It's probably caused by a race condition between receiving a notification and logging out of the account. I had this issue and the code fixed it.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/libraries/workmanager/impl/src/main/kotlin/io/element/android/libraries/workmanager/impl/DefaultWorkManagerScheduler.kt
+++ b/libraries/workmanager/impl/src/main/kotlin/io/element/android/libraries/workmanager/impl/DefaultWorkManagerScheduler.kt
@@ -11,6 +11,7 @@ package io.element.android.libraries.workmanager.impl
 import androidx.work.WorkManager
 import dev.zacsweers.metro.AppScope
 import dev.zacsweers.metro.ContributesBinding
+import dev.zacsweers.metro.SingleIn
 import io.element.android.libraries.matrix.api.core.SessionId
 import io.element.android.libraries.sessionstorage.api.observer.SessionListener
 import io.element.android.libraries.sessionstorage.api.observer.SessionObserver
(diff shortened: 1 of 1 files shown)


libraries/workmanager/impl/src/main/kotlin/io/element/android/libraries/workmanager/impl/DefaultWorkManagerScheduler.kt
```diff
@@ -8,23 +8,45 @@
...
+            .withPreviousValue()
+            .map { (prev, new) -> prev.orEmpty() - new.toSet() }
+            .onEach { removedSessions ->
+                for (sessionId in removedSessions) {
+                    Timber.d("Session removed for userId: $sessionId, cancelling associated workmanager requests")
+                    cancel(sessionId)
+                }
+            }
+            .launchIn(appCoroutineScope)
+    }
```
bmarty, line 41: Use `SessionObserver`?

jmartinesp: Good call! I had forgotten about it, to be honest 😅 .

PR #5858, opened 2025-12-05
Add SDK database vacuuming operations

jmartinesp:
## Content

Add a new `VacuumDatabaseWorker` operation which will be executed periodically, once a day when the device is idle.

## Motivation and context

Using the DB creates fragmentation in SQLite, so a manual `VACUUM` operation is needed from time to time to fix it and restore performance, otherwise it'll start to degrade.

## Screenshots / GIFs

image

(For this test I added a lower value than 1 day for periodicity, that's why there are several results)

## Tests

The only way to test this that I found is installing the APK and letting it run overnight. You can also manually vacuum from developer settings now.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/libraries/workmanager/impl/src/main/kotlin/io/element/android/libraries/workmanager/impl/DefaultWorkManagerScheduler.kt
+++ b/libraries/workmanager/impl/src/main/kotlin/io/element/android/libraries/workmanager/impl/DefaultWorkManagerScheduler.kt
@@ -8,6 +8,7 @@
 
 package io.element.android.libraries.workmanager.impl
 
+import androidx.work.WorkInfo
 import androidx.work.WorkManager
 import dev.zacsweers.metro.AppScope
 import dev.zacsweers.metro.ContributesBinding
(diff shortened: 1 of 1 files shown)


libraries/workmanager/impl/src/main/kotlin/io/element/android/libraries/workmanager/impl/DefaultWorkManagerScheduler.kt
```diff
@@ -51,6 +52,18 @@ class DefaultWorkManagerScheduler(
...
 
+    override fun hasPendingWork(sessionId: SessionId, requestType: WorkManagerRequestType): Boolean {
+        val workInfos = workManager.getWorkInfosByTag(workManagerTag(sessionId, requestType)).get().orEmpty()
+        return workInfos.any { info ->
+            val isPeriodic = info.periodicityInfo != null
+            val isCancelled = info.state == WorkInfo.State.CANCELLED
+            // It has pending work if:
+            // - It's not periodic and is not finished.
+            // - It's periodic and is not cancelled - since it'll be run again in a next iteration otherwise
+            !isPeriodic && !info.state.isFinished || isPeriodic && !isCancelled
```
bmarty, line 63: In case of failure of the worker, for instance when it returns `Result.failure()`, will `!isCancelled` return `true`?

jmartinesp: No, it will return `WorkInfo.State.FAILED` . `isFinished` will be true, but `isCancelled` won't.

Issue #5873, opened 2025-12-09
Optimise media quality has no impact on video size/resolution


mxandreas:
### Steps to reproduce

1. Make a 10s 4K video on the device - the size is 71MB and resolution 2160 p.
2. Turn "Optimize media quality" ON and upload the video.
3. Turn "Optimize media quality" OFF and upload the video.

#### What did you expect?

1. The resolution of the first video/upload is 720p.
2. The resolution of the second video/upload is 1080p. 

#### What happened instead?

The resolution of both video/uploads is 720p, with sizes 4,7MB.

### Your phone model

Samsung A36

### Will you send logs?

No

### Are you willing to provide a PR?

No

mxandreas:
@bmarty Let me know if you need logs.

PR #5892, opened 2025-12-11
Use the right video preset when sharing videos

jmartinesp:
## Content

- Makes `DefaultMediaOptimizationConfigProvider` take into account the feature flags to decide whether to use the separate video compression option or not, to select which video preset should be returned.
- Use this in several places instead of the raw `SessionPreferencesStore` to make sure the right preset is selected.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/5873

## Tests

1. In Settings -> Advanced settings, disable media compression.
2. Then share a 1080p+ video in a room.
3. Check the video is 1080p, not 720p.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/DefaultMediaOptimizationConfigProvider.kt
+++ b/libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/DefaultMediaOptimizationConfigProvider.kt
@@ -10,17 +10,28 @@ package io.element.android.libraries.mediaupload.impl
 
 import dev.zacsweers.metro.ContributesBinding
 import io.element.android.libraries.di.SessionScope
+import io.element.android.libraries.featureflag.api.FeatureFlagService
+import io.element.android.libraries.featureflag.api.FeatureFlags
 import io.element.android.libraries.mediaupload.api.MediaOptimizationConfig
 import io.element.android.libraries.mediaupload.api.MediaOptimizationConfigProvider
 import io.element.android.libraries.preferences.api.store.SessionPreferencesStore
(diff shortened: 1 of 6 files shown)


libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/DefaultMediaOptimizationConfigProvider.kt
```diff
@@ -10,17 +10,28 @@ package io.element.android.libraries.mediaupload.impl
...
-        videoCompressionPreset = sessionPreferencesStore.getVideoCompressionPreset().first(),
-    )
+    override suspend fun get(): MediaOptimizationConfig {
+        val compressImages = sessionPreferencesStore.doesOptimizeImages().first()
+        return MediaOptimizationConfig(
+            compressImages = compressImages,
+            videoCompressionPreset = if (featureFlagsService.isFeatureEnabled(FeatureFlags.SelectableMediaQuality)) {
+                sessionPreferencesStore.getVideoCompressionPreset().first()
+            } else {
+                if (compressImages) VideoCompressionPreset.STANDARD else VideoCompressionPreset.HIGH
```
bmarty, line 33: Small remark about the naming, not related to this PR.
We can read `VideoCompressionPreset.HIGH` like a High level of compression, so a low image quality. But actually this is the opposite, the quality will be high. We should maybe do some renaming in a dedicated PR. Maybe just rename `VideoCompressionPreset` to `VideoQualityPreset`?

jmartinesp: Sounds reasonable, yes.

PR #5910, opened 2025-12-16
Allow uploading extra data to Sentry when analytics is enabled

jmartinesp:
## Content

Allow adding extra analytic data to Sentry when analytics is enabled:

- Hashed homeserver value, so we can track issues on problematic ones while not exposing the value.
- Database sizes, which can help us understand if some performance issues may be caused by that.
- Make sure we can't upload session/user ids by mistake.
- Added some tests for `SentryAnalyticsProvider` too.

## Motivation and context

Improving the performance metrics.

## Tests

With analytics enabled, perform any action like opening a room or tapping on a notification, then check your interaction in Sentry uploaded those values.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/room/RustBaseRoomTest.kt
+++ b/libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/room/RustBaseRoomTest.kt
@@ -174,7 +174,6 @@ class RustBaseRoomTest {
                 dispatchers = dispatchers,
             ),
             roomMembershipObserver = roomMembershipObserver,
-            // Not using backgroundScope here, but the test scope
             sessionCoroutineScope = backgroundScope,
             roomInfoMapper = RoomInfoMapper(),
             initialRoomInfo = initialRoomInfo,


libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/room/RustBaseRoomTest.kt
```diff
@@ -175,7 +175,7 @@ class RustBaseRoomTest {
             ),
             roomMembershipObserver = roomMembershipObserver,
             // Not using backgroundScope here, but the test scope
-            sessionCoroutineScope = this,
+            sessionCoroutineScope = backgroundScope,
```
bmarty, line 177: Maybe remove the outdated comment above?

services/analyticsproviders/api/src/main/kotlin/io/element/android/services/analyticsproviders/api/AnalyticsUserData.kt
```diff
@@ -0,0 +1,20 @@
...
+
+package io.element.android.services.analyticsproviders.api
+
+object AnalyticsUserData {
+    const val HOMESERVER = "homeserver"
+
+    const val STATE_STORE_SIZE = "state_store_size"
+    const val EVENT_CACHE_SIZE = "event_cache_size"
+    const val CRYPTO_STORE_SIZE = "crypto_store_size"
+    const val MEDIA_STORE_SIZE = "media_store_size"
```
bmarty, line 16: Maybe for readability send the size in bytes, but in kilo bytes, or even in mega bytes?

jmartinesp: Unless I missed something, they should be uploaded in MBs.

libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/room/RustBaseRoomTest.kt
jmartinesp: Oops! Thanks!

Issue #5812, opened 2025-11-25
Wrong verify account step after creating a new account


srgcdev:
### Steps to reproduce

1. Open element x and create a new account
2. You are redirected to the mas web
3. You create a new account there
4. You are redirected again to element x
5. Element x shows a screen to verify the account with a recovery key or using another device and that's not possible because the account is new
6. When you skip the verify account step, element x ask you to create a recovery key

#### What did you expect?
When you create a new account I expect that the first thing element x ask you is to create a recovery key

#### What happened instead?
When you use element x for creating a new account, after that, element x ask you to use a recovery key or another device to verify the account. That is not possible because you are creating a new account.

### Will you send logs?

No

### Are you willing to provide a PR?

No

jmartinesp:
I can reproduce this. Using 'Can't confirm' dismisses the screen that's incorrectly displayed, but we'll try to figure out why this flow is displayed.

PR #5914, opened 2025-12-17
Fix unverified account after account creation

jmartinesp:
## Content

When we create an account either using OIDC or by importing a login and password one, we need to wait until the verification state is known (either verified or unverified). The problem is the verification service will return incorrect values until the E2EE tasks are initialized in the SDK, even if we add the state listeners after doing so.

So what we can do is initialize the E2EE setup, discard any invalid verification state received while it's not initialized, and take only those received after it's initialized. The `CreateAccountPresenter` for login and password account creation was already kind of doing this by waiting an extra time for the updated verified state values, but I moved it to `RustMatrixAuthenticationService` so it's centralised for both account creation flows.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/5812.

## Tests

Create a new account with an OIDC flow, check the 'verify your account' screen is not displayed.

You can also test the OIDC login and login and password account creation, if there is an available homeserver for that.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
@@ -36,6 +36,7 @@ import io.element.android.libraries.matrix.impl.keys.PassphraseGenerator
 import io.element.android.libraries.matrix.impl.mapper.toSessionData
 import io.element.android.libraries.matrix.impl.paths.SessionPaths
 import io.element.android.libraries.matrix.impl.paths.SessionPathsFactory
+import io.element.android.libraries.matrix.impl.toSession
 import io.element.android.libraries.sessionstorage.api.LoginType
 import io.element.android.libraries.sessionstorage.api.SessionStore
 import kotlinx.coroutines.CancellationException
@@ -172,9 +173,15 @@ class RustMatrixAuthenticationService(
                     passphrase = pendingPassphrase,
                     sessionPaths = currentSessionPaths,
                 )
+
+                // We restore the client using the just retrieved session data
+                client.restoreSession(sessionData.toSession())
                 val matrixClient = rustMatrixClientFactory.create(client)
+
+                // We wait for the verification state to be known
                 matrixClient.waitForKnownVerificationState()
 
+                // And once it's ready we share it and save the actual session data
                 newMatrixClientObservers.forEach { it.invoke(matrixClient) }
                 sessionStore.addSession(sessionData)
 
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
@@ -193,7 +193,7 @@ sealed interface ClientBuilderSlidingSync {
     data object Native : ClientBuilderSlidingSync
 }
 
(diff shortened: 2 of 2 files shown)


jmartinesp:
I've tried for almost a couple of hours to add some tests to this, but the `runTest` behaviour that automatically skips delays/timeouts made it impossible 🥲 .

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
```diff
@@ -160,16 +164,23 @@ class RustMatrixAuthenticationService(
...
-                currentClient ?: error("You need to call `setHomeserver()` first")
+                val client = currentClient ?: error("You need to call `setHomeserver()` first")
                 val currentSessionPaths = sessionPaths ?: error("You need to call `setHomeserver()` first")
                 val sessionData = externalSession.toSessionData(
                     isTokenValid = true,
                     loginType = LoginType.PASSWORD,
                     passphrase = pendingPassphrase,
                     sessionPaths = currentSessionPaths,
                 )
-                clear()
```
ganfra, line 171: Just to be sure, we don't need the clear anymore?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/verification/RustSessionVerificationService.kt
```diff
@@ -114,12 +129,14 @@ class RustSessionVerificationService(
     }
 
     override suspend fun requestCurrentSessionVerification() = tryOrFail {
+        ensureEncryptionIsInitialized()
```
ganfra, line 127: Can we group with the initVerificationController in a single method maybe?

ganfra (commented):
LGTM I guess, but wouldn't be possible to have this logic in the sdk? How iOS is managing this?

jmartinesp:
> How iOS is managing this?

Badly, they have the same issue 😅 .

> Wouldn't be possible to have this logic in the sdk? 

Maybe, but the crypto SDK is quite complex. I'd rather just fix it here and create an issue so they can maybe fix it in the SDK in the future.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/verification/RustSessionVerificationService.kt
jmartinesp: I wanted to keep them apart in case we didn't have to call them both at the same time, but so far they're always called together so it makes sense to have a single method.

jmartinesp: https://github.com/element-hq/element-x-android/pull/5914/commits/37a16852ce17f921abeb1893b0667d47bbad63e5

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
jmartinesp: The new code would have caused an issue, nice catch. https://github.com/element-hq/element-x-android/pull/5914/commits/1434d375801a00f510ccf9773679af14cad1ffbf should fix it and make this step work the same as the rest of the login/registration flows in this class. Sadly, I couldn't test it since I can't find a HS with this configuration 🫤 .

Issue #5890, opened 2025-12-11
[Task] EX and SDK performance monitoring


manuroe:
Please see and discuss the details in the meta issue.

manuroe:
@jmartinesp with https://github.com/element-hq/element-x-android/pull/6046 being merged, do you see other things to be done for Android?

jmartinesp:
If we don't detect any other outliers, I think we can close this task. If we still want to measure sync request/response times we can create a new one, since we'll need some collaboration from the SDK team for exposing these events.

PR #5937, opened 2025-12-18
Adjust metrics to the new specifications

jmartinesp:
## Content

- Add `TransactionDefinition` class as a way to describe the analytics transactions from [the <del>matrix</del> metrics spec](https://github.com/element-hq/element-meta/blob/manuroe/ex-performance-metrics/docs/ex_perfomance_metrics.md) (pun intended). Add them to the `AnalyticsTransactions` object.
- Base `AnalyticsLongRunningTransactions` on these definitions, add a `description` optional field to sentry transactions.
- Add `AnalyticsSendMessageWatcher` to track the last missing UX flow: sending a message and receiving it back from `/sync`.

## Motivation and context

Part of the work for https://github.com/element-hq/element-x-android/issues/5890.

## Tests

Perform some of these actions, see they are uploaded to Sentry with the proper values.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/services/analytics/noop/src/main/kotlin/io/element/android/services/analytics/noop/watchers/NoopAnalyticsSendMessageWatcher.kt
+++ b/services/analytics/noop/src/main/kotlin/io/element/android/services/analytics/noop/watchers/NoopAnalyticsSendMessageWatcher.kt
@@ -1,0 +1,18 @@
+/*
+ * Copyright (c) 2025 Element Creations Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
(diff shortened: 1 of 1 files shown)


manuroe:
image

Do you think it will be possible to include more spans from the SDK to the `Catch-up` metric? Having `/sync` request span would be a useful information.

jmartinesp:
I can double check if it's possible.

---

I don't think it is, since we have nothing in the SDK to tell us when the sync request has started. We also don't measure the time until the room list UI is changed, since that would be quite difficult to track and easy to measure incorrectly.

PR #5960, opened 2025-12-29
Ensure space feature is enabled

ganfra:
## Content
Remove the Space feature flag entirely.

On a follow-up PR, we might want to returns the default value (or `true`) when a feature is marked as finished, to avoid having previous values interfering.

## Motivation and context
Clean up code.
Makes sure space tab is visible.

## Screenshots / GIFs
No ui change.

## Tests

- Disable the space feature flag from an old version of the app
- Update to this PR version
- Check space tab is visible as long as you have joined at least one space.

## Tested devices

- [X] Physical
- [ ] Emulator
- OS version(s):

## Checklist


libraries/featureflag/impl/src/main/kotlin/io/element/android/libraries/featureflag/impl/DefaultFeatureFlagService.kt
```diff
@@ -25,10 +25,14 @@ class DefaultFeatureFlagService(
...
-            ?.isFeatureEnabledFlow(feature)
-            ?: flowOf(feature.defaultValue(buildMeta))
+        return if (feature.isFinished) {
+            flowOf(feature.defaultValue(buildMeta))
+        } else {
+            providers.filter { it.hasFeature(feature) }
+                .maxByOrNull(FeatureFlagProvider::priority)
+                ?.isFeatureEnabledFlow(feature)
+                ?: flowOf(feature.defaultValue(buildMeta))
+        }
```
bmarty, line 35: > On a follow-up PR, we might want to returns the default value (or true) when a feature is marked as finished, to avoid having previous values interfering.

Actually in the same PR then? Not a problem, as this change is in a dedicated commit, but just to confirm :)

ganfra: I totally forgot I had a commit for that, wanted to discuss with the team first

Issue #5944, opened 2025-12-19
Clear button on room list search is misaligned


frebib:
Confirmed on the latest nightly

![Screenshot_20251219-195859.png](https://github.com/user-attachments/assets/b4d156d4-0ca9-4600-80b5-aee584d8abb5)

I think it was caused by #5179

jmartinesp:
It's probably this issue in Compose: https://issuetracker.google.com/issues/449573623. We've already added some feedback in case it helps them fix the issue or provide a valid workaround.

PR #5969, opened 2026-01-02
Upgrade compose to 2025.12.01

bmarty:
Upgrade compose to 2025.12.01, not sure why Renovate does not handle this. 

Does not fix https://github.com/element-hq/element-x-android/issues/5944


frebib:
@bmarty this closed #5944 despite saying it doesn't fix it 😂
GitHub is trying to be too smart

manuroe:
Good catch @frebib ! I reopened it.

PR #5971, opened 2026-01-05
Sync Strings

ElementBot:


bmarty:
> LGTM I guess? Any French dev could double check the new strings @bmarty , @ganfra ?

I have checked the French translation (that I've done last week) and this seems fine.

FTR I have uniformised the translations of "QR code" to "code QR" in the existing French wording, but this will be in the next sync. This seems to be the correct way to name such code according to https://fr.wikipedia.org/wiki/Code_QR

PR #5980, opened 2026-01-08
[a11y] voice message improvements

bmarty:
## Content

Improve IconButton on the composer area regarding accessibility focus when the screen reader is enabled and when the user is navigating with a keyboard.

## Motivation and context

Closes https://github.com/element-hq/customer-success/issues/556

### Using the emulator

- Open a room timeline
- Using the keyboard arrows, navigate to the microphone icon to start recording a voice message.
- Pressing the space bar of the keyboard, you should be able to start recording, stop recording and send the voice message without losing the focus on the icon.

When a voice message as been recorded but not sent, you should be able to navigate to the play icon and press the space bar several times to start/pause the playback without losing the focus

When a voice message as been recorded but not sent, you should be able to navigate to bin icon and press the space bar to delete the pending voice message without losing the focus, which is still on the same icon area. The icon is now the "Attachement" icon.

### Using a real device with screen reader enabled

You should be able to perform the same actions using a real device with screen reader enabled.

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s):

## Checklist


libraries/textcomposer/impl/src/main/kotlin/io/element/android/libraries/textcomposer/TextComposer.kt
```diff
@@ -419,47 +427,74 @@ private fun StandardLayout(
...
         Row(verticalAlignment = Alignment.Bottom) {
-            if (voiceMessageState !is VoiceMessageState.Idle) {
-                if (voiceMessageState is VoiceMessageState.Preview || voiceMessageState is VoiceMessageState.Recording) {
-                    Box(
+            when (composerMode) {
+                is MessageComposerMode.Attachment -> {
+                    Spacer(modifier = Modifier.width(12.dp))
+                }
+                is MessageComposerMode.EditCaption -> {
+                    Spacer(modifier = Modifier.width(19.dp))
```
ganfra, line 454: I guess it's not new but those spacings are a bit...weird?

libraries/textcomposer/impl/src/main/kotlin/io/element/android/libraries/textcomposer/components/SendButton.kt
```diff
@@ -41,48 +41,40 @@ import io.element.android.libraries.textcomposer.model.MessageComposerMode
 @Composable
 internal fun SendButton(
```
ganfra: It feels like it shouldn't be called a `Button` anymore, as it's just an `Icon`. Can you rename? This one and all the others `Button`.

bmarty:
> Also the code is bit hard to follow as we are now checking the same state in multiple places...

I agree, but this is the price to pay to have only one `IconButton` in order to keep the accessibility focus when the icon (and the click callback) change. I'll try to find a better approach.

libraries/textcomposer/impl/src/main/kotlin/io/element/android/libraries/textcomposer/TextComposer.kt
bmarty: Yes, not new, and I think this matches the design for those composer modes.

bmarty:
@ganfra , PR updated, thanks for the remarks.

Issue #5996, opened 2026-01-12
Same room screen after sharing content to that room while it was in the background


bnjbvr:
### Steps to reproduce

- Starting from the room list, open a room's view
- Put the app in the background (without killing it)
- Share e.g. an image to this room, using Android's intent mechanism
- (I end up in the room view)
- Return to the previous screen

#### What did you expect?

That the previous screen be the room list.

#### What happened instead?

The previous screen is *also* the room's view, i.e. it was pushed twice in the navigation stack.

### Will you send logs?

No

### Are you willing to provide a PR?

No

PR #6001, opened 2026-01-13
When handling incoming share, reuse existing room screen if possible

jmartinesp:
## Content

- When attaching a room, reuse a previous node in the backstack if possible.
- When handling the success result for the incoming share node, dispose the node, wait for it to be disposed, then navigate to the room if needed. Otherwise, the attach room operation would override the disposal of the incoming share node.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/5996.

## Tests

1. Open a room.
2. Put the app in background.
3. Share some media to that same room from outside the app.
4. Check the room is displayed after sharing.
5. When going back, you'll land on the home screen, not a previous instance of this same room.

You can also test this with some other room/screen open and the back navigation should still make sense.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist


bmarty (approved):
Makes sense, thanks!

PR #6003, opened 2026-01-13
Ensure that log files are not too big.

### LINKED ISSUE #5983: Sending bug report sometimes fails: too many log lines
### Steps to reproduce

1. Try to send a bug report including logs
2. Observe this error:

Image

- The SDK is configured to store at max 7 * 24 files.
- We are ensuring that the total size of sent logs lines is not too high (`MAX_LOG_UPLOAD_SIZE = 50 * 1024 * 1024L`

We are not sure why this error is occurring now. Logs are more verbose and the limit is reach maybe faster, so the bug is maybe not new.

To mitigate this issue, we could reduce the number of logs sent to the latest 10 files, with an option to send all of them, which can still be useful to investigate issue related to cryptography. Generally, the last logs are the only useful to investigate a "regular" issue.

#### What did you expect?

The bug report to be sent

#### What happened instead?

The bug report is not sent

### Your phone model

Any

### Operating system version

Any

251200

### Homeserver

Any

### Will you send logs?

No

Yes

bmarty:
## Content

Ensure that log files are not too big, else the rageshake server will reject the request.

See https://github.com/element-hq/element-android/issues/9096#issuecomment-3480128082

## Motivation and context

Closes #5983

## Screenshots / GIFs

NA

## Tests

- Send a bug report.
- It should work as usual, no error should occurs.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/rageshake/impl/src/main/kotlin/io/element/android/features/rageshake/impl/reporter/DefaultBugReporter.kt
+++ b/features/rageshake/impl/src/main/kotlin/io/element/android/features/rageshake/impl/reporter/DefaultBugReporter.kt
@@ -158,6 +158,7 @@ class DefaultBugReporter(
                 }
                 if (withCrashLogs || withDevicesLogs) {
                     saveLogCat()
+                        ?.takeIf { it.length() < RageshakeConfig.MAX_LOG_CONTENT_SIZE }
                         ?.let { logCatFile ->
                             compressFile(logCatFile).also {
                                 logCatFile.safeDelete()
@@ -191,6 +192,7 @@ class DefaultBugReporter(
                     .addFormDataPart("label", buildMeta.versionName)
                     .addFormDataPart("label", buildMeta.flavorDescription)
                     .addFormDataPart("branch_name", buildMeta.gitBranchName)
+
                 userId?.let {
                     matrixClientProvider.getOrNull(it)?.let { client ->
                         val curveKey = client.encryptionService.deviceCurve25519()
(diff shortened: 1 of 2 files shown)


features/rageshake/impl/src/main/kotlin/io/element/android/features/rageshake/impl/reporter/DefaultBugReporter.kt
```diff
@@ -390,7 +392,11 @@ class DefaultBugReporter(
         ) {
             val logDirectory = logDirectory()
             logDirectory.listFiles()
-                ?.filter { it.isFile && !it.name.endsWith(LOG_CAT_FILENAME) }
+                ?.filter {
+                    it.isFile &&
+                        !it.name.endsWith(LOG_CAT_FILENAME) &&
+                        it.length() < RageshakeConfig.MAX_LOG_CONTENT_SIZE
```
jmartinesp, line 398: Shouldn't this be checked *after* the file is gzipped, not before?

bmarty: The server will decompress the file and index it, so the size has to be reasonable, else the server will reject the request.

bmarty: Long term fix should be to let the SDK rotate the file if it's too long, to avoid skipping long files when sending a RS.

jmartinesp (approved):
Thanks!

features/rageshake/impl/src/main/kotlin/io/element/android/features/rageshake/impl/reporter/DefaultBugReporter.kt
jmartinesp: That would need [this](https://github.com/tokio-rs/tracing/issues/1940) to be implemented. It seems like there are different implementations in forks, but none upstream.

Issue #5733, opened 2025-11-14
[Task] Add support for room selection and removal in the Space screen.


pixlwave:
Please see and discuss the details in the meta issue.

PR #6022, opened 2026-01-15
Space : manage rooms

ganfra:
## Content

Introduce the manage rooms action in the space screen, which allows to select multiple rooms (and only rooms) to be removed from the space graph.

## Motivation and context
Closes https://github.com/element-hq/element-x-android/issues/5733

## Screenshots / GIFs
See recorded 

## Tests

- Create a Space and add several rooms to it with EW
- Open the Space with the same account on EXA, click on the 3 dot menu to see the manage rooms action
- Play with the selection

## Tested devices

- [X] Physical
- [X] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/space/impl/src/main/kotlin/io/element/android/features/space/impl/root/SpacePermissions.kt
+++ b/features/space/impl/src/main/kotlin/io/element/android/features/space/impl/root/SpacePermissions.kt
@@ -12,6 +12,11 @@ import io.element.android.features.space.impl.settings.spaceSettingsPermissions
 import io.element.android.libraries.matrix.api.room.StateEventType
 import io.element.android.libraries.matrix.api.room.powerlevels.RoomPermissions
 
+/**
+ * Permissions needed for different actions in the Space screen.
+ * @param settingsPermissions Permissions related to space settings.
+ * @param canEditSpaceGraph Whether the user can edit the space graph (add/remove children).
+ */
 data class SpacePermissions(
     val settingsPermissions: SpaceSettingsPermissions,
     val canEditSpaceGraph: Boolean,


features/space/impl/src/main/kotlin/io/element/android/features/space/impl/root/SpacePermissions.kt
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package io.element.android.features.space.impl.root
+
+import io.element.android.features.space.impl.settings.SpaceSettingsPermissions
+import io.element.android.features.space.impl.settings.spaceSettingsPermissions
+import io.element.android.libraries.matrix.api.room.StateEventType
+import io.element.android.libraries.matrix.api.room.powerlevels.RoomPermissions
+
+data class SpacePermissions(
```
jmartinesp, line 20: Maybe add some docs about why this is needed? It seems a bit confusing to have `RoomPermissions`, `SpacePermissions`, then `SpaceSettingsPermissions`.

features/space/impl/src/main/kotlin/io/element/android/features/space/impl/root/SpacePresenter.kt
```diff
@@ -88,22 +89,46 @@ class SpacePresenter(
...
+                    .let { list ->
+                        if (isManageMode) {
+                            // In manage mode, only show rooms (not spaces)
+                            list.filter { !it.isSpace }
+                        } else {
+                            list
+                        }
+                    }
+                    .toImmutableList()
+            }
```
jmartinesp, line 129: I'm wondering if the remove spaces part would be better split into a separate presenter, to keep this one simple. Feel free to keep it here if you think it's not worth it though.

ganfra: I don't think it's worth it for now, let's keep it that way and iterate later if we are adding more code to it.

PR #6035, opened 2026-01-19
Add `MediaSource.safeUrl` for removing invalid fragment part from URLs

jmartinesp:
## Content

Add `MediaSource.safeUrl` which returns a valid MXC url without a fragment part or the original URL if it's not MXC.

## Motivation and context

We've seen some MXC urls in the wild having some `mxc://foo/bar#auto` fragment suffix, which is invalid, but the URL before that fragment part is valid and can be displayed.

## Tests

The only way I think we could test this is manually editing/sending some custom media event and adding the `#fragment` part to the MXC url.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/media/MediaSource.kt
+++ b/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/media/MediaSource.kt
@@ -9,7 +9,7 @@
 package io.element.android.libraries.matrix.api.media
 
 import android.os.Parcelable
-import androidx.core.net.toUri
+import kotlinx.parcelize.IgnoredOnParcel
 import kotlinx.parcelize.Parcelize
 
 @Parcelize
@@ -17,31 +17,20 @@ data class MediaSource(
     /**
      * Url of the media.
      */
-    val url: String,
+    private val url: String,
     /**
      * This is used to hold data for encrypted media.
      */
     val json: String? = null,
-) : Parcelable
-
-/**
- * Returns a new [MediaSource] with a valid URL.
- */
-fun MediaSource.withCleanUrl(): MediaSource {
-    val uri = this.url.toUri()
-    if (uri.scheme != "mxc") return this
-
-    // We've seen some MXC urls in the wild having some `mxc://foo/bar#auto` fragment suffix, which is invalid
-    val cleanedUrl = buildString {
-        append(uri.scheme)
-        if (!this.endsWith("://")) {
-            append("://")
-        }
-        append(uri.host)
-        if (uri.path != null) {
-            append(uri.path)
-        }
+) : Parcelable {
+    /**
+     * A URL with invalid parts (like `#fragment`, if it's an MXC url) removed.
+     */
+    @IgnoredOnParcel
+    val safeUrl = if (url.startsWith("mxc")) {
+        // We've seen some MXC urls in the wild having some `mxc://foo/bar#auto` fragment suffix, which is invalid
+        url.substringBefore("#")
+    } else {
+        url
     }
-
-    return this.copy(url = cleanedUrl)
 }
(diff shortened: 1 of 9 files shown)


jmartinesp:
I completely forgot about this draft PR 😬 .

libraries/matrixmedia/impl/src/main/kotlin/io/element/android/libraries/matrix/ui/media/CoilMediaFetcher.kt
```diff
@@ -28,14 +29,18 @@ internal class CoilMediaFetcher(
...
+            source.url.startsWith("mxc:") -> source.withCleanUrl()
+            else -> source
         }
         return when (val kind = mediaData.kind) {
-            is MediaRequestData.Kind.Content -> fetchContent(source)
-            is MediaRequestData.Kind.Thumbnail -> fetchThumbnail(source, kind)
-            is MediaRequestData.Kind.File -> fetchFile(source, kind)
+            is MediaRequestData.Kind.Content -> fetchContent(mediaSource)
+            is MediaRequestData.Kind.Thumbnail -> fetchThumbnail(mediaSource, kind)
+            is MediaRequestData.Kind.File -> fetchFile(mediaSource, kind)
```
bmarty, line 38: This change can be reverted with my proposal.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/media/MediaSource.kt
```diff
@@ -22,3 +23,25 @@ data class MediaSource(
      */
     val json: String? = null,
 ) : Parcelable
+
+/**
+ * Returns a new [MediaSource] with a valid URL.
+ */
+fun MediaSource.withCleanUrl(): MediaSource {
```
bmarty: Don't we want the url to be cleaned up in the constructor? It's maybe safer to move this code where we create MediaSource. Maybe something like:

```kotlin
@Parcelize
data class MediaSource(
    /**
     * Url of the media.
     */
    private val url: String,
    /**
     * This is used to hold data for encrypted media.
     */
    val json: String? = null,
) : Parcelable {
    @IgnoredOnParcel
    val safeUrl = if (url.startsWith("mxc")) {
        // We've seen some MXC urls in the wild having some `mxc://foo/bar#auto` fragment suffix, which is invalid
        url.substringBefore("#")
    } else {
        url
    }
}
```

or introduce a factory.

libraries/matrixmedia/impl/src/main/kotlin/io/element/android/libraries/matrix/ui/media/CoilMediaFetcher.kt
```diff
@@ -27,15 +27,15 @@ internal class CoilMediaFetcher(
     private val mediaData: MediaRequestData,
 ) : Fetcher {
     override suspend fun fetch(): FetchResult? {
-        val source = mediaData.source
-        if (source == null) {
+        val mediaSource = mediaData.source
```
bmarty, line 30: Fully revert the change?

jmartinesp:
> That said, I wonder if the fix should be in the SDK...

Fun fact: it seems like the SDK *should* already verify MXC urls. I'm not sure how we got those invalid ones in the client, but I just created a test with an mxc url with `#fragment` and it failed to validate, and that should prevent it from being returned to the client 🫤 .

libraries/matrixmedia/impl/src/main/kotlin/io/element/android/libraries/matrix/ui/media/CoilMediaFetcher.kt
bmarty: Ok, it's actually good like that.

bmarty (approved):
Thanks, let's merge this!

Issue #5901, opened 2025-12-12
Notifications don't show images any more


frebib:
This is a recent regression. To reproduce:

1. Receive a message that is an m.image either with or without a caption
2. Pull down the notification shade and observe that only the image name is displayed

bmarty:
Probably a regression from https://github.com/element-hq/element-x-android/pull/5684. I am having a look

bmarty:
I do not repro here. Can you post the event content of `m.image` anonymized so that I can see where the problem could come from?

With this content, it works as expected:

```json
{
  "type": "m.room.message",
  "content": {
    "body": "image.png",
    "info": {
      "size": 3204,
      "mimetype": "image/png",
      "w": 30,
      "h": 27,
      "xyz.amorgan.blurhash": "L04Ld$%N00WAx]j[WAj[00Rj_4WB",
      "org.matrix.msc4230.is_animated": false
    },
    "msgtype": "m.image",
    "m.mentions": {},
    "file": {
      "v": "v2",
      "key": {
        "alg": "A256CTR",
        "ext": true,
        "k": "REDACTED",
        "key_ops": [
          "encrypt",
          "decrypt"
        ],
        "kty": "oct"
      },
      "iv": "4N3Sga7vriYAAAAAAAAAAA",
      "hashes": {
        "sha256": "Z9yYyblXbgREDACTED"
      },
      "url": "mxc://matrix.org/uIdnwREDACTED"
    }
  }
}
```

frebib:
![Screenshot_20251215-102855.png](https://github.com/user-attachments/assets/ab51f849-f146-4f18-987d-48796366da09)

The photo is just of my desk so I didn't try to redact too hard. If you need logs or anything else, let me know 
```json
{
  "content": {
    "body": "photo.jpg",
    "file": {
      "hashes": {
        "sha256": "KUS+tCjiRwFX6ki0evx0kZt+19GwHoj5J6vNkSlh6Uc"
      },
      "iv": "\/...",
      "key": {
        "alg": "A256CTR",
        "ext": true,
        "k": "...",
        "key_ops": [
          "encrypt",
          "decrypt"
        ],
        "kty": "oct"
      },
      "url": "mxc:\/\/spritsail.io\/...",
      "v": "v2"
    },
    "info": {
      "h": 1020,
      "mimetype": "image\/jpeg",
      "size": 62706,
      "thumbnail_file": {
        "hashes": {
[...]

PR #6038, opened 2026-01-19
Use `MediaPreviewValue.Private` to check if media should be displayed in notifications

jmartinesp:
## Content

What the title says, we now take that option into account, inside a `when` so if a new option is added we'll also have to support it.

Also added `NotificationData.isPublicRoom` so we can use it to check if the room is public or not.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/5901.

## Tests

1. Go to Settings -> Advanced Settings -> Set 'Show media in timeline' to 'In private rooms'.
2. Backround the app.
3. From another device, in a private room, send an image. A notification should appear with a preview of the image.
4. From another device, in a public room, send an image. A caption or the filename of the image should be displayed instead.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
@@ -31,11 +31,12 @@ import io.element.android.libraries.matrix.api.core.SessionId
 import io.element.android.libraries.matrix.api.core.ThreadId
 import io.element.android.libraries.matrix.api.core.UserId
 import io.element.android.libraries.matrix.api.exception.NotificationResolverException
-import io.element.android.libraries.matrix.api.media.MediaPreviewValue
 import io.element.android.libraries.matrix.api.media.getMediaPreviewValue
+import io.element.android.libraries.matrix.api.media.isPreviewEnabled
 import io.element.android.libraries.matrix.api.notification.NotificationContent
 import io.element.android.libraries.matrix.api.notification.NotificationData
 import io.element.android.libraries.matrix.api.permalink.PermalinkParser
+import io.element.android.libraries.matrix.api.room.join.JoinRule
 import io.element.android.libraries.matrix.api.timeline.item.event.AudioMessageType
 import io.element.android.libraries.matrix.api.timeline.item.event.EmoteMessageType
 import io.element.android.libraries.matrix.api.timeline.item.event.EventType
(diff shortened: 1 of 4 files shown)


Susurrus:
Do you want to add a log message for if the URI ends up being null?

jmartinesp:
> Do you want to add a log message for if the URI ends up being null?

Good idea. Done in a05892465b1f953280530fb694dc138e4207a8c9.

frebib:
This is great. The only other remaining gripe I have in this area is that it's still possible to get empty or misleading notifications when the image shouldn't be displayed. I think it would make sense to mirror what the room list does in those cases.


Possibly something like:

When body/filename are empty/missing, display `Image`
When there's a filename but no caption, display: `Image: filename.jpg`
When there's a caption too, display:
`Image: filename.jpg\nThis is the caption`

At the very least, the first one should be uncontentious as it's better than an empty notification. The other two are more design decisions

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
```diff
@@ -140,10 +140,27 @@ class DefaultNotifiableEventResolver(
     ): Result<ResolvedPushEvent> = runCatchingExceptions {
         when (val content = this.content) {
             is NotificationContent.MessageLike.RoomMessage -> {
-                val showMediaPreview = client.mediaPreviewService.getMediaPreviewValue() == MediaPreviewValue.On
+                val showMediaPreview = when (client.mediaPreviewService.getMediaPreviewValue()) {
```
bmarty: Maybe reuse existing extension `isPreviewEnabled`:
```suggestion
                val showMediaPreview = client.mediaPreviewService.getMediaPreviewValue().isPreviewEnabled(joinRule) {
```

Instead of the new field `isPublicRoom`, the join rule will have to be provided

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
```diff
@@ -140,10 +140,27 @@ class DefaultNotifiableEventResolver(
...
-                val imageMimeType = if (showMediaPreview) content.getImageMimetype() else null
-                val imageUriString = imageMimeType?.let { content.fetchImageIfPresent(client, imageMimeType)?.toString() }
+                val imageMimeType = content.getImageMimetype()
+                val imageUriString = if (showMediaPreview && imageMimeType != null) {
+                    content.fetchImageIfPresent(client, imageMimeType)?.toString()
+                } else {
+                    null
+                }
+
+                if (content.getImageMimetype() != null && !showMediaPreview && isPublicRoom) {
```
bmarty:
```suggestion
                if (imageMimeType != null && !showMediaPreview && isPublicRoom) {
```

bmarty (approved):
Thanks for investigating and fixing the issue!

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
jmartinesp: You'd still have to pass the `joinRule` value anyway instead of the `isPublicRoom` one, right?

bmarty: Yes, this is what I am saying :)

> Instead of the new field isPublicRoom, the join rule will have to be provided

jmartinesp:
@frebib I haven't forgotten about your comment, but I think it's better if we implement that in a different PR.

Issue #6029, opened 2026-01-16
[RC v26.01.1]  - Can't send attachments if "Select media quality per upload is enabled"


bblacher:
### Steps to reproduce

Sending any attachment from within the app (image or document - doesn't matter) doesn't work. Pressing the "send arrow" just does nothing.

#### What did you expect?
The file/image should be sent.

#### What happened instead?
It wasn't sent.

#### Additional info
Sharing an attachment from another app into Element X works without issues.

### Your phone model

Pixel 8 Pro

### Operating system version

GrapheneOS - Android 16

### Application version and app store

v26.01.1

### Homeserver

Synapse v1.145.0

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

bblacher:
Update: this only appears if "Select media quality per upload is enabled"

Susurrus:
I'm also seeing this on 26.01.0, however turning off this setting (Advanced Settings > Optimize image upload quality) on my end doesn't change that. Adding a caption and turning off the "Optimize image quality" setting when you're viewing the image to send does nothing.

My current workaround is to take a photo normally, then share it using the normal Android approach. That does work.

PR #6039, opened 2026-01-19
Keep the child state in `AttachmentsPreviewPresenter` up to date

jmartinesp:
## Content

Wrap `mediaOptimizationSelectorState` in `AttachmentsPreviewPresenter` in a `rememberUpdatedState` call so it's now a `State<*>` value and the lambdas / functions that use it can access the updated state.

## Motivation and context

This makes the `handleEvents` function capture the reference and be able to check the updated state, unblocking the media sending flow.

Fixes https://github.com/element-hq/element-x-android/issues/6029

## Screenshots / GIFs

https://github.com/user-attachments/assets/b0130832-062e-464d-9026-5da7b268024a

## Tests

1. In developer settings, enable 'select media quality per upload'.
2. In any room, send an image.
3. If the preview screen actually processes and sends the image, it's working.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist


jmartinesp:
> Should we make sure we use that for every presenter states?

Probably, yeah. I can try doing a clean-up PR after this one.

---

In the end, it seems like indiscriminately using this pattern can also be a problem, so we're not doing that follow-up PR.

PR #6045, opened 2026-01-20
Add suggestions section to InvitePeopleView

ganfra:
## Content
  - Add suggestions section to the Invite People screen showing recent DM contacts
  - Users can quickly select people they've recently chatted with when inviting to a room

  New API : 

  - MatrixClient.getRecentDirectRooms(): Returns a Flow<RecentDirectRoom> of recently visited DM rooms, enabling lazy iteration and early termination
  - BaseRoom.getDirectRoomMember(): Gets the other member in a DM room.

## Motivation and context
Feature parity with EX iOS.

## Screenshots / GIFs
Check recorded screenshots.

## Tests

 - Open a room and tap "Invite"
  - Verify suggestions section appears with recent DM contacts
  - Verify users already in the room are not shown in suggestions
  - Select a suggestion and verify it appears in selected users
  - Deselect and verify state updates correctly
  - Start a search and verify suggestions hide
  - Close search and verify suggestions reappear

## Tested devices

- [X] Physical
- [X] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/tests/uitests/src/test/snapshots/images/features.invitepeople.impl_InvitePeopleView_Day_0_en.png
+++ b/tests/uitests/src/test/snapshots/images/features.invitepeople.impl_InvitePeopleView_Day_0_en.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:070f37faf63ee4bb7acabfbe19943181b367cb53d23978efe9ca669d6051e04c
-size 9137
+oid sha256:267f997e9f23c0745f8ac27216c125da7eb78db9f7fa42e670e5e2b40b8eeec2
+size 37777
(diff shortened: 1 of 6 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RustBaseRoom.kt
```diff
@@ -112,6 +113,20 @@ class RustBaseRoom(
         }
     }
 
+    override suspend fun getDirectRoomMember(): RoomMember? = withContext(roomDispatcher) {
+        runCatchingExceptions {
+            if (info().isDm) {
+                innerRoom.membersNoSync().use { members ->
+                    members.nextChunk(members.len())
+                        ?.map(RoomMemberMapper::map)
+                        ?.firstOrNull { roomMember -> roomMember.userId != sessionId && roomMember.membership.isActive() }
```
jmartinesp, line 122: This seems like an overkill. Shouldn't it be find, then map instead of map everything, then find?

ganfra: It's a dm, so we should only have 2 members, also I wanted to avoid repeating the logic for `isActive`

jmartinesp: Ah, true. Sorry, I was thinking about a room with lots of members.

Issue #5656, opened 2025-10-31
[Task] Add an empty state for spaces where the user has admin access.


pixlwave:
Please see and discuss the details in the meta issue.

PR #6064, opened 2026-01-22
Add an empty state for the space screen if the user can modify its graph

jmartinesp:
## Content

- Adds an empty state view to `SpaceView` that prompts the user to create a new room in the space if they can.
- Modifies `ConfigureRoomPresenter` and entry points to take an optional initial `parentSpaceId` value.
- When loading the editable spaces in `ConfigureRoomPresenter`, also try selecting the initial parent space from the list if present.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/5656.

## Tests

With the FF enabled:

- Open an empty space where you're an admin.
- Check the empty state appears.
- If you tap on the create room options, you should be taken to the create room UI.
- The parent space should be automatically pre-selected.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/features/createroom/api/src/main/kotlin/io/element/android/features/createroom/api/CreateRoomEntryPoint.kt
+++ b/features/createroom/api/src/main/kotlin/io/element/android/features/createroom/api/CreateRoomEntryPoint.kt
@@ -15,14 +15,13 @@ import io.element.android.libraries.architecture.FeatureEntryPoint
 import io.element.android.libraries.matrix.api.core.RoomId
 
 interface CreateRoomEntryPoint : FeatureEntryPoint {
-    fun setIsSpace(isSpace: Boolean): CreateRoomEntryPoint
-    fun setParentSpace(parentSpaceId: RoomId): CreateRoomEntryPoint
+    interface Builder {
+        fun setIsSpace(isSpace: Boolean): Builder
+        fun setParentSpace(parentSpaceId: RoomId): Builder
+        fun build(): Node
+    }
 
-    fun createNode(
-        parentNode: Node,
-        buildContext: BuildContext,
-        callback: Callback,
-    ): Node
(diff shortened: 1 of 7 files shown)


features/createroom/api/src/main/kotlin/io/element/android/features/createroom/api/CreateRoomEntryPoint.kt
```diff
@@ -15,8 +15,10 @@ import io.element.android.libraries.architecture.FeatureEntryPoint
 import io.element.android.libraries.matrix.api.core.RoomId
 
 interface CreateRoomEntryPoint : FeatureEntryPoint {
+    fun setIsSpace(isSpace: Boolean): CreateRoomEntryPoint
```
ganfra: Use Builder pattern instead? It's a bit confusing to return a `CreateRoomEntryPoint` here

features/space/impl/src/main/kotlin/io/element/android/features/space/impl/root/SpaceView.kt
```diff
@@ -259,61 +266,90 @@ private fun SpaceViewContent(
...
+                item {
+                    LoadingMoreIndicator(eventSink = state.eventSink)
+                }
             }
         }
     }
 }
 
+@Composable
+private fun EmptySpaceView(onCreateRoomClick: () -> Unit) {
```
ganfra, line 333: I can't find the design on figma, do you have a link?

jmartinesp: I think it's https://www.figma.com/design/kcnHxunG1LDWXsJhaNuiHz/ER-145--Spaces-on-Element-X?node-id=1679-45152&m=dev. It's odd, I was pretty sure there were several screens with this state before, but now I could only find this one.

PR #6067, opened 2026-01-22
Remove obfuscation in proguard

jmartinesp:
## Content

Remove obfuscation in proguard/R8 except for enterprise packages.

## Motivation and context

We don't really need obfuscation and it makes reading stacktraces in rageshakes a challenge sometimes.

## Tests

Build and install a release version with this config. It should work fine and the logs should contain the class and function names.

Some package names (not the `io.element` ones) may be minified for optimizations.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/app/common-proguard-rules.pro
+++ b/app/common-proguard-rules.pro
@@ -66,3 +66,11 @@
 -dontwarn androidx.window.sidecar.SidecarInterface
 -dontwarn androidx.window.sidecar.SidecarProvider
 -dontwarn androidx.window.sidecar.SidecarWindowLayoutInfo
+
+# Also needed after AGP 8.13.1 upgrade, it seems like proguard is now more aggressive on removing unused code
+-keep,allowoptimization,allowshrinking class org.matrix.rustcomponents.sdk.** { *;}
+-keep,allowoptimization,allowshrinking class uniffi.** { *;}
+-keep,allowoptimization,allowshrinking class io.element.android.x.di.** { *; }
+
+# Keep Metro classes
+-keep,allowoptimization,allowshrinking class dev.zacsweers.metro.** { *; }


app/proguard-rules.pro
```diff
@@ -5,6 +5,15 @@
 # For more details, see
 #   http://developer.android.com/guide/developing/tools/proguard.html
 
+# Increase optimizations passes to 3
+-optimizationpasses 3
+
+# Keep, optimize and shrink anything that's not enterprise-related
+-keep,allowoptimization,allowshrinking class !io.element.android.enterpriseconfig.**,!io.element.android.enterprise.features.enterprise.impl.** { *; }
```
bmarty: Why do we want the option `keep` here? The previous `keep` statement (`io.element.android.x.di.**`) had a smaller scope?

jmartinesp: Otherwise everything not mentioned will be obfuscated by default.

This is what the comment above says, keep everything but the enterprise-related packages from being obfuscated, but allow optimization and shrinking if needed.

jmartinesp: We can also forbid proguard from obfuscating everything, but I thought it could be useful to obfuscate the enterprise classes, in case we want to keep them a bit more 'secret'. I can replace this with a simple `-dontobfuscate` or some similar call.

bmarty: OK, my bad, I've read the doc again and it's fine!
I am checking one last point about the enterprise modules.

app/build.gradle.kts
```diff
@@ -132,8 +132,23 @@ android {
...
                 enable = true
                 keepRules {
-                    files.add(File(projectDir, "proguard-rules.pro"))
+                    files.add(File(projectDir, "common-proguard-rules.pro"))
                     files.add(getDefaultProguardFile("proguard-android-optimize.txt"))
+
+                    // Depending on whether the app flavor is enterprise or not we want to use different proguard rules.
+                    val flavorProguardFile = if (isEnterpriseBuild) {
+                        // Custom rules for enterprise builds
+                        File(projectDir, "enterprise-proguard-rules.pro")
```
bmarty, line 141: Maybe create an empty `enterprise-proguard-rules.pro` file? Else I understand that a warning will be output.

jmartinesp: I can add an empty one that can be overridden per pipeline flavor.

jmartinesp: Actually, adding that file would defeat the purpose of the warning, since there would always be an empty one unless you did something very wrong.

I think it's ok if we see the warning when building the pro-based apps, unless you know the app you're building should have a custom one. It's just a way of saying 'hey, this app isn't using any custom proguard rule, so if it should... this is why!'

Issue #6070, opened 2026-01-22
History Sharing: room header decoration for `shared` encrypted rooms


richvdh:
Per https://github.com/element-hq/element-meta/issues/3085, we'd like to add a decoration to the header for `shared` and E2EE rooms

richvdh:
Need to use a different icon for `history_visibility: world_readable`

PR #6090, opened 2026-01-26
Show an icon in the room header for shared history

richvdh:
## Content

Add a decoration to the header for encrypted rooms with `history_visibility:
{shared|public}`.

## Motivation and context

Fixes: #6070

## Screenshots / GIFs

image

## Tests

- Enable "Share encrypted history with new members" developer option
- Go to an encrypted room with shared history visibility
- Profit

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 15

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/topbars/MessagesViewTopBar.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/topbars/MessagesViewTopBar.kt
@@ -114,7 +114,7 @@ internal fun MessagesViewTopBar(
                     Icon(
                         imageVector = CompoundIcons.History(),
                         tint = ElementTheme.colors.iconInfoPrimary,
-                        contentDescription = null,
+                        contentDescription = stringResource(CommonStrings.common_shared_history),
                     )
                 }
             }
--- a/libraries/ui-strings/src/main/res/values/localazy.xml
+++ b/libraries/ui-strings/src/main/res/values/localazy.xml
@@ -334,6 +334,7 @@ Reason: %1$s."</string>
   <string name="common_server_url">"Server URL"</string>
   <string name="common_settings">"Settings"</string>
   <string name="common_share_space">"Share space"</string>
+  <string name="common_shared_history">"New members see history"</string>
   <string name="common_shared_location">"Shared location"</string>
   <string name="common_shared_space">"Shared space"</string>
   <string name="common_signing_out">"Signing out"</string>


jmartinesp:
GH actions is being quite uncooperative today.

jmartinesp:
@richvdh I'll merge this now since everything seems to pass, unless there's anything else you want to add to the PR?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/topbars/MessagesViewTopBar.kt
```diff
@@ -108,6 +109,14 @@ internal fun MessagesViewTopBar(
                     }
                     else -> Unit
                 }
+
+                if (showSharedHistoryIcon) {
+                    Icon(
+                        imageVector = CompoundIcons.History(),
+                        tint = ElementTheme.colors.iconInfoPrimary,
+                        contentDescription = null,
```
bmarty: This icon should have a `contentDescription`, for accessibility. But I see that the other Icons do not have any, so this is wrong...

Also maybe a click on the icon should display a dialog explaining what is it's aim, because at first sight it may not be obvious for the user.

This remark is also applicable to the other icons that may be displayed here...

richvdh: Yes, I just followed the examples of the existing icons.

What exactly is the `contentDescription` used for, given "hover" isn't really a thing that exists in a mobile UI? Is it used in screenreaders, or similar?

(Following the example of the web design, the description should probably be "New members see history".)

> Also maybe a click on the icon should display a dialog explaining what is it's aim, because at first sight it may not be obvious for the user.

This icon is in the top bar, and a tap on the top bar takes you to the room info card. The intention is that we'll have the same icon, with words next to it, and that will help users associate the icon with a given meaning:

image

jmartinesp: > What exactly is the contentDescription used for, given "hover" isn't really a thing that exists in a mobile UI? Is it used in screenreaders, or similar?

Text to speech for visually impaired users, mostly. That could read like 'Room title: Room name - history visibility shared' or something similar.

richvdh:
@jmartinesp thanks for fixing this up! Seems like maybe we should add a `contentDescription`, so let's hold off merging for a bit

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/topbars/MessagesViewTopBar.kt
bmarty: We also used to have long click with tooltip for toolbar actions, but it seems that it needs to be implemented manually with compose.

Anyway, since the accessibility issue is not a new thing, I'd say that it's not blocking this PR from being merged. That's also why I approved it.

richvdh: well, I've added a "contentDescription", since that was easy.

bmarty: Thanks, but that's not that easy 😅 . The string has to be added to Localazy, else it will be removed during the next sync.
I have added it, with the key `a11y.shared_history`. So you should sync the strings (running `./tools/localazy/downloadStrings.sh`), and replace the string key on the codebase.

bmarty: WOW sorry, you actually did add the string to Localazy, my bad!

bmarty: I have renamed the key to `a11y.shared_history`, since this is our convention for strings that are used exclusively for accessibility.

richvdh: hrm. I used `common` because we're going to need the same string in the Room Info, per the screenshot above.

jmartinesp: That's fine, `a11y` is also shared.

richvdh: [I ended up renaming the key back to `common.shared_history`]

PR #6092, opened 2026-01-27
Display a confirmation dialog when ending a poll from the event bottom sheet

bmarty:
## Content

89fac40265b7d436f991016161aa32c4cbfe37b7: Display a confirmation dialog when ending a poll from the event bottom sheet

Other commits: cleanup of the `:messages:impl` module.

## Motivation and context

Closes https://github.com/element-hq/customer-success/issues/703

## Tests

- In a room, create a poll
- In the timeline long click on the poll title
- In the bottom sheet, select "End poll"
- A confirmation dialog is now displayed, and user can confirm or cancel the action.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl/attachments/AttachmentsPreviewPresenterTest.kt
+++ b/features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl/attachments/AttachmentsPreviewPresenterTest.kt
@@ -11,9 +11,6 @@
 package io.element.android.features.messages.impl.attachments
 
 import android.net.Uri
-import app.cash.molecule.RecompositionMode
-import app.cash.molecule.moleculeFlow
-import app.cash.turbine.test
 import com.google.common.truth.Truth.assertThat
 import io.element.android.features.messages.impl.attachments.preview.AttachmentsPreviewEvent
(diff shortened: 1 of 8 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
```diff
@@ -341,9 +342,30 @@ fun MessagesView(
         maxBottomSheetContentHeight = maxComposerHeightPx.toDp(),
     )
 
+    var endPollConfirmingEvent: TimelineItem.Event? by remember { mutableStateOf(null) }
```
jmartinesp, line 345: It's kind of odd to have this in the view, but since the alternative is refactoring the whole way we handle these timeline events, I think it's fine.

bmarty: Yes... I used the same strategy than [here](https://github.com/element-hq/element-x-android/blob/eb31505dc71e44624ead84e0f8cda82d709422ce/features/poll/api/src/main/kotlin/io/element/android/features/poll/api/pollcontent/PollContentView.kt#L93), and in `PollContentView` this is odder in my opinion since a Dialog is added to the composition from an Event content view...

PR #6108, opened 2026-01-29
Remove `runBlocking` in `ThreadedMessagesNode`

jmartinesp:
## Content

Instead of using `runBlocking`, create the presenter asynchronously and display the UI when it's ready.

## Motivation and context

This should help reducing the number of reported ANRs.

## Tests

- Open a thread in a room. Everything should still work and not jank.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/threads/ThreadedMessagesNode.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/threads/ThreadedMessagesNode.kt
@@ -19,6 +19,7 @@ import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.lifecycle.Lifecycle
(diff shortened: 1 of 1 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/threads/ThreadedMessagesNode.kt
```diff
@@ -130,7 +140,10 @@ class ThreadedMessagesNode(
         super.onBuilt()
         lifecycle.subscribe(
             onCreate = {
-                sessionCoroutineScope.launch { analyticsService.capture(room.toAnalyticsViewRoom()) }
+                sessionCoroutineScope.launch {
```
ganfra: Use the lifecycleScope?

jmartinesp:
> The code is a bit convoluted, but I guess we can live with it

Better than adding yet another placeholder navigation node, I guess 😓 .

PR #6119, opened 2026-01-30
Ensure http 429 are retried 3 times before failing.

bmarty:
## Content

Ensure http 429 are retried 3 times before failing.

## Motivation and context

Give a chance for a request which is rate limited to succeed.

Android equivalent of https://github.com/element-hq/element-x-ios/pull/5022

## Tests

- Create a space
- Monitor the network
- Add lots of existing room to this space
- Observe that requests that fail with error 429 are retried 3 times.

Tested OK here, but at the end the requests are failing

image

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
@@ -161,7 +161,7 @@ class RustMatrixClientFactory(
             .requestConfig(
                 RequestConfig(
                     timeout = 30_000uL,
-                    // retryLimit must be non-zero for the SDK to retry API calls when rate-limited.
+                    // retryLimit must be non-zero for the SDK to retry API calls in case of error (including 429 Too Many Requests error).
                     retryLimit = 3u,
                     // Use default values for the rest
                     maxConcurrentRequests = null,


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -161,7 +161,8 @@ class RustMatrixClientFactory(
             .requestConfig(
                 RequestConfig(
                     timeout = 30_000uL,
-                    retryLimit = 0u,
+                    // retryLimit must be non-zero for the SDK to retry API calls when rate-limited.
```
ganfra: If I'm not mistaken, this is not just about rate limited error, this is a global retry configuration, so it'll also retry other errors. I think we said it was ok, but maybe change the comment?

ganfra (approved):

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
bmarty: [53e4846](https://github.com/element-hq/element-x-android/pull/6119/commits/53e48462a4911b9c77c0229ca1ccbd21ec673533)

Issue #6126, opened 2026-02-02
Bug: Unable to get back to the space after room creation


mxandreas:
### Steps to reproduce

1. Navigate to a space.
2. Create a new room.
3. Try to navigate back to the space using the back button (arrow) at the top left.

#### What did you expect?

I get back to the space where I started the room creation flow.

#### What happened instead?

The last step of the room creation flow (Invite people) is shown (and pressing Skip will take me to the room again).

Note: it works fine if you start the flow from the room list (and regardless if you select space or not).

### Will you send logs?

Yes

### Are you willing to provide a PR?

Yes

mxandreas:
Confirm this now works as expeced.

PR #6134, opened 2026-02-04
Fix back navigation after creating a room in a space

jmartinesp:
## Content

Previously, this displayed the 'invite users to the room' screen instead of going back to the parent space screen.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/6126

## Tests

- Go into a space.
- Create a new room in it.
- Go back from the newly created room, either by a back gesture/nav button or with the back button in the top app bar.
- The parent space screen should be displayed.

Note: the parent space screen will display the previous state where the new room is not present. This is a different issue.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/features/space/impl/src/main/kotlin/io/element/android/features/space/impl/SpaceFlowNode.kt
+++ b/features/space/impl/src/main/kotlin/io/element/android/features/space/impl/SpaceFlowNode.kt
@@ -20,7 +20,6 @@ import com.bumble.appyx.core.modality.BuildContext
 import com.bumble.appyx.core.node.Node
 import com.bumble.appyx.core.plugin.Plugin
 import com.bumble.appyx.navmodel.backstack.BackStack
-import com.bumble.appyx.navmodel.backstack.operation.newRoot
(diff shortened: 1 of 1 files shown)


features/space/impl/src/main/kotlin/io/element/android/features/space/impl/SpaceFlowNode.kt
```diff
@@ -175,6 +176,7 @@ class SpaceFlowNode(
                 val callback = object : CreateRoomEntryPoint.Callback {
                     override fun onRoomCreated(roomId: RoomId) {
                         callback.navigateToRoom(roomId, emptyList())
+                        backstack.newRoot(NavTarget.Root)
```
ganfra: Would be cleaner to call pop maybe?

jmartinesp: I think this is equivalent, but sure, I can use `pop()`.

ganfra (approved):

PR #6160, opened 2026-02-09
Sort audio device by device type before sending the list to Element Call

bmarty:
## Content

Sort audio device by device type before sending the list to Element Call

## Motivation and context

This ensure that the default audio device (the first one of the list)

From our internal room:

>We encounter a flicker issue with the earpiece overlay.
>
>In the logs I see:
>2026-02-09T10:34:14.018Z 2 matrix setAvailableAudioDevices called from native: [{"id":"2","name":"Built-in earpiece","isEarpiece":true,"isSpeaker":false,"isExternalHeadset":false},{"id":"3","name":"Built-in speaker","isEarpiece":false,"isSpeaker":true,"isExternalHeadset":false}]
>
>which makes element call use the first as the default until a specific one is selected by the controlledAudioApi (setAudioDevice)
>
>This one is selected: {"id":"2","name":"Built-in earpiece","isEarpiece":true,"isSpeaker":false,"isExternalHeadset":false} (The earpiece)
>
>very short (0.3-0.4seconds) later i see the following log:
>2026-02-09T10:34:14.396Z 2 matrix setAudioDevice called from native 3
the os will set the ouput to number three. At this point we update the device back to {"id":"3","name":"Built-in speaker","isEarpiece":false,"isSpeaker":true,"isExternalHeadset":false}
>
>Is there any reason we sort them like that? (earpiece first)
>Is it very easy to change the order of that array?
>
>The semantics of "first available device is the default" is the pattern the browser uses anyways so part of me thinks is elegant to keep that behaviour.
>
>If it is hard to do this on android the alternative of course is to never start with the earpiece on the EC widget. (this is easy)
But its a bit off an odd behaviour since we choose sth specific from the available array...
Another option would be to never select a default. There just will be no selected device until android sends us one.

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewAudioManager.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewAudioManager.kt
@@ -82,6 +82,13 @@ class WebViewAudioManager(
         AudioDeviceInfo.TYPE_BUILTIN_EARPIECE,
     )
 
+    private val audioDeviceComparator = Comparator<AudioDeviceInfo> { a, b ->
+        // If the device type is not in the wantedDeviceTypes list, we give it a high index, (i.e. low priority)
+        val indexOfA = wantedDeviceTypes.indexOf(a.type).let { if (it == -1) Int.MAX_VALUE else it }
+        val indexOfB = wantedDeviceTypes.indexOf(b.type).let { if (it == -1) Int.MAX_VALUE else it }
+        indexOfA.compareTo(indexOfB)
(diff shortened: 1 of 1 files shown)


toger5:
This fixes the issue: https://github.com/element-hq/element-call/issues/3729
And will allow to release v0.17.0 of element call.

This needs to be merged before EX android uses v0.17.0 embedded of EC which will be released once it made it through the release process: https://github.com/element-hq/voip-internal/issues/441

PR #6165, opened 2026-02-10
Use correct bg color.

bmarty:
Use correct bg color for the floating action button.

See update of color in https://www.figma.com/design/G1xy0HDZKJf5TCRFmKb5d5/Compound-Android-Components?node-id=2027-2043


jmartinesp (approved):
Thanks! There should not be any screenshot test changes because the underlying tokens are still the same, right?

bmarty:
> Thanks! There should not be any screenshot test changes because the underlying tokens are still the same, right?

This is what I believe, the CI will confirm.

Issue #6163, opened 2026-02-09
Crash when opening a room containing a malformed image message


kaeipnos:
### Steps to reproduce

I was debugging a bot using matrix nio library and was messing around trying to send malformed image messages. 

Using the room send function, I sent a message to a room with msgtype as m.image, an url with the uploaded image, and with a malformed info field with a width incorrectly set at 100000. The message was successfully accepted by the synapse server.

Something like:
```
await c.room_send(room_id="!YjIJYzUVDLiqCbbwjZ:...", message_type="m.room.message", content={"msgtype": "m.image", "
        ⋮ body": '', 'url':'mxc://.../cEudqDX...sqTUvkb', 'info':{'size':78911, 'mimetype':'image/png', 'w':100000, 'h'
        ⋮ :10}})

```

When opening the room in element x on android (26.01.2) containing this message, the app immediately crashed. 

### Outcome

I expected element x to not crash, and show that this message is not supported. 

Also, I think this would allow an attacker to DOS a room with write access for everyone using only element x as a client.

### Operating system version

android 16

### Application version and app store

26.01.2

### Homeserver

mine

### Will you send logs?

Yes

### Are you willing to provide a PR?

Yes

jmartinesp:
Thanks, I was able to reproduce it. The issue seems to be related to Compose and the layout constraints... again.

The whole stacktrace seems to be (I hope GH shrinks it a bit):

```
Thread: main, Exception: java.lang.IllegalArgumentException: Can't represent a width of 9900000 and height of 990 in Constraints
     	at androidx.compose.ui.unit.ConstraintsKt.throwInvalidConstraintException(Constraints.kt:401)
     	at androidx.compose.ui.unit.ConstraintsKt.createConstraints(Constraints.kt:425)
     	at androidx.compose.ui.unit.Constraints$Companion.fixed-JhjzzOo(Constraints.kt:225)
     	at androidx.compose.foundation.layout.AspectRatioNode.measure-3p2s80s(AspectRatio.kt:115)
     	at androidx.compose.ui.node.LayoutModifierNodeCoordinator.measure-BRTryo0(LayoutModifierNodeCoordinator.kt:190)
     	at androidx.compose.foundation.layout.SizeNode.measure-3p2s80s(Size.kt:865)
     	at androidx.compose.ui.node.LayoutModifierNodeCoordinator.measure-BRTryo0(LayoutModifierNodeCoordinator.kt:190)
[...]

PR #6168, opened 2026-02-10
Ensure aspect ratio of images in the timeline is restricted

jmartinesp:
## Content

Make sure we `coerce` the possible values for image widths and heights and generated aspect ratio values to something reasonable.

Otherwise, this could cause a crash in Compose since the width and height values could become way too large.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/6163

## Tests

Kind of difficult to test, you need to send an image event with incorrect width/height values. You can test it by maybe copying an existing image event, modifying it and sending it as a new one.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentMessageFactory.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentMessageFactory.kt
@@ -50,6 +50,11 @@ import kotlinx.collections.immutable.toImmutableList
 import org.jsoup.nodes.Document
 import kotlin.time.Duration
 
+private const val MIN_IMAGE_SIZE = 1L
+private const val MAX_IMAGE_SIZE = 10_000L
+private const val MIN_ASPECT_RATIO = 0.001f
+private const val MAX_ASPECT_RATIO = 10f
+
 @Inject
 class TimelineItemContentMessageFactory(
     private val fileSizeFormatter: FileSizeFormatter,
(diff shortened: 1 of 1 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentMessageFactory.kt
```diff
@@ -83,7 +83,10 @@ class TimelineItemContentMessageFactory(
                 val dom = messageType.formattedCaption?.toHtmlDocument(permalinkParser = permalinkParser)
                 val formattedCaption = dom?.let(::parseHtml)
                     ?: messageType.caption?.withLinks()
-                val aspectRatio = aspectRatioOf(messageType.info?.width, messageType.info?.height)
+                // Coerce the image sizes and prevent invalid aspect ratios, which can cause crashes
+                val width = messageType.info?.width?.coerceIn(1, 10_000)
```
ganfra: Use constants too?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemContentMessageFactory.kt
```diff
@@ -83,7 +83,10 @@ class TimelineItemContentMessageFactory(
...
+                val height = messageType.info?.height?.coerceIn(1, 10_000)
+                val aspectRatio = aspectRatioOf(width, height)?.coerceAtMost(10f)
```
ganfra: why not coerceIn here?

jmartinesp: If width and height are already in 1..10000, there's no way the min value is `<= 0`, AFAICT.

PR #6174, opened 2026-02-11
Filter out MSC4171 service_members for DM rooms

### LINKED ISSUE #4034: https://github.com/element-hq/element-x-android/blob/b2a9ebb2d25da51a13627cca06ef7e86a4697ab8/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/RoomIsDmCheck.kt#L18
Have DM rooms (2 users + a bridge bot) show up in the `People` filter just like they do in other Matrix clients

Because bridged users are people too. It provides a consistent (and more importantly) intuitive experience

Adjust the `isDm()` check here: https://github.com/element-hq/element-x-android/blob/b2a9ebb2d25da51a13627cca06ef7e86a4697ab8/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/RoomIsDmCheck.kt#L18
to use the recently implemented `service_members` functionality in Rust SDK added in https://github.com/matrix-org/matrix-rust-sdk/pull/4335. Whilst https://github.com/matrix-org/matrix-rust-sdk/pull/4335 fixes the room name for bridged rooms, it doesn't expose the required information through the SDK FFI layer for the app to know which users in a room are service members.

I'm asking here if this seems like a reasonable thing to do given how Element wants to frame DM chats, and whether approach is considered acceptable. I'm happy to attempt some of the implementation for this

I have a hacky patch that achieves roughly the same thing, but it's awful and not a good way to solve this https://github.com/frebib/element-x-android/commit/e5a5acc1b6a82954c8036a73b6819db3df3aa029

### Additional context

_No response_

Yes

frebib:
Fixes avatars for DM rooms with a 3rd user that is marked as a service member. Also fixes avatars for shortcuts to bridge rooms.

Depends on https://github.com/matrix-org/matrix-rust-sdk/pull/6159
 
## Content

FIXME

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/4034

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomInfoExt.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomInfoExt.kt
@@ -18,7 +18,8 @@ import org.matrix.rustcomponents.sdk.RoomInfo
  */
 fun RoomInfo.elementHeroes(): List<MatrixUser> {
     return heroes
-        .takeIf { isDirect && activeMembersCount.toLong() == 2L }
+        .takeIf { isDirect && (activeMembersCount - serviceMembers.size.toULong()) == 2UL }
+        ?.filter { !serviceMembers.contains(it.userId) }
         ?.takeIf { it.size == 1 }
         ?.map { it.map() }
         .orEmpty()
--- a/features/space/impl/src/main/kotlin/io/element/android/features/space/impl/root/SpaceStateProvider.kt
+++ b/features/space/impl/src/main/kotlin/io/element/android/features/space/impl/root/SpaceStateProvider.kt
@@ -119,6 +119,7 @@ private fun aSpaceInfo(
         activeMembersCount = 5,
         invitedMembersCount = 0,
         joinedMembersCount = 5,
+        serviceMembers = persistentListOf(),
         roomPowerLevels = null,
         highlightCount = 0,
         notificationCount = 0,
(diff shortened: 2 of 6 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomInfoExt.kt
```diff
@@ -18,7 +18,8 @@ import org.matrix.rustcomponents.sdk.RoomInfo
  */
 fun RoomInfo.elementHeroes(): List<MatrixUser> {
     return heroes
-        .takeIf { isDirect && activeMembersCount.toLong() == 2L }
+        .takeIf { isDirect && (activeMembersCount - serviceMembers.size.toULong()) == 2UL }
```
frebib, line 21: So this has the possibility to be inaccurate, unfortunately. There's no guarantee that the users listed in `serviceMembers` are joined to the room (so wouldn't be accounted for in `activeMembersCount`). It's a big guess. Avoiding this check would obviously solve this. Also getting the full member list would also work. The last option I'm considering is having the homeserver calculate this number correctly in the sync response so the app doesn't need to

frebib:
The other major issue with this is that it's still broken for the room list in space view. The `SpaceRoom` struct doesn't contain service members, and it almost seems silly to add it there too just so we can duplicate this logic. Ideally the homeserver would filter out these service members from heroes in sync, but according to MSC4171 that's not mandatory, so rust-sdk should probably do this for us. Raised https://github.com/matrix-org/matrix-rust-sdk/issues/6176 to do that

bmarty:
Some changes have been done around the definition of "DM", this PR needs to be re-created but I believe that the issue it is fixing is gone.

PR #6185, opened 2026-02-11
Remove `NavigationState.Space`.

bmarty:
We do not have the concept of current space, this code was created when the Space was not a thing in EXA.


jmartinesp:
That said we technically *can* use these APIs now since we can open spaces, I guess we can just remove them because they aren't useful at the moment?

bmarty:
> That said we technically _can_ use these APIs now since we can open spaces, I guess we can just remove them because they aren't useful at the moment?

Yes, the Space state is unused. We mainly use the application state to filter notifications, and removing the Space state is simplifying the code. Another PR is coming regarding notification.

PR #6187, opened 2026-02-12
Try to fix common issue with the top app bar

jmartinesp:
## Content

Make sure the size to measure the layout in `TopAppBarScrollBehaviorLayout` can't be negative.

## Motivation and context

The exception is:

> Size(720 x -{height}) is out of range. Each dimension must be between 0 and 16777215.

So forcing the height to at least 0 should fix it (although it might cause UI glitches?)

## Tests

To be honest, I don't know how to test this since I could never reproduce the issue to begin with.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist


ganfra (approved):
LGTM I guess!

PR #6193, opened 2026-02-12
Remove `runBlocking` call to restore sessions when the app starts

jmartinesp:
## Content

This should take care of a few ANRs, although it may make loading the initial state a bit slower (not that I noticed that myself).

This is touching the most critical part of the app, so we need to be extra careful when testing it.

## Motivation and context

This `runBlocking` was a constant source of ANR issues. With this maybe the ANR rate will drop significantly.

## Tests

- Start the app. The logged in UI should load just fine.
- Open a different app and use `adb shell am kill io.element.android.x.debug`. The logs in AS should display the app has been killed.
- Reopen the app, the previous session should be restored shortly after.

I also tried this 'kill the app, then open it' flow with a share intent, so it seems like it can work on different scenarios and not just the base 'app opened' scenario.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/RootFlowNode.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/RootFlowNode.kt
@@ -154,7 +154,8 @@ class RootFlowNode(
                         switchToNotLoggedInFlow(null)
                     }
                 }
-            }.launchIn(lifecycleScope)
+            }
+            .launchIn(lifecycleScope)
     }
 
     /**


jmartinesp:
FWIW I've been using a release version with the latest changes of this branch for the past couple of days and I've had no issues with restoration either when coming back to the app through the launcher, or when using incoming share or opening notifications.

jmartinesp:
I've been using this for a week in a day-to-day scenario and had no issues, so it seems fine to merge this if we're comfortable with the changes.

appnav/src/main/kotlin/io/element/android/appnav/RootFlowNode.kt
```diff
@@ -134,6 +150,41 @@ class RootFlowNode(
         }.launchIn(lifecycleScope)
     }
 
+    /**
+     * Restore the saved state for navigation in the current backstack.
+     *
+     * **WARNING:** this is an unsafe operation abusing the internals of the Appyx library, but it's the only way allow async state
```
bmarty, line 164: Maybe add a this warning in a comment above the line where Appyx version is set ([here](https://github.com/element-hq/element-x-android/blob/develop/gradle/libs.versions.toml#L45)) too? Else I am pretty sure we may miss to check this during the next Appyx lib upgrade (if there is any)

PR #6194, opened 2026-02-12
request audio focus when recording voice messages

### LINKED ISSUE #5997: Recording an audio message should pause the current media playing
### Your use case

When a media is playing in the background (e.g. listening to some music on deezer), starting to record an audio message should pause the media playback, and resume it after the recording is done. This is the behavior of WhatsApp and Signal, and I find it superior to the behavior of EXA, which is to let the playback in the background.

nope

### Additional context

_No response_

No :(

vmfunc:
recording a voice message now requests audio focus; background media
(spotify, deezer, etc.) gets paused while recording and resumes when
you stop. matches the behavior of signal and whatsapp.

the AudioFocus infrastructure was already in place for playback
(DefaultMediaPlayer uses it), just wasn't wired into the recording path.

fixes #5997

### DIFF
--- a/libraries/audio/impl/src/main/kotlin/io/element/android/libraries/audio/impl/DefaultAudioFocus.kt
+++ b/libraries/audio/impl/src/main/kotlin/io/element/android/libraries/audio/impl/DefaultAudioFocus.kt
@@ -100,10 +100,8 @@ private fun AudioFocusRequester.willPausedWhenDucked(): Boolean {
     return when (this) {
         // (note that for Element Call, there is no action when the focus is lost)
         AudioFocusRequester.ElementCall,
-        AudioFocusRequester.VoiceMessage -> true
-        // no audio output to duck when recording, and we don't want notification
-        // sounds to interrupt a recording via transient focus loss
-        AudioFocusRequester.RecordVoiceMessage -> false
+        AudioFocusRequester.VoiceMessage,
+        AudioFocusRequester.RecordVoiceMessage -> true
         // For the MediaViewer, we let the system automatically handle the ducking
         // https://developer.android.com/media/optimize/audio-focus#automatic-ducking
         AudioFocusRequester.MediaViewer -> false


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/voicemessages/composer/DefaultVoiceMessageComposerPresenter.kt
```diff
@@ -246,19 +249,23 @@ class DefaultVoiceMessageComposerPresenter(
 
     private fun CoroutineScope.startRecording() = launch {
         try {
+            audioFocus.requestAudioFocus(AudioFocusRequester.VoiceMessage) {}
```
bmarty: In the case of a voice recording, `willPausedWhenDucked` should maybe return `false`?
Maybe create a new `AudioFocusRequester.RecordVoiceMessage` to handle this case.
Also the second parameter `onFocusLost` should maybe do something like stopping the recording?

Maybe worth checking the behavior when a voice message is getting recorded and an audio stream is (re)started (from a notification for instance).

vmfunc: yeah good call on all of these, pushed two more commits.

added `RecordVoiceMessage` as a separate requester with `willPausedWhenDucked = false` - no audio output to duck when recording, and this way notification sounds won't trigger transient focus loss and kill the recording. only actual focus loss (phone call etc) stops it.

`onFocusLost` now finishes the recording and leaves the partial result in preview state so nothing gets lost.

haven't tested the notification audio restart on device yet but with `willPausedWhenDucked = false` those events shouldn't reach us at all.

libraries/audio/impl/src/main/kotlin/io/element/android/libraries/audio/impl/DefaultAudioFocus.kt
```diff
@@ -99,6 +101,9 @@ private fun AudioFocusRequester.willPausedWhenDucked(): Boolean {
         // (note that for Element Call, there is no action when the focus is lost)
         AudioFocusRequester.ElementCall,
         AudioFocusRequester.VoiceMessage -> true
+        // no audio output to duck when recording, and we don't want notification
+        // sounds to interrupt a recording via transient focus loss
+        AudioFocusRequester.RecordVoiceMessage -> false
```
bmarty: I am always beaten by this API. Reading the [doc](https://developer.android.com/media/optimize/audio-focus#automatic-ducking) again I think we need to return `true`, in order to have the callback invoked.

vmfunc: changed it to true and grouped it with the existing VoiceMessage

vmfunc:
hey @bmarty seems ready for review - codecov fails but after digging i doubt it comes from my end

vmfunc:
sorry for bumping this. any issues with this PR?

bmarty:
> sorry for bumping this. any issues with this PR?

Sorry for the delay, I was off last week. Thanks again for the PR!

vmfunc:
all good, thank you!

PR #6195, opened 2026-02-12
Fix call button color and ensure call can always be declined from the notification

bmarty:
## Content

Use correct color for the incoming call screen buttons. Also improve the UI rendering in wide screen / landscape.

Ensure that the call can be declined from the notification. It was not working in some cases.

## Motivation and context

Nicer UI, working notification action

## Screenshots / GIFs

In landscape, the incoming call screen is less ugly

|Before|After|
|-|-|
|Screenshot_20260212_160505|Screenshot_20260212_160945|

## Tests

- Make sure the phone is not locked, the screen is on.
- Receive an Element Call
- When the notification to accept/decline the call is displayed
- From the task switcher, kill Element X
- Attempt to decline the call from the notification.

If the call is declined and the notification is dismissed, the bug is fixed.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/call/impl/src/test/kotlin/io/element/android/features/call/utils/DefaultActiveCallManagerTest.kt
+++ b/features/call/impl/src/test/kotlin/io/element/android/features/call/utils/DefaultActiveCallManagerTest.kt
@@ -199,6 +199,34 @@ class DefaultActiveCallManagerTest {
         }
     }
 
+    @Test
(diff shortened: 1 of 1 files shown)


jmartinesp:
God, it looked terrible in landscape. Now it's better, but it still feels weird, maybe we should ask design to enforce portrait mode here 😅 .

PR #6212, opened 2026-02-16
Add free disk space action to CI actions

jmartinesp:
## Content

Add `jlumbroso/free-disk-space` action to free up space.

- With `large-packages` enabled (removal of these enabled), 14GB of disk space are saved, but this can take 1-2 minutes to run in every GH flow.
- With `large-packages` disabled, we save 8.7GB of disk space, which seems good enough and this takes ~10s.

In the end we went with option 2 for speed, 8.7 extra free GB seem good enough for now.

## Motivation and context

I don't want to see another 'no space left on device' message ever again.

## Tests

Tested with everything but Android and Swap removed and it worked, but was slow. Then tried with `large-packages` and it seemed good enough.

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,14 +28,14 @@ jobs:
       - name: Free Disk Space (Ubuntu)
         uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be
         with:
-          # this might remove tools that are actually needed,
-          # if set to "true" but frees about 6 GB
+          # This might remove tools that are actually needed, if set to "true" but frees about 6 GB
           tool-cache: true
-          # all of these default to true, but we should only need the 'android' one (and maybe swap-storage?)
+          # All of these default to true, but we should only need the 'android' one (and maybe swap-storage?)
           android: false
           dotnet: true
           haskell: true
-          large-packages: true
+          # This takes way too long to run (~2 minutes) and it saves only ~5.5GB
+          large-packages: false
           docker-images: true
           swap-storage: false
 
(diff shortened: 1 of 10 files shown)


.github/workflows/build.yml
```diff
@@ -25,6 +25,20 @@ jobs:
       group: ${{ github.ref == 'refs/heads/develop' && format('build-develop-{0}-{1}', matrix.variant, github.sha) || format('build-{0}-{1}', matrix.variant, github.ref)  }}
       cancel-in-progress: true
     steps:
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be
```
ganfra, line 29: Why not using @main ?

.github/workflows/build.yml
```diff
@@ -25,6 +25,20 @@ jobs:
...
+        with:
+          # This might remove tools that are actually needed, if set to "true" but frees about 6 GB
+          tool-cache: true
+          # All of these default to true, but we should only need the 'android' one (and maybe swap-storage?)
```
ganfra, line 33: The comment is weird, "should only need the android one" but then it's `android:false`

jmartinesp: The logic is inverted: `true` means remove, `false` means keep. Yeah, I set them all backwards the first time 😅 .

.github/workflows/build.yml
jmartinesp: Sonar warned us (correctly) that using a branch or even a tag is dangerous, as someone could compromise the repo and upload one or several commits and we'd just run whatever code it contains without realising it.

PR #6215, opened 2026-02-17
Limit the max number of opened rooms in the backstack

jmartinesp:
## Content

When opening a new room in `LoggedInFlowNode`, make sure we discard old room navigation nodes so we don't end up with an infinite back stack of nodes when using permalinks (Room A -> Room B -> Room A -> Room B -> ...).

This should help with the `TransactionTooLargeExceptions` we are seeing, since every one of these nodes and their sub-nodes would be saved to the instance state.

Also, make sure we use `LoggedInFlowNode.attachRoom` as much as possible to ensure this check is used.

## Motivation and context

Fix crash that usually happens when backgrounding the app.

## Tests

It's difficult to test other than having a permalink loop as described above and after going back and forth in those rooms for 10-15 iterations checking the app doesn't crash after backgrounding it.

Without this, that would cause a crash most of the time.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist


appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -352,7 +353,9 @@ class LoggedInFlowNode(
                     }
 
                     override fun navigateToRoomSettings(roomId: RoomId) {
-                        backstack.push(NavTarget.Room(roomId.toRoomIdOrAlias(), initialElement = RoomNavigationTarget.Details))
+                        sessionCoroutineScope.launch {
+                            backstack.push(NavTarget.Room(roomId.toRoomIdOrAlias(), initialElement = RoomNavigationTarget.Details))
```
ganfra: Use `attachRoom`?

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -323,12 +323,13 @@ class LoggedInFlowNode(
             NavTarget.Home -> {
                 val callback = object : HomeEntryPoint.Callback {
                     override fun navigateToRoom(roomId: RoomId, joinedRoom: JoinedRoom?) {
-                        backstack.push(
-                            NavTarget.Room(
+                        sessionCoroutineScope.launch {
```
ganfra: Use the `nodeLifecycle` instead?

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -673,16 +691,50 @@ private class AttachRoomOperation(
                 val roomNavTarget = it.key.navTarget as? LoggedInFlowNode.NavTarget.Room
                 roomNavTarget?.roomIdOrAlias == roomTarget.roomIdOrAlias
             }
+
+            // Make sure the backstack of rooms can't grow indefinitely when opening permalinks.
+            // Having 5 rooms in the backstack seems reasonable and shouldn't grow the saved state size too much.
+            val maxRoomCount = 5
```
ganfra: Define as constant?

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -673,16 +691,50 @@ private class AttachRoomOperation(
...
+            val roomElementCount = elements.count { it.key.navTarget is LoggedInFlowNode.NavTarget.Room }
+
+            Timber.d("Current room nodes: $roomElementCount/$maxRoomCount")
+            val currentElements = if (roomElementCount + 1 > maxRoomCount) {
```
ganfra: Maybe could create a small extension method instead?
```
   fun <T> List<T>.takeLast(n: Int, predicate: (T) -> Boolean): List<T> {
        val matchingIndices = indices.filter { predicate(this[it]) }
        val indicesToRemove = matchingIndices.dropLast(n).toSet()
        return filterIndexed { index, _ -> index !in indicesToRemove }
    }
```

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
jmartinesp: We can't really use it unless we refactor `attachRoom` to take an `initialElement` instead of the event id, and that means having to do extra checks in the method.

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/6215/commits/9aabffceae89ff24b25ec36cabd7452bf3171680

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/6215/commits/6ec06a188aae5d0ca641b793422dac5a3893a97e as well

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/6215/commits/6ec06a188aae5d0ca641b793422dac5a3893a97e

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
ganfra: Then remove the coroutine scope here?

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/collections/ListExt.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright (c) 2026 Element Creations Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.androidutils.collections
+
+/**
+ * Keep all the items in the list except for the last [count] items that match [predicate].
```
ganfra: Mmm, it's more 
`/*
Returns a list containing last [count] elements that match [predicate] while preserving other elements. ` 
And so I don't understand your naming

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -673,16 +716,39 @@ private class AttachRoomOperation(
...
+            // Make sure the backstack of rooms can't grow indefinitely when opening permalinks.
+            val roomElementCount = elements.count { it.key.navTarget is LoggedInFlowNode.NavTarget.Room }
+
+            Timber.d("Current room nodes: $roomElementCount/$MAX_ROOM_NODE_COUNT")
+            val currentElements = if (roomElementCount + 1 > MAX_ROOM_NODE_COUNT) {
+                elements.takeExceptLast(MAX_ROOM_NODE_COUNT) { element ->
+                    element.key.navTarget is LoggedInFlowNode.NavTarget.Room
+                }
+                    // Then reverse the order again after removing the extra room nodes
+                    .asReversed()
```
ganfra: Remove the reversal I think

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
jmartinesp: I ended up doing the refactor, I was a bit worried that something may break but as far as I can tell it's working fine: https://github.com/element-hq/element-x-android/pull/6215/commits/9dc8fc73a86068f8a476d53e21f7ba7e54f57868

Issue #6209, opened 2026-02-16
Simplify and improve the notification fetching flow


jmartinesp:
## The issue

When we receive a push notification for a message in a room we need to actually fetch the associated event from the homeserver, since the notification itself only contains the associated room id and event id to ensure we don't disclose any private information. Until recently, we did that immediately after the notification was received, assuming that if we were able to receive a notification, the app would have network access and be able to fetch the events even if it was on background. However:

- This started failing for FCM on Android 15, which introduced more limits to background processing, specially for internet connectivity. This *shouldn't* happen, as a received FCM message should wake the app with network access, but it seems like something wasn't working as expected.
- For UnifiedPush on Android 15, the OS doesn't have this mechanism for detecting a new push has been received, so they created their own workaround, consisting in launching the registered app in a temporary foreground state. Note this can still mean the app is foregrounded but has no internet connectivity.

To try to fix this, we switched to using `WorkManager` to schedule notification event fetching, since `Worker` instances are supposed to have network connectivity if launched with the right constraints. However, this also means fetching notifications may not be as immediate as it used to, but it should be more reliable, although in general they're launched just after they're scheduled, since the app is temporary in foreground.

However, there are 2 issues with `WorkManager` at the moment:
1. You are only guaranteed to get network access if you request it as a constraint when building the worker request. However, this constraint makes the OS check the url `https://google.com/generate_204` to understand if it has internet connectivity and blocks VPN usage by default. While the VPN check can (in theory) be manually disabled, the internet connectivity one can be tricky, specially in air-gapped environments.
2. While adding the network connectivity request in theory should be enough to ensure the worker only runs when we can access the internet, it turns out there is [an issue](https://issuetracker.google.com/issues/445324855#comment28) in WorkManager at the moment that starts the worker when connectivity isn't available.

Besides that, there are other associated issues to the current notification fetching approach:
- The use workmanager/legacy flow behaviour behind a feature flag, so we have the same notification fetching code twice.
- The retry mechanism is too simple, and could potentially mean workers being retried indefinitely, although we haven't seen that in the wild.
- Since we wanted to keep the API similar for the previous approach and the WorkManager based one, we pass too much data around, which is poorly grouped and contains duplicate info.
[...]

PR #6219, opened 2026-02-18
Add max retry checks for `FetchNotificationsWorker`

jmartinesp:
## Content

Add a maximum of 3 retries for fetching notifications with `WorkManager`.

## Motivation and context

Not having this can potentially result in a work request being re-scheduled indefinitely.

Part of https://github.com/element-hq/element-x-android/issues/6209

## Tests

I don't think this can be easily tested.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/workmanager/FetchNotificationsWorker.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/workmanager/FetchNotificationsWorker.kt
@@ -24,6 +24,8 @@ import io.element.android.libraries.core.extensions.runCatchingExceptions
 import io.element.android.libraries.di.annotations.ApplicationContext
 import io.element.android.libraries.matrix.api.auth.SessionRestorationException
 import io.element.android.libraries.matrix.api.core.SessionId
+import io.element.android.libraries.matrix.api.exception.ClientException
+import io.element.android.libraries.matrix.api.exception.isNetworkError
 import io.element.android.libraries.push.api.push.NotificationEventRequest
 import io.element.android.libraries.push.api.push.SyncOnNotifiableEvent
 import io.element.android.libraries.push.impl.notifications.NotifiableEventResolver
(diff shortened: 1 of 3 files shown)


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/workmanager/FetchNotificationsWorker.kt
```diff
@@ -53,22 +56,27 @@ class FetchNotificationsWorker(
...
         if (!hasNetwork) {
-            Timber.w("No network, retrying later")
-            return@withContext Result.retry()
+            if (workerParams.runAttemptCount < MAX_RETRY_ATTEMPTS) {
+                Timber.tag(TAG).w("No network, retrying later")
+                return@withContext Result.retry()
+            } else {
+                Timber.tag(TAG).w("No network available and reached max retry attempts (${workerParams.runAttemptCount}/$MAX_RETRY_ATTEMPTS)")
+                return@withContext Result.failure()
+            }
```
ganfra: Maybe if there is no network we should not count as a retry?

jmartinesp: I don't think we can make an exception just for this case, to be honest 🫤 .

ganfra: Maybe we could let the worker wait for "network" instead of retrying?

jmartinesp: I don't think the workers can be alive in background for long. We could try, but I'm sure that would have some kind of penalty with the future scheduling.

jmartinesp: Maybe we can make the retry backoff exponential instead of linear: my thoughts on this was to have it linear so it's retried after a short while, but I was thinking about a temporary issue in the HS, not the connection failing.

jmartinesp:
Closing this for https://github.com/element-hq/element-x-android/pull/6258

PR #6231, opened 2026-02-20
Disable the cross-process lock in the SDK

jmartinesp:
## Content

What the title says.

## Motivation and context

We don't use multiple process as iOS does, so we don't need it. It should improve DB performance a bit and overall waste less resources.

## Tests

Use the app normally, check you can receive push notifications and they get processed as usual. Battery usage should decrease too, but that's a bit more difficult to measure.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/ClientBuilderProvider.kt
```diff
@@ -20,5 +21,7 @@ interface ClientBuilderProvider {
 class RustClientBuilderProvider : ClientBuilderProvider {
     override fun provide(): ClientBuilder {
         return ClientBuilder()
+            // Make sure all built clients use the single process cross-process lock config
+            .crossProcessLockConfig(CrossProcessLockConfig.SingleProcess)
```
bmarty: `RustClientBuilderProvider` is a way to hide the usage of `org.matrix.rustcomponents.sdk.ClientBuilder`.

I'd move the configuration in `RustMatrixClientFactory.getBaseClientBuilder` instead.

PR #6258, opened 2026-02-26
Simplify push notification flow by using locally stored values for pending pushes

jmartinesp:
## Content

- Creates new `PushRequest` to replace `NotificationEventRequest`. These will be stored in the push history DB.
- Rework a bit the `WorkManagerWorkRequest` class and changed it to `WorkManagerRequestBuilder`. Its `build` function is now suspend, and it returns a new `WorkManagerRequestWrapper` that contains info about the request and how it should be enqueued.
- Replace `FetchNotificationsWorker` with `FetchPendingNotificationsWorker`, which queries the pending notifications from the DB instead of receiving them from a parameter. This ensures we're always querying the latest values. Only a single one of these workers can run at a time for each session id to avoid data races.
- Modify `PushHistoryService` to allow adding, replacing, querying and removing `PushRequest` items.
- Split `PushHandler` into a new `NotificationResultProcessor` component, so the handler only handles the receiving part of the push flow and the new component handles the post-processing for the notification results.

## Motivation and context

Addresses several points in https://github.com/element-hq/element-x-android/issues/6209:

- How we group the data.
- How we retry the work requests.
- Removes the previous flow to directly fetch notifications without WorkManager.

## Tests

Test push notifications in several scenarios, including receiving notifications while the device was offline.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist


libraries/workmanager/api/src/main/kotlin/io/element/android/libraries/workmanager/api/WorkManagerScheduler.kt
```diff
@@ -11,9 +11,21 @@ package io.element.android.libraries.workmanager.api
...
 
 interface WorkManagerScheduler {
-    fun submit(workManagerRequest: WorkManagerRequest)
+    /**
+     * Submits a new work request built from [workManagerRequestBuilder] to run in `WorkManager`.
+     */
+    suspend fun submit(workManagerRequestBuilder: WorkManagerRequestBuilder)
+
+    /**
+     * Checks if there are any pending requests scheduled for the provieded [sessionId] and [requestType].
```
bmarty:
```suggestion
     * Checks if there are any pending requests scheduled for the provided [sessionId] and [requestType].
```

libraries/push/impl/src/main/sqldelight/io/element/android/libraries/push/impl/db/PushRequest.sq
```diff
@@ -0,0 +1,25 @@
...
+selectAllPendingForSession:
+SELECT * FROM PushRequest WHERE status = 0 AND sessionId = ? AND pushDate > ? ORDER BY pushDate ASC;
+
+insertPushRequest:
+INSERT OR REPLACE INTO PushRequest VALUES ?;
+
+removeAll:
+DELETE FROM PushRequest;
+
+-- add query to keep only the last x entries
```
bmarty: Maybe remove (or update) the AI prompt 😛?

libraries/push/impl/src/main/sqldelight/migrations/1.sqm
```diff
@@ -0,0 +1,14 @@
...
+    providerInfo TEXT NOT NULL,
+    eventId TEXT NOT NULL,
+    roomId TEXT NOT NULL,
+    sessionId TEXT NOT NULL,
+    status INTEGER NOT NULL DEFAULT 0,
+    retries INTEGER NOT NULL DEFAULT 0,
+    PRIMARY KEY(sessionId, eventId)
+);
+
+CREATE INDEX PushRequestSessionAndStatus ON PushRequest (sessionId, status);
```
bmarty, line 14: Not for this PR, but we could add a database migration test as per for the session database.

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/history/PushHistoryService.kt
```diff
@@ -26,12 +29,33 @@ interface PushHistoryService {
...
+
+    /**
+     * Adds or replaces an existing [PushRequest] in the local database.
+     */
+    suspend fun enqueuePushRequest(pushRequest: PushRequest): Result<Unit>
+
+    /**
+     * Replace a list of [PushRequest] in the database.
+     */
+    suspend fun replacePushRequests(pushRequests: List<PushRequest>): Result<Unit>
```
bmarty: Maybe keep the database semantic and rename to `insertOrUpdatePushRequests`?

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/history/PushHistoryService.kt
```diff
@@ -26,12 +29,33 @@ interface PushHistoryService {
         includeDeviceState: Boolean,
         comment: String?,
     )
+
+    /**
+     * Adds or replaces an existing [PushRequest] in the local database.
+     */
+    suspend fun enqueuePushRequest(pushRequest: PushRequest): Result<Unit>
```
bmarty: Maybe keep the database semantic and rename to `insertOrUpdatePushRequest`?

libraries/push/impl/src/main/sqldelight/io/element/android/libraries/push/impl/db/PushRequest.sq
jmartinesp: I guess I should remove it from `PushHistory.sq` too, since I copied the base schema from there 😬 . I didn't even realise it was an LLM prompt, but now that you mention it it's pretty obvious 😅 .

libraries/push/impl/src/main/sqldelight/migrations/1.sqm
jmartinesp: I already added migration verification and used `./gradlew :libraries:push:impl:verifySqlDelightMigration`, I thought that checked exactly these kind of issues.

bmarty: We're supposed to have some db files (as per in https://github.com/element-hq/element-x-android/tree/develop/libraries/session-storage/impl/src/main/sqldelight/databases). The recipe is https://github.com/element-hq/element-x-android/blob/develop/libraries/session-storage/impl/build.gradle.kts#L41 for the other base.

bmarty: And yes we have some. Sorry I missed it!

jmartinesp: I think we may be missing the v1 DB file though 👀 .

jmartinesp: https://github.com/element-hq/element-x-android/pull/6258/commits/7b9ecce7be1221713adfc01186666d44d86d7bce

PR #6266, opened 2026-02-27
Low priority room filtering and tagging

### LINKED ISSUE #3821: Support "Low Priority" Rooms

I would like Low Priority rooms to be removed from the main app list (possibly configurable).

I would like the avoid having low priority rooms and conversations down out my more important conversations in the main conversation list. For example I am in many "community" rooms for general topics that I occasionally read, ask questions or respond to mentions. However these rooms tend to be much more active than my more involved conversations with many less people.

Respect the `m.lowpriority` room setting and hide away these rooms from the default/main view.

1. Continue using Element non-X. This is my current "solution".
2. Favorite my more important rooms. This solves some of the issues but has a very different UX, for example the default is not favourite so it is easy to lose new rooms or one-off messages from others.

### Additional context

Element X iOS issue: https://github.com/element-hq/element-x-ios/issues/2478

No

TheJJ:
## Content

Implements filtering and tagging low priority rooms

In the room context menu, you can (next to setting it as favorite) now tag it as low priority.
The room list hides the room then, unless you select the new low priority filter.

## Motivation and context

low priority rooms are mainly write-only / occasional reading rooms that shouldn't obstruct your room view.
this implements filtering them out by default, just like element web and element android classic does.

to make this work, i've also added low prio room info to the rust sdk in matrix-org/matrix-rust-sdk#6183, so other rust-sdk clients can follow!

fixes #3821

## Tests

- `./gradlew :app:compileGplayDebugKotlin`
- `./gradlew :features:home:impl:testDebugUnitTest :libraries:matrix:impl:testDebugUnitTest` (didn't touch other tests)
- `./tools/quality/check.sh`

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 15

## Checklist


TheJJ:
The change in [tests/uitests/src/test/snapshots/images/features.home.impl_HomeView_Day_16_en.png](https://github.com/element-hq/element-x-android/pull/6266/changes#diff-055624c136aea688ee7ad57ded426e7391ce290921f463f30c0b719c97a53ba5) happens because the order in `HomeStateProvider` is `[RoomListStateProvider + aHomeState(empty space)]`, and we had to add a new state in the `RoomListStateProvider`. This could be changed by swapping those two to be future-proof, but we'd end up with a screenshot diff anyway, sorry.
All other screenshots seem to match up! :smile_cat:

tests/uitests/src/test/snapshots/images/features.home.impl.roomlist_RoomListModalBottomSheetContent_Day_0_en.png
bmarty, line 1: Just one first comment: not sure that we want to give so much visibility to the "Low priority" flag. Having an entry in the room detail screen (and maybe at the very bottom of the screen) is probably enough. No need to have it in the bottom sheet displayed when long clicking on a room.

I let @mxandreas having a look at this.

mxandreas:
> The room list hides the room then, unless you select the new low priority filter.

The way it currently works on Element Web is that it moves the room to the bottom of the room list, instead of hiding it. Therefore, I think the default is to do the same on mobile.

Note that EW also adds a down arrow to the avatar of the low priority room.
Screenshot 2026-03-02 at 11 43 50

cc @amshakal to confirm.

tests/uitests/src/test/snapshots/images/features.home.impl.roomlist_RoomListModalBottomSheetContent_Day_0_en.png
jmartinesp: About this, I think it's fine if we have favorite and low priority as opposites in the context menu, since they *are* opposite actions and you could argue the same for not including the favorite action then 😅 . It's also in the context menu in EW.

TheJJ: i would also say that low prio and favorite are equally important.
if we would allow users custom "tags" as groups just like mattermost does, that would be even better, and could be handled through a single entry then.

bmarty: Low priority and favorite are just special room tags, they are not strictly opposite actions ([specs](https://spec.matrix.org/latest/client-server-api/#events-14)). A room can have both tags. Element Web considers that the tags are exclusive, i.e. setting a favorite room as low priority will remove the favorite tag (and vice versa). I do not see such code in this PR, maybe this will have to be handled in the SDK?

Anyway I still consider that setting a room as low priority is an advanced feature, and should not be in the quick actions of rooms. And if we keep favorite and low priority there, using 2 toggles that are inter-dependent is not the ideal UX.

TheJJ: yes, the filtering and tagging is done through the sdk.

shouldn't we keep the context menu ui consistent accross official clients then?
the element-web client also shows this in the room context menu directly:
element-web-room-context-menu

i mean as long as we get low prio room filtering shipped in element x android i'm happy - what would you like me to change so we can progress?

bmarty: I am fine with these changes, waiting for @amshakal to answer @mxandreas' question.

TheJJ: great, thank you!

mxandreas: I am fine with showing low-priority and favourite toggles side by side similarly to web. I think if we don't do that then people conclude it does not exist on mobile. And I think it is not worth opening the discussion if it should be designed in some other way on all platforms.

What about the other question about not showing the low-priority room vs. moving it to the bottom? Are we doing it like on web, e.g. moving to the bottom?

TheJJ: currently low-prio rooms are hidden if not selected (we're just passing the rust sdk to use non-lowprio rooms).
i think this is a good approach since the roomlist remains shorter, and if you search for rooms you'll find the low-prio ones anyway. but i hope this is something we can iterate on afterwards, especially if it needs sdk support?

if we wanted to improve the UX there, i'd suggest adding the possibility to set arbitrary tags for rooms, just like mattermost does. then for each tag, you can decide how to order and whether or not to display this tag by default?

TheJJ:
What would you like me to update? I'm also happy if the real maintainers take over, as long as we get a first version of low priority room filtering delivered soon :)

TheJJ:
Hey! I'd really love to see this feature make it, and its even on the feature wishlist. What's needed to proceed, please?

mxandreas:
> What would you like me to update? I'm also happy if the real maintainers take over, as long as we get a first version of low priority room filtering delivered soon :)

Sorry for the delay. I think what we need is the room still remain visible in the room list, simply moved to the bottom like on Web. Otherwise there is also no option to make it visible again, if needed.

TheJJ:
Alright, I can add that quite easily (if the sdk isn't missing another feature).
Although i see little use in listing the low prio rooms in the main overview, which just gets even longer then although the low prio chats are kind of archived, Web should also change that in my opinion.

And I'd be happy if a maintaier could carry this over the finish line instead of me (although if nobody pucks it up I'll keep driving).

mxandreas:
> although the low prio chats are kind of archived, Web should also change that in my opinion

If you hide the room, the question is how the user gets it back - because the room is no longer visible in their roomlist (the user made a mistake or something changed).

Potentially, the Low Priority filter could show such rooms, but this makes the filter to work inconsistently with others. I think we can take one step at a time here and just move it to the bottom, then think if we want to start hiding it as well.

TheJJ:
if you search for the name you'll find it again, or if you enable the low prio filter button with this patch you see all low prio rooms. and the low prio settings toggle will also remain open until dismissal if setting a room low-prio was a quick mistake you want to undo.

i'm just asking because if we decided to hide and treat low prio as "archival" (Mathew suggested that in some issue) then we don't have to go back and forward here.

mxandreas:
> if you search for the name you'll find it again, or if you enable the low prio filter button with this patch you see all low prio rooms.

It is not consistent with how search or filter work today and how they generally work. The common approach is that both search and filter leave certain items in the list visible, not make new items appear that you do not see without the filter or the search.

> and the low prio settings toggle will also remain open until dismissal if setting a room low-prio was a quick mistake you want to undo.

Which setting toggle? Was there a plan to add any new ones as part of this work?

TheJJ:
Ok then, let's make the low prio rooms just be sorted at the bottom.  Would be great if we could add arbitrary tags to them, and then display/group/hide accordingly (cf Mattermost).

In the room context menu theres a new toggle to set its low priority state.

TheJJ:
I fear I don't have time to work on this for some duration, so I'd be very glad if official maintainers or somebody else carries it over the finish line!

PR #6274, opened 2026-03-03
Use `ShareIntentHandler` early to avoid distributing the whole intent

jmartinesp:
## Content

We now pass a new `ShareIntentData` class around, containing the minimum amount of data needed. We also have a new `OnSharedData` post-processor to revoke uri access after they've been shared.

## Motivation and context

The previous behaviour would make the intent be serialized as part of `NavTarget.IncomingShare` and could potentially lead to `TransactionTooLargeException`s if the intent was quite large.

## Tests

Try sending text and files, everything should still work as before.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/features/share/api/src/main/kotlin/io/element/android/features/share/api/ShareIntentHandler.kt
+++ b/features/share/api/src/main/kotlin/io/element/android/features/share/api/ShareIntentHandler.kt
@@ -8,17 +8,8 @@
 package io.element.android.features.share.api
 
 import android.content.Intent
-import android.net.Uri
-import android.os.Parcelable
-import kotlinx.parcelize.Parcelize
 
 interface ShareIntentHandler {
-    @Parcelize
-    data class UriToShare(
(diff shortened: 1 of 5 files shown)


features/share/api/src/main/kotlin/io/element/android/features/share/api/ShareIntentHandler.kt
```diff
@@ -0,0 +1,30 @@
...
+package io.element.android.features.share.api
+
+import android.content.Intent
+import android.net.Uri
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+interface ShareIntentHandler {
+    @Parcelize
+    data class UriToShare(
```
bmarty: Maybe move this declaration to `ShareIntentData` where the model lives?

PR #6296, opened 2026-03-06
Sign in with element classic final

bmarty:
## Content

Let Element X import details from any existing session from Element Classic to sign in and verify the session automatically.

Note: require https://github.com/element-hq/element-android/pull/9136 to merged and released.

## Motivation and context

Help users of Element Classic migrate smoothly to Element X.

## Screenshots / GIFs

See recorded ones.

Demo of the happy path (case 5 below):

https://github.com/user-attachments/assets/7117da60-97ff-45c8-8ea9-3d0ee45747f4

Demo of case 11:

https://github.com/user-attachments/assets/7d4ed7a7-5366-45d6-b08b-ef390e91ecc7

## Tests

There are several cases to consider, here is a summary:

|Case|Element Classic state|Effect on Element X|
|-|-|-|
|1|Element Classic is not installed|No effect|
|2|Element Classic is installed, but no session exist|No effect|
|3|Element Classic is installed, a session exists, but the session is not verified, or some secrets (private keys) are missing|Pre-fill the homeserver and the userId, but will not be able to verify the session|
|4|Element Classic is installed, a session exists, is verified, all the secrets are known, but a session with the same userId exists on Element X.This can happen since Element X supports multi accounts.|No effect.|
|5|Happy path:Element Classic is installed, a session exists, is verified, all the secrets are known, and no session with the same userId exists on Element X|The flow to import a session from Element Classic is possible|
|6|Error in the communication between the 2 applications (Android) / accessing the shared storage (iOS) to retrieve the data.|No effect, no error displayed. Only logs.|
|7|Edge case:Element Classic is installed, a session exists, is verified, all the secrets are known, and no session with the same userId exists on Element X. But once on the MAS page displayed by Element X, user logs in to a different account|No effect, the regular login flow continues, i.e. the user will have to manually verify the session. The secrets exported by EC will not be used and just deleted.|
|8|Edge case:Element X provides the “mxid:” hint to MAS, but MAS just ignores it if it is connected to another account. It could warn the user. MAS issue: The login_hint is ignored when already logged in.|If the user continues, the secrets will not be used (mismatched userId) and the user will have to verify manually.|
|9|Edge case:The user does not use the button to sign in with Element Classic, but does a manual login. At the end the user signs in to the same account as the Element Classic one.|It is acceptable that EX requests the secret from EC to perform the verification step automatically.|
|10|Element Classic is installed, a session exists, is verified, all the secrets are known, and no session with the same userId exists on Element X. But EC is logged in using SSO that Element X does not support|No effect|
|11|Element Classic is installed, a session exists, is verified, the cross-signing keys are known but backups nor recovery have been set up. Running the migration will transfer the cross-signing keys but no backup key. Since EX is set up to automatically create a backup, it will do so. This in turn enables Element Classic to upload backups but not download. If the user decides to log out of EX it will not get a warning about setting up recovery since it is not the last device the user has. This would leave Classic in a broken state where it uploads to the backup but has no ability to download backups.|Display a dialog telling the user that they should first enable keybackup on Element Classic, before login using Element X. When the user is done with Element Classic, when Element X is resumed, the data are loaded again|
|12|Element Classic has a verified session, had a key backup, but key backup is now deleted. Element Classic still exports the old key backup key to Element X|Handle the same as 11.|

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist


libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/appyx/FaderOrSliderTransitionHandler.kt
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * A TransitionHandler that uses fade transition when the operation is Replace or NewRoot,
+ * and slide transition for all other cases.
+ */
+private class FaderOrSliderTransitionHandler<NavTarget>(
+    private val slider: ModifierTransitionHandler<NavTarget, BackStack.State>,
+    private val fader: ModifierTransitionHandler<NavTarget, BackStack.State>,
+) : ModifierTransitionHandler<NavTarget, BackStack.State>() {
+    override fun createModifier(
```
ElementBot: - :warning: Modifier factory functions should be extensions on Modifier

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/classic/ElementClassicConnection.kt
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    override fun requestSession() {
+        Timber.tag(loggerTag.value).w("requestSession()")
+        coroutineScope.launch {
+            val finalMessenger = messenger
+            if (finalMessenger == null) {
+                Timber.tag(loggerTag.value).w("The messenger is null, can't request data")
+                // Do not emit error, else the regular on boarding flow will be displayed
+                // emitState(ElementClassicConnectionState.Error("The messenger is null, can't request data"))
```
jmartinesp: Should this still be around?

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/classic/ElementClassicConnection.kt
```diff
@@ -0,0 +1,382 @@
...
+    ) : ElementClassicConnectionState
+
+    data class Error(val error: String) : ElementClassicConnectionState
+}
+
+private val loggerTag = LoggerTag("ECConnection")
+
+@ContributesBinding(AppScope::class)
+@SingleIn(AppScope::class)
+class DefaultElementClassicConnection(
```
jmartinesp, line 66: Maybe some diagram or explanation of how the flow should happen would be helpful as docs.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/classic/ElementClassicConnection.kt
```diff
@@ -0,0 +1,382 @@
...
+            Timber.tag(loggerTag.value).d("onServiceDisconnected")
+            // This is called when the connection with the service has been
+            // unexpectedly disconnected&mdash;that is, its process crashed.
+            messenger = null
+            bound = false
+        }
+    }
+
+    override fun start() {
+        Timber.tag(loggerTag.value).w("start()")
```
jmartinesp: Should this be a debug log now?

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/classic/ElementClassicConnection.kt
```diff
@@ -0,0 +1,382 @@
...
+                    // disconnected (and then reconnected if it can be restarted)
+                    // so there is no need to do anything here.
+                    Timber.tag(loggerTag.value).e(e, "RemoteException")
+                }
+            }
+        }
+    }
+
+    private val mutableStateFlow = MutableStateFlow<ElementClassicConnectionState>(ElementClassicConnectionState.Idle)
+    override val stateFlow = mutableStateFlow.asStateFlow()
```
jmartinesp: Nit: should this be at the start of this component?

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/classic/ElementClassicConnection.kt
```diff
@@ -0,0 +1,382 @@
...
+        matrixAuthenticationService.setElementClassicSession(
+            session = (state as? ElementClassicConnectionState.ElementClassicReady)?.elementClassicSession
+        )
+        mutableStateFlow.emit(state)
+    }
+
+    private fun getElementClassicComponent() = ComponentName(
+        BuildConfig.elementClassicPackage,
+        ELEMENT_CLASSIC_SERVICE_FULL_CLASS_NAME,
+    )
```
jmartinesp: Nit: this could probably be a `val` instead?

jmartinesp (commented):
Some minor comments about the code, just nitpicks so feel free to ignore them. 

I'll give it another check with the IDE later and then test it. How should I test this, btw? Do I need versions of EXA and EA signed with some specific keys?

bmarty:
> How should I test this, btw? Do I need versions of EXA and EA signed with some specific keys?

The code should run with debug versions and release version of the applications (EC debug <-> EX debug) and (EC release <-> EX release). But I need to check if nightly build could communicate together I do not think this is the case.

You can install Element Classic from develop (now that https://github.com/element-hq/element-android/pull/9136 has been merged) and EX from this branch (using diawi QR code should be OK)

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/classic/ElementClassicConnection.kt
bmarty: Is using a getter better to you?
`get() = mutableStateFlow.asStateFlow()`

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/classic/ElementClassicConnection.kt
bmarty: Oh yes, the function was different before.

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/classic/ElementClassicConnection.kt
jmartinesp: Oh, I just meant other properties are at the top of the component (~L75), so this should probably be there too?

bmarty: Sure. I was confused because this class has a `fun start()` 😅

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/appyx/FaderOrSliderTransitionHandler.kt
ElementBot: - :warning: Modifier factory functions should be extensions on Modifier

jmartinesp (approved):
It works really well! Just a couple of things I noticed:

1. I had a couple of sessions on EXA that I've been using for a while (so multi-account enabled). When I logged out of both accounts to test this, EXA wouldn't display the new screen to use EA's session data. After I cleared the app's data this was working, so maybe the logout didn't get rid of some data in the local DBs and case 4 happened, where EXA thought it didn't need to display anything because the same session already existed? Or maybe it was the force-kill that helped with this? It's not a blocking issue IMO, and I can't reproduce it anymore.
2. Another weird thing I experienced is I had EA with an account without key storage and EXA already opened in the initial screen. I then verified EA and went back to EXA, and when trying to log in it still asked me to verify my account, maybe because the session data had been shared before I verified EA (in the new initial screen)? Restarting the app used the right flow and I was logged in and verified directly.

bmarty:
For point 1, I am not sure what happened.
For point 2, we EX is resumed on the Sign in with Classic screen, the data is not refreshed on this screen, only in the MissingKeyBackup screen ([here](https://github.com/element-hq/element-x-android/blob/8c5caabed488e28ae725d515a185c1bce49d702b/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/classic/missingkeybackup/MissingKeyBackupPresenter.kt#L34)). This could be changed and the data could be refreshed each time `ClassicFlowNode` is resumed maybe. Not sure if it could have some side effect, but I can give it a try

bmarty:
Thanks @jmartinesp. I have added [fd3c4c2](https://github.com/element-hq/element-x-android/pull/6296/commits/fd3c4c2b2b8eff400541550ba237b25cbb93a54b) that should ensure that Element X is always up to date with Element Classic. So now:
- start EX. The screen to login to Element Classic is displayed
- go to background and to foreground. Nothing particular happen (in particular no glitch in the avatar that got fetched again)
- switch to Element Classic
- change the avatar of the account
- switch to Element X: the avatar should be updated
- switch to Element Classic
- Sign out
- switch to Element X: the screen to login with Classic is replaced by the on boarding screen

I have also checked that there is no regression when the screen to go and enable key backup on Element Classic is displayed.

Also this change is simplifying the code, so all good!

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/LoginFlowTransitionHandler.kt
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * A TransitionHandler that uses fade transition when OnBoarding is replacing the current screen,
+ * and slide transition for all other cases.
+ */
+private class LoginFlowTransitionHandler(
+    private val slider: ModifierTransitionHandler<LoginFlowNode.NavTarget, BackStack.State>,
+    private val fader: ModifierTransitionHandler<LoginFlowNode.NavTarget, BackStack.State>,
+) : ModifierTransitionHandler<LoginFlowNode.NavTarget, BackStack.State>() {
+    override fun createModifier(
```
ElementBot, line 31: - :warning: Modifier factory functions should be extensions on Modifier

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/classic/ElementClassicConnection.kt
```diff
@@ -0,0 +1,411 @@
...
+                }
+            }
+        }
+    }
+
+    /**
+     * Handler of incoming messages from service.
+     */
+    @Suppress("DEPRECATION")
+    inner class IncomingHandler : Handler() {
```
ElementBot, line 206: - :warning: This `Handler` class should be static or leaks might occur (io.element.android.features.login.impl.classic.DefaultElementClassicConnection.IncomingHandler)

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/appyx/FaderOrSliderTransitionHandler.kt
ElementBot, line 31: - :warning: Modifier factory functions should be extensions on Modifier

PR #6302, opened 2026-03-09
Sync Strings

ElementBot:


features/linknewdevice/impl/src/main/res/values-tr/translations.xml
```diff
@@ -4,7 +4,7 @@
   <string name="screen_link_new_device_desktop_step1">"%1$s uygulamasını bir dizüstü veya masaüstü bilgisayarda açın"</string>
   <string name="screen_link_new_device_desktop_step3">"QR kodunu bu cihazla tarayın"</string>
   <string name="screen_link_new_device_desktop_submit">"Taramaya hazır"</string>
-  <string name="screen_link_new_device_desktop_title">"QR kodu almak için %1$s uygulamasını masaüstü bilgisayarda açın%1$s%1$s"</string>
+  <string name="screen_link_new_device_desktop_title">"QR kodu almak için %1$s uygulamasını masaüstü bilgisayarda açın"</string>
```
jmartinesp, line 7: I'm surprised we didn't get any crashes about this 😬 .

screenshots/de/libraries.textcomposer_TextComposerVoiceNotEncrypted_Day_0_de.png
jmartinesp, line 1: Not really related to this PR, but is the stop icon too large now?

bmarty: I had checked in [Figma](https://www.figma.com/design/Ni6Ii8YKtmXCKYNE90cC67/Timeline--new-?node-id=201-2154), there is no design with a stop icon. But we are now using the stop icon from compound.

features/linknewdevice/impl/src/main/res/values-tr/translations.xml
bmarty: No crash, but I think that the template value will appears 3 times.

screenshots/de/libraries.textcomposer_TextComposerVoiceNotEncrypted_Day_0_de.png
jmartinesp: Yeah, it's just that it look larger than the pause or play icons, I think. Anyway, not really related to this PR, it's just that I first noticed it here.

Issue #6298, opened 2026-03-06
Element Call may not function correctly on certain devices.


YamatoRyou:
### Steps to reproduce

Enter a room that supports Element Call and initiate a call.

#### What did you expect?
n/a

#### What happened instead?
Element Call fails to prepare as expected, instead displaying "Join as a guest". 
A few seconds later, an EXA pop-up appears with an error message. 
**The embedded version of Element Call was used during testing.**

A video demonstration:
https://youtube.com/shorts/oKfGPcV8ezc

Logs:
[{5F842310-1EE1-88B2-040F-9BDC9C3E62CA}.zip](https://github.com/user-attachments/files/25803356/5F842310-1EE1-88B2-040F-9BDC9C3E62CA.zip)
Password: `***`
**Please download this file as soon as possible.**

WebView version: 118.0.5993.80

----
**The test device is not mine, and it appears this issue only occurs on a limited number of specific devices. All my own Android devices are working fine.**

Related: https://github.com/element-hq/element-call/issues/3519

### Your phone model

Redmi Note 12 Tutbo

### Operating system version

HyperOS 1.0.18 (Android 14)

### Application version and app store

Element 26.3.0-nightly (67d90492)

### Homeserver

Synapse 1.148.0

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

jmartinesp:
From the logs:

```
// These look fine
03-07 03:40:01.294 D org.matrix.rust.sdk: properties: {"widgetId":"57146930-2398-41f0-a79d-85259b494582","parentUrl":"https://appassets.androidplatform.net/element-call/index.html","roomId":"!TxkGysRFDuDmqKPRpR:dl.example.com","password":null,"userId":"@test8:dl.example.com","displayName":"test8","deviceId":"...","baseUrl":"https://matrix-synapse.example.com:18009/","lang":"zh-CN","fonts":[],"fontScale":null,"theme":"light","viaServers":null,"homeserver":null,"posthogApiHost":null,"posthogApiKey":null,"posthogUserId":null,"rageshakeSubmitUrl":"https://element.io/bugreports/submit","sentryDsn":null,"sentryEnvironment":"DEBUG","e2eEnabled":true} configuration: {"preload":false,"perParticipantE2EE":false,"noiseSuppression":true,"echoCancellation":true} | 
...
// I don't see these when trying to open a call myself. I'm not sure why the register endpoint is used by EC though
[...]

PR #6304, opened 2026-03-09
Fix element call lobby error

bmarty:
## Content

Avoid the application to consider that EC fails to load when the lobby is displayed.
Last commits are our regular maintenance.

## Motivation and context

Part of #6298 

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/pip/PictureInPictureEvent.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/pip/PictureInPictureEvent.kt
@@ -10,8 +10,8 @@ package io.element.android.features.call.impl.pip
 
 import io.element.android.features.call.impl.utils.PipController
 
-sealed interface PictureInPictureEvents {
-    data class SetPipController(val pipController: PipController) : PictureInPictureEvents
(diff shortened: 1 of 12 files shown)


toger5 (changes requested):
We should Always get a `content_loaded` action.
Independent if we show the lobby or if we join the call directly.

Hence
```
  Timber.w("The call took too long to load. Displaying an error before exiting.")

  // This will display a simple 'Sorry, an error occurred' dialog and force the user to exit the call
  webViewError = ""
```
should not be reached...

If you are seeing some timeout and error screen in the lobby there problably is some additional logic that checks for an ongoing rtc session or the join widget action (that should be the only difference between lobby and direct join calls)

What was the exact setup where you could see the lobby getting intercepted by the error screen?

looking at: https://youtube.com/shorts/oKfGPcV8ezc it seems like it is not showing the lobby. The video indicates that they are running into a case, where they start the webview in NON widget mode but in spa mode.

What we would need to investigate is what their url params are. Why it was not showing the widget.

bmarty:
OK, thanks. Let's close the PR then.

PR #6305, opened 2026-03-09
Add network constraints for fetching notifications with `WorkManager`

jmartinesp:
## Content

- Add `isNetworkBlocked` and `isInAirGappedEnvironment` to `NetworkMonitor`, which are populated from `NetworkCallback`. We can remove `NetworkBlockedChecker` now.
- Improve the DI of `SyncPendingNotificationsRequestBuilder` so we use assisted injection now and don't have to add its dependencies to the instantiator.
- Expose the air-gapped state in `EnterpriseService.isInAirGappedEnvironment`.
- Use that to decide if we should add the `NetworkCapabilities.NET_CAPABILITY_VALIDATED` constraint to the notification fetching workers (which checks Google's `204` endpoints to know if the network has internet connectivity).
- Add a feature flag to disable this behaviour.

## Motivation and context

Part of https://github.com/element-hq/element-x-android/issues/6209.

## Tests

Things that can be checked:

- There should be logs for network being disabled when the app goes to background for a while, and of it being enabled when bringing it back to foreground. This means the callback works as the checker did before.
- If you set up a MITM proxy and block `https://www.google.com/generate_204` and `http://connectivitycheck.gstatic.com/generate_204`, your device will display your network not having internet connectivity, and the app should know it's in an air-gapped env. If using this you receive a notification, it should still be fetched.
- If for any reason it didn't, disabling the new feature flag and testing it again should make it work as expected.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist


features/networkmonitor/api/src/main/kotlin/io/element/android/features/networkmonitor/api/NetworkMonitor.kt
```diff
@@ -24,5 +24,11 @@ interface NetworkMonitor {
     /**
      * Checks if the active network is being blocked by Doze, even if it's available.
      */
-    fun isNetworkBlocked(): Boolean
+    val isNetworkBlocked: StateFlow<Boolean>
+
+    /**
+    * A flow indicating whether the app is running in an air-gapped environment.
```
bmarty:
```suggestion
     * A flow indicating whether the app is running in an air-gapped environment.
```

libraries/push/impl/src/test/kotlin/io/element/android/libraries/push/impl/push/DefaultPushHandlerTest.kt
```diff
@@ -238,8 +238,16 @@ class DefaultPushHandlerTest {
             analyticsService = analyticsService,
             systemClock = systemClock,
             workManagerScheduler = workManagerScheduler,
-            buildVersionSdkIntProvider = buildVersionSdkIntProvider,
             resultProcessor = resultProcessor,
+            syncPendingNotificationsRequestFactory = object : SyncPendingNotificationsRequestBuilder.Factory {
```
bmarty: Maybe create a `FakeSyncPendingNotificationsRequestFactory` class?

libraries/push/impl/src/test/kotlin/io/element/android/libraries/push/impl/workmanager/DefaultSyncPendingNotificationsRequestBuilderTest.kt
```diff
@@ -0,0 +1,151 @@
+/*
+ * Copyright (c) 2025 Element Creations Ltd.
+ * Copyright 2025 New Vector Ltd.
```
bmarty: `Copyright (c) 2026 Element Creations Ltd.` should be enough for new files.

features/networkmonitor/impl/src/main/kotlin/io/element/android/features/networkmonitor/impl/DefaultNetworkMonitor.kt
```diff
@@ -39,13 +42,13 @@ import java.util.concurrent.atomic.AtomicInteger
...
-    appCoroutineScope: CoroutineScope,
+    @AppCoroutineScope appCoroutineScope: CoroutineScope,
+    private val buildMeta: BuildMeta,
 ) : NetworkMonitor {
     private val connectivityManager: ConnectivityManager = context.getSystemService(ConnectivityManager::class.java)
-    private val blockedNetworkBlockedChecker = NetworkBlockedChecker(connectivityManager)
 
-    override fun isNetworkBlocked(): Boolean = blockedNetworkBlockedChecker.isNetworkBlocked()
+    override val isNetworkBlocked = MutableStateFlow(false)
+    override val isInAirGappedEnvironment = MutableStateFlow(false)
```
bmarty, line 51: Is it fine to initialize the 2 stateflow with `false`? Maybe read values as it was done before in `NetworkBlockedChecker`?

libraries/push/impl/src/test/kotlin/io/element/android/libraries/push/impl/workmanager/DefaultSyncPendingNotificationsRequestBuilderTest.kt
jmartinesp: Agh, this is because the file was renamed although git thinks it's new. I'll change it anyway.

features/networkmonitor/impl/src/main/kotlin/io/element/android/features/networkmonitor/impl/DefaultNetworkMonitor.kt
jmartinesp: When the app is started, there's no chance network can be blocked by Doze, so `isNetworkBlocked` will always be false. And for the air-gapped env check we need to receive the callback to actually know what's its status, I think.

bmarty: > When the app is started, there's no chance network can be blocked by Doze

The app can be started by a Push, so are you sure this is true?

jmartinesp: Hmm, good call. It *should* always be exempted from network restrictions in that case, according to the docs but well... we all know how reliable Google docs are 😅 .

bmarty: That's "true" for Firebase, but not really for UnifiedPush...

jmartinesp: https://github.com/element-hq/element-x-android/pull/6305/commits/24ac3640967fe2586bbf66c8755c23494a9b67d0 should fix it

PR #6321, opened 2026-03-11
Add a foreground service with a wakelock for fetching push notifications

jmartinesp:
## Content

- Creates `PushHandlingWakeLock` to centralise how wake locks for push notification handling are used: it'll control when we start the foreground service that will keep the device awake when scheduling push notification resolving. This foreground service will only be started if the device is not in 'interactive' mode, which means the CPU can enter deep sleep at any time.
- Acquire this wakelock for 3 minutes as soon as we receive a push notification in this service. The 3 minute limit comes from the type of foreground service, which has a matching timeout value.
- When we're starting processing the notifications for a session, stop the service and release the wakelock early if possible.
- When we try to stop the service, if there are still pending operations, restart the service so those are also processed without the CPU going into deep sleep.

## Motivation and context

This ensures the notification fetching is done as soon as possible when the device is locked/idle:

1. The async code that would previously skip some extra time while waiting for continuations (i.e. 10s timeout turning into 1-2 min) will behave as expected.
2. Same for coroutines launched with the wakelock enabled.
3. The expedited workers scheduled with the wakelock acquired should execute immediately: previously this could take several seconds, or even 10-20s.

## Tests

Everything should still work as expected, but notifications should be displayed earlier.

I tested this both with Firebase and UnifiedPush and could notice how pushes that could take 10-20s to display before when the device was locked displayed in at most 5s now.

If you want to test it yourself, just disconnect and lock your device so it can go into deep sleep, wait for 10-20s, then trigger some action that will make you receive a push notification. Having a EW session open can be useful too, to verify the timing: if you have another device to compare to, you'll see the one using this branch will display a notification at most 1-2s after EW does, if not simultaneously, while the one using the current `develop` can take >10s if it's hit with a particularly bad deep sleep state. Using UnifiedPush makes the tests more reliable as they don't have the FCM wakelock to prevent deep sleep, although using Firebase does not always prevent this behaviour, it's way more unlikely, but it can still happen if we don't start the wakelock/foreground service.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/DefaultPushHandlingWakeLock.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/DefaultPushHandlingWakeLock.kt
@@ -8,15 +8,13 @@
 package io.element.android.libraries.push.impl.push
 
 import android.content.Context
-import android.os.PowerManager
-import androidx.core.content.getSystemService
 import dev.zacsweers.metro.AppScope
 import dev.zacsweers.metro.ContributesBinding
 import dev.zacsweers.metro.SingleIn
 import io.element.android.libraries.di.annotations.ApplicationContext
 import io.element.android.libraries.push.api.push.PushHandlingWakeLock
 import timber.log.Timber
-import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicInteger
 import kotlin.time.Duration
 
 @ContributesBinding(AppScope::class)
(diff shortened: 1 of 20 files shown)


ganfra (approved):
Let's see how it improves things, thanks!

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/DefaultPushHandlingWakeLock.kt
```diff
@@ -0,0 +1,45 @@
...
+        Timber.d("Releasing wakelock used for push handling.")
+        FetchPushForegroundService.stop(context)
+        if (count.decrementAndGet() <= 0) {
+            Timber.d("No more wakelock needed for push handling, stopping service.")
+            count.set(0)
+        } else {
+            Timber.d("Wakelock still needed for push handling, restarting service | count: ${count.get()}.")
+            FetchPushForegroundService.startIfNeeded(context)
+        }
+    }
```
bmarty, line 44: This code needs to be read again I think

PR #6346, opened 2026-03-13
Add room URL previews to the timeline

### LINKED ISSUE #2267: URL Preview for both normal and E2EE Rooms/DMs

Have ability to Preview URLs in both non-E2EE and E2EE Rooms/DMs (example: YouTube video links) so
you can see the Video Description and the Video thumbnail

Most of the time people only copy/paste url links, they
seldom bother to also copy and paste a youtube video name/title

Searching through posted video links with video thumbnails immediately makes it s lot easier and much
faster to quickly locate the desired video to watch

_No response_

### Additional context

_No response_

Yes

kalix127:
## Content

Add URL previews for room timeline messages.

This change adds:
- a room-level setting to enable or disable link previews
- URL preview rendering for previewable links in message timelines
- authenticated preview fetching from the homeserver
- preview support in regular messages, pinned messages, and threads
- compact preview card layout adjustments so the card renders above the URL and fits better in the message bubble
- Compose previews for the URL preview card (with and without image) for Paparazzi screenshot coverage
- unit tests for URL parsing utility functions (`isPreviewableUrl`, `hostNameFromUrl`)

## Motivation and context

Implements an open issue for link previews in the Android timeline.

Closes #2267

Desktop was already able to show previews with the homeserver configured correctly, but Android was missing the feature and the room-level control for it.

## Screenshots / GIFs

https://github.com/user-attachments/assets/5d53f8e6-b489-4ab1-be50-d5a95b512deb

## Tests

- Added Compose `@PreviewsDayNight` previews for `TimelineItemUrlPreviewView` (with image and without image) automatically picked up by Paparazzi for screenshot tests
- Added unit tests for `isPreviewableUrl` (http, https, ftp, mailto, malformed) and `hostNameFromUrl` (hostname extraction, www prefix removal, invalid input fallback)
- Built and deployed the debug app to a physical device
- Verified on-device that:
  - the debug app installs side by side with the daily app
  - YouTube links render a preview card
  - the room setting enables and disables previews
  - the preview card renders above the URL
  - the preview sizing and timestamp layout behave correctly

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): GrapheneOS, Android 15

## Checklist


Skeleton22121:
FINALLY, thanks for you work, i've been waiting for a long time for these not to be broken

kalix127:
@americanrefugee could you take a look at this PR when you get a chance?

americanrefugee:
It's looking good to me, but I think @amshakal was working on the actual designs...

mxandreas:
Before going any further with the development, it should be confirmed if we technically, e.g. from privacy/sustainability point of view want to proceed with homeserver providing the preview. AFAIK, the desired solution is that the preview will be fetched and attached to the message by the sender, and the sender can of course remove the the preview before they send.

I also would avoid adding any settings (in particular at a room level), I guess one reason you might want a setting is again because how it is currently implemented.

cc @dkasak

mxandreas:
> the preview will be fetched and attached to the message by the sender

Adding to this, that the corresponding MSC is https://github.com/matrix-org/matrix-spec-proposals/pull/4095 and it seems some clients have implemented it already.

kalix127:
I'm officially back, any updates on this PR?

mxandreas:
> I'm officially back, any updates on this PR?

I learned that our integrations team is also looking into this feature (they started from the Web side), but it is yet unclear what technology choice will be used. I hope we can provide an update soon on this one.

Fmstrat:
@mxandreas was a choice made? Been watching this one as it's my final requirement before switching off Classic. Thanks!

mxandreas:
> I'm officially back, any updates on this PR?

First, for context, we have agreed that the [MSC4095](https://github.com/beeper/matrix-spec-proposals/blob/bundled-url-previews/proposals/4095-bundled-url-previews.md) which bundles the preview by the sender is a long term solution that we want to apply as it eliminates the privacy concerns and the settings that come with it. While implementing on a single platform is not a huge effort, we need it on all 3 platforms as otherwise an Android user would not see any previews if they have friends using iOS, etc.

But as URL previews are valuable, we'd like to ship this PR for the time being. In terms of settings we want to copy the model that Element Web current has and not have a room-level setting (as it may be hard to get rid of it later when we switch to the MSC4095). Thus in advanced settings there would be radio buttons to:
* Swich off previews everywhere.
* Show previews only in unencrypted rooms (this is the default).
* Show previews in all rooms.

The setting is local.

@amshakal Will provide the detailed designs for the setting as well as she may adjust how the previews themselves look visually. cc @daniellekirkwood

kalix127:
Perfect

manfrommedan:
Nice feature, thanks for this! Link previews have been missing for a while.

While testing I hit a bug: mentions and permalinks also get a website preview, which is not intended. A mention like `@alice:server` renders as a `URLSpan` pointing at `https://matrix.to/#/@alice:example.org`, and the current classifier only looks at the scheme:

```kotlin
internal fun isPreviewableUrl(url: String): Boolean {
    return tryOrNull { URI(url).scheme?.lowercase() } in setOf("http", "https")
}
```

A matrix.to link is `https`, so it passes the check and the timeline fetches a preview for the matrix.to page. Same for room links (`#/!room:server`) and event permalinks (`#/$event`). So a plain `@nickname` shows an unexpected matrix.to preview card.

The robust fix is to classify the link with the SDK's `PermalinkParser` instead of by URL shape: anything it resolves to something other than a `FallbackLink` is a Matrix identifier and should be skipped:

```kotlin
if (permalinkParser.parse(url) !is PermalinkData.FallbackLink) return false
```

This also covers custom permalink base URLs, not just matrix.to.

manfrommedan:
I did some more testing and the mention case is not fully closed. Sending a message that is just a mention of `@someone:my.homeserver` shows a preview card for `my.homeserver` (its landing page), even though I never posted a link.

What happens: a mention renders as a pill, but its underlying text is the full matrix id, and Linkify turns the `my.homeserver` tail into a plain `http://my.homeserver` URLSpan. By the time `isPreviewableUrl` sees it, it is just a domain, so the parser check lets it through.

That URLSpan belongs to the mention, not to real content, so the smallest fix is to drop any URLSpan overlapping a mention when collecting preview candidates, reusing the existing `getMentionSpans()`:

```diff
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/urlpreview/UrlPreviewParser.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/urlpreview/UrlPreviewParser.kt
@@ imports @@
 import io.element.android.libraries.matrix.api.permalink.PermalinkParser
+import io.element.android.libraries.textcomposer.mentions.getMentionSpans
 import org.jsoup.nodes.Document
@@ private fun CharSequence.extractUrlSpans(): List<String> @@
     return spanned.getSpans(0, spanned.length, URLSpan::class.java)
         .orEmpty()
         .sortedBy { spanned.getSpanStart(it) }
+        // Drop URLSpans that fall inside a mention: Linkify auto-links the ":server" tail of the id.
+        .filter { spanned.getMentionSpans(spanned.getSpanStart(it), spanned.getSpanEnd(it)).isEmpty() }
         .map { it.url }
```

It is read-only over the text, so it does not touch the mention's styling or click handling, and `getMentionSpans()` already covers both `MentionSpan` and the RTE `CustomMentionSpan`.

kalix127:
Did some manual testing with all the cases I could think of and now it looks right.

bmarty:
Closing as we will only implement to the new Url Preview solution, aka [bundled preview](https://github.com/matrix-org/matrix-spec-proposals/pull/4095) on Element X.

PR #6349, opened 2026-03-15
Do not show "Not Encrypted" badge for public rooms

bxdxnn:
## Content

Do not show "Not Encrypted" badge for public rooms

## Motivation and context

There's no point in warning that a public room isn't encrypted, because anyone can still join it and view the messages, even if encryption is in place.

## Tests

- See a public room
- It shouldn't show "Not Encrypted" badge above the message composer

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
@@ -87,7 +87,7 @@ class AttachmentsPreviewPresenter(
 
         val markdownTextEditorState = rememberMarkdownTextEditorState(initialText = null, initialFocus = false)
         val textEditorState by rememberUpdatedState(
-            TextEditorState.Markdown(markdownTextEditorState, isRoomEncrypted = null)
+            TextEditorState.Markdown(markdownTextEditorState, isRoomEncrypted = null, isRoomPublic = null)
         )
 
         val ongoingSendAttachmentJob = remember { mutableStateOf<Job?>(null) }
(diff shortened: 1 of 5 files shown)


jmartinesp:
@mxandreas I think we need your feedback here, but I think the current behavior is the expected one, right?

mxandreas:
> @mxandreas I think we need your feedback here, but I think the current behavior is the expected one, right?

Is, it is expected. The primary purpose of this is to prevent maliciously spoofing rooms (e.g. via similar names) which aren't encrypted without the user knowing about that. We want this to be visible at all times when it is not encrypted.

jmartinesp:
Thanks. Also, I wanted to add:

> There's no point in warning that a public room isn't encrypted, because anyone can still join it and view the messages, even if encryption is in place.

That's true only if room history is set up to allow reading the full history and not just the messages since you were invited, if I'm not mistaken. And for encrypted rooms I think shared room history isn't implemented yet in Element (I think).

Thank you for your PR @bxdxnn , but we want to keep the current behaviour for the reasons mentioned above.

bxdxnn:
@jmartinesp @mxandreas

Just my two cents, I’d really appreciate if this could be reconsidered. Right now, it feels like we're falling back into the same "message shields" UX problem from the previous Element Classic client:


I understand and respect this decision, but the "Not Encrypted" warning is really confusing for an average person (what not encrypted? is my data safe? etc.) and this badge is mostly a technical noise in 99.9% cases. Even for the "public + encryption + history since joined" case, I can't think of why would someone want to differentiate whether the public room is encrypted or not by seeing (or not seeing) a persistent badge above the message composer.

Someone trying to spoof an encrypted public room seems like a very niche edge case and I don't understand why would the attacker who is trying to typosquat wouldn't just select the same encryption settings as the original room, and showing this badge wouldn't make the user more aware of some security concerns of such rooms and in some cases might rather allude him into a false sense of security (e.g. "if this badge is not shown, the communication is 100% secure"). If someone wants to know the encryption status of the room, it's always shown in the room's profile card (even in this PR).

mxandreas:
@bxdxnn 

Thanks for the thoughts and feedback. It is one of those cases where it is hard to make it work well for everyone. It does not matter if it is a niche attack or not - we have to account for the worst case scenario, unfortunately in here. If we had unlimited amount of resources, we could invest more to figure out if there is a better compromise but for the moment if there is a "friction" we optimize for the use cases where people always use encrypted rooms. Knowing that this is not ideal for others.

Regarding shields - we're expecting somewhat similar friction there as well since not all users are used to verifying their devices which now becomes mandatory soon. They also do not see why this is necessary if they just participate in public rooms.

PR #6352, opened 2026-03-16
Sync Strings - iterate on wording about crypto identity

ElementBot:


features/securebackup/impl/src/main/res/values/localazy.xml
```diff
@@ -2,16 +2,16 @@
 <resources xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
   <string name="screen_chat_backup_key_backup_action_disable">"Delete key storage"</string>
   <string name="screen_chat_backup_key_backup_action_enable">"Turn on backup"</string>
-  <string name="screen_chat_backup_key_backup_description">"Store your cryptographic identity and message keys securely on the server. This will allow you to view your message history on any new devices. %1$s."</string>
+  <string name="screen_chat_backup_key_backup_description">"This will allow you to view your chat history on any new devices and is required for backup of chats and digital identity. %1$@."</string>
```
bmarty: There is a problem in the pattern here. Fixed on Localazy.
```suggestion
  <string name="screen_chat_backup_key_backup_description">"This will allow you to view your chat history on any new devices and is required for backup of chats and digital identity. %1$s."</string>
```

tests/uitests/src/test/snapshots/images/features.home.impl.components_RoomListContentView_Day_4_en.png
bmarty, line 1: @mxandreas with the new version:

Image

I am not sure that the user will understand what is happening. With the title "Back up your chats" they can understand that the backup is not effective yet. But the description suggests the opposite. WDYT?

tests/uitests/src/test/snapshots/images/features.messages.impl.crypto.identity_IdentityChangeStateView_Day_1_en.png
bmarty, line 1: @mxandreas not from this PR, but I have the feeling that `has been reset` is more correct that `was reset`. Do you agree?

tests/uitests/src/test/snapshots/images/features.home.impl.components_RoomListContentView_Day_4_en.png
jmartinesp: I also find it way more confusing than the previous one, to be honest. The user might not understand what the chats being end-to-end encrypted means. What about something like:

'Your chats are automatically backed up, but they're encrypted. To be able to read them on other devices or when the app is reinstalled you will need to create a key so you can recover your digital identity and decrypt them.'

---

Now that I read my suggestion above, I'm not sure if it's actually easier to understand for a non-techie user 😅 . Maybe just changing the title as @bmarty suggested would be better.

mxandreas: Thanks for the feedback. We'll keep it currently as it is with the potential to make another iteration if needed. The main goal for now was to make the users aware that this is backup (something that they know better than recovery) and when do they need the recovery key.

tests/uitests/src/test/snapshots/images/features.messages.impl.crypto.identity_IdentityChangeStateView_Day_1_en.png
mxandreas: I was curious :) so I looked up and seems like actually "was" seems to be the right one in this context. https://prowritingaid.com/has-been-vs-was

I would not change it anyway right now to avoid changing too many things at once.

tests/uitests/src/test/snapshots/images/features.ftue.impl.sessionverification.choosemode_ChooseSelfVerificationModeView_Day_0_en.png
bmarty, line 1: @mxandreas I do not think that this change is expected, but maybe. Can you confirm please?

mxandreas: > [@mxandreas](https://github.com/mxandreas) I do not think that this change is expected, but maybe. Can you confirm please?

The rest is expected but "Enter recovery key" should stay "Use recovery key". Before I made the changes in Localazy, I applied them locally in my iOS development environment, and there I still see "Use recovery key". It uses the key `screen_identity_confirmation_use_recovery_key` in there it seems.

tests/uitests/src/test/snapshots/images/features.messages.impl.crypto.identity_IdentityChangeStateView_Day_1_en.png
bmarty: OK, thanks

tests/uitests/src/test/snapshots/images/features.ftue.impl.sessionverification.choosemode_ChooseSelfVerificationModeView_Day_0_en.png
bmarty: Sorry, I was not specific, and there are several changes here. I was talking about the change "Sign out" -> "Remove this device". In this context (the user just signed in), I find the action "Remove this device" a bit weird, but I understand that this is a global replacement of action.

For the "Use recovery key" wording, I cant fix that once this PR is merged (I will merge it now)

bmarty (approved):

PR #6356, opened 2026-03-16
Fix long messages not being clickable

### LINKED ISSUE #6347: Long messages are not tappable
### Steps to reproduce

1. See a long message (https://matrix.to/#/!Ky5AhQMtLqvpRaZi7d2Br86i9hY_7gLFMQxV6MaCaQY/$EKSDwktAXO_-GBxvLpVVxdBi8V_73fNWCFS9JnyWp54?via=matrix.org)
2. Try to tap/hold the message in the bottom or middle. The top of it is tappable.

#### What did you expect?
I can hold or tap the message anywhere

#### What happened instead?
Message is not actionable after some paragraphs

### Your phone model

It occurs on any phone, but the starting point where you can no longer tap it varies between different DPIs.

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

No

No

jmartinesp:
## Content

As @bmarty found out, `clip = true` causes the click event to be ignored in some cases. Since we have the shape we want to draw and we're using a custom `onDraw` modifier anyway to cut-out part of the path, we can just draw everything using the modifier and avoid using `clip = true`.

This seems to fix the issue.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/6347

## Tests

With a very long message (as in, very 'tall' one with lots of new lines and line wraps, a long lorem ipsum works) coming from another user, try tapping on the bottom of it.

If it works, it's fixed.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist


tests/uitests/src/test/snapshots/images/features.messages.impl.timeline.components_TimelineItemEventRowForDirectRoom_Day_0_en.png
bmarty, line 1: It seems that there is a regression here :/

bmarty (changes requested):
There is a regression when rendering images :/

tests/uitests/src/test/snapshots/images/features.messages.impl.timeline.components_TimelineItemEventRowForDirectRoom_Day_0_en.png
jmartinesp: Ah damn, I didn't realise images were broken 🫤 . Let me check if there's a better way to handle it.

jmartinesp: [There](https://github.com/element-hq/element-x-android/pull/6356/commits/0aa9f8a1c589b898d295f74e3ad5ed3a1fbcf180) I fixed it (I think 😅 ).

jmartinesp: The new screenshots for these look fine, *\*phew\**.

tests/uitests/src/test/snapshots/images/features.messages.impl.timeline.components_TimelineItemEventRow_Night_0_en.png
bmarty, line 1: Not sure if it is visible IRL, but on the new version, the top right corner is not perfect:

Image

It seems that we can see the background below the image.

bmarty (commented):
Maybe we can apply the fix for long text only? The code is really convoluted now :/

tests/uitests/src/test/snapshots/images/features.messages.impl.timeline.components_TimelineItemEventRow_Night_0_en.png
jmartinesp: It's probably some issue with the anti-aliasing, I'm not sure if we can make it perfect, to be honest...

---

It seems like it is visible in the deployed app in dark mode 🫤 .

jmartinesp:
> Maybe we can apply the fix for long text only? The code is really convoluted now :/

I tried to wrap all the complexity in a single method which wraps all the drawing. Not having several composition strategies also fixes the issues with the aliasing in corners.

bmarty (approved):
No diff in the screenshot record anymore, nice!

PR #6359, opened 2026-03-16
Filter empty days

bxdxnn:
## Content

There may be day dividers that include only filtered events. This change will eliminate those empty day dividers.

## Motivation and context

See #6360 "Tests"

## Tests

- See a day that only has filtered events

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist


bxdxnn:
It looks like empty days can be also reproduced if the only event of a day was a self-join event

jmartinesp:
I might be wrong, but I think this algorithm checks the wrong thing, if we have:

| Index # | Item |
| -- | -- |
| 0 | Message A |
| 1 | --Today-- |
| 2 | Message B |
| 3 | --Yesterday-- |
| 4 | --Day before yesterday-- |

A happened today, B happened yesterday and no message was found for the day before, so we can remove it. If I add this test case to your code, it returns `A, Today, B`, removing the day separators for both yesterday and the day before:

image

Also,  `aProfileTimelineDetailsReady` doesn't seem to exist AFAICT.

---

I have an alternative algorithm that seems to work though, maybe I can add it to https://github.com/element-hq/element-x-android/pull/6360 directly and close this PR?

bxdxnn:
OK if you add this directly to #6360, this PR is vibecoded and for me it seemed to be a general bug with the post-processor.

PR #6367, opened 2026-03-17
Add compatibility with rustls

jmartinesp:
## Content

A new `rustls-platform-verifier-android` library has to be added to the project, it'll be called from Rust to get access to the certificates on Android.

Originally, this was supposed to be added as a local maven repo pointing to the rust crate that publishes the AAR, but that's just plain terrible (more details [here](https://github.com/rustls/rustls-platform-verifier#android)).

Instead, what we can do is use a script that uses `cargo-download` to download the latest crate or a specified version, unzip it and add the `aar` file to the `:libraries:matrix:impl` module.

## Motivation and context

Android counterpart of https://github.com/matrix-org/matrix-rust-sdk/pull/6306.

## Tests

**Without any self-signed certificates** - those are still failing - try a fresh login to any homeserver. When I tried this before, every request was failing because rustls couldn't fetch Android's credentials storage.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist


.github/workflows/sonar.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         with:
           cache-read-only: ${{ github.ref != 'refs/heads/develop' }}
       - name: Build debug code and test fixtures
-        run: ./gradlew assembleDebug createFullJarDebugTestFixtures :app:createFullJarGplayDebugTestFixtures $CI_GRADLE_ARG_PROPERTIES
+        run: ./gradlew assembleGplayDebug createFullJarDebugTestFixtures :app:createFullJarGplayDebugTestFixtures $CI_GRADLE_ARG_PROPERTIES
```
ganfra, line 57: Why changing this here?

ganfra (approved):
LGTM I guess, I trust you on this! 

One question, when should we call the script exactly?

.github/workflows/sonar.yml
jmartinesp: It seems like `assembleDebug` causes `:libraries:matrix:impl:bundleDebugAar` to run, in turn causing this issue:

```
* What went wrong:
Execution failed for task ':libraries:matrix:impl:bundleDebugAar'.
> Error while evaluating property 'hasLocalAarDeps' of task ':libraries:matrix:impl:bundleDebugAar'.
   > Direct local .aar file dependencies are not supported when building an AAR. The resulting AAR would be broken because the classes and Android resources from any local .aar file dependencies would not be packaged in the resulting AAR. Previous versions of the Android Gradle Plugin produce broken AARs in this case too (despite not throwing this error). The following direct local .aar file dependencies of the :libraries:matrix:impl project caused this error: /Users/jorge/Developer/Element/element-x-android/libraries/matrix/impl/libs/rustls-platform-verifier-android.aar
```

AFAICT, `assembleGplayDebug` still builds the source Sonarqube needs to scan, but it won't create this 'fat AAR' file.

jmartinesp:
> LGTM I guess, I trust you on this!

You don't have to trust me, you can just test the branch 😅 . Without these changes and removing the `ClientBuilder.disableSslVerification()` call, no network requests to HTTPS homeservers will work with the latest SDK.

> One question, when should we call the script exactly?

When a new rustls version is used in the Rust SDK which comes with a new `platform-verifier` version. They should ping us in that case (I have added some docs [here](https://github.com/element-hq/element-x-android/pull/6367/commits/e49a20935986a38948a763fbcad57cb82d8e50a4)).

PR #6374, opened 2026-03-18
Move filters to the top bar filters

bxdxnn:
## Motivation and context

With the addition of the "Your spaces" button, it makes sense to have all filters in one place for a more consistent experience and to simplify the home screen.

## Tests

- See if filters work
- See if resetting them from the home screen works

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/HomeView.kt
+++ b/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/HomeView.kt
@@ -26,8 +26,10 @@ import androidx.compose.material3.FabPosition
 import androidx.compose.material3.TopAppBarDefaults
 import androidx.compose.material3.rememberTopAppBarState
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.input.nestedscroll.nestedScroll
 import androidx.compose.ui.platform.LocalLayoutDirection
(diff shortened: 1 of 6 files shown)


jmartinesp:
Thanks for your contribution and the effort you put into it, but we can't accept it for several reasons:

1. The UI is not working great when you have lots of spaces: it needs an extra click to access the filters which were immediately accessible before, and then it obscures the underlying room list so you can't see what's the result until you dismiss the bottom sheet (I attached a video).
2. The room list no longer scrolls at all in my device for some reason (you can check it at the end of that video).
3. When making significant UI changes in one platform, we need to do the same on the other one, so if we wanted to add this, we'd need to implement an equivalent PR on iOS.

https://github.com/user-attachments/assets/cc92d9f2-b434-4529-827e-cba97ec2bbb4

Also, UI changes need to be approved by the product and design team before being merged. For this reason, it's highly recommended to first open an issue exposing the problem and your proposal to fix it, then have product and design give feedback about it, address it, and once we've agreed on something it should be implemented: otherwise, this usually ends up with some unnecessary frustration for the contributor who put some effort in trying to improve the client and then their changes are rejected because they don't match what we expect.

I'll ping @mxandreas here for his product feedback, but I believe this doesn't match the behaviour we expect from this screen even if the scrolling bug was fixed.

mxandreas:
> I'll ping @mxandreas here for his product feedback, but I believe this doesn't match the behaviour we expect from this screen even if the scrolling bug was fixed.

It is not a bad idea as such to explore, but because the spaces filter is new (meaning that we're still collecting feedback) as well as we're planning to also potentially implement "sections" for the purpose of grouping rooms just for your own needs (for which spaces is an overkill), we're currently not making ad-hoc changes in this area. Thanks though for the idea, we'll take into consideration.

cc @americanrefugee

jmartinesp:
I'm closing this, but thanks again for the contribution.

PR #6409, opened 2026-03-18
Switch to earpiece if device is held to ear when playing voice messages

### LINKED ISSUE #2424: Add ability to listen voice messages through ear speaker
### Your use case

https://github.com/element-hq/element-android/issues/7575

I would like to have to ability to listen to voice messages through the phone speaker 

Because like that only I can listen to them and not everybody around.

There was an open issue in the old element client repo already
https://github.com/element-hq/element-android/issues/7575

But nobody worked on that feature so far. It is crucial in my opinion. Please work on that!

_No response_

### Additional context

_No response_

No

bxdxnn:
## Content

Be able to listen to voice message using the earpiece when a device is held to ear

## Motivation and context

Fixes #2424

This is a highly desired feature and the standard behavior in all messaging apps. The earpiece sensor feature remains available even when external devices are connected, due to privacy considerations. Here's why:

1. Android does not provide information on whether a connected Bluetooth device is a speaker or a headset.  
2. Even if we were certain (we are with wired headsets) the connected device is a private headset:  
   - It could be used as a speaker at maximum volume.  
   - It might be in use by another person.  
   - The user might forget he have an unused Bluetooth headset connected, which could lead to confusion or inconvenience when audio is unexpectedly absent.

## Tests

- Test that switching to earpiece mode occurs when the device is held to the ear.
- Ensure the output device priority order:  
  1. Wired headset  
  2. Bluetooth device  
  3. Speakers  
  4. Ensure that the earpiece is used if the device is held to the ear, overriding the above options.  
- Test that audio playback works over external devices.  
- Test that audio doesn't unexpectedly switches to external device if it's currently on the earpiece and a device gets connected (e.g. over Bluetooth).
- Test that audio switches to an external device as expected when it gets connected and the device is not held to the ear.  
- Test that audio is not played through the earpiece unless the device is explicitly held to the ear.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist


bxdxnn:
@ganfra any updates?

bxdxnn:
@bmarty can this be reviewed?

mxandreas:
> This is a highly desired feature and the standard behavior in all messaging apps.

So, again, I did some testing on both iOS and Android using Signal and WhatsApp. The results are:

1) If you do not have a BT device connected, it does indeed switch to earpiece when you move your phone to your ear.
2) However, if the BT device was connected, then it kept playing on the BT device.

The results were consistent across both platforms and apps. I am not sure if this can be really used as a "safety feature", I see it more as a convenience feature and thus, I'd rather keep playing on the BT if that is connected.

@fkwp what are we currently doing for Element Call calls regarding this?

fkwp:
> > This is a highly desired feature and the standard behavior in all messaging apps.
> 
> So, again, I did some testing on both iOS and Android using Signal and WhatsApp. The results are:
> 
> 1. If you do not have a BT device connected, it does indeed switch to earpiece when you move your phone to your ear.
> 2. However, if the BT device was connected, then it kept playing on the BT device.
> 
> The results were consistent across both platforms and apps. I am not sure if this can be really used as a "safety feature", I see it more as a convenience feature and thus, I'd rather keep playing on the BT if that is connected.
> 
> @fkwp what are we currently doing for Element Call calls regarding this?

I do agree with @mxandreas a headset (independent of bt or wired) is overruling earpiece)

bxdxnn:
@mxandreas

It will play on BT/external devices, but if held to an ear, it will switch to the earpiece, i.e. it won't disable the sensor feature like in WhatsApp in case you still need it.

WhatsApp and other apps disable the sensor feature when external devices are connected, which is personally frustrating in cases like when the phone is connected to a car. I could be out of a car and on parking, and with the sensor disabled I will have to disable Bluetooth first, which is also very inconvenient if I have Android Auto connected and the passengers need it at the same time. With this PR, I can just hold the device to ear if I need for it to play on the earpiece instead of speakers without disconnecting BT.
Another case, I might also have music playing on the external device, and I'd like to be able to listen to the voice message on the earpiece while the music is still playing and without having to disconnect anything first.

If you don't want it to play to the ear (keep playing on BT), don't hold the device to the ear  ̄\\_ (ツ)_/ ̄

This is a privacy feature, so IMO we should prioritize privacy here instead of copying other apps. If you want to disable this feature if an external device is connected, please tell.

mxandreas:
Thanks for providing the example scenarios.

My previous point was primarily about the initial claim that this is "standard in messaging apps" while it actually does not seem to be so. At least not in WhatsApp and Signal which Element is compared to and has overlapping use cases with.

When we consider changing the behavior, we have to gain some certainty that this is a step in the right direction, and something that is appreciated by most users. Without the ability to conduct a ad-hoc UX research, using other apps as a reference is the first low-hanging fruit. If this is a privacy concern, then it should equally to apply to messages sent via WhatsApp or Signal which are also secure messengers.

What we do not know well is if the WA and Signal behavior is intentional or accidental. And if it is (or was) accidental, then why haven't they done anything to change it.

Having said that - I kind of agree that it may be safe to try overriding BT, because indeed one can simply not move the phone to his ear when they do not want to use earpiece. @fkwp do you have any specific arguments against, like situation where this could be non-ideal?

bxdxnn:
@bmarty 2 months since the PR was created, I assume the current flow is OK with the product team, can we move forward with this PR?

bmarty:
### Some remarks

> Even if we were certain (we are with wired headsets)

This is not very true, you can have a loud speaker (a hi-fi system, a car, etc.) connected to the device with a wire.

> Another case, I might also have music playing on the external device, and I'd like to be able to listen to the voice message on the earpiece while the music is still playing and without having to disconnect anything first.

Are you sure that the device will be able to play music on the BT speaker and in the same time play a voice message on the earpiece? I doubt it, but I may be wrong

### My 2 cents

Now regarding the proposal itself, I sort of agree with @mxandreas , the use case is limited. If a user plays a voice message and the voice message is "accidentally" played loudly, then the user will have to stop the playback, and disconnect the BT headset for instance, that I agree is annoying. But will it happen that often? Another point is discoverability, will user will have the reflex to put the phone next their ear when this happen? Especially not easy if the phone is mounted on a car holder.

### Alternative solutions

- Maybe we could have an audio selector, as per Element Call but for voice message?
- Maybe we could have a setting to *always* play voice message in the earpiece for privacy reason?

bxdxnn:
> This is not very true, you can have a loud speaker (a hi-fi system, a car, etc.) connected to the device with a wire.

You might be lucky (the Android API is not very good regarding externally connected device types, this is also the reason why I recommend to not disable this feature even when external devices are connected) to distinguish between a loud speaker and a headset using `AudioDeviceInfo.TYPE_WIRED_HEADSET`/`AudioDeviceInfo.TYPE_WIRED_HEADPHONES` and     `AudioDeviceInfo.TYPE_LINE_ANALOG`/`AudioDeviceInfo.TYPE_AUX_LINE`/`AudioDeviceInfo.TYPE_USB_DEVICE` enums.



> Are you sure that the device will be able to play music on the BT speaker and in the same time play a voice message on the earpiece? I doubt it, but I may be wrong

I don't remember testing it, but it should be possible, though obviously might vary in different ROMs.

> But will it happen that often?

Well, as often as one would use external audio devices and receive voice messages at the same time

> Another point is discoverability, will user will have the reflex to put the phone next their ear when this happen? Especially not easy if the phone is mounted on a car holder.

AFAICT this is a very popular feature, and there's already a small delay (while the audio is downloading?) before the voice message gets played, so most users should reflex it. User might also reflex to put the phone out of the holder before starting playing the voice message to be able to hold the phone next to the ear faster.

bmarty:
For the first point, what I am saying is that a wired headset is not necessarily a private headset

bxdxnn:
> For the first point, what I am saying is that a wired headset is not necessarily a private headset

I've listed that it might actually not be a private headset after that part in the PR description:
> * It could be used as a speaker at maximum volume.
> * It might be in use by another person.
> * The user might forget he have an unused Bluetooth headset connected, which could lead to confusion or inconvenience when audio is unexpectedly absent.

bmarty:
To clarify: one can plug anything on the female jack socket, the OS cannot know what's connected, the technology does not allow this.

app/src/main/kotlin/io/element/android/x/di/AppGraph.kt
```diff
@@ -27,6 +29,11 @@ interface AppGraph : NodeFactoriesBindings {
     val workerProviders:
         Map<KClass<out ListenableWorker>, MetroWorkerFactory.WorkerInstanceFactory<*>>
 
+    @Provides
+    fun provideVoiceMessageAudioManager(
+        @ApplicationContext context: Context
+    ): VoiceMessageAudioManager? = DefaultVoiceMessageAudioManager(context)
+
```
bmarty: This code is not necessary, using `@ContributesBinding` on `DefaultVoiceMessageAudioManager` is enough, please remove.

libraries/mediaplayer/impl/src/main/kotlin/io/element/android/libraries/mediaplayer/impl/SimplePlayer.kt
```diff
@@ -49,7 +49,18 @@ object SimplePlayerModule {
     @Provides
     fun simplePlayerProvider(
         @ApplicationContext context: Context,
-    ): SimplePlayer = DefaultSimplePlayer(ExoPlayer.Builder(context).build())
+    ): SimplePlayer {
+        val exoPlayer = ExoPlayer.Builder(context)
+            .setAudioAttributes(
+                androidx.media3.common.AudioAttributes.Builder()
```
bmarty: Please use import statement
```
import androidx.media3.common.AudioAttributes
import androidx.media3.common.C
```

libraries/voiceplayer/api/src/main/kotlin/io/element/android/libraries/voiceplayer/api/VoiceMessageAudioManager.kt
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (c) 2025 Element Creations Ltd.
+ * Copyright 2023-2025 New Vector Ltd.
```
bmarty: Just `Copyright (c) 2026 Element Creations Ltd.` is enough

libraries/voiceplayer/impl/src/main/kotlin/io/element/android/libraries/voiceplayer/impl/DefaultVoiceMessageAudioManager.kt
```diff
@@ -0,0 +1,443 @@
+/*
+ * Copyright (c) 2025 Element Creations Ltd.
+ * Copyright 2023-2025 New Vector Ltd.
```
bmarty: 1 line with 2026 please.

libraries/voiceplayer/impl/src/main/kotlin/io/element/android/libraries/voiceplayer/impl/DefaultVoiceMessageAudioManager.kt
```diff
@@ -0,0 +1,443 @@
...
+ * - Proximity sensor always active (switches to earpiece when device is held to ear)
+ * - Turns off screen when device is held to ear
+ * - Uses TYPE_BLUETOOTH_SCO for Bluetooth communication (TYPE_BLUETOOTH_A2DP is not valid for
+ *   setCommunicationDevice() as it's for media playback only)
+ * - Proximity sensor remains active even when Bluetooth is connected
+ */
+@ContributesBinding(RoomScope::class)
+public class DefaultVoiceMessageAudioManager constructor(
+    @ApplicationContext private val context: Context,
+) : VoiceMessageAudioManager {
```
bmarty, line 49: The class declaration should look like:
```
class DefaultVoiceMessageAudioManager(
    @AppCoroutineScope private val appCoroutineScope: CoroutineScope,
    coroutineDispatchers: CoroutineDispatchers,
) : VoiceMessageAudioManager {
```

libraries/voiceplayer/impl/src/main/kotlin/io/element/android/libraries/voiceplayer/impl/DefaultVoiceMessageAudioManager.kt
```diff
@@ -0,0 +1,443 @@
...
+
+    private val proximitySensorMutex = Mutex()
+    private val audioDeviceCallbackMutex = Mutex()
+    private val proximityListenerMutex = Mutex()
+    private val deviceSelectionMutex = Mutex()
+
+    private var isRoutingActive = false
+    private var currentDeviceType: Int? = null
+
+    private val mainScope = MainScope()
```
bmarty: Please remove and use the `appCoroutineScope` provided in the constructor instead.

libraries/voiceplayer/impl/src/main/kotlin/io/element/android/libraries/voiceplayer/impl/DefaultVoiceMessageAudioManager.kt
```diff
@@ -0,0 +1,443 @@
...
+    private val audioDeviceCallbackScope = CoroutineScope(Dispatchers.Main)
```
bmarty:
```suggestion
    private val audioDeviceCallbackScope = appCoroutineScope.childScope(coroutineDispatchers.main, "VoiceMessageAudioManager.AudioDeviceCallback")
```

libraries/voiceplayer/impl/src/main/kotlin/io/element/android/libraries/voiceplayer/impl/DefaultVoiceMessageAudioManager.kt
```diff
@@ -0,0 +1,443 @@
...
+    /**
+     * Start proximity sensor monitoring.
+     */
+    private fun startProximityMonitoring() {
+        val sensor = sensorManager?.getDefaultSensor(Sensor.TYPE_PROXIMITY) ?: run {
+            Timber.w("Proximity sensor not available")
+            return
+        }
+
+        mainScope.launch(Dispatchers.Main) {
```
bmarty: No need to wrap in a coroutine here.

libraries/voiceplayer/impl/src/main/kotlin/io/element/android/libraries/voiceplayer/impl/DefaultVoiceMessageAudioManager.kt
```diff
@@ -0,0 +1,443 @@
...
+                SensorManager.SENSOR_DELAY_NORMAL
+            )
+        }
+    }
+
+    /**
+     * Stop proximity sensor monitoring.
+     */
+    private fun stopProximityMonitoring() {
+        mainScope.launch(Dispatchers.Main) {
```
bmarty: No need to wrap in a coroutine here.

libraries/voiceplayer/impl/src/main/kotlin/io/element/android/libraries/voiceplayer/impl/DefaultVoiceMessageAudioManager.kt
```diff
@@ -0,0 +1,443 @@
...
+
+            mainScope.launch {
+                proximityListenerMutex.withLock {
+                    if (!isRoutingActive) return@launch
+
+                    event?.let {
+                        val distance = it.values[0]
+                        val isClose = distance < 5.0f
+
+                        if (isClose) {
```
bmarty, line 418: `if` or `else` block will be run every time the sensor report a change. You should keep the previous value of `isClose` and act only if the valu change.

libraries/voiceplayer/impl/src/test/kotlin/io/element/android/libraries/voiceplayer/impl/DefaultVoiceMessagePlayerTest.kt
```diff
@@ -290,7 +292,9 @@ private fun createDefaultVoiceMessagePlayer(
         json = null
     ),
     mimeType = MimeTypes.Ogg,
-    filename = "someBody.ogg"
+    filename = "someBody.ogg",
+    voiceMessageAudioManager = null,
+    sessionCoroutineScope = CoroutineScope(coroutineContext)
```
bmarty: You can use `backgroundScope` here.

bmarty (changes requested):
I have noted several problems:
- stopRouting() is not invoked when the voice message reach the end of the playback.
- tested on my device, the proximity sensor does not work well, I'll check again.

On the form: please do not force push the branch, and push new commits, this way it is easier to review. Thanks!

bxdxnn:
> please do not force push the branch, and push new commits, this way it is easier to review.

BTW, IIRC if you press "Viewed" on the files when reviewing, you will see the difference between what you've reviewed previously and what has changed in the PR, regardless whether it was force-pushed or not.

bxdxnn:
> * tested on my device, the proximity sensor does not work well, I'll check again.

Works perfectly on a Pixel device, maybe you can share what doesn't work exactly?

bxdxnn:
@bmarty any updates? I won't be able to do an iOS PR for this, as it would require a physical iPhone for testing, unless a blindly vibecoded slop would be enough...

PR #6410, opened 2026-03-18
Fix `TransactionTooLargeExceptions` caused by Appyx

jmartinesp:
## Content

After a long debugging session, we discovered the code Appyx uses to clear the saved state of nodes that have been removed is not working because of a race condition, causing this saved state to grow indefinitely.

To fix it, we need to wait until the node has been disposed, which will call `SaveableStateHolder.removeState` once, removing the associated `SaveableStateRegistry`, and *then* call `removeState` again when we detect the node has been removed from the navigation graph.

Since these classes and APIs are private in Appyx, we had to copy and modify and use these copies.

## Motivation and context

Fix the long-standing issue causing `TransactionTooLargeExceptions` and crashes when using the app for a very long time.

## Tests

There's no easy way to test this, but you can check the logs to see when nav keys are disposed and its state removed, which should always be in that order.

Disposed keys may not have a removal operation after it, and this is caused because it's a child of another node that's being removed, which in turns removes the whole stack under it.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

### DIFF
--- a/libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/appyx/SafeChildrenTransitionScope.kt
+++ b/libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/appyx/SafeChildrenTransitionScope.kt
@@ -1,0 +1,267 @@
+/*
+ * Copyright (c) 2026 Element Creations Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.architecture.appyx
+
+import android.annotation.SuppressLint
+import androidx.compose.animation.ExperimentalSharedTransitionApi
+import androidx.compose.animation.SharedTransitionLayout
+import androidx.compose.foundation.layout.Box
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.Immutable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.derivedStateOf
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.key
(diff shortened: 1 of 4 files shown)


libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/appyx/SafeChildrenTransitionScope.kt
```diff
@@ -0,0 +1,267 @@
...
+        noinline block: @Composable ChildTransitionScope<S>.(
+            child: ChildRenderer,
+            transitionDescriptor: TransitionDescriptor<T, S>
+        ) -> Unit,
+    ) {
+        safeChildren(V::class, block)
+    }
+
+    @Composable
+    inline fun <reified V : T> ParentNode<T>.children(
```
ElementBot: - :warning: Composable functions that return Unit should start with an uppercase letter

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/appyx/SafeChildrenTransitionScope.kt
```diff
@@ -0,0 +1,267 @@
...
+
+@Immutable
+class SafeChildrenTransitionScope<T : Any, S>(
+    val transitionHandler: TransitionHandler<T, S>,
+    val transitionParams: TransitionParams,
+    val navModel: NavModel<T, S>
+) {
+
+    @Composable
+    inline fun <reified V : T> ParentNode<T>.children(
```
ElementBot, line 67: - :warning: Composable functions that return Unit should start with an uppercase letter

libraries/architecture/src/main/kotlin/io/element/android/libraries/architecture/appyx/SafeChildrenTransitionScope.kt
ElementBot, line 77: - :warning: Composable functions that return Unit should start with an uppercase letter

bmarty (approved):
Thanks!

PR #6412, opened 2026-03-18
Fix keyboard not auto-opening when editing a message

kalix127:
## Content

Remove the unreliable isKeyboardVisible() guard in SoftKeyboardEffect so the keyboard always opens when entering a special composer mode. Also request focus explicitly when setting text in edit mode.

## Motivation and context

When a user selects "Edit" on a message, the composer enters edit mode and the text is populated correctly, but the soft keyboard does not open automatically and the user has to manually tap the input field. The root cause was InputMethodManager.isAcceptingText returning true after the action sheet dismissed (even though the keyboard wasn't visible), causing showKeyboard() and onRequestFocus() to be skipped entirely.

## Screenshots / GIFs

None 

## Tests

1.  Send a text message in a room
2.  Long-press the message and select "Edit"
3. Verify the keyboard opens automatically without needing to tap the input field                                                                                                   
4. Send an image with a caption
5. Long-press the image and select "Edit caption"                                                                                                                                   
6. Verify the keyboard opens automatically                                                                       
7. Reply to any message and verify the keyboard opens automatically

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
@@ -706,14 +706,14 @@ class MessageComposerPresenter(
                     val draft = createDraftFromState(markdownTextEditorState, richTextEditorState)
                     updateDraft(draft, isVolatile = true).join()
                 }
-                setText(newComposerMode.content, markdownTextEditorState, richTextEditorState)
+                setText(newComposerMode.content, markdownTextEditorState, richTextEditorState, requestFocus = true)
             }
             is MessageComposerMode.EditCaption -> {
                 if (currentComposerMode.isEditing.not()) {
                     val draft = createDraftFromState(markdownTextEditorState, richTextEditorState)
                     updateDraft(draft, isVolatile = true).join()
                 }
-                setText(newComposerMode.content, markdownTextEditorState, richTextEditorState)
+                setText(newComposerMode.content, markdownTextEditorState, richTextEditorState, requestFocus = true)
             }
             else -> {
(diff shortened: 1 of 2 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
```diff
@@ -706,14 +706,14 @@ class MessageComposerPresenter(
...
-                setText(newComposerMode.content, markdownTextEditorState, richTextEditorState)
+                setText(newComposerMode.content, markdownTextEditorState, richTextEditorState, requestFocus = true)
             }
             is MessageComposerMode.EditCaption -> {
                 if (currentComposerMode.isEditing.not()) {
                     val draft = createDraftFromState(markdownTextEditorState, richTextEditorState)
                     updateDraft(draft, isVolatile = true).join()
                 }
-                setText(newComposerMode.content, markdownTextEditorState, richTextEditorState)
+                setText(newComposerMode.content, markdownTextEditorState, richTextEditorState, requestFocus = true)
```
jmartinesp, line 716: When I tested this in my test device, it focused on the composer, but it never displayed the keyboard, so it feels a bit broken: either we don't focus or, if we focus, we should display the keyboard IMO.

kalix127: I've updated the approach and now instead of calling showKeyboard on the root view (which I suspect might be ignored on some devices for non-input views), it now focuses the editor first, then uses view.post {} to call showKeyboard on the actually focused view via findFocus(). 
Could you try again and let me know if it helps?
  
Also, could you share what device/ROM you tested on and whether you were using the rich text editor or markdown mode?

jmartinesp (approved):
Thanks, those latest changes work fine for me.

PR #6420, opened 2026-03-19
Fix: "Reset identity" flow leaves backup disabled #5075

### LINKED ISSUE #5075: "Reset identity" flow leaves backup disabled
I think this probably requires you to be using a MAS homeserver.

STR (seems to be intermittent, may require a couple of attempts to reproduce)
 1. Sign in
 2. On "Confirm your identity" page, click "Can't confirm?" 
 3. Click "Continue reset"
 4. Click "Yes, reset now"
 5. On MAS page, click "Finish reset" 
    Image
 6. Observe "Identity reset successfully"
 7. Click X button
 8. Observe "Help improve" dialog. Click "OK".
At this point, the app is logged in, but keys are not being backed up.

andybalaam:
Fix https://github.com/element-hq/element-x-android/issues/5075

Don't cancel the resetOidc job in onStart or onDestroy of ResetIdentityFlowNode

Sometimes, when using MAS to log in, then choosing to reset identity, the reset process was being cancelled mid-way, preventing the `resetOidc` job from turning on Key Storage.

In my investigation of this, I found that the `onStart` and `onDestroy` methods were _both_ sometimes called when the `resetOidc` job was still running.

So this change prevents us from cancelling the `resetOidc` job in `onStart` and `onDestroy` (and therefore removes `onDestroy` completely).

I am concerned that we might leave `resetOidc` running if the user does not choose "Finish Reset" in MAS, but I think the right fix for this is to do something inside `resetOidc` (in the Rust) because I don't think we have anything in the Android world where we know it's time to kill `resetOidc` rather than just wait a little longer for it to complete. If anyone knows better, please advise me.

## Tests

I followed the testcase in https://github.com/element-hq/element-x-android/issues/5075 and observed that "Key storage" in Settings was sometimes disabled after login and reset (1 out of 5 times for me).

After the fix I ran through the same test 10 times and it worked every time: log in, reset identity, check Key storage in settings: it is enabled.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 15

## Checklist

- [n/a] UI change has been tested on both light and dark themes

### DIFF
--- a/features/securebackup/impl/src/main/kotlin/io/element/android/features/securebackup/impl/reset/ResetIdentityFlowNode.kt
+++ b/features/securebackup/impl/src/main/kotlin/io/element/android/features/securebackup/impl/reset/ResetIdentityFlowNode.kt
@@ -40,9 +40,7 @@ import io.element.android.libraries.matrix.api.encryption.IdentityOidcResetHandl
 import io.element.android.libraries.matrix.api.encryption.IdentityPasswordResetHandle
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.flow.collectLatest
-import kotlinx.coroutines.flow.filter
-import kotlinx.coroutines.flow.take
+import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.launch
 import kotlinx.parcelize.Parcelize
 import timber.log.Timber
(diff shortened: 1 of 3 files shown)


jmartinesp:
Looking at your changes made me realise how we could improve and simplify this, so I'll upload a few changes that should take care of only cancelling the reset job if the user navigates back or starts a new attempt.

jmartinesp:
@andybalaam can you test https://github.com/element-hq/element-x-android/pull/6420/commits/d9ead9972717a9ae0a186f5de60f8996013e53ad still works as expected? AFAICT it did when I tested it, but I might have missed some corner case.

Also it seems like this PR will need an extra review from someone in @element-hq/element-android-reviewers  now 😅 .

andybalaam:
> Looking at your changes made me realise how we could improve and simplify this, so I'll upload a few changes that should take care of only cancelling the reset job if the user navigates back or starts a new attempt.

Ah brilliant, thank you - I was hoping someone who understood what was going on would properly engage with this.

Do you know why the tests are failing? I don't know how to run them :-(

jmartinesp:
> Do you know why the tests are failing? I don't know how to run them :-(

Maestro is very unstable at the moment, but it's not a required check so we can merge even if it fails, don't worry. We just don't have the energy to fix it every time it breaks 😅 .

andybalaam:
> @andybalaam can you test [d9ead99](https://github.com/element-hq/element-x-android/commit/d9ead9972717a9ae0a186f5de60f8996013e53ad) still works as expected? AFAICT it did when I tested it, but I might have missed some corner case.

@jmartinesp I just tried it and it seems not quite right.

Here is the flow I am doing:

* Sign in manually
* "You're about to sign in" -> Continue
* MAS: "Continue to Element X dbg?" -> Continue
* (Wait for "You're about to sign in" to disappear)
* Can't confirm
* "Can't confirm? You'll need to reset..." -> Continue reset
* Yes, reset now
* MAS: "Reset your identity..." -> Finish reset
* MAS: "Identity reset successfully..." -> X in top left

What should happen now is that I am in the app (seeing "Help improve Element X dbg").

But instead I see "Can't confirm? You'll need to reset..." and it never goes away.

If I click the left arrow (back) and then "Can't confirm?" again, sometimes that immediately puts me into the app (seeing "Help improve Element X dbg").

jmartinesp:
After a long debugging session, it seems like https://github.com/element-hq/element-x-android/pull/6420/commits/8e7f16033968f73af063c23dd13b5f47d2b59676 should fix the issue. @bmarty  or @ganfra  could you take a look at this PR?

features/securebackup/impl/src/main/kotlin/io/element/android/features/securebackup/impl/reset/ResetIdentityFlowNode.kt
```diff
@@ -122,7 +107,13 @@ class ResetIdentityFlowNode(
     }
 
     private fun CoroutineScope.startReset() = launch {
+        // Instead of cancelling the reset job on every ON_START, we can do it before starting a new attempt
+        cancelResetJob()
         resetIdentityFlowManager.getResetHandle()
+            // We're only interested in the success/failure case, and we need this flow to stop by itself
+            // since each call to `startReset` will create a new one
+            .filter { it.isSuccess() || it.isFailure() }
+            .take(1)
```
ganfra: Maybe a `first { it.success() || it.isFailure()}` would make it even clearer?

features/securebackup/impl/src/main/kotlin/io/element/android/features/securebackup/impl/reset/ResetIdentityFlowNode.kt
```diff
@@ -135,21 +126,35 @@ class ResetIdentityFlowNode(
...
                             is IdentityPasswordResetHandle -> backstack.push(NavTarget.ResetPassword)
                         }
                     }
                     else -> Unit
                 }
-            }
+        }
+    }
+
+    override fun performUpNavigation(): Boolean {
```
ganfra, line 140: Is it different than using the lifecycle `onDestroy`?

jmartinesp: It seems like it is: I wasn't able to reproduce it, but @andybalaam mentioned that, when resetting the app like this, he sometimes saw the `onDestroy` reset operation being called if I'm not mistaken (config change maybe?). I can't really understand why, but using the up navigation instead seems be a good workaround.

andybalaam: Yes, that's right.

ganfra (approved):

PR #6424, opened 2026-03-20
Fix wakelock not stopping early when notifications are disabled

jmartinesp:
## Content

If notifications for a device are disabled when there is no connection with the HS, the push registration will still exist, so the device can still receive push notifications.

In that cases, we were running into an issue where the wakelock for push notifications was started immediately after receiving a push but was never stopped and it ran for 3 minutes until its timeout, keeping the device awake for no reason.

This patch changes `DefaultPushHandler` so if we don't need the wakelock it returns `false` and we can stop the wakelock early.

## Motivation and context

Fixes an issue reported by @bmarty  (rageshake [here](https://github.com/element-hq/element-x-android-rageshakes/issues/10530)).

## Tests

There are unit tests, but if you want to test it, receiving a push in this case should not keep the device awake long enough for it to get warm. Also, there should be extra logs.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist


bmarty (approved):
Thanks!

PR #6431, opened 2026-03-21
Natural media viewer swiping order

### LINKED ISSUE #4296: Strange swipe direction between media in room

### LINKED ISSUE #6350: Inconsistent swipe navigation in media gallery: left swipe should go forward, right swipe backward
### Your use case

**Describe the bug / Problem**
When viewing images in the media gallery (opened by tapping an image in the timeline), the swipe navigation direction is opposite to the standard convention used in virtually all other Android apps and galleries.

> Expected behavior (industry standard): Swiping right to left (←) should navigate to the next (forward) image. Swiping left to right (→) should navigate to the previous (backward) image.

> Current behavior in Element X: Swiping right to left (←) navigates to the previous image (backward), and swiping left to right (→) navigates to the next image (forward). This is the reverse of what users expect.

**Steps to reproduce**

- Open any room with multiple images in the timeline.
- Tap on an image to open the media gallery view.
- Try swiping left and right to navigate between images.

**Expected behavior**
Swiping should feel intuitive and match the user's mental model from other apps (gallery apps, Instagram, WhatsApp, etc.), where the direction of swipe matches the direction of navigation through the timeline.

**Actual behavior**
The navigation feels "backwards" and disorienting, requiring users to relearn muscle memory specifically for Element X.

**Why this matters**
This is a fundamental UX inconsistency. When every other app behaves the same way, deviating from that standard creates friction and makes the app feel less polished. Users should not have to think about which way to swipe.

Environment:
    OS: Android [15]
    App version: Element X [26.03.3]

_No response_

### Additional context

_No response_

Yes

bxdxnn:
## Content

Change the swiping order in the media viewer and update tests:
* Swipe from right to left goes to a newer media
* Swipe from left ro right goes to an older media

## Motivation and context

Fixes #6350
Fixes #4296

Right now, the swiping order in the media viewer is wrong:
* Swiping from right to left goes to an older media
* Swiping from left to right goes to a newer media

The new bevahior in this PR is more natural and a standard behavior in most of the apps.

## Tests

* Swipe from right to left goes to a newer media
* Swipe from left ro right goes to an older media

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
+++ b/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
@@ -120,11 +120,25 @@ class MediaViewerDataSource(
      */
     private fun buildMediaViewerPageList(groupedItems: List<MediaItem>) = buildList {
         // Filter out DateSeparator items, we do not need them for the media viewer
-        val groupedItemsNoDateSeparator = groupedItems.filterNot { it is MediaItem.DateSeparator }
-        pagerKeysHandler.accept(groupedItemsNoDateSeparator)
-        groupedItemsNoDateSeparator.forEach { mediaItem ->
+        val itemsNoDateSeparator = groupedItems.filterNot { it is MediaItem.DateSeparator }
+        // Separate loading indicators and media events
+        val loadingIndicators = itemsNoDateSeparator.filterIsInstance<MediaItem.LoadingIndicator>()
+        val mediaEvents = itemsNoDateSeparator.filterIsInstance<MediaItem.Event>()
+        // Determine backward and forward loading indicators
+        val backwardLoading = loadingIndicators.find { it.direction == Timeline.PaginationDirection.BACKWARDS }
+        val forwardLoading = loadingIndicators.find { it.direction == Timeline.PaginationDirection.FORWARDS }
+        // Build ordered list: backward loading, media events (oldest first), forward loading
+        // Media events are currently newest first, reverse to get oldest first
+        val orderedEvents = mediaEvents.reversed()
+        // Create new list of MediaItem in order: backwardLoading, orderedEvents, forwardLoading
+        val orderedItems = buildList {
+            backwardLoading?.let { add(it) }
+            addAll(orderedEvents)
+            forwardLoading?.let { add(it) }
+        }
+        pagerKeysHandler.accept(orderedItems)
+        orderedItems.forEach { mediaItem ->
             when (mediaItem) {
-                is MediaItem.DateSeparator -> Unit
                 is MediaItem.Event -> {
                     val sourceUrl = mediaItem.mediaSource().safeUrl
                     val localMedia = localMediaStates.getOrPut(sourceUrl) {
@@ -148,6 +162,7 @@ class MediaViewerDataSource(
                         pagerKey = pagerKeysHandler.getKey(mediaItem),
                     )
                 )
+                is MediaItem.DateSeparator -> Unit // already filtered out
             }
         }
     }.toImmutableList()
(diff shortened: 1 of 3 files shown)


jmartinesp:
I think this addresses https://github.com/element-hq/element-x-android/issues/6350 or https://github.com/element-hq/element-x-android/issues/4296, right? I'm not sure if reversing the order is the right call when opening the media viewer from both the timeline and the gallery though. 

@bmarty I see you assigned to one of these issues, WDYT?

bxdxnn:
> I think this addresses #6350 or #4296, right? I'm not sure if reversing the order is the right call when opening the media viewer from both the timeline and the gallery though.
> 
> @bmarty I see you assigned to one of these issues, WDYT?

Correct, this PR fixes these issues. Just added.

jmartinesp:
I wanted some feedback from @bmarty since it seems like he planned to work on a related task, but maybe I should ask @mxandreas or @americanrefugee instead: should we just reverse the order of items in the media viewer, or maybe change it conditionally depending on where this screen is started from?:

- From selecting a media item in the timeline: items to the left are older, items to the right are newer.
- From opening some media in the gallery of a room: items to the left are newer, items to the right are older.

---

Looking at other clients:
- Whatsapp keeps a consistent order: left is older, right is newer. It's what's done in this PR.
- Telegram on the other hand does what I proposed above. This makes more sense to me, but it's a subjective preference.

americanrefugee:
This is a tricky one...

I agree that when viewing media from the timeline that right=newer / left=older makes total sense for both platforms.

The sticky bit is when viewing the Media Browser on iOS vs. Android.
- On iOS, the gallery shows the most recent at the bottom of the list. Therefore, right=newer / left=older still makes sense
- But on Android, the gallery list items are reversed - the most recent is at the top of the list. Therefore, right=older / left=newer would be more natural

So my personal opinion is:
- Both platforms have a common swipe before for viewing media from the timeline (right=news / left=older)
- When viewing from the Media Browser, the swipe pattern depends on whether most recent items are at the bottom of the list (iOS) vs. the top (Android)

Would love to get thoughts from @mxandreas

americanrefugee:
To clarify, when I say for example "right=newer / left=older" I mean...

If viewing an image, the image to the right is newer and the image to the left is older. To see the image to the right you swipe left. To see the image to the left you swipe right.

Hopefully that makes sense 😅

mxandreas:
> When viewing from the Media Browser, the swipe pattern depends on whether most recent items are at the bottom of the list (iOS) vs. the top (Android)

Why is the ordering different in Android gallery? If we changed that order we'd have straightforward solution exactly like we currently have on iOS.

americanrefugee:
> Why is the ordering different in Android gallery? If we changed that order we'd have straightforward solution exactly like we currently have on iOS.

Because that's the native pattern on Android in general. To make them match would be a refactoring of the media gallery on Android. I'm not opposed to that, but I'm also not sure what the reaction would be from Android users.

In general we try and follow native patterns when possible.

mxandreas:
> Because that's the native pattern on Android in general.

Ok, if we do not want to change this, then I have no better ideas than what we discussed:

1) Opening from timeline should certainly follow the ascending order, e.g. older images on the left. It would be very weird for this not to be the case.
2) Opening from the gallery - it could follow the gallery order (e.g. older on the right), but I have a slight preference of doing the same as for timeline. Because once the user has the image open they probably forget about the previous context and if sometimes swiping on the same direction gives you newer and sometimes older image, then I think this may be more confusing.

americanrefugee:
FWIW I agree with you @mxandreas

jmartinesp:
So, just to confirm: are we ok with the behaviour change in this PR, reversing the order in both cases?

jmartinesp (approved):
Ok, this implementation was verified by product and design and works fine. Let me merge this into a branch of mine for some minor tweaks (I'd upload it to your forked repo, but I can't 🫤, maybe you need to enable [this](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)) and then it can be merged.

jmartinesp:
Actually, a rebase that fixes the failing screenshot tests would probably be good enough to get it merged. Let's see if it works automatically.

bxdxnn:
> Ok, this implementation was verified by product and design and works fine. Let me merge this into a branch of mine for some minor tweaks (I'd upload it to your forked repo, but I can't 🫤, maybe you need to enable [this](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)) and then it can be merged.

I have this enabled.
Another maintainer couldn't upload his changes too, but he [solved](https://github.com/element-hq/element-x-ios/pull/5384#issuecomment-4251142263) it by disabling Git LFS push.

PR #6433, opened 2026-03-22
Add floating date indicator while scrolling the timeline

### LINKED ISSUE #2270: Show Date on scroll
It would be great if element X shows the date of the messages while scrolling.

currently it's almost impossible to know when messages were sent in large groups.

In similar style to other apps like telegram.

Schildi chat next already does this, it shouldn't be too difficult to just merge the patch back to element X.
https://github.com/SchildiChat/schildichat-android-next

_No response_

### Additional context

_No response_

No

kalix127:
## Content
                                                                                                                          
Add a floating date badge that appears while scrolling the chat timeline, similar to WhatsApp. The badge shows the date  
of the currently visible messages, auto-updates as the user scrolls to different days, and fades out ~1.2s after
scrolling stops.                                                                                                         
                                                                                                                    
## Motivation and context          

Closes https://github.com/element-hq/element-x-android/issues/2270                                                       

When scrolling through a long chat history, users have no way to know which day's messages they're viewing without       
finding an inline day separator. This adds a floating overlay that provides constant date context while scrolling. 
                                                                                                                          
## Screenshots / GIFs                                                                                              
                                    
https://github.com/user-attachments/assets/97c32fba-315f-4826-93a9-20e6120af4eb

## Tests                                                                                                                 
                                                                                                                    
- Step 1: Open a room with messages spanning multiple days                                                               
- Step 2: Scroll up — badge appears centered below the top bar showing the current date
- Step 3: Stop scrolling — badge fades out after ~1.2s                                                                   
- Step 4: Resume scrolling — badge fades back in                                                                         
- Step 5: Scroll to a different day — badge text updates                                                                 
- Step 6: Scroll back to bottom (live) — badge does not appear                                                           
                                                                                                                          
## Tested devices                                                                                                        
                                                                                                                          
- [x] Physical                                                                                                     
- [ ] Emulator                     
- OS version(s):

## Checklist

- [x] Accessibility has been taken into account. See                                                                     
https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility                               


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
```diff
@@ -0,0 +1,139 @@
...
+                    return@derivedStateOf item.model.formattedDate
+                }
+            }
+            null
+        }
+    }
+
+    val isAtBottom by remember {
+        derivedStateOf {
+            lazyListState.firstVisibleItemIndex < 3 && isLive
```
jmartinesp, line 83: Was 3 picked at random or does it have to do with the 'typing...' and other virtual items?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+
+    var isBadgeVisible by remember { mutableStateOf(false) }
+
+    LaunchedEffect(Unit) {
+        snapshotFlow { lazyListState.isScrollInProgress }
+            .distinctUntilChanged()
```
jmartinesp: `distinctUntilChanged` shouldn't be necessary when using `snapshotFlow` AFAICT.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
```diff
@@ -0,0 +1,139 @@
...
+            .collectLatest { isScrolling ->
+                if (isScrolling) {
+                    isBadgeVisible = true
+                } else {
+                    delay(1200.milliseconds)
```
jmartinesp: That's an oddly specific delay 😅 .

jmartinesp:
From my personal viewpoint, I'd love to have this included. In fact, I tried to start a discussion about this some time ago, but it didn't reach a conclusion.

@mxandreas , what do you think?

mxandreas:
> what do you think

Yes, this makes sense and looks straightforward to me. @americanrefugee any concerns from the design side?

americanrefugee:
Is the text in the floating pill the same size as the text in the timeline? It's hard to tell in the video.

Otherwise it looks great!

jmartinesp:
> Is the text in the floating pill the same size as the text in the timeline? It's hard to tell in the video.

It's actually smaller, it uses `fontBodySmMedium` while the one in the timeline uses `fontBodyMdMedium`

americanrefugee:
> It's actually smaller, it uses `fontBodySmMedium` while the one in the timeline uses `fontBodyMdMedium`

Can you please change it to the same size as in the timeline? So, fontBodyMdMedium instead of Sm

kalix127:
> Can you please change it to the same size as in the timeline? So, fontBodyMdMedium instead of Sm

Done it

americanrefugee:
Actually, one more change (very sorry)
- Please change the background color to **bg/canvas/default** instead of bg/subtle-secondary

Meanwhile, here's the component in Compound: https://www.figma.com/design/G1xy0HDZKJf5TCRFmKb5d5/Compound-Android-Components?node-id=4984-6676&t=J4n38AxGm5xjoO2I-4

americanrefugee:
Ugg, please don't hate me...

- Make sure the final badge is 36px tall
- That means there should be 8px between the text and top/bottom edges, and 16px on the left/right edges

I promise that this is the last thing!!!

kalix127:
> I promise that this is the last thing!!!

 dont worry about it at all! whatever it takes to get it merged, im using the app every day, everyone i know uses it too, and ill for sure keep using it for years. also ill go back later and re-check all my other UI PRs to make sure they actually match the compound guidelines

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
```diff
@@ -0,0 +1,139 @@
...
+
+@Composable
+internal fun FloatingDateBadge(
+    dateText: String,
+    modifier: Modifier = Modifier,
+) {
+    Surface(
+        modifier = modifier,
+        shape = RoundedCornerShape(16.dp),
+        color = ElementTheme.colors.bgCanvasDefault.copy(alpha = 0.85f),
```
jmartinesp: @americanrefugee just to double check, are we ok with using this color at 85% opacity for the background of the badge?

jmartinesp (commented):
A couple of comments, looks fine otherwise. I'll leave the final review to @bmarty.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
kalix127: Not random, it's the same check used for canAutoScroll in TimelineView.kt:293

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
kalix127: Fixed

americanrefugee:
Also, this should show in the Thread view.

kalix127:
> Also, this should show in the Thread view.

To be honest, I was thinking of implementing a proper thread section. The UI should be like Element Desktop, where it looks like a separate chat instead of just saying "thread" above the message, which is pretty useless and also ugly (in my opinion).

What do you think?
I could start working on it as early as tomorrow.

manuroe:
Thanks @kalix127 for all the changes.

>  The UI should be like Element Desktop, where it looks like a separate chat 

This kind of display is already there but not enabled by default. If you enable the threads feature from the labs settings, the app will display threads in a separate timeline view.

jmartinesp:
@americanrefugee I applied some of the changes in the Figma file locally, but are you sure you want to use the same background than on the actual timeline UI? For the light mode it's ok because the shadow makes the floating label stand out, but on dark mode the shadow can't be seen at all, so everything looks plain and it's sometimes impossible to tell you're seeing 2 overlapping elements:


https://github.com/user-attachments/assets/795ef98f-8dcf-4bab-887e-4199d492d1b7

americanrefugee:
> @americanrefugee I applied some of the changes in the Figma file locally, but are you sure you want to use the same background than on the actual timeline UI? For the light mode it's ok because the shadow makes the floating label stand out, but on dark mode the shadow can't be seen at all, so everything looks plain and it's sometimes impossible to tell you're seeing 2 overlapping elements:

Fair point :) Please change the bg for **Dark mode** to **bg/subtle-primary**. So it will be an asymmetrical color between light and dark modes.

americanrefugee:
Also, a couple notes based on what we're going to do on iOS:
- By default do NOT show the floating date indicator
- Once the user starts scrolling the timeline, then show the indicator (if there are pinned messages, the pinned message banner will scroll away)
- When scrolling has stopped, the indicator should then fade away after 2 seconds

This way the indicator is only shown on scroll and isn't always in your face / covering content :)

americanrefugee:
One more point just to be clear...

- Although the pinned messages banner slides away when scrolling up, it slides back in when scrolling down. The floating date indicator should therefore move below the pinned messages banner when scrolling down.

Does that make sense?

kalix127:
> Does that make sense?

Yes it does, I will work on it later

kalix127:
> This kind of display is already there but not enabled by default. If you enable the threads feature from the labs settings, the app will display threads in a separate timeline view.

Why is the thread view disabled by default?

kalix127:
Based on the feedback, here's what was addressed:                                                                              
                                                                                                                                 
  - Changed the dark mode background to bgSubtlePrimary so the badge stands out against the timeline, while keeping bgCanvasDefault with 0.85 alpha in light mode where the shadow provides enough contrast                                        
  - Badge now positions dynamically below the pinned messages banner when it's visible, and moves back to the top when the banner slides away                                                                                                         
  - Fade timeout changed from 1.2s to 2s to match iOS behavior

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
```diff
@@ -119,7 +121,11 @@ internal fun FloatingDateBadge(
     Surface(
         modifier = modifier,
         shape = RoundedCornerShape(16.dp),
-        color = ElementTheme.colors.bgCanvasDefault.copy(alpha = 0.85f),
+        color = if (ElementTheme.isLightTheme) {
+            ElementTheme.colors.bgCanvasDefault.copy(alpha = 0.85f)
+        } else {
+            ElementTheme.colors.bgSubtlePrimary
+        },
```
jmartinesp: We usually keep this in a list of custom color aliases in `io.element.android.libraries.designsystem.theme.ColorAliases`. While it's not mandatory, it'd be nice to have this added there as another `val SemanticColors.***` (`floatingDateBadgeBackground` maybe?).

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesView.kt
```diff
@@ -464,6 +465,9 @@ private fun MessagesViewContent(
             val scrollBehavior = PinnedMessagesBannerViewDefaults.rememberScrollBehavior(
                 pinnedMessagesCount = (state.pinnedMessagesBannerState as? PinnedMessagesBannerState.Visible)?.pinnedMessagesCount() ?: 0,
             )
+            var pinnedBannerHeightPx by remember { mutableIntStateOf(0) }
+            val pinnedBannerHeightDp = with(LocalDensity.current) { pinnedBannerHeightPx.toDp() }
```
jmartinesp: Nit: I think you can probably just store the `pinnedBannerHeightDp` using `by remember { mutableStateOf(0.dp) }`, keeping a reference to `LocalDensity.current` in a variable and then using `with(density) { pinnedBannerHeightDp = it.height.toDp() }` inside `.onSizeChanged { ... }`, removing `pinnedBannerHeightPx` completely.

I'm not sure if it's worth it, but it seems a tiny bit more efficient since this would only be calculated when the size changes instead of on every composition.

jmartinesp (approved):
I added a couple of nitpicks you can address if you want to, but not mandatory IMO. @bmarty any other feedback since you were the original reviewer?

jmartinesp:
iOS decided to merge their equivalent behind a FF, so we'll do the same. I'll merge this into a branch of mine where I'll add that feature flag, then merge that one when it's ready. Thanks again!

bmarty (commented):
LGTM, thanks! Waiting for Jorge PR which will add the FF.
I'll also double check what happen when the screen reader is enabled: in this case the timeline is rendered in the reverse order.

PR #6434, opened 2026-03-22
Fix media seeking flicker

bxdxnn:
## Content

Fix media player cursor flickering

## Motivation and context

When seeking media, the player cursor will flicker to the previous position, and then seek.

## Tests

- Try to seek media

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/local/player/MediaPlayerControllerState.kt
+++ b/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/local/player/MediaPlayerControllerState.kt
@@ -18,7 +18,16 @@ data class MediaPlayerControllerState(
     val durationInMillis: Long,
     val canMute: Boolean,
     val isMuted: Boolean,
+    val seekingToMillis: Long?,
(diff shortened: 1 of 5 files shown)


libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/local/player/MediaPlayerControllerState.kt
```diff
@@ -18,7 +18,11 @@ data class MediaPlayerControllerState(
     val durationInMillis: Long,
     val canMute: Boolean,
     val isMuted: Boolean,
+    val seekingToMillis: Long? = null,
```
jmartinesp: We have some rules to avoid default values in data classes, so the quality checks are failing. Could you make this always take a value and initialize it with `null`?

jmartinesp (changes requested):
There are some Konsist issues. Otherwise LGTM. Also, it would be nice to have some comments about what `seekingToMillis` is and how it's used or when it's reset.

jmartinesp (approved):
Thanks!

PR #6440, opened 2026-03-23
Add media player notification widget and allow background playback

bxdxnn:
## Content

Add a MediaStyle notification for the media player and allow background playback

## Motivation and context

Media playback stops when the app goes to background because MediaVideoView and MediaAudioView explicitly pause/release ExoPlayer on lifecycle events. This adds a MediaSessionService that owns the ExoPlayer so playback survives backgrounding.

## Tests

- Test that all notification buttons are working
- Test background playback in-app
- Test background playback when not in-app
- Test that you can go back to the currently played file: tap on the notification or on the file in the timeline
- Test that the in-app player and the notification are synced and are correctly reflecting the current state 

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist


bxdxnn:
@bmarty FYI, this was done almost fully by LLM

bmarty:
Thanks for the PR. I am not sure if we want to have such feature in the application. Can you post a video demoing it so I can ask the product team? Thanks

bxdxnn:
> Thanks for the PR. I am not sure if we want to have such feature in the application. Can you post a video demoing it so I can ask the product team? Thanks

Hello, I can't make a demo video as there are some technical bugs and minor design questions regarding the metadata in the notification since the last rebase, but I can continue if the main behavior will be accepted by the product team.
The main behavior changes will be:
1. Show an Android media notification widget that allows to control the player and the output device, similar to VLC and other audio players (see the example screenshot)
2. The player will automatically continue to the next media in the timeline when the app is in background or the player is closed, creating a playlist experience
3. The playback continues if you close the player or the app goes background

If this feature won't be accepted, can we allow background playback at least just when the app goes background? It'll be also great to be able to listen to media while you're in the app and the player is hidden, similar to Telegram, but this leads to the need of this feature because there would be no player controls without this PR.

Example screenshot of the widget (no cover art, in a DM):
Screenshot

bmarty:
Thanks for your reply. We are building a messaging application and not a media player. The application is able to play media (video, voice message) when in the foreground as a facility, but I believe that users expect all media playing to just stop when the app is put into background. But I let @mxandreas give the final words here.

mxandreas:
I have checked the behavior on WhatsApp and Signal, and they do continue playing whatever was playing when you lock the screen or switch to another app, but they stop once the current file is finished. Thus, this as a behavior is fine as long as it can be done with what the OS/system provides and we do not need to design for any of the audio players, etc.

I am curious though to understand in which situations are you @bxdxnn seeing this as a friction personally - what are you listening to, why is it challenging to keep the app in the foreground, etc. And in particular why would you even like to continue with the next/previous file in the timeline?

Also, for the next PRs which are about new features (visual or not), please open a a Enhancement Request issue in the https://github.com/element-hq/element-meta/issues repo - describing the the use case and the why behind the feature. This way you can get feedback earlier if/what can we have in the product before we start reviewing the code which is always more effort.

bxdxnn:
@mxandreas 

> I have checked the behavior on WhatsApp and Signal, and they do continue playing whatever was playing when you lock the screen or switch to another app, but they stop once the current file is finished. Thus, this as a behavior is fine as long as it can be done with what the OS/system provides and we do not need to design for any of the audio players, etc.

Do you want me to only implement the background widget (allowing background playback), without any other behavioral/UI changes to the app (cover arts, metadata, automatic forward/previous while in background, etc.)?

> I am curious though to understand in which situations are you @bxdxnn see this as a friction personally - what are you listening to, why is it challenging to keep the app in the foreground, etc. And in particular why would you even like to continue with the next/previous file in the timeline?

I want to create a read-only channel in a space where I publish my (~1 hour duration) audio books. Continuing it makes more convenient to listen to all medias in the room.
It's also challenging to listen to any media which's length exceeds screen timeout.

> Also, for the next PRs which are about new features (visual or not), please open a a Enhancement Request issue in the https://github.com/element-hq/element-meta/issues repo - describing the the use case and the why behind the feature. This way you can get feedback earlier if/what can we have in the product before we start reviewing the code which is always more effort.

* Many issues from outside contributors don't get feedback until there's a PR for it.
  * For example, one of the most liked features (`max_selections` polls https://github.com/element-hq/element-meta/issues/2694) didn't get any feedback for almost 4 years until I've made a PR for it.
* Making a PR makes it faster to proceed on an issue, any discussions can be easily resolved later.
* It makes it easier to discuss when you can already test the implementation and how it'll work.

bmarty:
> It's also challenging to listen to any media which's length exceeds screen timeout.

We should have a wake lock when playing media to mitigate this. I though we had one, but it's apparently not the case.

bxdxnn:
The widget has an ability to show the cover art, but what should be shown when there's no cover art for an audio?
* Nothing (current behavior)
* Room's avatar
* Sender's avatar

mxandreas:
> We should have a wake lock when playing media to mitigate this. I though we had one, but it's apparently not the case.

Good point, @bmarty. Do you have any strong arguments for "wake lock" vs "continue in the background" for this scenario?

> The widget has an ability to show the cover art, but what should be shown when there's no cover art for an audio?

I'd say the first iteration would be fine probably without anything, but if it easy to make, then probably sender's avatar. @americanrefugee do you agree?

americanrefugee:
> @mxandreas : I'd say the first iteration would be fine probably without anything, but if it easy to make, then probably sender's avatar. @americanrefugee do you agree?

Yes, that's fine with me.

mxandreas:
> I have checked the behavior on WhatsApp and Signal, and they do continue playing whatever was playing when you lock the screen or switch to another app, but they stop once the current file is finished.

I have to correct myself regarding this, in fact they continue playing, but there are some differences in behaviour, see details in https://github.com/element-hq/element-x-android/pull/6454#issuecomment-4213417299

bxdxnn:
I have removed the playlist (play next timeline media after ending) feature and the notification now uses sender's avatar for the cover art, addressing the product's decision review. Waiting for code review now.

About WA/Signal behavior: I'm not sure what it means here, auto-playing voice messages is a different case from auto playing media or playing it in background.

bmarty:
I checked and we actually have a screen wake lock only when playing a video. The Composable is `KeepScreenOn()`. It is not used for audio playback (voice messages or audio files), so yes, this is annoying. We could also keep the screen on when playing audio files, even if I recognize that this is a bit wrong to keep the screen on only for audio playback, but this is maybe a more simpler fix.

bxdxnn:
> We could also keep the screen on when playing audio files, even if I recognize that this is a bit wrong to keep the screen on only for audio playback, but this is maybe a more simpler fix.

IMO, this would still won't work if I'd like to listen with the screen off or while in another app.

This PR just utilizes Android media features as said, I haven't modified the existing player UI.

bmarty:
To summarize the discussion above, I think what we expect is:

- a service that handle the playback of any voice message and audio attachment, so that the playback continue while the screen goes off, or the app is moved into background
  - no playlist management. At the end of the current audio playback, the service stops
  - There is also the sign out case to handle when this happen during an audio playback
  - other case, maybe stop the playback when the user switches between accounts?
- no change on the current behavior of video files: screen is kept on, but playback stop if the user lock the phone or navigate away

I have also noticed this behavior of WhatsApp: when a voice message is played from a channel then a user navigates to the channel list, a player UI is rendered at the top of the channel list. But in this case the playback is paused when the screen goes of. But maybe we can keep it simple for now and just continue the playback in all cases.

In the mean time I will start adding comments to the PR

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/local/player/MediaPlaybackContext.kt
```diff
@@ -0,0 +1,20 @@
...
+package io.element.android.libraries.mediaviewer.impl.local.player
+
+import androidx.compose.runtime.compositionLocalOf
+import io.element.android.libraries.matrix.api.media.MediaSource
+
+data class MediaPlaybackContext(
+    val sessionId: String = "",
+    val roomId: String = "",
+    val eventId: String = "",
+    val thumbnailSource: MediaSource? = null,
```
bmarty, line 20: Use type `SessionId`, `RoomId`, and `EventId` instead of String? This is a global remark for the PR, we avoid as much as possible using String for matrix ids, using typealias reduce errors and help for readability

bmarty (commented):
Looks like this PR also contains the code from #6545. Can you cleanup please?

Also the code is not compiling, I let you check the errors, I have triggerd the CI (probably due to rebase)

I have test the code with a mp3, and when the playback is started, app in put into background, the audio file keeps playing. This is fine. When when I click on the notification widget, the playback is paused. I think that it should keep playing.

On the form, can I ask you to not squash commits and force push once you take into account the remarks? That way we can easily see the diff with the previous review. Thanks!

bxdxnn:
> On the form, can I ask you to not squash commits and force push once you take into account the remarks? That way we can easily see the diff with the previous review. Thanks!

Sorry, the last rebase seems to went totally wrong, causing compilation errors and that random commit. I'll have to force push, but I'll address your review comments in a separate commit, so the base code that you've reviewed should remain the same.

> * no playlist management. At the end of the current audio playback, the service stops

This makes it inconvenient to be able to go back to the room/message of the file that was playing, or to press previous/forward to play other files, or to just play the file again.

Since we shouldn't have "playlist management", I assume that I should totally remove the previous/forward buttons from the notification then? They're the only "playlist management" feature. The idea was to be able to "swipe" media from the notification, similar how you can already do it from the app's media viewer.

mxandreas:
> Since we shouldn't have "playlist management", I assume that I should totally remove the previous/forward buttons from the notification then?

As per my comment in here: https://github.com/element-hq/element-x-android/pull/6454#issuecomment-4213417299

* We can continue playing the next file automatically or do this later.
* Regardless of the above, we can have the previous/next buttons or do this later.

(I'd rather keep the PR smaller and then do another on top)

bxdxnn:
> * We can continue playing the next file automatically or do this later.

That was the original behavior, but I've removed it. I'll revert it to play the next file automatically then.

> * Regardless of the above, we can have the previous/next buttons or do this later.

Already done, so there's nothing to do.

PR #6453, opened 2026-03-24
Increase icon size of audio and files in the timeline

bmarty:
Increase icon size of audio and files in the timeline to 36 for container and 24 for the icon.

Requested by @americanrefugee .

Also remove the rotation applied to the attachement when it's been replied to, see https://www.figma.com/design/G1xy0HDZKJf5TCRFmKb5d5/Compound-Android-Components?node-id=2019-6286


jmartinesp (approved):
It looks way better now, thanks!

PR #6454, opened 2026-03-24
Add auto-play consecutive voice messages in timeline

### LINKED ISSUE #4618: Auto-play voice messages, if there are multiple of them

When I receive several voice messages at the same time, and I start playing the first one: I would like the second one to start playing as soon as the first one ended, etc. up to the last audio message.

If there's a text message in between two audio messages, I'd suggest implementing the same behavior as Signal: interrupt the auto play of the next message.

It's nice to not have to press play for every single audio message I receive.

_No response_

### Additional context

_No response_

No

kalix127:
## Content

Add auto-play for consecutive voice messages in the timeline. When a voice message finishes playing, if the next chronological message is also a voice message, it automatically starts playing. The chain continues until a non-voice message is encountered or the end of the timeline is reached.

Key changes:
- New `VoiceMessageAutoplayManager` interface (voiceplayer API) and `DefaultVoiceMessageAutoplayManager` implementation (voiceplayer impl, RoomScope singleton)
- Manager observes `MediaPlayer.state` for `isEnded` transitions and triggers playback of the next consecutive voice message
- `VoiceMessagePresenter` listens for reset requests and cancels autoplay on manual pause
- `TimelinePresenter` feeds current timeline items to the autoplay manager on each update
- Virtual items (day separators, read markers) are skipped when checking adjacency

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/4618

This is a standard feature in almost every modern messaging app.

## Tests

- 1. Send 2+ voice messages in a row in a test room
- 2. Play the first one, when it ends, the next should auto-start
- 3. Insert a text message between voice messages, autoplay should stop at the boundary
- 4. During autoplay, manually tap pause, chain should stop
- 5. During autoplay, tap a different voice message, chain should cancel and play the tapped one
- 6. Replay the first voice message after a completed chain, autoplay should work again
- 7. Scroll voice messages off-screen during autoplay, playback should continue unaffected

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 15

## Checklist

---

## Open questions

@jmartinesp

1. **Transition sound:** Currently there is no sound effect played between consecutive voice messages during autoplay. Open to suggestions on whether we should add one (e.g. a short tick/beep) and if so, what would fit best.
2. **User setting:** Autoplay is always enabled for now. Open to feedback on whether we should add an on/off toggle in user settings.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
@@ -32,6 +32,7 @@ import io.element.android.features.messages.impl.timeline.factories.TimelineItem
 import io.element.android.features.messages.impl.timeline.factories.TimelineItemsFactoryConfig
 import io.element.android.features.messages.impl.timeline.model.NewEventState
 import io.element.android.features.messages.impl.timeline.model.TimelineItem
+import io.element.android.features.messages.impl.timeline.model.event.TimelineItemVoiceContent
 import io.element.android.features.messages.impl.timeline.model.virtual.TimelineItemTypingNotificationModel
 import io.element.android.features.messages.impl.typing.TypingNotificationState
 import io.element.android.features.messages.impl.userEventPermissions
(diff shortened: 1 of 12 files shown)


bxdxnn (approved):
LGTM, but the common behavior is to also play a sound between messages before playing them.

kalix127:
> LGTM, but the common behavior is to also play a sound between messages before playing them.

I know, in fact I asked for feedback and suggestions on which sound to use. Honestly, the options are endless.

kalix127:
In the meantime, I've added code to play a sound when it moves on to the next audio recording. Right now, it uses the same default notification sound, but all you have to do is replace it with a different sound, and then it's all set.

@americanrefugee , do you have any suggestions for a sound to use?
I've tried changing the notification sound in a few different ways, but with little success.

americanrefugee:
> @americanrefugee , do you have any suggestions for a sound to use?

Actually I think @amshakal would be a better person to ask since she reviewed the notification sounds in general recently.

mxandreas:
> This is a standard feature in almost every modern messaging app.

Regarding this, I checked WhatsApp and Signal, and this is what you have in there:

1. If the app is in the foreground, they both continue playing the next voice message.
2. If the app is not in the foreground, Signal continues playing but not WhatsApp.

None of them seem to have any preferences to control or change that behavior.

There is not a ton of evidence and specific references but it seems like at least some WhatsApp users are quite unhappy that they do not get continuous playing while the app is not in the foreground, especially as it used to be this way in the past.

bmarty:
FTR iOS PR has been created at https://github.com/element-hq/element-x-android/pull/6454

bxdxnn:
> FTR iOS PR has been created at #6454

https://github.com/element-hq/element-x-ios/pull/6089 actually...

mxandreas:
Created a story for this: https://github.com/element-hq/element-meta/issues/3308

PR #6469, opened 2026-03-25
Allow pinning rooms to the top of the room list

### LINKED ISSUE #6466: Allow pinning rooms to the top of the room list
Pin specific rooms so they always appear at the top of the room list, with the most recently pinned room first.

The Favourites filter hides everything that isn't a favourite when active. Pinning is a different thing: pinned rooms just stay at the top no matter what you're looking at. Filter by "People"? Your pinned DMs are at the top. Filter by "Rooms"? Your pinned groups are at the top. No filter? Still at the top. Favourites is about marking rooms you like, pinning is about keeping rooms you need right now always visible.

Long-press a room and there's a "Pin to top" toggle in the context menu, right next to the Favourite toggle. You can also pin/unpin from the Room Details screen. Pinned rooms show a small pin icon next to the timestamp so you can tell at a glance. The most recently pinned room goes to the very top of the pinned ones. Storage is local per session using DataStore since the Matrix SDK doesn't support custom room tags.

Using the existing Favourites feature, but it's a filter (hides non-favourite rooms) not a sort modifier. The two serve different purposes and should coexist.

### Additional context

_No response_

Yes

kalix127:
## Content

Add the ability to pin rooms to the top of the room list. Pinning only affects sort order, not visibility: pinned rooms that match the active filter appear first, but pinned rooms that don't match the filter are still hidden. The most recently pinned room appears at the very top. A small pin icon shows next to the timestamp on pinned rooms.

Pin/unpin is available from:
- Long-press context menu on the room list (next to the Favourite toggle)
- Room Details screen (next to the Favourite toggle)

Pinned room IDs are stored locally per session using DataStore since the Matrix Rust SDK does not expose custom room tags.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/6466

## Screenshots / GIFs

https://github.com/user-attachments/assets/5a98f8f9-2123-4952-ac64-694382053bed

## Tests

- Long-press a room, tap "Pin to top", verify it moves to the top of the list with a pin icon
- Pin multiple rooms, verify the most recently pinned one is at the very top
- Apply filters (People, Rooms, Unread), verify pinned rooms float to the top within the filtered set
- Apply a filter that excludes a pinned room (e.g. filter "Rooms" when the pinned room is a DM), verify it doesn't show
- Unpin a room via context menu, verify it goes back to its normal position
- Pin/unpin from Room Details screen, verify it works the same way
- Verify the pin toggle switch updates immediately in the context menu (no need to close/reopen)

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 15

## Checklist


kalix127:
Updates on this PR?

bmarty:
Yes, see @mxandreas comment here: https://github.com/element-hq/element-x-android/issues/6466#issuecomment-4214014013 , we will address this topic using sections.

PR #6470, opened 2026-03-25
Fix `ForegroundServiceDidNotStartInTimeException`

jmartinesp:
## Content

- Move starting the foreground service to `FetchPushForegroundService.onCreate` so it's called ASAP.
- When stopping the service, wait for it to be in foreground for up to 5s: another cause of this exception is the service being stopped when it was running but not in foreground yet.

## Motivation and context

Fix crash spotted in the wild with `ForegroundServiceDidNotStartInTimeException`.

## Tests

You can maybe test everything else works as expected?

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

    - [x] No.

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/FetchPushForegroundService.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/FetchPushForegroundService.kt
@@ -25,6 +25,7 @@ import io.element.android.libraries.ui.strings.CommonStrings
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
+import timber.log.Timber
 import kotlin.time.Duration.Companion.minutes
 
 private const val NOTIFICATION_ID = 1001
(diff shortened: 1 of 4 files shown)


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/FetchPushForegroundService.kt
```diff
@@ -107,9 +125,34 @@ class FetchPushForegroundService : Service() {
...
+                    while (!isInForeground) {
+                        delay(50)
+                        val updatedServiceInfo = getRunningServiceInfo(context)
+                        if (updatedServiceInfo == null) {
+                            Timber.d("FetchPushForegroundService is no longer running, no need to stop it.")
+                            return@withTimeoutOrNull
+                        }
+                        isInForeground = updatedServiceInfo.foreground == true
+                    }
+                }
```
bmarty: Maybe log something when there has been a timeout?

bmarty (approved):
Thanks for hunting this bug!

PR #6482, opened 2026-03-27
Add support for slash commands (under Feature Flag)

bmarty:
## Content

This is adding support for a few slash command.
There is a feature flag, which is not enabled by default.
Some command are working, some are not supported.
Some command may behave wrongly when the rich text editor is used.

## Motivation and context

Implement slash commands

## Screenshots / GIFs

2 new screenshot for the command suggestion.

## Tests

- Enable the flag
- Send a few command:
   - "/me is sending an emote"
   - "/leave" to leave the room
   - etc.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

    - [ ] No.


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
```diff
@@ -125,10 +132,15 @@ class MessageComposerPresenter(
...
     private val notificationConversationService: NotificationConversationService,
+    private val slashService: SlashService,
 ) : Presenter<MessageComposerState> {
     @AssistedFactory
     interface Factory {
-        fun create(timelineController: TimelineController, navigator: MessagesNavigator): MessageComposerPresenter
+        fun create(
+            timelineController: TimelineController,
+            navigator: MessagesNavigator,
+            isInThread: Boolean,
```
bmarty, line 142: Note: isInThread here seems to be safer than the current way to detect if the MessageComposer is writing to a thread. We may want to fix potential issue regarding that.

jmartinesp:
Does the WIP part in the title mean I should not review it yet? 😅

features/location/impl/src/test/kotlin/io/element/android/features/location/impl/share/ShareLocationPresenterTest.kt
```diff
@@ -6,6 +6,8 @@
  * Please see LICENSE files in the repository root for full details.
  */
 
+@file:OptIn(ExperimentalCoroutinesApi::class)
```
jmartinesp, line 9: This is probably some leftover code?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/util/MessageEventContent.kt
```diff
@@ -10,19 +10,52 @@ package io.element.android.libraries.matrix.impl.util
...
-            messageEventContentFromHtml(body, htmlBody)
-        } else {
-            messageEventContentFromMarkdown(body)
-        }.withMentions(intentionalMentions.map())
+    fun from(
+        body: String,
+        htmlBody: String?,
+        intentionalMentions: List<IntentionalMention>,
+        asEmote: Boolean = false,
+        asPlainText: Boolean = false,
```
jmartinesp: Maybe it should be 2 methods instead? `text` and `emote`?

libraries/slashcommands/api/src/main/kotlin/io/element/android/libraries/slashcommands/api/SlashCommand.kt
```diff
@@ -0,0 +1,73 @@
...
+
+package io.element.android.libraries.slash.api
+
+import io.element.android.libraries.matrix.api.core.RoomIdOrAlias
+import io.element.android.libraries.matrix.api.core.UserId
+
+/**
+ * Represent a slash command.
+ */
+sealed interface SlashCommand {
```
jmartinesp, line 16: I haven't fully tested it yet, so I'm not sure if this is done elsewhere and I might have missed it, but should we have an `isValid` property or something similar to know if a command can be used in EX?

libraries/slash/api/src/main/kotlin/io/element/android/libraries/slash/api/SlashService.kt
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.slash.api
+
+import io.element.android.libraries.matrix.api.timeline.Timeline
+
+interface SlashService {
```
jmartinesp: `SlashCommandService`? Also, the module being `slash` is not that descriptive, IMO.

libraries/slashcommands/impl/src/main/kotlin/io/element/android/libraries/slashcommands/impl/rainbow/RainbowGenerator.kt
```diff
@@ -0,0 +1,113 @@
...
+import kotlin.math.pow
+import kotlin.math.roundToInt
+import kotlin.math.sin
+
+/**
+ * Inspired from React-Sdk
+ * Ref: https://github.com/matrix-org/matrix-react-sdk/blob/develop/src/utils/colour.js
+ */
+@Inject
+class RainbowGenerator {
```
jmartinesp, line 21: 🌈

libraries/slash/impl/src/main/kotlin/io/element/android/libraries/slash/impl/CommandExecutor.kt
```diff
@@ -0,0 +1,219 @@
...
+        val formattedText = "<span data-mx-spoiler>${slashCommand.message}</span>"
+        return timeline.sendMessage(
+            body = text,
+            htmlBody = formattedText,
+            intentionalMentions = emptyList(),
+        )
+    }
+
+    private suspend fun sendShrug(slashCommand: SlashCommand.SendShrug, timeline: Timeline): Result<Unit> {
+        return sendPrefixedMessage(timeline, "¯\\_(ツ)_/¯", slashCommand.message)
```
jmartinesp: I think you need another `\` before the 2nd `_`, at least that's what I always end up doing myself 😁 . Maybe it's not strictly necessary because it doesn't have a closing one, but it's more correct like that.

libraries/slashcommands/impl/src/main/kotlin/io/element/android/libraries/slashcommands/impl/CommandParser.kt
```diff
@@ -0,0 +1,426 @@
...
+            }
+            when {
+                Command.PLAIN.matches(slashCommand) -> {
+                    if (message.isNotEmpty()) {
+                        SlashCommand.SendPlainText(message = message)
+                    } else {
+                        syntaxError(Command.PLAIN)
+                    }
+                }
+                Command.CHANGE_DISPLAY_NAME.matches(slashCommand) -> {
```
jmartinesp, line 84: We're adding some HS capability checks in the SDK for changing the display name or avatar, so maybe we'll have to apply them here too (or maybe by disabling the command rows in the UI instead?).

jmartinesp (commented):
I haven't checked this in depth because, well, 2.4kLoC diff, but here are some remarks:

features/location/impl/src/test/kotlin/io/element/android/features/location/impl/share/ShareLocationPresenterTest.kt
bmarty: The compiler was complaining because of the warning. It could have been a in separate commit, sorry.

libraries/slash/impl/src/main/kotlin/io/element/android/libraries/slash/impl/CommandExecutor.kt
bmarty: I will check. I have another issue, the first arm (`\`)  is not rendered...

libraries/slash/api/src/main/kotlin/io/element/android/libraries/slash/api/SlashService.kt
bmarty: Yes, but it should not conflict with another module name I think. I do not like module name (lower case) with several words...

libraries/slashcommands/impl/src/main/kotlin/io/element/android/libraries/slashcommands/impl/CommandParser.kt
bmarty: Good point. I guess the server will reject the request and a (very technical) error dialog will be displayed. Maybe acceptable for slash commands?

libraries/slash/api/src/main/kotlin/io/element/android/libraries/slash/api/SlashService.kt
jmartinesp: `:libraries:slashcommands` seems fine to me, even if it's in a single word style.

libraries/slash/impl/src/main/kotlin/io/element/android/libraries/slash/impl/CommandExecutor.kt
jmartinesp: Oh, it's probably just escaping the 2nd one using the 1st one 😬 .

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/util/MessageEventContent.kt
bmarty: I will introduce a new MsgType enum to support more cases (snowfall, confetti) in the future

libraries/slash/impl/src/main/kotlin/io/element/android/libraries/slash/impl/CommandExecutor.kt
bmarty: Oh yes, that was it. Thanks!

libraries/slashcommands/api/src/main/kotlin/io/element/android/libraries/slashcommands/api/SlashCommand.kt
bmarty: Added `isSupported: Boolean`

ganfra (approved):
Thanks!

rajil:
The slash command which i was looking forward to was `/myroomnick`, and it says 'Not Yet Implemented'. Will that get support in the near future?

PR #6496, opened 2026-03-31
Add floating/sticky date badge in the timeline

jmartinesp:
## Content

Based on https://github.com/element-hq/element-x-android/pull/6433 (thanks @kalix127 !).

Adds the floating badge created in that PR, and a disabled by default feature flag to actually display it.

## Motivation and context

The equivalent [Element X iOS PR](https://github.com/element-hq/element-x-ios/pull/5313/changes#diff-e59f20a145020d88d30221b07336f5a5b5bac2df8650af2e9d2d9cc58c043cfbR440) added this feature disabled at the moment so we can test it for a bit longer.

## Screenshots / GIFs

In the original PR.

## Tests

Enable the feature flag, the badge should be visible. Disable it and it should be gone.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

    - [x] No.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
@@ -45,7 +45,6 @@ internal fun BoxScope.FloatingDateBadgeOverlay(
     lazyListState: LazyListState,
     timelineItems: ImmutableList<TimelineItem>,
     isLive: Boolean,
-    useReverseLayout: Boolean,
     topOffset: Dp = 0.dp,
 ) {
     val currentDateText by remember(timelineItems) {
@@ -54,11 +53,7 @@ internal fun BoxScope.FloatingDateBadgeOverlay(
             if (visibleItems.isEmpty()) return@derivedStateOf null
 
             // In reverse layout, the last visible item is at the top of the screen
-            val topVisibleIndex = if (useReverseLayout) {
-                visibleItems.last().index
-            } else {
-                visibleItems.first().index
-            }
+            val topVisibleIndex = visibleItems.last().index
 
             // Search forward (toward older items) for the nearest day separator
             for (i in topVisibleIndex until timelineItems.size) {
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
@@ -22,6 +22,7 @@ import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.runtime.setValue
 import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.Alignment
@@ -48,16 +49,19 @@ internal fun BoxScope.FloatingDateBadgeOverlay(
     isLive: Boolean,
     topOffset: Dp = 0.dp,
 ) {
+    // This needs to be a state to trigger a `derivedState` recalculation
+    val updatedTimelineItems by rememberUpdatedState(timelineItems)
+
     // Look for the last visible item with a timestamp, starting from the last visible item and going backwards until we find one or reach the start of the list
-    val lastVisibleItemWithTimestamp by remember(timelineItems) {
+    val lastVisibleItemWithTimestamp by remember {
         derivedStateOf {
             var index = lazyListState.layoutInfo.visibleItemsInfo.lastOrNull()?.index ?: return@derivedStateOf null
             while (index >= 0) {
-                when (val item = timelineItems.getOrNull(index)) {
+                when (val item = updatedTimelineItems.getOrNull(index)) {
                     is TimelineItem.Event -> return@derivedStateOf item
                     is TimelineItem.Virtual -> if (item.model is TimelineItemDaySeparatorModel) return@derivedStateOf item
                     is TimelineItem.GroupedEvents -> return@derivedStateOf item.events.firstOrNull()
-                    else -> Unit
+                    null -> Unit
                 }
                 index--
             }
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
@@ -47,30 +47,28 @@ internal fun BoxScope.FloatingDateBadgeOverlay(
     isLive: Boolean,
     topOffset: Dp = 0.dp,
 ) {
-    val currentDateText by remember(timelineItems) {
+    // Look for the last visible item with a timestamp, starting from the last visible item and going backwards until we find one or reach the start of the list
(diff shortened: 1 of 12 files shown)


libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
```diff
@@ -156,10 +156,17 @@ enum class FeatureFlags(
...
-        title = "validate internet connectivity when scheduling notification fetching",
+        title = "Validate internet connectivity when scheduling notification fetching",
         description = "Only fetch events for push notifications when the device has internet connectivity. " +
             "Enabling this can be problematic in air-gapped environments.",
         defaultValue = { true },
         isFinished = false,
     ),
+    FloatingDateBadge(
+        key = "feature.floating_date_badge",
+        title = "Display a sticky date badge in the timeline",
```
bmarty: Maybe

```suggestion
        title = "Display sticky date headers in the timeline",
```

is a bit better (user POV)?

bmarty (commented):
Thanks for adding the FF.

I have noticed a few issues:

1. When a lot of messages have been sent on the same day, the date badge does not appear until I scroll to the past. I guess we need more data in the timeline to have a Date item.
2. When the screen reader is enabled, the timeline items order is reversed, and in this case the date badge is still rendered at the top of the screen, and the date does not match what is displayed on screen. See below:

Image

We should maybe display date badge only when the timeline is not reversed?

3. In a room with pinned messages, when the pinned messages banner is displayed, the date header has a vertical offset and so does not change when the date in the timeline is just below the date header.

Sorry, I should probably have reported this issues in the original PR. We may fix them separately since the feature is behind a FF.

jmartinesp:
> When a lot of messages have been sent on the same day, the date badge does not appear until I scroll to the past. I guess we need more data in the timeline to have a Date item.

Oh, I hadn't thought about that, nicely spotted. I think the iOS PR just generates a new formatted date from the timestamp of the first visible event. I'm not sure if we should do the same for performance reasons, or just use that as a fallback maybe?

> We should maybe display date badge only when the timeline is not reversed?

Sounds reasonable, especially given how easy it is to break the talkback navigation 🫤 .

> In a room with pinned messages, when the pinned messages banner is displayed, the date header has a vertical offset and so does not change when the date in the timeline is just below the date header.

Oh, that one is tricky. I wonder if the OS team is handling this somehow.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
```diff
@@ -0,0 +1,140 @@
...
+            }
+            // Fallback: search backward (toward newer items)
+            for (i in topVisibleIndex - 1 downTo 0) {
+                val item = timelineItems[i]
+                if (item is TimelineItem.Virtual && item.model is TimelineItemDaySeparatorModel) {
+                    return@derivedStateOf item.model.formattedDate
+                }
+            }
+            null
+        }
```
ganfra: This feels a bit hacky honestly to search for the DaySeparator? We could just get the timestamp of the next event and format it?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
```diff
@@ -0,0 +1,140 @@
...
+
+@Composable
+internal fun BoxScope.FloatingDateBadgeOverlay(
+    lazyListState: LazyListState,
+    timelineItems: ImmutableList<TimelineItem>,
+    isLive: Boolean,
+    useReverseLayout: Boolean,
+    topOffset: Dp = 0.dp,
+) {
+    val currentDateText by remember(timelineItems) {
```
ganfra: We are recreating the `derivedState` each time the timeline items change here, not ideal... use `rememberUpdatedState(timelineItems)` instead?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
jmartinesp: Formatting the timestamp as you scroll could affect performance, and also we don't have access to `DefaultDateFormatter` in the compose code or an easy way to provide one (using some `LocalDateFormatter` maybe?).

But `TimelineItem.Event` is created with `sentTime: String` (formatted) and `sentTimeMillis`. Since we're already formatting the datetime when we create this, we could also add a `sentDate: String` and format the date only in this field.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
jmartinesp: Sorry, I'm not sure I understand what you want to change... wouldn't that work the same? `rememberUpdatedState` just wraps whatever input you have in a mutable state so it can be referenced later in some lambda/effect.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/6496/commits/bb9debf62658576bfa6dda4ea60053af774048da

jmartinesp:
> In a room with pinned messages, when the pinned messages banner is displayed, the date header has a vertical offset and so does not change when the date in the timeline is just below the date header.

Looking at this again, I can't think of an easy solution to be honest. If the banner animates we'd change the badge contents with a simple 1-2px scroll i.e. which is also quite bad.  Maybe we can just live with this? FWIW, I didn't find anything related to this in [the iOS PR](https://github.com/element-hq/element-x-ios/pull/5313).

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/FloatingDateBadge.kt
ganfra: Yeah that's the point, we need timelineItems to be a state so it can be read in the `derivedState` lambda

jmartinesp: Agh, I always forget that. Thanks for the reminder!

jmartinesp: https://github.com/element-hq/element-x-android/pull/6496/commits/3714a6070286efbaa37b5c8e893a46da2d931210

jmartinesp: Actually, with that we can remove the whole `remember(updatedTimelineItems)` part and use a simple `remember`,  I guess.

ganfra: No need to pass it in the remember of the derivedState then :P 
I'm even wondering if we should use snapshotFlow here..

jmartinesp: https://github.com/element-hq/element-x-android/pull/6496/commits/99823f784447490d7756d770a738f2f7a03bba1b

ganfra (approved):
Thanks! Let's merge and see how it goes I guess!

PR #6498, opened 2026-03-31
Make media captions scrollable

bxdxnn:
## Content

Make media captions scrollable and add a shadow when it's possible to scroll

## Motivation and context

ATM, it's not possible to scroll captions or know that there's more text.

## Screenshots / GIFs

https://github.com/user-attachments/assets/b4e63c1b-6d1a-410c-ac32-ea6ba5d842d2

## Tests

- Test that caption is scrollable
- Test that the shadow goes away when there's nothing to scroll

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.
- [x] UI change has been tested on both light and dark themes (**note: I'm not sure if that's expected, but the media player seems to be always in a dark theme, even though there are screenshot tests for a light-themed media player.**)

### DIFF
--- a/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerStateProvider.kt
+++ b/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerStateProvider.kt
@@ -29,6 +29,14 @@ import io.element.android.libraries.mediaviewer.impl.details.aMediaDeleteConfirm
 import io.element.android.libraries.mediaviewer.impl.details.aMediaDetailsBottomSheetState
 import kotlinx.collections.immutable.toImmutableList
 
+private const val LONG_CAPTION = "This is a very long caption that should be scrollable in the media viewer. " +
+    "It contains multiple lines of text to demonstrate the scrolling behavior. " +
+    "Line 1: Lorem ipsum dolor sit amet, consectetur adipiscing elit. " +
+    "Line 2: Sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. " +
+    "Line 3: Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris. " +
+    "Line 4: Duis aute irure dolor in reprehenderit in voluptate velit esse cillum. " +
+    "Line 5: Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia."
+
 open class MediaViewerStateProvider : PreviewParameterProvider<MediaViewerState> {
     override val values: Sequence<MediaViewerState>
         get() = sequenceOf(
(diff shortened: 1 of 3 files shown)


jmartinesp:
@americanrefugee this seems like a good solution to me for too long captions since I don't think these have any hard limit, what do you think? 

What other IMs do:

- Telegram limits the amount of text and allows you to drag the caption over the media instead of scrolling.
- WhatsApp limits the text you can send too and just has a 'read more' button that overlays everything at full length on top of the media.

jmartinesp:
@ganfra mentioned another option would be a 'read more...' action like WhatsApp does, but instead displaying a bottom sheet in that case with the full content, similar to what we do for the room topic in the room details screen.

Personally, I don't like that too much, since this solution seems simpler to both implement and to access from a user POV.

bxdxnn:
I agree, IMO the current solution is the simplest.

americanrefugee:
We should set a max-number of lines allowed before the content scrolls. Maybe 9 lines of text? 

Can we then perhaps get screenshots of what that would look like for each media type and orientation (e.g. portrait vs landscape video with controls, portrait vs landscape image, doc, audio file)?

If everything still looks good, then I'm fine with the scrolling as implemented. Let's see how users react.

jmartinesp:
> We should set a max-number of lines allowed before the content scrolls. Maybe 9 lines of text?

Right now I think it's done based on the available screen space. If you look at the video in the PR description, on portrait in my test device it's 9 lines. I think on landscape it doesn't look as good though, we'd definitely need to adjust that height to maybe 2-3 lines 😅 

image

> Can we then perhaps get screenshots of what that would look like for each media type and orientation (e.g. portrait vs landscape video with controls, portrait vs landscape image, doc, audio file)?

For other media types I think it would look mostly the same, just without the media control bar. Also, you can always type on the media itself to hide all this UI and see only the media.

---

This is what it looks like after some adjustments for landscape orientation:

image

| Type | Portrait | Landscape |
|---|---|---|
| Image |image|image|
| Video | <video src=https://github.com/user-attachments/assets/9c57ed6b-3473-4b24-a359-5674a418366e />  | image |
| File |image| image|
| Audio | image| image|

americanrefugee:
Maybe for landscape mode the max # of lines should be 5. I think 2-3 max lines will be too hard to scroll.

jmartinesp:
I've added some screenshots to my previous message for different media types.

> Maybe for landscape mode the max # of lines should be 5. I think 2-3 max lines will be too hard to scroll.

That could take too much space for videos or audio since they have an extra media control bar, but maybe if we make it go below the navigation UI of the OS (which we should anyway) we could recover part of that space.

americanrefugee:
Ok, I see the screenshots now. 3 max lines in landscape mode is fine.

mxandreas:
👍 from my side as well. However, as it is tedious to track such discrepancies between iOS and Android, is it possible to immediately add it to iOS as well? It is not a show stopper but I'd like to avoid such discrepancies between iOS and Android. In particular as when the scrolling works, users are probably going to use longer captions.

jmartinesp (approved):
I'll merge this branch into one of mine so I can tweak it a bit, then merge it to `develop`. Thanks!

PR #6502, opened 2026-03-31
Fix crash when using `View.hideKeyboardAndAwaitAnimation`

jmartinesp:
## Content

Remove the `View.OnApplyWindowInsetsListener` used in modern Android versions to detect if the insets changed after they do the first time: this is a single use operation and the listener will be called every time the insets change.

Also, replace `Mutex` with `CompletableDeferred`: it will have the same effect, but it won't throw an exception if for some reason we have several `future.complete` calls.

## Motivation and context

Fix a crash observed live.

## Tests

I'm not sure how to test this, to be honest. On my case, it crashed automatically.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

    - [x] No.

### DIFF
--- a/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/ui/View.kt
+++ b/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/ui/View.kt
@@ -16,8 +16,8 @@ import android.view.ViewTreeObserver
 import android.view.WindowInsets
 import android.view.inputmethod.InputMethodManager
 import androidx.core.content.getSystemService
+import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.suspendCancellableCoroutine
-import kotlinx.coroutines.sync.Mutex
 import kotlin.coroutines.resume
 
 fun View.hideKeyboard() {
@@ -28,11 +28,13 @@ fun View.hideKeyboard() {
 suspend fun View.hideKeyboardAndAwaitAnimation() {
     val imm = context?.getSystemService<InputMethodManager>()
 
-    val mutex = Mutex()
+    val future = CompletableDeferred(Unit)
     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
         setOnApplyWindowInsetsListener { view, insets ->
             if (!insets.isVisible(WindowInsets.Type.ime())) {
-                mutex.unlock()
+                future.complete(Unit)
+                // Remove the listener now, it's a single use operation
+                setOnApplyWindowInsetsListener(null)
             }
             insets
         }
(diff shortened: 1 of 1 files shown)


libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/ui/View.kt
```diff
@@ -41,14 +43,15 @@ suspend fun View.hideKeyboardAndAwaitAnimation() {
...
+                if (resultCode == InputMethodManager.RESULT_UNCHANGED_HIDDEN || resultCode == InputMethodManager.RESULT_HIDDEN) {
+                    future.complete(Unit)
                 }
             }
         })
     }
-    mutex.lock()
+
+    // Await the future to ensure the keyboard hide animation has completed before proceeding
+    future.await()
```
bmarty: This should maybe never happen, and it's not new, but I think that we will have a dead lock if `imm` is null. Maybe check the value of `imm`?

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/ui/View.kt
```diff
@@ -28,11 +28,13 @@ fun View.hideKeyboard() {
 suspend fun View.hideKeyboardAndAwaitAnimation() {
     val imm = context?.getSystemService<InputMethodManager>()
 
-    val mutex = Mutex()
+    val future = CompletableDeferred(Unit)
```
bmarty:
```suggestion
    val future = CompletableDeferred<Unit>()
```

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/ui/View.kt
bmarty: The future will not complete if the user manually close the keyboard. We probably need to check the keyboard state before waiting for it to become hidden.

bmarty (changes requested):

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/ui/View.kt
jmartinesp: Done in https://github.com/element-hq/element-x-android/pull/6502/commits/2e6f0238374a3fc5a7b43a366d7580e0309e0adb

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/ui/View.kt
jmartinesp: I think this should fix it: https://github.com/element-hq/element-x-android/pull/6502/commits/57f4b93ae79234a5fb181866a73bf28353cde7ba

At least it seemed to work fine for me locally. I also added a timeout in case something really weird happens and the future is never completed.

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/ui/View.kt
```diff
@@ -16,39 +16,54 @@ import android.view.ViewTreeObserver
...
             if (!insets.isVisible(WindowInsets.Type.ime())) {
-                mutex.unlock()
+                future.complete(Unit)
+                // Remove the listener now, it's a single use operation
+                setOnApplyWindowInsetsListener(null)
             }
             insets
         }
-        imm?.hideSoftInputFromWindow(windowToken, 0)
+        imm.hideSoftInputFromWindow(windowToken, 0)
```
bmarty, line 48: Could have `val requested = ` here too, no? If yes, maybe extract this from the `if` block and simplify the code with:
```
    if (requested) {
        // Await the future to ensure the keyboard hide animation has completed before proceeding
        withTimeoutOrNull(1.seconds) { future.await() }
    }
```
?

bmarty (approved):
Thanks for the update. Tested OK. One last remark maybe?

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/ui/View.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/6502/commits/5d20b23b52b7b7f38f0ff9f340e71f64fece8e05

bmarty (approved):
Thanks!

PR #6507, opened 2026-04-01
Take into account homeserver capabilities

### LINKED ISSUE #6491: [Task] Prevent users from changing their profile based on HS capabilities
Please see and discuss the details in the meta issue (https://github.com/element-hq/element-meta/issues/3167).

---

Depends on https://github.com/matrix-org/matrix-rust-sdk/pull/6371. The implementation in the clients should just make sure we disable the UI that would allow the users to change their profile (display name, avatar) based on the HS capabilities.

jmartinesp:
## Content

Add `HomeserverCapabilitiesProvider` to check if the HS allows changing the user's display name or avatar.

Filter out `/nick` and `/myavatar` slash commands based on the capabilities. Add `SlashCommand.CHANGE_AVATAR` too.

Also, modify the edit user profile screen to reflect these values.

## Motivation and context

Closes #6491 

## Screenshots / GIFs

Edit profile screen with disabled options:

image

## Tests

I used a test HS with these capabilities disabled and checked the actual UI was disabled for changing my own avatar and display name.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

    - [x] No.

### DIFF
--- a/libraries/slashcommands/impl/src/main/kotlin/io/element/android/libraries/slashcommands/impl/DefaultSlashCommandService.kt
+++ b/libraries/slashcommands/impl/src/main/kotlin/io/element/android/libraries/slashcommands/impl/DefaultSlashCommandService.kt
@@ -11,6 +11,7 @@ import dev.zacsweers.metro.ContributesBinding
 import io.element.android.libraries.di.RoomScope
 import io.element.android.libraries.featureflag.api.FeatureFlagService
 import io.element.android.libraries.featureflag.api.FeatureFlags
+import io.element.android.libraries.matrix.api.HomeserverCapabilitiesProvider
 import io.element.android.libraries.matrix.api.timeline.Timeline
 import io.element.android.libraries.preferences.api.store.AppPreferencesStore
 import io.element.android.libraries.slashcommands.api.SlashCommand
@@ -18,6 +19,8 @@ import io.element.android.libraries.slashcommands.api.SlashCommandService
 import io.element.android.libraries.slashcommands.api.SlashCommandSuggestion
 import io.element.android.services.toolbox.api.strings.StringProvider
 import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.withTimeoutOrNull
+import kotlin.time.Duration.Companion.seconds
 
 @ContributesBinding(RoomScope::class)
 class DefaultSlashCommandService(
@@ -26,6 +29,7 @@ class DefaultSlashCommandService(
     private val stringProvider: StringProvider,
     private val appPreferencesStore: AppPreferencesStore,
     private val featureFlagService: FeatureFlagService,
+    private val capabilitiesProvider: HomeserverCapabilitiesProvider,
 ) : SlashCommandService {
     override suspend fun getSuggestions(
         text: String,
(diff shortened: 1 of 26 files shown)


libraries/slashcommands/impl/src/main/kotlin/io/element/android/libraries/slashcommands/impl/DefaultSlashCommandService.kt
```diff
@@ -26,26 +29,47 @@ class DefaultSlashCommandService(
...
-        }.filter {
-            !it.isDevCommand || isDeveloperModeEnabled
-        }.map {
-            SlashCommandSuggestion(
-                command = it.command,
-                parameters = it.parameters,
-                description = stringProvider.getString(it.description),
-            )
-        }
+        return Command.entries
```
ganfra, line 40: The list is not that big, but I'd avoid chaining filter as it's creating a new list after each filter operation... Maybe converts first the list to a sequence, or just apply one filter?

ganfra (approved):
One remark otherwise LGTM!

libraries/slashcommands/impl/src/main/kotlin/io/element/android/libraries/slashcommands/impl/DefaultSlashCommandService.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/6507/commits/14d6d05c41654d4097e023fcac790e3af8d298b9

libraries/slashcommands/impl/src/main/kotlin/io/element/android/libraries/slashcommands/impl/DefaultSlashCommandService.kt
```diff
@@ -26,26 +29,49 @@ class DefaultSlashCommandService(
...
+            // Don't include the change display name commands if the user can't change their display name
+            .run {
+                val canUserChangeDisplayName = withTimeoutOrNull(5.seconds) {
+                    capabilitiesProvider.canChangeDisplayName().getOrNull()
+                } ?: false
+                if (!canUserChangeDisplayName) {
+                    filterNot { it == Command.CHANGE_DISPLAY_NAME || it == Command.CHANGE_DISPLAY_NAME_FOR_ROOM }
+                } else {
+                    this
+                }
```
bmarty, line 54: There is maybe room for optimization since at this point, Command.CHANGE_DISPLAY_NAME and Command.CHANGE_DISPLAY_NAME_FOR_ROOM may already have been filtered out. So no need to invoke `canChangeDisplayName()`.
Same remark for the next filter.

Also one question, I see that you are using a timeout, does it means that `canChangeDisplayName()` can be long and delay the apparition of the suggestions?

jmartinesp: > There is maybe room for optimization since at this point, Command.CHANGE_DISPLAY_NAME and Command.CHANGE_DISPLAY_NAME_FOR_ROOM may already have been filtered out. So no need to invoke canChangeDisplayName().

You'd check if those are present first, then?

> Also one question, I see that you are using a timeout, does it means that canChangeDisplayName() can be long and delay the apparition of the suggestions?

It's very unlikely, since this should be refreshed as soon as you get network connectivity and then cached so you read it from memory instead in the SDK, but it *could* happen that fetching this value needs you to perform a network request.

bmarty: OK, if the method is fast, I guess there is no real issue, even if some request will be done even when the sequence is already fully filtered out.

PR #6509, opened 2026-04-01
Tentative fix for `ForegroundServiceStartNotAllowedException`

jmartinesp:
## Content

A couple of tentative fixes to this crash:

- When failing to start the service in foreground, don't crash. This is a helper to speed up the scheduling by keeping the CPU awake, not a critical part that must succeed.
- Don't try to restart the service when there are pending push notification fetches, as it sometimes fails.

Either of these changes would fix the issue (or both). Another possible solution is to remove the attempt to restart and change the kind of foreground service used to keep the CPU awake, but then we need to be extra careful about always stopping it after a short time, since we won't have the default 3 minute timeout `SHORT_SERVICE` adds.

## Motivation and context

Fix https://github.com/element-hq/element-x-android-rageshakes/issues/11044

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


ganfra (approved):
Let's see how it goes

bmarty:
OK thanks, merged, I'll trigger a nightly build.

PR #6519, opened 2026-04-02
Support gallery messages

bxdxnn:
## Content

Support sending and rendering gallery messages (a message with multiple photos)

## Screenshots / GIFs

Screenshot

## Tests

- Test that sending works
- Test that rendering works

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bxdxnn:
@americanrefugee please review the design. Particularly:
* How to make a better border between tiles?
* What should be the border radius of the message? You can see on the screenshot, the bottom-right corner that "connects" messages doesn't look good.
* More complex, but please test the logic of handling landscape/portrait messages in a gallery and if any changes are needed.
* Tapping on a picture in the gallery will allow viewing/swiping only the pictures of the gallery, but in a non-gallery message, you can see all pictures.

americanrefugee:
I think we're are already working on this feature for iOS. cc @Velin92

americanrefugee:
@bmarty we should coordinate to make sure the same thing is applied to Android.

mxandreas:
How has been this accomplished? To make it work cross-platform and interoperable with other clients, it needs to use a MSC like this https://github.com/matrix-org/matrix-spec-proposals/pull/4274 ? Is this the intention?

Also, this clearly needs support on all 3 platforms Web, iOS and Android before it can be generally released.

bxdxnn:
> How has been this accomplished? To make it work cross-platform and interoperable with other clients, it needs to use a MSC like this [matrix-org/matrix-spec-proposals#4274](https://github.com/matrix-org/matrix-spec-proposals/pull/4274) ? Is this the intention?
> 
> Also, this clearly needs support on all 3 platforms Web, iOS and Android before it can be generally released.

SDK already supports gallery messages:
https://github.com/matrix-org/matrix-rust-sdk/pull/4977
https://github.com/matrix-org/matrix-rust-sdk/pull/5125

mxandreas:
> SDK already supports gallery messages:

Ok, that clarifies the technical approach. My suggestion is that if we want to move forward, we need to first confirm, if there is capacity to implement this also for Web. If Rust SDK supports it, I am guessing it may not be such a lot of effort for iOS. However, Web is not using Rust SDK, thus needs to be checked first.

If all of that looks feasible then we still need to figure out the designs, but I think we're able to do that if there if there is confirmation it to be done on all 3 platforms.

bxdxnn:
> If all of that looks feasible then we still need to figure out the designs, but I think we're able to do that if there if there is confirmation it to be done on all 3 platforms.

I was able to do a POC Web implementation, do I need an iOS PR too for this feature to proceed with the design?

mxandreas:
> I was able to do a POC Web implementation, do I need an iOS PR too for this feature to proceed with the design?

It is sufficient if we know that you want and can do it on iOS. Which I understand is the case, correct? @americanrefugee can you look into the designs?

A couple of notes:

1. The MSC is built so that you can have one caption per gallery.
2. The MSC works for images/videos but also files (attachments). It would be great to support it for all types - there's less confusion, e.g. another client may send you a gallery of files, and if you do not support rendering it (while we support rendering images), it is confusing.
3. Currently if you open an image/video or an attachment from timeline - you can swipe left/right to go to the previous/next in the room - if now the item is part of a gallery, should such browsing be limited to the gallery items only, or you would be able to "exit" the gallery and see items which are before and after? I think yes.

americanrefugee:
Here's some direction for the media gallery in a timeline...

THE EXAMPLE BELOW IS ONLY FOR ANDROID
- For iOS, the logic is exactly the same, but the fonts are specific to iOS
- This is NOT for Web (yet to be designed)

### Context

Here's an example of multiple media items uploaded to a timeline, with and without a caption:

Timeline - Images uploaded - Sent - With caption
Timeline - Images uploaded - Sent - No caption

### Gallery item preview

Here's an example of images or video gallery items, including overflow, in Light and Dark themes:
Media Preview

### Full gallery

And here's an example of the various gallery groupings with 1, 2, 3, 4, 5, and 5+ items:

Media group gallery

- Spacing between items = 4px
- Corner radius for the whole group = 6px

americanrefugee:
We will also need a **version of a gallery message in the timeline for files**. 

Since files may or may not contain a preview, for this version we should show a list:
 
Timeline - Files uploaded - Sent
Timeline - Files uploaded - Received


**The logic is as follows:**
- If the user uploads files that all have a preview (such as a bunch of PDFs), then we can still show them like you've already built for images and videos.
- But if the user uploads a mixture of any files that do not have a preview, then all the files are rendered together as a list (shown above).
   - The icon for any document (PDF, ZIP, DOC, etc.) is "Attachment" (paperclip)
   - Audio files show a play button, length, and sound bars
   - Images and videos show a thumbnail of the image, file name, file type, and size.

**When replying to a gallery,** the bubbles should look like this:

Timeline - Reply to media uploaded
Timeline - Files uploaded - Sent

Does this make sense?

americanrefugee:
I accidentally added images above for iOS. I just updated them to show Android. Sorry!

mxandreas:
> If the user uploads files that all have a preview (such as a bunch of PDFs), then we can still show them like you've already built for images and videos.

@americanrefugee At least currently we do not render any preview for a PDF. Let's keep it the same way now.

bxdxnn:
@americanrefugee

I've implemented the designs, but can you send/tell more precise style attributes for the designs you've sent if possible? Particularly, I want to know the size of the video icon, the gradient attributes, and the timestamp position attributes.

> And here's an example of the various gallery groupings with 1, 2, 3, 4, 5, and 5+ items:

FYI, this example is too simple as it doesn't take into account landscape/vertical mixed media grid handling. I've left the current grid compaction logic without changes, as it should already shows landscape-only galleries like in the example.

> * Audio files show a play button, length, and sound bars

The waveform will need to be generated each time you encounter an audio file in a gallery message, which I believe can lead to big performance issues as the audio file will need to be downloaded before generating the waveform. Imagine if someone would send a 1 GB audio file, it'd take a while and might cost a lot if the user is on a metered connection.
Alternatively, we can add the pre-generated waveform for each audio file in the gallery, the same way how we're doing it right now with the voice messages. But I guess that wasn't defined in the MSC and additionally we'll need SDK support too.
Do you really want to basically convert audio files to voice messages? Besides the technical reasons, this is inconsistent with how audio files are normally displayed, and it will also make it impossible to save/share the audio file, unless we'll make a bottom sheet for the attacment rows.

americanrefugee:
> this example is too simple as it doesn't take into account landscape/vertical mixed media grid handling

Regardless of the original images's aspect ratio (landscape/vertical/square), the image previews should be displayed in that grid. Each image should be centered and fill the box. So if you have a tall image, it would appear as a horizontal preview if in the top position. In other words, the grid does not reflect the actual image aspect ratios.

The reason is because 1.) we want to keep this super simple and 2.) we're trying to conserve vertical space in the timeline. A gallery shouldn't take up the whole screen.

> Do you really want to basically convert audio files to voice messages?

No, you're right. In the timeline an audio message should display an audio icon, file name, file type, and size - NOT the play button and sound waves. I've updated the design, see below:

Timeline - Files uploaded - Received
Timeline - Files uploaded - Sent
Timeline - Reply to files uploaded

> I want to know the size of the video icon, the gradient attributes, and the timestamp position attributes.

- Icon = 16x16, positioned 8px from left and bottom edges
- Timestamp = font/body/sm/medium, positioned 8px from right and bottom edges
- Single video preview = Gradient from 0% opacity to 100% color/bg/canvas/default (top to bottom)
- Overflow preview (with "+X") = color/gray/100 at 70% opacity (no gradient)

Let me know if you need any other values.

bxdxnn:
> Let me know if you need any other values.

Please tell the details (like size) for the tiles in the "Full gallery" example.

> Each image should be centered and fill the box

FYI, I'll assume that if the image is narrow it should be zoom-in-to-fit, and if too tall/wide, crop-to-fit, both go to the center of the image.

americanrefugee:
> Please tell the details (like size) for the tiles in the "Full gallery" example.

It depends on the size of the message bubble. Let's say that the overall message bubble width is 292px...
- There is an 8px margin on the left / right side between the message bubble edge and its contents
- That leaves 276px
- There is a 4px gutter between image previews
- If the bottom row is 1 image, then the height is 50% of the width
- If the bottom row is 2 square images, and with a gutter of 4px, they squares would be 136x136
   - If the top row is also 2 squares, then they would be the same size for a total of 136 + 4 + 136 =276 
   - If the top row is 1 rectangle, then it would be 276w x 136h
- If the bottom row is 3 square images, and with a 4px gutter, the squares would be 89.3x89x3
   - If the top row is 2 squares, then they are 136x136
   - f the top row is 1 rectangle, then it would be 276w x 136h

Or, another way to describe the sizing...
- All square images are exactly square with a 4px gutter between them
- All rectangular images are the same size as 2 squares width and 1 square hight
- There is always a 4px gutter between rows

Does that help?

bxdxnn:
The implementation is done and the designs were fully implemented. You can check the [preview screenshots](https://github.com/element-hq/element-x-android/pull/6519/changes#diff-46ebaa68723539c725aea2d3ce24b0de3d99ed164e069ac3d6987e53e203c918) to compare.
I'll look into implementing this for iOS. Meanwhile, it'll be great to have the design for Web.

mxandreas:
I have taken a quick look at the current version. One question is the swiping between the images - are you limited to this gallery only or can you "go out" to previous/next timeline items as well. Both have pros and cons and that's why there are also different approaches out there in other apps, for example:

1. Signal allows you go out to any previous/next image and I think this is what we should do to keep things simple. The only downside is that the user may not realize the went out and think that the image was part of this gallery.
2. WhatsApp will constrain you to the current gallery if 2+ photos are sent together. Which is confusing because if there was just one image, you can go to others. (I guess) to mitigate that, they have up/down wiping within gallery and left/right across timeline items.

cc @americanrefugee

bxdxnn:
@mxandreas

I agree that this behavior might need some discussion. Currently, opening a gallery image will let you swipe only the images of the gallery. I.e.: you can't "go out" if you pressed on a gallery image, but you can "go in" and "go out" into/out of a gallery if you press on a non-gallery image.

My reasoning was:
* Whatsapp's "gallery scroll":
  * Pros:
    1. Familiar behavior to a user, compared to the current behavior (everyone has used Whatsapp at least once)
  * Cons:
    1. More complex
    2. Can't go to the media viewer right away
* Simply allow scrolling the entire timeline (Signal):
  * Pros:
    1. Simplest
    2. Not confusing
  * Cons:
    1. Less privacy (use case: if you let someone else to scroll the gallery to see only the photos of the gallery)
    2. Yes, it'll be not clear whether the photo is a part of a gallery or not
* Current behavior:
  * Pros:
    1. Simple
    2. Goes into the media viewer right away
    3. You know when the gallery starts and when it finishes
  * Cons:
    1. Not a familiar behavior
    2. [Therefore] might be confusing at first

bxdxnn:
@americanrefugee How the preview when sending should look like? ATM, it's just the first image with a count badge.

americanrefugee:
@bxdxnn can you post a screenshot of what you have right now? I'm not sure what you mean by "preview when sending"

bxdxnn:
> @bxdxnn can you post a screenshot of what you have right now? I'm not sure what you mean by "preview when sending"

The media preview screen, where you can add a caption before sending media. It looks like this when sending a gallery:

preview

Another way to do it could be just allowing to swipe through the images. This is what will be done for the iOS as a PoC. It'll be great if you could clarify the approach for consistency.

americanrefugee:
Ahh, that's what you mean. Then I have some designs to share...

## There are 2 scenarios
1. **Uploading media from system Photos & videos library** (images and videos only)
2. **Uploading attachments from system Files library** (any other files, of which some may be images or videos)

### Upload media

1. Composer > "+" button > Select "Photo & video library"
01-Timeline-Options

---

2. Open system media library and multi-select up to 10 items
02-Media Library

---

3. Enter media preview mode
   - Carousel: Swipe left to advance to the next media item; swipe right to go back
   - Images have an "Edit" button (this is being addressed in another issue, so don't worry about it)
03-Media Item-Image

---

4. Videos have additional controls, but no ability to edit
04-Media item-Video

---

5. Add a caption like you would for any single media item
05-Add Caption

---

### Upload files

6. If select "Attachment" from the "+" menu, then open system file library and multi-select up to 10 items
06-File Library

---

7. Enter file preview mode
   - Files without a preview (e.g. ZIP) are shown on a "card" with icon, file name, file type + size
07-Attchment-No preview

---

8. Files with a preview
   - PDF, DOC, or other files with previews show the content itself and allows the user to scroll through any pages
08-Attachment-Preview

---

9. Audio files
   - Similar presentation to files without a preview, expect there are controls to play/pause/scrub the audio content
09-Attachment-Audio

---

10. Images and videos
   - There may of course be images or videos uploaded from the Files library
10-Attachment-Media

---

I know that's a lot of screens, but hopefully everything is pretty straightforward.

bxdxnn:
@americanrefugee
> It depends on the size of the message bubble. Let's say that the overall message bubble width is 292px...

The iOS doesn't seems to have width constraints for media, can you please clarify the tile/bubble sizes details for iOS?

americanrefugee:
> The iOS doesn't seems to have width constraints for media, can you please clarify the tile/bubble sizes details for iOS?

Let's assume an overall bubble with of 292 as in my designs:
- 8px margins left, top, right
- Only 2 rows of images, 4px between each row
- Row with 1 image is 276 x 136
- Row with 2 images is 276 x 136, images are 136x136 with a 4px gutter
- Row with 3 images is 276 x 89, images are 89x89

americanrefugee:
Btw I'm realizing that all the designs I've been providing are for Android, not iOS. Is that an issue?

bxdxnn:
> Let's assume an overall bubble with of 292 as in my designs:

The main problem is that iOS doesn't seems to limit width on media (each message type has it's own max width constraint, except media, BTW this also makes iOS to suffer from the same "narrow-media caption" bug), and there's no such thing like `fillMaxWidth` (i.e. 3/4 of the screen width) contraints on iOS that I'm using for the gallery bubbles on Android, so I'd need to clarify what should be the width of the bubble. Should I go with the 292px (please say the needed value in `pt` points) width as you say?

> Btw I'm realizing that all the designs I've been providing are for Android, not iOS. Is that an issue?

No, it's not an issue ATM, everything else seems to be similar on iOS.

americanrefugee:
> Should I go with the 292px... ?

Actually, I changed my mind. To make sure things still look ok on small screens, let's go with:

**Bubble width = 280w**
- 8px margins left, top, right
- Only 2 rows of images, 4px between each row
- Row with 1 image is 264w x 130h
- Row with 2 images is 264w x 130h, images are 130x130 with a 4px gutter
- Row with 3 images is 264 x 85, images are 85x85

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/GalleryMediaGalleryDataSource.kt
```diff
@@ -0,0 +1,102 @@
...
+import io.element.android.libraries.mediaviewer.impl.datasource.MediaGalleryDataSource
+import io.element.android.libraries.mediaviewer.impl.model.GroupedMediaItems
+import io.element.android.libraries.mediaviewer.impl.model.MediaItem
+import kotlinx.collections.immutable.toImmutableList
+import kotlinx.coroutines.flow.flowOf
+
+class GalleryMediaGalleryDataSource(
+    private val data: GroupedMediaItems,
+) : MediaGalleryDataSource {
+    override fun start() = Unit
```
frebib:
```suggestion
    override fun start(coroutineScope: CoroutineScope) = Unit
```
otherwise this doesn't compile

bxdxnn: Thanks, I'm currently rebasing this PR, and the rebase already includes this fix. I'll push later once I implement the designs.

bxdxnn:
> **Bubble width = 280w**
> 
> * 8px margins left, top, right
> * Only 2 rows of images, 4px between each row
> * Row with 1 image is 264w x 130h
> * Row with 2 images is 264w x 130h, images are 130x130 with a 4px gutter
> * Row with 3 images is 264 x 85, images are 85x85

@americanrefugee can you please send screenshot examples for this?

americanrefugee:
@bxdxnn Here's a screenshot of all configurations for both iOS and Android:
- 2 images
- 3 images
- 4 images
- 5 images
- 6+ images
- Gallery without caption

iOS-Gallery_configurations

Android-Gallery_configurations

bxdxnn:
iOS implementation is done: https://github.com/element-hq/element-x-ios/pull/5601
I'll update the Android implementation according with the new designs. Meanwhile, it'll be great to have the design for Web.

americanrefugee (commented):
What is the font size you're using for the overflow number (e.g. "+4")? It should be font/headline/sm/medium but it looks really big here - perhaps because the snapshots are zoomed in?

bxdxnn:
Implemented the latest designs + cleanup/fixes.

Ready to review/test/merge

bmarty:
Sorry @bxdxnn , https://github.com/element-hq/element-x-android/pull/6842 has been merged first, this branch now have conflicts to fix. Then we can review the PR and merge it.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
```diff
@@ -161,12 +158,11 @@ class AttachmentsPreviewPresenter(
...
         LaunchedEffect(maxUploadSize) {
-            // Check file upload size if the media won't be processed for upload
-            val isImageFile = mediaAttachment.localMedia.info.isImageAttachment()
-            val isVideoFile = mediaAttachment.localMedia.info.mimeType.isMimeTypeVideo()
+            val isImageFile = firstMediaAttachment.localMedia.info.isImageAttachment()
+            val isVideoFile = firstMediaAttachment.localMedia.info.mimeType.isMimeTypeVideo()
             if (maxUploadSize != null && !(isImageFile || isVideoFile)) {
                 // If file size is not known, we're permissive and allow sending. The SDK will cancel the upload if needed.
-                val fileSize = mediaAttachment.localMedia.info.fileSize ?: 0L
+                val fileSize = firstMediaAttachment.localMedia.info.fileSize ?: 0L
```
bmarty: I think that we should iterate on all the attachments here, instead of only the first one.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
```diff
@@ -215,26 +213,18 @@ class AttachmentsPreviewPresenter(
...
+                            sendGalleryPreProcessed(
+                                mediaUploadInfos = allMediaUploadInfos,
+                                caption = caption,
+                                sendActionState = sendActionState,
+                                inReplyToEventId = inReplyToEventId,
+                            )
+
+                            // Clean up the pre-processed media after it's been sent
+                            mediaSender.cleanUp()
+                            editedTempFileToDelete?.safeDelete()
```
bmarty: I think that we can have a list of `editedTempFileToDelete`, when several images are edited.

bmarty (changes requested):
Thanks for the rebase! 2 first remarks

bxdxnn:
@bmarty do I need to resolve the merge conflict (by merging, not rebasing/force-pushing), or they should be resolved only after the final review?

bmarty:
I am testing the feature right now, you can do the merge now or later, as soon as you do not force push the branch. Thanks!

americanrefugee:
I have a small change request:

- Please replace the text button "Edit" with an icon button using the new "Crop" icon

Android - Add media

bxdxnn:
> * Please replace the text button "Edit" with an icon button using the new "Crop" icon

Are you sure this needs to be a part of this PR? It seems like it should be done in another PR, editing images is not specific to gallery messages.

libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/DefaultPinnedMessagesBannerFormatter.kt
```diff
@@ -104,6 +105,9 @@ class DefaultPinnedMessagesBannerFormatter(
             is OtherMessageType -> {
                 messageType.body
             }
+            is GalleryMessageType -> {
+                messageType.body.prefixWith(CommonStrings.common_gallery)
+            }
```
bxdxnn, line 110: Should we have another prefix (like "Files: caption") for attachment galleries? i.e. "file" galleries that have at least one attachment without a preview available.

bmarty:
@bxdxnn I checked, the image edition does not work when you edit the second image. It affects the first. Can you check please?

bmarty:
FTR, for now Element Web does not render the event of `msgtype` = `dm.filament.gallery`, it will render a blank event:

image

So either we wait for all Element client to be able to render such event, or we add a warning, as per the caption to warn the user.

Maybe we should first have a PR which renders such events in all platforms, then enable the sending. CC @mxandreas

bxdxnn:
There's already a PR to support galleries in EW: https://github.com/element-hq/element-web/pull/33111
A design for Web is needed to proceed with that PR, adding sending isn't too complex.

manuroe:
@bmarty thanks for the check on EW.

The MSC actually [states](https://github.com/Johennes/matrix-spec-proposals/blob/johannes/msgtype-galleries/proposals/4274-inline-media-galleries.md#potential-issues) there is no backward compatibiliy: 
*Additionally, this proposal doesn't prescribe a fallback mechanism for clients that don't support m.gallery msgtypes. Clients that care about backwards compatibility could encode media links or a generic fallback text into body or formatted_body, respectively. It is unclear, however, if a fallback is actually desirable or not because unwrapping 60 images into somebody else's timeline might be considered inappropriate if not spammy.*

We need to split the rendering from the sending part. The rendering can be always on. The sending behind a feature flag disabled by default for the the moment.

bxdxnn:
It seems like there should be a general redesign of "preview before sending" in Element Web, because currently it can't attach captions either.
I'll lock multi-attachment selection behind a feature flag for now I guess (though it's not clear when we would remove it), will push with a few other fixes.

bmarty:
Sorry edition is still not working:
- Add 3 images
- In the preview edit image `#2` by setting a crop area and save
- Edit image `#3`
- The crop area of image `#2` is applied instead of the default crop area
- At the end the send button is not working.

Another thing to change:
- Send a gallery with 2 images.
- In the timeline click on the image at the top (image 1)
- The the media viewer is opened to image 2

bxdxnn:
@bmarty I'm not able to reproduce any of these (app version: `feature/gallery-messages (6370a7c0)` debug build), are you sure you have the latest branch pulled-in and the app freshly built?

bmarty:
@bxdxnn here is a demo of the issue. You can see that when I edit the second photo, the crop is already set to the value set to the first image.


https://github.com/user-attachments/assets/b4c83dfa-ef32-4f11-8f3c-0632a5cddd78

If I rotate the image, the same problem occurs. All the photo are rotated when editing.

bxdxnn:
@bmarty OK, fixed

bmarty:
@bxdxnn I have pushed a few commits to fix some issue. I have done tests and it seems to work as expected.

bmarty (approved):
We will merge this WIP feature (FF is disabled) to land the code to develop. I have created issues to track the remaining work, and they may be other issues to fix.

Thanks for the initial work @bxdxnn !

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/GalleryMediaGalleryDataSource.kt
```diff
@@ -0,0 +1,101 @@
...
+                    fileExtension = galleryItem.filename.substringAfterLast('.', ""),
+                    senderId = mediaInfo.senderId,
+                    senderName = mediaInfo.senderName,
+                    senderAvatar = mediaInfo.senderAvatar,
+                    dateSent = mediaInfo.dateSent,
+                    dateSentFull = mediaInfo.dateSentFull,
+                    waveform = null,
+                    duration = null,
+                )
+                val id = UniqueId("${eventId?.value ?: "gallery"}_$index")
```
ganfra, line 59: Is it ok to have the same id for all the items?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/event/TimelineItemGalleryContent.kt
```diff
@@ -0,0 +1,58 @@
...
+import io.element.android.libraries.matrix.api.media.MediaSource
+import io.element.android.libraries.matrix.ui.media.MediaRequestData
+import kotlin.time.Duration
+
+data class TimelineItemGalleryContent(
+    val body: String,
+    val caption: String?,
+    val formattedCaption: CharSequence?,
+    override val isEdited: Boolean,
+    val items: List<GalleryItem>,
```
ganfra: `ImmutableList`?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
```diff
@@ -123,52 +136,55 @@ class AttachmentsPreviewPresenter(
...
-            isApplyingImageEdits,
-        ) {
-            // If the media optimization selector is not displayed, we can pre-process the media
-            // to prepare it for sending. This is done to avoid blocking the UI thread when the
-            // user clicks on the send button.
-            @Suppress("ComplexCondition")
+        LaunchedEffect(attachments) {
+            currentIndex = 0
+        }
+
```
ganfra: Mmmm, this code makes no sense, attachments is an `ImmutableList`, so will not change

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemAttachmentsListView.kt
```diff
@@ -0,0 +1,234 @@
...
+                    AttachmentListItem(
+                        attachment = attachment,
+                        onClick = onContentClick?.let { { it(index) } },
+                        onContentLayoutChange = onContentLayoutChange,
+                    )
+                }
+            }
+        }
+
+        if (content.showCaption) {
```
ganfra, line 86: Why not using same logic than for GalleryView? 

```
if (content.showCaption) {
            val caption = if (LocalInspectionMode.current) {
                SpannedString(content.caption)
            } else {
                content.formattedCaption ?: SpannedString(content.caption)
            }
            CompositionLocalProvider(
                LocalContentColor provides ElementTheme.colors.textPrimary,
                LocalTextStyle provides ElementTheme.typography.fontBodyLgRegular
            ) {
                EditorStyledText(
                    modifier = Modifier
                        .padding(horizontal = 4.dp)
                        .widthIn(min = 120.dp),
                    text = caption,
                    style = ElementRichTextEditorStyle.textStyle(),
                    onLinkClickedListener = onLinkClick,
                    onLinkLongClickedListener = onLinkLongClick,
                    releaseOnDetach = false,
                    onTextLayout = ContentAvoidingLayout.measureLegacyLastTextLine(onContentLayoutChange = onContentLayoutChange),
                )
            }
        }
    }
```

libraries/mediaviewer/api/src/main/kotlin/io/element/android/libraries/mediaviewer/api/GalleryItemData.kt
```diff
@@ -0,0 +1,23 @@
...
+
+@Parcelize
+data class GalleryItemData(
+    val filename: String,
+    val mimeType: String,
+    val mediaSource: MediaSource,
+    val thumbnailSource: MediaSource?,
+    val isVideo: Boolean,
+    val isAudio: Boolean,
+    val isFile: Boolean,
```
ganfra: Not really fluent, can we use at least an enum?

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/media/GalleryItemInfo.kt
```diff
@@ -0,0 +1,47 @@
...
+
+package io.element.android.libraries.matrix.api.media
+
+import io.element.android.libraries.matrix.api.timeline.item.event.FormattedBody
+import java.io.File
+
+sealed interface GalleryItemInfo {
+    val file: File
+    val caption: String?
+    val formattedCaption: FormattedBody?
```
ganfra: Can we really have the caption on individual item?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesFlowNode.kt
```diff
@@ -685,6 +702,59 @@ class MessagesFlowNode(
...
+                        event.sentTimeMillis,
+                        mode = DateFormatterMode.Day,
+                    ),
+                    dateSentFull = dateFormatter.format(
+                        timestamp = event.sentTimeMillis,
+                        mode = DateFormatterMode.Full,
+                    ),
+                    waveform = null,
+                    duration = null,
+                )
```
ganfra: Maybe we should refactor this `MediaInfo` or not use it for Gallery, because there is a lot of useless/wrong data

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesFlowNode.kt
```diff
@@ -685,6 +702,59 @@ class MessagesFlowNode(
...
+                    MediaViewerEntryPoint.MediaViewerMode.TimelineImagesAndVideos(timelineMode)
+                } else {
+                    MediaViewerEntryPoint.MediaViewerMode.TimelineFilesAndAudios(timelineMode)
+                }
+                NavTarget.MediaViewer(
+                    mode = mode,
+                    eventId = event.eventId,
+                    mediaInfo = mediaInfo,
+                    mediaSource = item.mediaSource,
+                    thumbnailSource = item.thumbnailSource,
```
ganfra: feels like we should rework this, as it makes no sense to have this with gallery...

libraries/mediaviewer/api/src/main/kotlin/io/element/android/libraries/mediaviewer/api/MediaViewerEntryPoint.kt
```diff
@@ -42,6 +42,7 @@ interface MediaViewerEntryPoint : FeatureEntryPoint {
         val mediaSource: MediaSource,
         val thumbnailSource: MediaSource?,
         val canShowInfo: Boolean,
+        val galleryItems: List<GalleryItemData> = emptyList(),
```
ganfra: This feels weird to have this along side mediaSource/thumbnailSource. We should probably rework this...

ganfra (commented):
Some more remarks after a quick look

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/GalleryMediaGalleryDataSource.kt
bmarty: `$index` is used, so I think that the ids are not the same?

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/media/GalleryItemInfo.kt
bmarty: Yes, even if we do not use them, the specs claims that the format for each gallery item should be identical than for a single item. See https://github.com/Johennes/matrix-spec-proposals/blob/johannes/msgtype-galleries/proposals/4274-inline-media-galleries.md

The Rust SDK has the data, so they are mapped. I have removed the fields for code clarity.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesFlowNode.kt
bmarty: I have done a refactor

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesFlowNode.kt
bmarty: I have done a refactor

libraries/mediaviewer/api/src/main/kotlin/io/element/android/libraries/mediaviewer/api/MediaViewerEntryPoint.kt
bmarty: I have done a refactor

bmarty:
Merging on my branch so that I can record screenshots.

PR #6059, opened 2026-01-21
Add support for Android direct share targets

Xant3s:
## Content

Added implementation for Android Direct Share targets.

## Motivation and context

Addresses #3835.
Allows users to share content directly to Element X rooms from the Android system share sheet, improving the sharing experience.
This feature exists in the legacy Android app.

## Screenshots / GIFs

AndroidDirectShareTargets

## Tests

- Verified that rooms appear as Direct Share targets in the Android system share sheet
- Tested sharing text and one or more images to a room via direct targets.
- Verified that regular sharing is unchanged
- Validated functionality on API 24 and API 34 emulators.

## Tested devices

- [ ] Physical
- [X] Emulator
- OS version(s): API 24 (Android 7.0), API 34 (Android 14)

## Checklist

- [X] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [X] Pull request is based on the develop branch
- [X] Pull request title will be used in the release note, it clearly define what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [ ] You've made a self review of your PR

PR #6545, opened 2026-04-07
Show preview when sharing media from intent

### LINKED ISSUE #2987: Preview before sharing

#2984 is adding a way to receive elements (file / text) from other application but once the room is selected there is no preview, and the data is sent to the room.

This can be a problem since it does not give a way to validate the data which will be sent.

Text could be previewed, or if only one room is selected (note: it's not possible to select multiple rooms right now), the text composer could be fill with the text and the message not sent, so that the user can edit and then send or just cancel the operation. Note that this can interfere if there is a not sent draft in the room (support for draft is coming).

Image and other files: a preview of the data can be displayed, with a way to remove some items. Also image could be cropped, rotated, but this can be handled separately.

Better user experience

Create design for screen preview and implement it.

_No response_

### Additional context

_No response_

No

### LINKED ISSUE #6540: When sharing a picture with the share intent, allow adding a caption before sending
### Steps to reproduce

From another app (e.g. photos), share an image to ElementX. Select a room; then submit.

#### What did you expect?

I'd like to see the preview send screen, where I can add a caption to the image, before sending it to the room

#### What happened instead?

Image is directly sent to the room, without a caption

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

No

No

bxdxnn:
## Motivation and context

Fixes #6540
Fixes #2987

## Tests

- Share a media using the Android sharing intent

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bxdxnn:
@mxandreas @americanrefugee

Should we show the preview before selecting the user to send (current behavior) or after?

frebib:
Note that this also adds closer parity with iOS too

bxdxnn:
> Note that this also adds closer parity with iOS too

Good! What's the preview/send order in iOS?

frebib:
> > Note that this also adds closer parity with iOS too
> 
> Good! What's the preview/send order in iOS?

On the iPad we have it's user selection then preview

features/share/impl/src/main/kotlin/io/element/android/features/share/impl/preview/ShareMediaPreviewStateProvider.kt
```diff
@@ -0,0 +1,32 @@
...
+            aShareMediaPreviewState(),
+            aShareMediaPreviewState(
+                textEditorState = aTextEditorStateMarkdown(initialText = "This is a caption"),
+            ),
+        )
+}
+
+fun aShareMediaPreviewState(
+    text: String? = null,
+    uris: List<io.element.android.features.share.api.UriToShare> = emptyList(),
```
bmarty: Could be nice to tell the agent to not use full qualified name, but use import statements instead.

features/share/impl/build.gradle.kts
```diff
@@ -40,8 +40,11 @@ dependencies {
     implementation(projects.libraries.uiStrings)
     implementation(projects.libraries.testtags)
     implementation(projects.services.appnavstate.api)
-    api(libs.statemachine)
+    implementation(projects.libraries.mediaviewer.api)
+    implementation(projects.libraries.textcomposer.impl)
```
bmarty, line 44: Generally `impl` module should not depend on other `impl` modules. Can you double check this please?

bmarty (commented):
Thanks. We already have a media preview screen, but it is in the `:features:messages:impl` module. Maybe a first step would be to extract this existing code to its own module and then reuse it. It has more features, and it will reduce code duplication.

features/share/impl/build.gradle.kts
bxdxnn: It doesn't work if I change it to `api` or remove this line

bxdxnn:
> Thanks. We already have a media preview screen, but it is in the `:features:messages:impl` module. Maybe a first step would be to extract this existing code to its own module and then reuse it. It has more features, and it will reduce code duplication.

Is it required for this PR? It seems to be very complex to fully extract that functionality then share it, or do you want to only extract the View?

bmarty:
> > Thanks. We already have a media preview screen, but it is in the `:features:messages:impl` module. Maybe a first step would be to extract this existing code to its own module and then reuse it. It has more features, and it will reduce code duplication.
> 
> Is it required for this PR? It seems to be very complex to fully extract that functionality then share it, or do you want to only extract the View?

Ideally yes. Maybe do a separate PR that just extract the code with the Preview to its own `api` / `impl` module. The `messages` module is too big anyway. Moreover, features like editing image before sending would land to the new `:features:mediapreview:impl` module. I can take a quick look on how hard this could be.

features/share/impl/build.gradle.kts
bmarty: yes, ok my bad, this module is a bit special.

mxandreas:
FWIW, on iOS the preview screen to enter the caption is always shown.

When sharing, there are 2 paths:

1. You can immediately select a DM or a room from the OS share sheet (it is populated with the most frequent/recent ones across your different apps, e.g. Element, WhatsApp, etc.). In this case you do not need any further DM/room selection.
2. You can also just choose the app, in which case EXI will provide the user with a list of rooms to choose from. See screenshot below.

IMG_8992

(It seems like this is very similar on Android as well, although I could not test path #1 as the OS did not offer any Element chats, perhaps because I haven't used it frequently).

bxdxnn:
> (It seems like this is very similar on Android as well, although I could not test path #1 as the OS did not offer any Element chats, perhaps because I haven't used it frequently).

This is implemented in #6059 but not merged yet.

bxdxnn:
I've fully moved the media preview into it's own module in the last pushes, is there anything else that needs to be done before merging?

PR #6548, opened 2026-04-08
Reduce log level of activity lifecycle from warning to debug.

bmarty:
Warning was used to highlight the log in the logcat view, but warning can be interpreted as real warning by other tools, so we're now using the debug level for the lifecycle event of the MainActivity.


manuroe (approved):
🚀🚀🚀
Thank you!

PR #6561, opened 2026-04-10
Remove space announcement

bmarty:
## Content

Remove space announcement, but keep the code that manages announcement and improve it.
Keep the Space announcement for the preview, but stop using the translated string to that they can be removed from Localazy.

## Motivation and context

We do not need the space announcement and it was outdated.

## Tests

- Install the app
- Log in to an account
- open the Space tab
- No space announcement is displayed

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

    - [ ] No.


features/announcement/impl/src/main/kotlin/io/element/android/features/announcement/impl/store/DefaultAnnouncementStore.kt
```diff
@@ -52,6 +52,6 @@ class DefaultAnnouncementStore(
 }
 
 private fun Announcement.toKey() = when (this) {
-    Announcement.Space -> spaceAnnouncementKey
+    is Announcement.Fullscreen -> intPreferencesKey("fullscreen_" + this::class.simpleName)
```
ganfra: I'd avoid this as it can break a bit easily with refactoring, obfuscation..

features/announcement/api/src/main/kotlin/io/element/android/features/announcement/api/Announcement.kt
```diff
@@ -8,7 +8,14 @@
...
 
-enum class Announcement {
-    Space,
-    NewNotificationSound,
+import androidx.compose.runtime.Immutable
+
+@Immutable
+sealed interface Announcement {
+    @Immutable
+    sealed interface Fullscreen : Announcement {
```
ganfra: Can we use an enum instead? This would make it easier to iterate over, and use as key

ganfra (commented):
Since we can only display one fullscreen announcement at a time, could we expose a nextFullscreenAnnouncementFlow(): Flow<Announcement.Fullscreen?> and keep that prioritization in
 one place? If Fullscreen became an enum class, this gets simpler too because we can iterate over entries instead of wiring each announcement separately in the
 presenter.

Exemple of what it could be

```
fun nextFullscreenAnnouncementFlow(): Flow<Announcement.Fullscreen?> =
        nextFullscreenAnnouncementFlow(Announcement.Fullscreen.entries)

    private fun nextFullscreenAnnouncementFlow(announcements: List<Announcement.Fullscreen>,
    ): Flow<Announcement.Fullscreen?> {
        val next = announcements.firstOrNull() ?: return flowOf(null)
        return announcementStatusFlow(next).flatMapLatest { status ->
            if (status == AnnouncementStatus.Show) {
                flowOf(next)
            } else {
                nextFullscreenAnnouncementFlow(announcements.drop(1))
            }
        }
    }
```

bmarty:
The code you are proposing cannot really work since the method will wait for the first `Fullscreen` announcement to be shown, but actually any arbitrary full screen announcement could be rendered at any time. For spaces for instance, the announcement was shown when the user entered the space tab. If there were other not shown Fullscreen announcements declared before the `Spaces` once, the `Spaces` announcement would not have been rendered.

ganfra (approved):

PR #6573, opened 2026-04-14
Fix `isInAirGappedEnvironment` check for older APIs

### LINKED ISSUE #6551: Element X still requires access to `fcm.googleapis.com` for UnifiedPush notifications to work
### Environment

- **Element X version:** 26.03.04 (from GitHub releases, not Play Store)
- **ntfy version:** 1.24.0 (F-Drod build)
- **Android:**
- **Synapse** latest (self-hosted)
- **UnifiedPush distributor:** ntfy (self-hosted server)

### Description

I have a fully self-hosted Matrix setup + MAS and wanted to achieve completely Google-free push notifications using UnifiedPush + a self-hosted ntfy server. Both Element X and the ntfy app were installed from sources **outside of Google Play** (Element X from GitHub releases, ntfy from F-Droid), specifically to avoid any Google dependency.

The server-side setup is complete and verified working:

- A self-hosted ntfy instance is running and accessible at a public HTTPS domain
- Element X correctly registered a UnifiedPush topic with ntfy
- Synapse is successfully delivering push payloads to the ntfy push gateway (HTTP 200 responses confirmed in Synapse logs)
- The ntfy topic is receiving the payloads, confirmed by subscribing via curl:

```bash
curl -s "https://push.example.com/.../json"
# returns the push payloadfrom Synapse immediately when a message is recieved
```

- The Element X notification troubleshooter completes **all steps successfully** with no errors

Despite the entire pipeline working end to end, Synapse → ntfy server → ntfy app → Element X, **no notification appears** when Element X is in the background.

The only way to receive notifications is if the device has access to `fcm.googleapis.com`. As soon as a VPN is used that allows access to `fcm.googleapis.com`, notifications start arriving immediately. Blocking `fcm.googleapis.com` again stops them. This makes it clear that Element X is still depending on FCM at some layer even when UnifiedPush is configured and working.

The push payload received at ntfy shows `"format":"event_id_only"`, meaning Element X needs to wake up, connect to Synapse, and fetch the event. It appears Element X is waking up from the UnifiedPush signal delivered by ntfy, I can see "Syncing Notifications..." notification appearing on my phone sometimes but no actual notifications appears unless FCM is also reachable.

### Expected behavior

When a UnifiedPush distributor (ntfy) is installed, configured, and successfully receiving push payloads, Element X should wake up and display notifications **without any connectivity to Google services**. This is the entire point of UnifiedPush.

### Why this matters

The ability to send notifications without relying on Google or Apple infrastructure is not just a privacy preference, it can be critical. In some countries, internet access is periodically restricted or blocked entirely, often with Google services being among the first to become unreachable. Matrix is a natural choice for communication during these situations precisely because it is decentralized and self-hostable. However, if push notifications only work when Google's servers are reachable, then users lose notifications exactly when they need Matrix the most.

If the Github build of Element X requires FCM to be reachable even when UnifiedPush is fully configured, it raises a serious question: **what is the actual purpose of UnifiedPush support?** Users go through the effort of self-hosting a push server, configuring ntfy, and installing outside of Google Play specifically to remove this dependency, only to find it still exists at the application layer.

If there is something misconfigured on the client by me side that causes this fallback to FCM, please clarify what it is.

_No response_

### Additional context

_No response_

No

jmartinesp:
## Content

Use `networkCapabilities.hasCapability` instead of `networkCapabilities.capabilities.contains`, which only works on Android 12 and newer versions.

Also, enable validating the network in `WorkManager` only in enterprise builds by default.

## Motivation and context

This check wasn't working on API 30 and lower. I didn't realise `hasCapability` existed, so I incorrectly assumed if you could only check the capability in API 31 and greater it meant it was only enforced on those APIs too, and the tests I ran were on a device with API 36 🫤 .

Should fix https://github.com/element-hq/element-x-android/issues/6551.

## Tests

It's kind of difficult to test, but try using a device with Android 11 or lower in a network with no access to google APIs (or at least to the connectivity check endpoint). You should see some logs saying it's in air gapped env, and you should still be able to receive push notifications using unified push.

## Tested devices

I don't have a setup to test these changes at the moment, I did try the previous check with a real device using Android 16 and it worked, so I assume generalising the same check for all Android APIs should work the same.

## Checklist

    - [x] No.

### DIFF
--- a/libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
+++ b/libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
@@ -138,7 +138,7 @@ enum class FeatureFlags(
         title = "Validate internet connectivity when scheduling notification fetching",
         description = "Only fetch events for push notifications when the device has internet connectivity. " +
             "Enabling this can be problematic in air-gapped environments.",
-        defaultValue = { it.isEnterpriseBuild },
+        defaultValue = { true },
         isFinished = false,
     ),
     FloatingDateBadge(
--- a/libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
+++ b/libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
@@ -138,7 +138,7 @@ enum class FeatureFlags(
         title = "Validate internet connectivity when scheduling notification fetching",
         description = "Only fetch events for push notifications when the device has internet connectivity. " +
             "Enabling this can be problematic in air-gapped environments.",
-        defaultValue = { true },
+        defaultValue = { it.isEnterpriseBuild },
         isFinished = false,
     ),
     FloatingDateBadge(
--- a/features/networkmonitor/impl/src/main/kotlin/io/element/android/features/networkmonitor/impl/DefaultNetworkMonitor.kt
+++ b/features/networkmonitor/impl/src/main/kotlin/io/element/android/features/networkmonitor/impl/DefaultNetworkMonitor.kt
@@ -15,7 +15,6 @@ import android.net.ConnectivityManager
 import android.net.Network
 import android.net.NetworkCapabilities
 import android.net.NetworkRequest
-import android.os.Build
 import dev.zacsweers.metro.AppScope
 import dev.zacsweers.metro.ContributesBinding
 import dev.zacsweers.metro.SingleIn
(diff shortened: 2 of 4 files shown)


libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
```diff
@@ -138,7 +138,7 @@ enum class FeatureFlags(
         title = "Validate internet connectivity when scheduling notification fetching",
         description = "Only fetch events for push notifications when the device has internet connectivity. " +
             "Enabling this can be problematic in air-gapped environments.",
-        defaultValue = { true },
+        defaultValue = { it.isEnterpriseBuild },
```
bmarty: Why this change? It can become tricky to guess the expected behavior of an app variant / version, especially because the user can also change the settings.

bmarty (approved):
Thanks!

libraries/featureflag/api/src/main/kotlin/io/element/android/libraries/featureflag/api/FeatureFlags.kt
jmartinesp: I suspect this could be negatively affecting unified push users in networks where the google connectivity check endpoint is blocked. We could also remove the `if (!buildMeta.isEnterpriseBuild) return` check in the network monitor so the air-gapped check value is correct for these users.

jmartinesp: Done in https://github.com/element-hq/element-x-android/pull/6573/commits/820c2c369c7c5e039c5fd6b6eefc59c7c3de5770. @mxandreas these are the changes I mentioned in our discussion before.

PR #6586, opened 2026-04-15
Take into account the value of FeatureFlags.SignInWithClassic

bmarty:
After many reworks, it appears that the feature flags ended up not being used. This PR should fix that.

Related to #6296


jmartinesp (approved):
Thanks!

PR #6589, opened 2026-04-15
Add extra logs for timeline pagination

jmartinesp:
## Content

Add some new logs for what happens when we update the pagination status of different timelines.

Also, add the current timeline mode as part of the log tag so we can be sure about which timeline is being updated.

## Motivation and context

Try to debug issues with pagination more easily.

## Tests

Check the logs.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/RustTimeline.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/RustTimeline.kt
@@ -130,6 +130,8 @@ class RustTimeline(
         Timeline.PaginationStatus(isPaginating = false, hasMoreToLoad = mode is Timeline.Mode.FocusedOnEvent)
     )
 
+    private val loggerTag = "Timeline($mode)"
+
     init {
         when (mode) {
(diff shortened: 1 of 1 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/timeline/RustTimeline.kt
```diff
@@ -177,10 +179,11 @@ class RustTimeline(
     }
 
     private fun updatePaginationStatus(direction: Timeline.PaginationDirection, update: (Timeline.PaginationStatus) -> Timeline.PaginationStatus) {
-        when (direction) {
+        val result = when (direction) {
             Timeline.PaginationDirection.BACKWARDS -> backwardPaginationStatus.getAndUpdate(update)
             Timeline.PaginationDirection.FORWARDS -> forwardPaginationStatus.getAndUpdate(update)
         }
+        Timber.tag(loggerTag).d("updatePaginationStatus $direction: $result")
```
bmarty, line 186: `.also` could be used to, but it's OK like that!

bmarty (approved):
Thanks!

Issue #6563, opened 2026-04-10
Defaulting to speaker in handset mode


rajil:
### Steps to reproduce

EC from one device to another using the handset mode

#### What did you expect?
Ear piece will be used
#### What happened instead?
Audio comes out of speaker

Rageshake sent with the title same as github title.

### Your phone model

Samsung S23 Ultra

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

rajil:
@BillCarsonFr let me know if you need anything more.

rajil:
@BillCarsonFr I can confirm that with v26.04.4 the audio is routed correct to the earpiece as expected on Samsung S23 (Android 16). However, an older generation Samsung device (Galaxy M31s) running Android 12  still routes to the speaker. I have submitted logs today using the title 'Audio through speaker in handset mode'.

Issue #6422, opened 2026-03-20
Talk only through the loudspeaker


Ownerofdump:
### Steps to reproduce

1. Start Call (1:1 / group / there is no difference)
2. I switch to the phone conversation mode (the handset is on the top right
3. The sound continues to be played through the speaker

### Your phone model

Google Pixel 3a

### Operating system version

Android 12

### Application version and app store

Element X Android v26.03.X

### Homeserver

Yes

### Will you send logs?

No (sorry, but is not my phone)

### Are you willing to provide a PR?
No (sorry, but is not my phone)

bmarty:
Fixed by https://github.com/element-hq/element-x-android/pull/6609

PR #6609, opened 2026-04-17
audio: Let EC decide alone what communication device to use

### LINKED ISSUE #3886: ElementX android: Audio routing for voice calls is wrong (but app is set to audio mode)
Reproduce:
 - start a call on android using the audio call button
 - see the Back to speaker overlay (and the correct button for earpiece mode)
 - BUT the actual audio output is the loudspeaker
Workaround:
 - Press the speaker button (or go back to speaker)
 - Press the earpiece button again -> you will have sound through earpice mode and the same visual interface as when starting the call

BillCarsonFr:
## Content

We are going for a new strategy were EC alone controls what device to use.
So this PR is removing all the code were EX was trying to set the device.

Now EC decides (based on the ordered list of devices provided by EX) , and EX enforces that the wanted device is selected.

We have occurence (on android S+) of the communication device automatically changed by the OS(?) jsut after EX did pick a device, when this happens EX will now re-apply the EC choice.
It technically could create a loop where the system repeatidly revert the EC selection, but I haven't found a scenario reproducing that.

This is the alternative fix of https://github.com/element-hq/element-call/pull/3888

Should fix 

- https://github.com/element-hq/element-x-android/issues/6563
- https://github.com/element-hq/element-x-android/issues/6422

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


toger5:
Checking the logs they are indeed very confident, that Android does not anymroe call any of the native switch commands. But in a real world test it still switches to the speaker output device. So I think (if all android select methods are remove) android might do some magic we dont get noticed about.

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewAudioManager.kt
```diff
@@ -144,39 +128,20 @@ class WebViewAudioManager(
...
             val audioDevices = (listAudioDevices() + validNewDevices).distinctBy { it.id }.sortedWith(audioDeviceComparator)
             setAvailableAudioDevices(audioDevices.map(SerializableAudioDevice::fromAudioDeviceInfo))
-            // This should automatically switch to a new device if it has a higher priority than the current one
-            selectDefaultAudioDevice(audioDevices)
         }
 
         override fun onAudioDevicesRemoved(removedDevices: Array<out AudioDeviceInfo>?) {
             // Update the available devices
+            // Element Call will then decide to switch devices if needed
             setAvailableAudioDevices()
```
toger5, line 147: Unrelated but have we tested, that `listAudioDevices` will be up to date for the remove callback? Or do we need to do a similar thing to the `onAudioDevicesAdded` callback?

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewAudioManager.kt
```diff
@@ -113,23 +113,7 @@ class WebViewAudioManager(
...
-                    // Remove the expected id so we only retry once
-                    expectedNewCommunicationDeviceId = null
-                    audioManager.selectAudioDevice(expectedDeviceId.toString())
-                }
-            } else {
-                Timber.d("Audio device cleared")
-                expectedNewCommunicationDeviceId = null
-                audioManager.selectAudioDevice(null)
-            }
+            Timber.d("Audio device changed, type: ${device?.id}")
```
toger5, line 121: The existence of such a callback already implies, that the OS might do some funky auto switching?

I suspect (if we want the: control everything from EC) approach we need to do some manual work here to make sure, we revert back to the prev selected device if the change results in sth we did not initiate?
Could that be why its not working for me?

toger5 (changes requested):

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewAudioManager.kt
BillCarsonFr: Yes the OS might do, but when a device is added or removed. So this would be handled on EC side because this would have trigger a available device change event, and then EC will do the selection.
So there could be a period where the OS fallback to a device and EC havent yet processed the available update to choose the device. But that is fine?

The other case would be that another app is competing with us and switching, but that mean this app is taking audio focus and has more priority (native call?). But then I think in that case we should not do anything but instead listen to focus changes, and reselect our device when we have the focus back

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewAudioManager.kt
BillCarsonFr: not sure, have you seen something in your logs?

toger5 (changes requested):
We also have to switch from the deprecated methods to the new ones:
```

  setAvailableAudioDevices(devices: OutputDevice[]): void;
  setAudioDevice(id: string): void;
  onAudioDeviceSelect?: (id: string) => void;
  onAudioPlaybackStarted?: () => void;
  setAudioEnabled(enabled: boolean): void;
  showNativeAudioDevicePicker?: () => void;
  onBackButtonPressed?: () => void;

  /** @deprecated use  setAvailableAudioDevices instead*/
  setAvailableOutputDevices(devices: OutputDevice[]): void;
  /** @deprecated use  setAudioDevice instead*/
  setOutputDevice(id: string): void;
  /** @deprecated use  onAudioDeviceSelect instead*/
  onOutputDeviceSelect?: (id: string) => void;
  /** @deprecated use  setAudioEnabled instead*/
  setOutputEnabled(enabled: boolean): void;
  /** @deprecated use  showNativeAudioDevicePicker instead*/
  showNativeOutputDevicePicker?: () => void;
```

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewAudioManager.kt
```diff
@@ -410,16 +358,16 @@ class WebViewAudioManager(
                     return
                 }
                 setAudioEnabled(true)
+                @Suppress("DEPRECATION")
                 isSpeakerphoneOn = device.type == AudioDeviceInfo.TYPE_BUILTIN_SPEAKER
                 isBluetoothScoOn = device.type == AudioDeviceInfo.TYPE_BLUETOOTH_SCO
             } else {
+                @Suppress("DEPRECATION")
```
jmartinesp, line 360: Is it better having these 2 annotations than the one for the whole method?

jmartinesp (approved):
Thanks, the PR seems to work fine for the most part. I wasn't able to test the case when you add or remove audio devices (I don't have BT headphones around at the moment), so maybe someone else could test that.

We still have the long standing issue where the first selected device always uses the media volume instead of the call one when EXA starts the call, but that's nothing new.

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewAudioManager.kt
BillCarsonFr: Makes it clearer what is exaclty deprecated exactly

PR #6620, opened 2026-04-18
Fix crash when going back to threads list

bxdxnn:
## Content

The fix wraps both calls with `runCatchingExceptions` to gracefully handle the destroyed state. If the inner is already destroyed, it starts with empty items and lets the subscription updates populate the list instead of crashing.

## Motivation and context

The crash occurs because `inner.items()` and `inner.paginationState()` are called in the `onStart` block, which can fail if the Rust `ThreadListService` has been destroyed before the subscription tries to use it (e.g., when navigating back from a thread detail view).

## Tests

- Go to the threads list, then open any thread, then go back. It shouldn't crash anymore.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

    - [ ] No.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/threads/list/ThreadsListPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/threads/list/ThreadsListPresenter.kt
@@ -97,12 +97,6 @@ class ThreadsListPresenter(
 
         val roomInfo by room.roomInfoFlow.collectAsState()
 
-        DisposableEffect(Unit) {
-            onDispose {
-                threadsListService.destroy()
-            }
-        }
-
         fun handleEvent(event: ThreadsListEvents) {
             when (event) {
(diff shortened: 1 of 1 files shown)


jmartinesp:
Thanks for the fix (and the report)! While it's a valid workaround, I think the actual fix should be removing these lines instead:

https://github.com/element-hq/element-x-android/blob/80470b37920a4f8443eb5f09d5be7ceedfe71378/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/threads/list/ThreadsListPresenter.kt#L100-L104

In the beginning, a new `RustThreadListService` was being instantiated every time you opened this presenter because it was an asynchronous process. Now it's synchronous and we keep a single instance around, which is destroyed along with the room pointer in Rust. It seems like we forgot to remove this previous cleanup step which is now causing the crash 😬 .

Could you remove that `DisposableEffect` block and update the PR?

bxdxnn:
@jmartinesp I've tested your fix, works too.

jmartinesp (approved):
LGTM, thanks again!

PR #6626, opened 2026-04-20
Sync Strings

ElementBot:


screenshots/de/libraries.mediaviewer.impl.viewer_MediaViewerViewLandscape_12_de.png
jmartinesp, line 1: Not from here, but it seems like this screen is broken in landscape 🫤 .

jmartinesp: So is the info bottom sheet, it's not scrollable.

jmartinesp (approved):
LGTM.

jmartinesp (changes requested):
... or not:

```
element-x-android/element-x-android/libraries/ui-strings/src/main/res/values-uz/translations.xml:413: Error: Inconsistent number of arguments in formatting string crypto_event_key_forwarded_unknown_profile_dialog_content; found both 2 here and 1 in values/localazy.xml [StringFormatCount]
    <string name="crypto_event_key_forwarded_unknown_profile_dialog_content">"%1$s (%2$s) bu xabarni ulashdi, chunki u yuborilganda siz xonada emas edingiz."</string>
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
element-x-android/element-x-android/libraries/ui-strings/src/main/res/values/localazy.xml:413: Conflicting number of arguments (1) here
```

features/invitepeople/impl/src/main/res/values-fr/translations.xml
```diff
@@ -2,4 +2,12 @@
...
   <string name="screen_invite_users_already_a_member">"Déjà membre"</string>
   <string name="screen_invite_users_already_invited">"Déjà invité(e)"</string>
+  <plurals name="screen_invite_users_confirm_dialog_subtitle">
+    <item quantity="one">"Vous n’avez actuellement aucune conversation avec ce contact. Veuillez confirmer son invitation à rejoindre ce salon avant de continuer."</item>
+    <item quantity="other">"Vous n’avez actuellement aucune conversation avec ces contacts. Veuillez confirmer leur invitation à rejoindre ce salon avant de continuer."</item>
+  </plurals>
+  <plurals name="screen_invite_users_confirm_dialog_title">
+    <item quantity="one">"Inviter un nouveau contact dans ce salon ?"</item>
+    <item quantity="other">"Inviter de nouveaux contacts dans ce salon ?"</item>
+  </plurals>
```
jmartinesp: These strings seem to be problematic:

```
Lint found 6 errors and 4 warnings. First failure:
/home/runner/work/element-x-android/element-x-android/features/invitepeople/impl/src/main/res/values-fr/translations.xml:6: Error: The quantity 'one' matches more than one specific number in this locale (0, 1), but the message did not include a formatting argument (such as %d). This is usually an internationalization error. See full issue explanation for more. [ImpliedQuantity]
    <item quantity="one">"Vous n’avez actuellement aucune conversation avec ce contact. Veuillez confirmer son invitation à rejoindre ce salon avant de continuer."</item>
```

Any idea how we could fix this, @bmarty, @ganfra ? I don't think we can manually suppress this check anywhere.

bmarty: This patch fixes it in French: 
[lintConfig.patch](https://github.com/user-attachments/files/26887475/lintConfig.patch)

but other languages (hu and ru) are in error, and I do not think we can really ignore the error the same way:

>  /Users/bmarty/workspaces/element-x-android/features/invitepeople/impl/src/main/res/values-hr/translations.xml:6: Error: The quantity 'one' matches more than one specific number in this locale (1, 21, 31, 41, 51, 61, 71, 81, 101, 1001, …), but the message did not include a formatting argument (such as %d). This is usually an internationalization error. See full issue explanation for more. [ImpliedQuantity]

```xml
<item quantity="one">"Trenutno nemate razgovora s ovim kontaktom. Potvrdite da ste ga pozvali u ovu sobu prije nego što nastavite."</item>
```

11 could potentially (hardly, but still) happen.

In this case I would just delete the incorrect translation, with the risk that the error come back...

jmartinesp: Localazy doesn't allow you to add custom plural values for those languages though... just 'one', 'few', 'other'.

bmarty: The problem here is that the string itself does not contain `%d` and the string will be used by the system when the integer is 1, 21, 31, etc. (so not only 1)

bmarty: Basically the "one" item is used for 1 and some other numbers ending with 1

bmarty: Maybe the fix is to use a single `string` like `You currently don’t have any chats with this (these) contact(s). Confirm inviting them to this room before continuing.`.

And try to remember to never use plurals without `%d` in the English strings...

jmartinesp: https://github.com/element-hq/element-x-android/pull/6626/commits/3fc126172e369d17961ecab0a840dc63e6551c41

jmartinesp (approved):
Ok, should be fixed now.

PR #6631, opened 2026-04-21
Fix media viewer bottom sheets not being scrollable

jmartinesp:
## Content

Make the bottom sheets for the media viewer screen scrollable.

## Motivation and context

In some screenshot tests, the landscape screenshots for these bottom sheets looked weird. It turns out it wasn't just an issue in the screenshot tests, but a real one in the app.

Both the info and confirm deletion bottom sheets were not scrollable, which made them useless on devices with low resolution or landscape orientation in most devices.

## Screenshots / GIFs

No changes, I hope.

## Tests

Open the media viewer in landscape and then tap on the 'more info' button, which should open a bottom sheet that's now scrollable. Using it, scroll to the 'delete' action and use it, which should display a confirmation bottom sheet that should also be scrollable now.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

    - [x] No.

### DIFF
--- a/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/details/MediaDeleteConfirmationBottomSheet.kt
+++ b/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/details/MediaDeleteConfirmationBottomSheet.kt
@@ -17,8 +17,11 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.verticalScroll
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.Text
+import androidx.compose.material3.rememberModalBottomSheetState
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
@@ -54,10 +57,12 @@ fun MediaDeleteConfirmationBottomSheet(
     ModalBottomSheet(
         modifier = modifier,
         onDismissRequest = onDismiss,
+        sheetState = rememberModalBottomSheetState(skipPartiallyExpanded = true)
     ) {
         Column(
             modifier = Modifier
                 .fillMaxWidth()
+                .verticalScroll(rememberScrollState())
                 .padding(horizontal = 16.dp),
         ) {
             IconTitleSubtitleMolecule(
--- a/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/details/MediaDetailsBottomSheet.kt
+++ b/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/details/MediaDetailsBottomSheet.kt
@@ -15,6 +15,8 @@ import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.verticalScroll
 import androidx.compose.material3.ExperimentalMaterial3Api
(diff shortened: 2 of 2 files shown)


libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/details/MediaDeleteConfirmationBottomSheet.kt
```diff
@@ -54,10 +57,12 @@ fun MediaDeleteConfirmationBottomSheet(
     ModalBottomSheet(
         modifier = modifier,
         onDismissRequest = onDismiss,
+        sheetState = rememberModalBottomSheetState(skipPartiallyExpanded = true)
     ) {
         Column(
             modifier = Modifier
                 .fillMaxWidth()
+                .verticalScroll(rememberScrollState())
```
bmarty, line 65: Maybe using a `LazyColumn` is better?

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/details/MediaDetailsBottomSheet.kt
```diff
@@ -62,7 +64,8 @@ fun MediaDetailsBottomSheet(
     ) {
         Column(
             modifier = Modifier
-                .fillMaxWidth(),
+                .fillMaxWidth()
+                .verticalScroll(rememberScrollState()),
```
bmarty, line 68: Maybe using a `LazyColumn` is better?

bmarty (commented):
Thanks. I have found other `ModalBottomSheet`s that cannot scroll:
- RoomListContextMenu
- RoomListDeclineInviteMenu
- InvitePeopleConfirmModal (there is a `LazyColumn`, but not sure we can access the button below)
- ResolveVerifiedUserSendFailureView
- AttachmentsBottomSheet
- ChangeOwnRoleBottomSheet
- RoomMemberActionsBottomSheet
- JoinRoomByAddressView
- CreateDmConfirmationBottomSheet
- CaptionWarningBottomSheet

Maybe an opportunity to fix them all? Not sure how to ensure that the issue does not happen again in the next Modal we will write. Any ideas?

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/details/MediaDeleteConfirmationBottomSheet.kt
jmartinesp: I don't think a `LazyColumn` is worth it, the amount of UI components to display is very small. It would make sense for displaying 50-100 rows or something like that.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/details/MediaDetailsBottomSheet.kt
jmartinesp: Ditto ^

jmartinesp:
> Maybe an opportunity to fix them all? 

Good idea.

> Not sure how to ensure that the issue does not happen again in the next Modal we will write. Any ideas?

No idea either, I was thinking about a Konsist check, but that's pretty much impossible to enforce 🫤 .

bmarty:
> No idea either, I was thinking about a Konsist check, but that's pretty much impossible to enforce 🫤 .

Maybe `content` parameter of `fun ModalBottomSheet` could be of type `LazyListScope.() -> Unit`? But we will lose the ability to have a "header" in the content. Or we add an optional one: `header: @Composable ColumnScope.() -> Unit = {}`

jmartinesp:
Maybe we could pass also a 'container' `@Composable`, like:

```kotlin
@Composable
fun ModalBottomSheet(
    ...
    container: @Composable: (content: @Composable ColumnScope.() -> Unit) -> Unit = {  content ->
        Column(modifier = Modifier.verticalScroll(rememberScrollState()), content = content) 
    },
    content: @Composable ColumnScope.() -> Unit,
)
```

ganfra (approved):

jmartinesp:
I'll merge this as is, then create a separate PR for generalising a fix, since we're still exploring different options.

PR #6633, opened 2026-04-21
Update wording of deactivate account screen

### LINKED ISSUE #6608: Update deactivate account screen
See details and discussion here: https://github.com/element-hq/element-x-ios/issues/5432

bmarty:
## Content

First commit is adding helper methods to use generic plurals (without mentioned the actual quantity) in a handy way.
Second commit is update wording of deactivate account screen.

## Motivation and context

Closes #6608 

## Tests

- See recorded screenshots.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/ui-strings/src/main/res/values/localazy.xml
+++ b/libraries/ui-strings/src/main/res/values/localazy.xml
@@ -84,8 +84,8 @@
   <string name="action_create">"Create"</string>
   <string name="action_create_room">"Create room"</string>
   <string name="action_create_space">"Create space"</string>
-  <string name="action_deactivate">"Deactivate"</string>
-  <string name="action_deactivate_account">"Deactivate account"</string>
+  <string name="action_deactivate">"Delete"</string>
+  <string name="action_deactivate_account">"Delete account"</string>
   <string name="action_decline">"Decline"</string>
   <string name="action_decline_and_block">"Decline and block"</string>
   <string name="action_delete_poll">"Delete Poll"</string>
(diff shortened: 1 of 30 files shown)


libraries/ui-strings/src/main/res/values/localazy.xml
```diff
@@ -84,8 +84,8 @@
   <string name="action_create">"Create"</string>
   <string name="action_create_room">"Create room"</string>
   <string name="action_create_space">"Create space"</string>
-  <string name="action_deactivate">"Deactivate"</string>
-  <string name="action_deactivate_account">"Deactivate account"</string>
+  <string name="action_deactivate">"Delete"</string>
+  <string name="action_deactivate_account">"Delete account"</string>
```
pixlwave: Just observing that this goes against the rule that `action/common` keys should match the contents and will likely be confusing if someone tries to use `action_deactivate` elsewhere.

We should probably rename the keys too in this instance (or create new a new `action_delete` string and just rename the `action_deactivate_account` to `action_delete_account` as deactivate might be useful in the future.)

bmarty: I agree. But I wanted to keep the existing key names to highlight the fact that the keys are used in the context of account deactivation. Maybe I could just delete the 2 `action_deactivate` keys and use longer key name specifically for this feature.

bmarty: @pixlwave I have created `action_delete` and `action_delete_account`. Later (when both Android and iOS are not using when anymore), we could just delete the keys `action_deactivate` and `action_deactivate_account` on Localazy.

jmartinesp (approved):
LGTM, thanks!

libraries/ui-strings/src/main/res/values/localazy.xml
pixlwave: Super, thanks 🙏

PR #6634, opened 2026-04-21
Make Element Call screen work edge-to-edge

jmartinesp:
## Content

- Remove `Scaffold` component from `CallScreenView`. We didn't use it anymore and it consumed the window insets.
- Consume the window insets for the webview. This seems to allow further customisation on the webview side of things.
- Add immersive mode when the screen is in 'compact height' mode (aka landscape) so the call is truly edge-to-edge.

Needs https://github.com/element-hq/element-x-android/pull/6766.

## Motivation and context

Requested by @robintown from the VOIP team.

## Screenshots / GIFs

https://github.com/user-attachments/assets/b64e74cf-4b66-44d7-b78c-b1428c0686d9

## Tests

Start a call:

- When in portrait mode, the screen should look the same.
- On landscape, it should display edge-to-edge.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
@@ -32,6 +32,9 @@ import androidx.compose.runtime.rememberUpdatedState
 import androidx.core.app.PictureInPictureModeChangedInfo
 import androidx.core.content.IntentCompat
 import androidx.core.util.Consumer
+import androidx.core.view.WindowCompat
+import androidx.core.view.WindowInsetsCompat
+import androidx.core.view.WindowInsetsControllerCompat
 import androidx.lifecycle.Lifecycle
 import dev.zacsweers.metro.Inject
 import io.element.android.compound.colors.SemanticColorsLightDark
@@ -52,6 +55,7 @@ import io.element.android.libraries.audio.api.AudioFocusRequester
 import io.element.android.libraries.core.log.logger.LoggerTag
 import io.element.android.libraries.core.meta.BuildMeta
 import io.element.android.libraries.designsystem.theme.ElementThemeApp
+import io.element.android.libraries.designsystem.utils.hasCompactHeightWindowSize
 import io.element.android.libraries.featureflag.api.FeatureFlagService
 import io.element.android.libraries.preferences.api.store.AppPreferencesStore
 import timber.log.Timber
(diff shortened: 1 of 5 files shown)


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
```diff
@@ -111,6 +115,27 @@ class ElementCallActivity :
             val colors by remember(webViewTarget.value?.sessionId) {
                 enterpriseService.semanticColorsFlow(sessionId = webViewTarget.value?.sessionId)
             }.collectAsState(SemanticColorsLightDark.default)
+
+            // When the height is compact, hide the system bars by default to maximize the space for the call, using immersive mode
+            val hasCompactHeight = hasCompactHeightWindowSize()
+            DisposableEffect(hasCompactHeight) {
+                if (hasCompactHeight) {
+                    val window = this@ElementCallActivity.window ?: return@DisposableEffect onDispose {}
```
bmarty, line 123: Could be simplified even if I think that `window` can never be null here (but who knows with the Android API 😅 )


```suggestion
                    window ?: return@DisposableEffect onDispose {}
```

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
```diff
@@ -111,6 +115,27 @@ class ElementCallActivity :
             val colors by remember(webViewTarget.value?.sessionId) {
                 enterpriseService.semanticColorsFlow(sessionId = webViewTarget.value?.sessionId)
             }.collectAsState(SemanticColorsLightDark.default)
+
+            // When the height is compact, hide the system bars by default to maximize the space for the call, using immersive mode
+            val hasCompactHeight = hasCompactHeightWindowSize()
+            DisposableEffect(hasCompactHeight) {
+                if (hasCompactHeight) {
```
bmarty: `hasCompactHeight` is true when in PIP mode, which is not causing any issue but the code to hide the bars is a bit useless, since they are already hidden in the pip window. But maybe change the test to:

```kotlin
DisposableEffect(hasCompactHeight, pipState.isInPictureInPicture) {
    if (hasCompactHeight && !pipState.isInPictureInPicture) {
```

Tested OK on my side.

Also worth noting that when in landscape mode, the header of Element Call is not displayed, so the button to enter PiP is not displayed too. Pressing back can enter PiP mode though.

bmarty (approved):
Tested OK, thanks.
Feel free to ignore my remarks.

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/ElementCallActivity.kt
jmartinesp: In `Activity` there are several checks for `mWindow != null` so I guess it can be null sometimes. Not when this is running, probably, but better safe than sorry, I think 😅 .

Issue #6632, opened 2026-04-21
Calls broken on Huawei devices (WebView 114): Promise.withResolvers + addWebMessageListener issues


manfrommedan:
## Description

Element X v26.02 calls don't work on Huawei devices — the WebView shows "Join as guest" page instead of the call UI. Tested on Huawei with system WebView based on Chromium 114.

v26.01 worked fine on the same device.

## Root Causes

Two separate issues were identified by debugging on-device:

### 1. `Promise.withResolvers` not available in older WebViews

The bundled Element Call JS module (`assets/element-call/assets/index-*.js`) uses `Promise.withResolvers()` (10 occurrences). This API was introduced in **Chrome 119**. Huawei ships a system WebView based on **Chromium 114**, so the ES module silently fails to load (`<script type="module">` errors don't trigger `window.onerror`). Element Call never initializes → "Join as guest" page is shown instead of the call UI.

**Affected devices:** Any device with system WebView older than Chrome 119 (Huawei, some older Samsung devices, etc.)

**Suggested fix:** Add a polyfill in `element-call/index.html` before the module script tag, or configure the bundler (Vite/esbuild) to target a lower Chrome version:

```javascript
if (!Promise.withResolvers) {
    Promise.withResolvers = function () {
        var resolve, reject;
        var promise = new Promise(function (a, b) { resolve = a; reject = b; });
        return { promise: promise, resolve: resolve, reject: reject };
    };
}
```

### 2. `addWebMessageListener` reports as supported but doesn't deliver messages on Huawei

In `WebViewWidgetMessageInterceptor.kt`, when `WebViewFeature.isFeatureSupported(WebViewFeature.WEB_MESSAGE_LISTENER)` returns `true`, the code uses `WebViewCompat.addWebMessageListener()`. On Huawei WebView 114, this feature **reports as supported** but messages are never actually delivered to the native listener. The JS proxy object created by `addWebMessageListener` also shadows the `addJavascriptInterface` object (both registered as `"elementX"`), breaking the fallback path too.

Result: `content_loaded` from Element Call JS never reaches native code → widget driver never sends auth tokens → deadlock → call never starts.

**Suggested fix options:**
- Add a runtime ping/pong check to verify `addWebMessageListener` actually delivers messages before relying on it
- Or use `addJavascriptInterface` universally — it works on all devices and there's no security or feature downside for this use case (local assets, trusted content, `setOf("*")`)
[...]

PR #6640, opened 2026-04-22
Fix calls on Huawei devices: skip addWebMessageListener on Chromium < 119

manfrommedan:
## Problem

Calls do not work on Huawei and other devices with system WebView based on Chromium < 119. The call screen shows "Join as guest" instead of the call UI.

Huawei WebView reports `WEB_MESSAGE_LISTENER` as supported via `WebViewFeature.isFeatureSupported()`, but silently drops messages sent through `addWebMessageListener`. This is a known issue across the Android ecosystem. Capacitor, Home Assistant, Flutter InAppWebView and Tauri have all documented workarounds for Huawei WebView (see ionic-team/capacitor#6555, ionic-team/capacitor#6258).

Huawei does not allow users to change the WebView provider on their devices.

## Fix

- Register `addJavascriptInterface` unconditionally as the baseline message channel (works on all WebView implementations)
- Only use `addWebMessageListener` when the WebView is based on Chromium 119+, where it is known to work reliably
- On Chrome 119+ both are registered, `addWebMessageListener` takes priority (same behavior as before)
- On Chromium < 119 only `addJavascriptInterface` is used, avoiding the broken `addWebMessageListener`

## Note

There is a second issue affecting these devices: Element Call JS bundle uses `Promise.withResolvers` (Chrome 119+), which causes the ES module to silently fail on older WebViews. A polyfill fix for that has been submitted to Element Call: element-hq/element-call#3905

## Test plan

- Tested on Huawei device with WebView Chromium 114: calls work correctly
- Built from source, full debug APK verified on device
- `compileDebugKotlin` passes, full `assembleDebug` passes

Related: #6632

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewWidgetMessageInterceptor.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/WebViewWidgetMessageInterceptor.kt
@@ -140,26 +140,32 @@ class WebViewWidgetMessageInterceptor(
             }
         }
 
-        // Create a WebMessageListener, which will receive messages from the WebView and reply to them
-        val webMessageListener = WebViewCompat.WebMessageListener { _, message, _, _, _ ->
-            onMessageReceived(message.data)
-        }
+        // Always register JavascriptInterface as the baseline message channel.
+        // This works on all WebView implementations including Huawei.
+        webView.addJavascriptInterface(object {
+            @JavascriptInterface
+            fun postMessage(json: String?) {
+                onMessageReceived(json)
+            }
+        }, LISTENER_NAME)
 
-        // Use WebMessageListener if supported, otherwise use JavascriptInterface
-        if (WebViewFeature.isFeatureSupported(WebViewFeature.WEB_MESSAGE_LISTENER)) {
+        // Additionally register WebMessageListener on WebViews that reliably support it.
+        // Huawei WebView (Chromium < 119) reports WEB_MESSAGE_LISTENER as supported
+        // but silently drops messages, so we only trust it on Chrome 119+.
+        // See: https://github.com/element-hq/element-x-android/issues/6632
+        val webViewVersion = WebViewCompat.getCurrentWebViewPackage(webView.context)
+            ?.versionName?.split(".")?.firstOrNull()?.toIntOrNull() ?: 0
+
+        if (webViewVersion >= 119 &&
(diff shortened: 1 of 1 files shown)


manfrommedan:
I have read the CLA Document and I hereby sign the CLA

jmartinesp (approved):
Thanks, the feature still works well in a device not affected by the issue. It would be nice to double check with some other affected user if this fixes their problems, but I guess the EC PR needs to be merged and integrated first.

maksimkurb:
Can confirm that applying this fix as well as https://github.com/element-hq/element-call/pull/3905 fix problems with calling on Huawei devices (tested at least on two different phones). Thanks a lot!

bestperson-free:
> Подтверждаю, что применение этого исправления, а также решения [ проблемы element-hq/element-call#3905, ](https://github.com/element-hq/element-call/pull/3905) устраняет проблемы со звонками на устройствах Huawei (проверено как минимум на двух разных телефонах). Большое спасибо!

Какую версию сейчас можно использовать?

maksimkurb:
> Какую версию сейчас можно использовать?

Until these two fixes are merged into the main branch you can clone repositories and compile them manually: firstly, patch element-call and publish it to the local maven repository under some version like `0.19.999-SNAPSHOT` to avoid collisions with real versions, then patch element-x-android with this fix and replace element-call version from 0.19.x to `0.19.999-SNAPSHOT`. Also, you would need to add `mavenLocal()` to the `repositores` section in the `build.gradle` files.

I can share [my built APKs](https://plik.cubly.ru/#/?id=B0mwcceoJNFvNCPy) with you, **but use them at your own risk**: I am just a stranger from the Internet, and you don't have to trust my APKs as they can contain backdoors and etc (I'm not a bad guy, I just tell you to be cautious). Ideally, you can use these APKs on some non-production matrix server with two demo accounts just to test that this fix works and then compile Element X by yourself or wait until this fix will be merged and released.

jmartinesp (approved):
Approving this since I could confirm the 'normal' scenario works as expected and some affected users have confirmed this fixes the issue for them when applied with another fix in the Element Call side.

dreamzxb:
After testing Element X for Android v26.05.0, it appears this issue has not yet been resolved; the problem persists on Android devices exactly as it did before.

maksimkurb:
MR on the element-call side has not been merged yet, and it is also required to fix calls on Huawei.
You should wait until https://github.com/element-hq/element-call/pull/3905 is merged and then wait until element-x-android updates its element-call dependency to the fixed version.

PR #6642, opened 2026-04-22
Change native back button behavior in EC view (close settings in EC with os native back)

### LINKED ISSUE #3909: EX android send esc on native back button

toger5:
Fixes: https://github.com/element-hq/element-call/issues/3909

 
## Content
this changes the back button behavior to:
 - inject escape into webview instead of going back.
 - the webview will call back when no other modal is open.

## Motivation and context
https://github.com/element-hq/element-call/issues/3385

## Screenshots / GIFs

[Screen_recording_20260428_155256.webm](https://github.com/user-attachments/assets/6c923472-b26f-4bcb-98ec-9ef0f7988262)

(At the end it looks like things don't react. This is a slow emulator. Just wait it will react eventually)
## Tests

- Run the correct EX version
- Setup the correct EC version in the devtools: https://github.com/element-hq/element-call/pull/3927 (using `https://pr3927--element-call.netlify.app/room` as the url)
- start a call
- open settings
- press native back (or use gesture)
- see the settings closing
- press back again
- observe that we move into pip if nothing is focussed

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

    - [x] Yes. In this case, please request a review by Copilot. (all code was written by hand but AI helped with api usage AND the Robolectric tests for the CallScreenView.kt has been fully generated based on the exact test cases)
    - [ ] No.

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenView.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenView.kt
@@ -67,12 +67,12 @@ internal fun CallScreenView(
     var callWebView by remember { mutableStateOf<WebView?>(null) }
 
     fun handleBack(fromNative: Boolean = false) {
-        when (CallScreenBackPressPolicy.resolve(supportPip = pipState.supportPip, hasWebView = callWebView!=null,fromNative)) {
+        when (CallScreenBackPressPolicy.resolve(supportPip = pipState.supportPip, hasWebView = callWebView != null, fromNative)) {
             CallScreenBackPressAction.EnterPictureInPicture ->
                 pipState.eventSink(PictureInPictureEvents.EnterPictureInPicture)
             CallScreenBackPressAction.DispatchEscapeToWebView ->
                 callWebView?.dispatchEscKeyEvent()
-            null-> Timber.d("Back press with unsupported pip is a no-op")
+            null -> Timber.d("Back press with unsupported pip is a no-op")
         }
     }
 
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenView.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenView.kt
@@ -75,8 +75,9 @@ internal fun CallScreenView(
     Scaffold(
         modifier = modifier,
     ) { padding ->
+        var callWebView by remember { mutableStateOf<WebView?>(null) }
         BackHandler {
-            handleBack()
+            callWebView?.dispatchEscKeyEvent()
         }
         if (state.webViewError != null) {
             ErrorDialog(
@@ -111,6 +112,7 @@ internal fun CallScreenView(
                 },
                 onConsoleMessage = onConsoleMessage,
                 onCreateWebView = { webView ->
+                    callWebView = webView
                     webView.addBackHandler(onBackPressed = ::handleBack)
                     val interceptor = WebViewWidgetMessageInterceptor(
                         webView = webView,
@@ -135,6 +137,9 @@ internal fun CallScreenView(
                     pipState.eventSink(PictureInPictureEvents.SetPipController(pipController))
                 },
                 onDestroyWebView = {
+                    if (callWebView === it) {
+                        callWebView = null
+                    }
                     // Reset audio mode
                     webViewAudioManager?.onCallStopped()
                 }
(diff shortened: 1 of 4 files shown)


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenView.kt
```diff
@@ -257,6 +271,11 @@ private fun WebView.addBackHandler(onBackPressed: () -> Unit) {
     )
 }
 
+private fun WebView.dispatchEscKeyEvent() {
+    dispatchKeyEvent(android.view.KeyEvent(android.view.KeyEvent.ACTION_DOWN, android.view.KeyEvent.KEYCODE_ESCAPE))
+    dispatchKeyEvent(android.view.KeyEvent(android.view.KeyEvent.ACTION_UP, android.view.KeyEvent.KEYCODE_ESCAPE))
+}
```
jmartinesp, line 269: This is an interesting way to pass the action to the webview 😅 .

jmartinesp:
@toger5 the provided test url (`https://pr3908-sdk--element-call.netlify.app/room`) doesn't seem to be working, I always get a 404.

toger5:
@jmartinesp you need to use the one without the sdk.https://pr3908--element-call.netlify.app/room (failiour on my side. Copied the wrong one. Update in the description as well)

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenView.kt
toger5: The alternative would be to do some javascript injection. Which i did not like...
this was what I found in the webView android docs and sounded like it was what i needed... What would you have preferred?

Did you expect a dedicated action?
I think in this case a generic button press is less overhead since dismiss on escape often is for free on web.

jmartinesp:
Thanks, I can test it now. I think this needs a bit more work:

1. Tapping the back button doesn't seem to work in the lobby, previously it would exit this screen..

https://github.com/user-attachments/assets/d58fc13d-9a27-45fc-a707-e3a966873b33

2. When opening the overflow menu and then closing it, the back key and gesture stops working until I open some other menu.

https://github.com/user-attachments/assets/c4ba45c3-1a93-4b3f-98d1-738488f8ea13

The videos look a bit weird because I tested it with the back button in the emulator, the pointer in the screen mirroring and the actual device touchscreen, so it's a mix of everything 😅 .

features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenView.kt
jmartinesp: It's fine like that IMO, just a bit unorthodox.

toger5:
> Tapping the back button doesn't seem to work in the lobby, previously it would exit this screen..

this is an oversight in the EC app

toger5:
> When opening the overflow menu and then closing it, the back key and gesture stops working until I open some other menu.

This also is a mistake in the EC app. It still treats the ... button as active/focused and wont let the esc button drop down to the background.

So i think this PR will be untouched fixing those issues.

toger5:
We updated the logic in EC: with this PR: https://github.com/element-hq/element-call/pull/3927

(there was a better condition for when we ignore the esc button and when we do not)

jmartinesp:
Sorry, does this mean we should wait until the EC code is integrated? I was trying to test the branch with the new changes you mentioned, but it seems like the env is now gone.

toger5:
@jmartinesp I will ping you once it is ready. There is another issue that occured. (the drawer modals behave different to the normal modals)

toger5:
@jmartinesp now the EC side it in good shape to try again:
`https://pr3927--element-call.netlify.app/room`

I also update the screen share video in the original post.
It includes the edgecases that were not working before.

jmartinesp (approved):
Everything seems to be working now, thanks. Let me know if you need help fixing the lint issues.

Can this be merged as is, or does it need a new EC embedded version to work properly?

toger5:
Imo it can get merged as is. It will only work like it is shown in the next EC version however.
But merging it now would result in:
 - No pip with OS back button (that now needs a call from EC)
 - Back button would close modals in EC (since this PR now sends esc keys)
 - The only way to get into pip is with the dedicated minimise button in the top left corner.
 
 But I think this is an acceptable intermediate state?
 It is similarly broken as right now:
 You will always go into pip on back button press  but cannot close modals in EC VS you can close modals in EC but never get to pip with the OS back button.

PR #6652, opened 2026-04-24
Allow forwarding messages to multiple rooms

### LINKED ISSUE #6520: Forward / share to several persons
### Steps to reproduce

1. Long press a photo/message
2. Share / forward to someone else or to another room
3. Select Element X
4. On the list, only radio button to send to ONE person/room

#### What did you expect?
Be able to select only SEVERAL person/room like on Element Classic (1.6.52):

Image

#### What happened instead?
Element X, only radio button so we can select only ONE person/room:

Image

Despite on 

### Your phone model

Samsung S24+

### Operating system version

Android 16.0

Element X 26.03.3

### Homeserver

Synapse 1.150.0

### Will you send logs?

No

No

bxdxnn:
## Motivation and context

Fixes #6520

The functionality already existed but it was commented-out.

## Tests

- Test that sharing (either from intent or using "Forward" for messages) sends to multiple rooms

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [x] No.


libraries/roomselect/impl/src/main/kotlin/io/element/android/libraries/roomselect/impl/RoomSelectView.kt
```diff
@@ -73,7 +73,6 @@ fun RoomSelectView(
 ) {
     @Suppress("UNUSED_PARAMETER")
     fun onRoomRemoved(roomInfo: SelectRoomInfo) {
-        // TODO toggle selection when multi-selection is enabled
```
bmarty, line 76: You should not use `RoomSelectEvents.RemoveSelectedRoom` anymore, it will deselect all the rooms. `RemoveSelectedRoom` can just be removed (there is a TODO about it).

libraries/roomselect/impl/src/main/kotlin/io/element/android/libraries/roomselect/impl/RoomSelectView.kt
```diff
@@ -148,7 +147,6 @@ fun RoomSelectView(
                 LazyColumn(state = lazyListState) {
                     item {
                         SelectedRoomsHelper(
-                            // TODO state.isForwarding
```
bmarty, line 151: Maybe remove the parameter `isForwarding`?

libraries/roomselect/impl/src/main/kotlin/io/element/android/libraries/roomselect/impl/RoomSelectView.kt
```diff
@@ -169,12 +167,14 @@ fun RoomSelectView(
...
-//                )
-                Spacer(modifier = Modifier.height(20.dp))
+                if (state.selectedRooms.isNotEmpty()) {
+                    SelectedRoomsHelper(
+                        isForwarding = false,
+                        selectedRooms = state.selectedRooms
+                    )
+                } else {
+                    Spacer(modifier = Modifier.height(20.dp))
+                }
```
bmarty, line 188: The screen will blink when we add the first room or remove the last room. This is not ideal, but I am not sure how we could improve that without having a blank strip before the user select the first room. Maybe use `AnimatedVisibility` to improve the rendering?

bmarty (commented):
Thanks for the PR @bxdxnn .

I have spotted an issue, when several rooms are selected, if you deselect one, all will be deselected.

@jmartinesp I see that multi selection has been disabled when the forward feature was introduced. What was the reason?

@mxandreas is it OK to have multi selection of rooms for both user stories:
- sharing content from a third party application
- forwarding a message from another room

Thanks.

jmartinesp:
> I see that multi selection has been disabled when the forward feature was introduced. What was the reason?

IIRC it was a product decision because it didn't work well on iOS at the time so we decided not to add it on Android... and then iOS fixed it out of the planning and we never aligned 🫠 .

https://github.com/element-hq/element-x-android/pull/635

libraries/roomselect/impl/src/main/kotlin/io/element/android/libraries/roomselect/impl/RoomSelectView.kt
bxdxnn: Refactored it

libraries/roomselect/impl/src/main/kotlin/io/element/android/libraries/roomselect/impl/RoomSelectView.kt
bxdxnn: I'm not sure, I don't have any blinking, but I've added an animation as you've requested anyway.
BTW, please clarify what kind of animation is wanted, e.g. fade-in/out or smoothly expand in height (and potentially width), etc.

mxandreas:
> and then iOS fixed it out of the planning and we never aligned

@jmartinesp It seems iOS currently does not enable forwarding to multiple rooms.
@bmarty Yes, it is fine to select multiple rooms in both cases.

jmartinesp:
> @jmartinesp It seems iOS currently does not enable forwarding to multiple rooms.

Oh sorry, I must have been thinking about sharing media/files to multiple rooms. I think the root issue was the same.

mxandreas:
> > @jmartinesp It seems iOS currently does not enable forwarding to multiple rooms.
> 
> Oh sorry, I must have been thinking about sharing media/files to multiple rooms. I think the root issue was the same.

You mean via the share sheet? There, too, you can only select a single room on iOS.

jmartinesp:
Ah true, sorry for the confusion again. It seems like I was thinking about sharing several files/media to the same room instead: https://github.com/element-hq/element-x-ios/pull/4358

PR #6654, opened 2026-04-25
Add clipping to RoomSummaryRow

bxdxnn:
## Content

Add clipping to the name and the latest event's text to room summary

## Motivation and context

If someone sends a very tall [Zalgo text](https://en.wikipedia.org/wiki/Zalgo_text) to a room, the room list becomes unreadable as the diacritics will span across the entire screen.

## Tests

- Test that Zalgo text clips
- Test that normal messages still render properly
- Test that normal diacritics still render properly

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

    - [x] No.

### DIFF
--- a/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/components/RoomSummaryRow.kt
+++ b/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/components/RoomSummaryRow.kt
@@ -29,6 +29,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clipToBounds
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.AnnotatedString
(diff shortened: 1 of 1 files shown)


features/home/impl/src/main/kotlin/io/element/android/features/home/impl/components/RoomSummaryRow.kt
```diff
@@ -222,20 +223,17 @@ private fun NameAndTimestampRow(
...
-                fontStyle = FontStyle.Italic.takeIf { name == null },
-                color = ElementTheme.colors.roomListRoomName,
-                maxLines = 1,
-                overflow = TextOverflow.Ellipsis
-            )
-        }
+        Text(
+            modifier = Modifier
+                .weight(1f)
+                .graphicsLayer { clip = true },
```
bmarty: There is an extension `Modifier.clipToBounds()` that can be used and do the same thing. We already use it at some places. Maybe use it so that we can quickly search for usage and see every place where we ensure that text does not mess up the screen.

features/home/impl/src/main/kotlin/io/element/android/features/home/impl/components/RoomSummaryRow.kt
```diff
@@ -282,7 +280,9 @@ private fun MessagePreviewAndIndicatorRow(
     ) {
         if (room.isTombstoned) {
             Text(
-                modifier = Modifier.weight(1f),
+                modifier = Modifier
+                    .weight(1f)
+                    .clipToBounds(),
```
bmarty: The text is not user generated content (UGC), it's `stringResource(R.string.screen_roomlist_tombstoned_room_description)`, so I think that using `clipToBounds()` is not necessary here.

features/home/impl/src/main/kotlin/io/element/android/features/home/impl/components/RoomSummaryRow.kt
```diff
@@ -303,7 +303,9 @@ private fun MessagePreviewAndIndicatorRow(
                 )
                 Spacer(modifier = Modifier.width(6.dp))
                 Text(
-                    modifier = Modifier.weight(1f),
+                    modifier = Modifier
+                        .weight(1f)
+                        .clipToBounds(),
```
bmarty: Same remark. Please only use `clipToBounds()` only on `Text`s which render UGC.

bmarty (changes requested):
Thanks for the update. See my new comments.

bmarty (approved):
Thanks!

Issue #6534, opened 2026-04-04
Room loading infinitely


bxdxnn:
### Steps to reproduce

1. Press on a room

#### What did you expect?

It will show a loading circle forever

### Your phone model

Reproducable on any phone

### Homeserver

Synapse 1.150.0

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

bxdxnn:
The issue is gone after clearing cache, but it's not clear if this will happen again.

bxdxnn:
It happened again today, for multiple users.
Sending rageshakes linking this issue.

bxdxnn:
One of the users got a more useful log (sent a rageshake linking this issue):
```
DEBUG matrix_sdk::http_client::native: Sending request num_attempt=1 | crates/matrix-sdk/src/http_client/native.rs:78 | spans: fetch_members > send{request_id="REQ-52" method=GET uri="https://<homeserver>/_matrix/client/v3/rooms/<room>/members"}
ERROR elementx: SDK background task failed
io.element.android.libraries.matrix.impl.core.SdkBackgroundTaskError: SDK background task 'timeline::encryption_state_changes' failure:
Early termination
  at org.matrix.rustcomponents.sdk.uniffiCallbackInterfaceClientDelegate$onBackgroundTaskErrorReport$callback$makeCall$1.invoke(r8-map-id-da36705454fffc50b4fe1debb995cced5a6f617df0a9e83730051b1c67d621ec:66)
  at org.matrix.rustcomponents.sdk.uniffiCallbackInterfaceClientDelegate$onBackgroundTaskErrorReport.callback(r8-map-id-da36705454fffc50b4fe1debb995cced5a6f617df0a9e83730051b1c67d621ec:38)
  at java.lang.reflect.Method.invoke(Native Method)
  at com.sun.jna.CallbackReference$DefaultCallbackProxy.invokeCallback(r8-map-id-da36705454fffc50b4fe1debb995cced5a6f617df0a9e83730051b1c67d621ec:60)
  at com.sun.jna.CallbackReference$DefaultCallbackProxy.callback(r8-map-id-da36705454fffc50b4fe1debb995cced5a6f617df0a9e83730051b1c67d621ec:1)
```

Indeed, I also see `stream reset by client (CANCEL)` after this URL in PCAPDroid.

jmartinesp:
Maybe it's this:

```
[...]

PR #6674, opened 2026-04-28
Mitigate a deadlock when loading room timelines

jmartinesp:
## Content

Ensure we destroy the media focused timeline when closing the media viewer/gallery screens, not when the room is disposed.

## Motivation and context

This may be happening because we were not destroying focused event timelines used for the media viewer/gallery when necessary, and having several of those in memory back paginating *may* have caused a deadlock in the event cache.

At least I can't reproduce the deadlock anymore, while previously I could reproduce it at least 1/3 of the times I used the media viewer in some rooms.

The symptoms match https://github.com/element-hq/element-x-android/issues/6534, but not the way to reproduce it. That may have been a different deadlock.

## Tests

Since this is a deadlock, it wasn't possible to consistently reproduce it. What I did was opening an image in the timeline, closing it, then opening another one and checking the logs and the screen state: if it's loading forever and you see no back pagination logs, then you have a deadlock. If you then exit the room and try to open it again, you'll be stuck in the home screen with an infinite loading dialog.

**EDIT:** the changes don't fully fix the issue, but they mitigate it, making it way less likely to happen. I've had some success on reproducing it with this branch by opening the media viewer, paginating and then closing and opening it quickly, with quick iterations.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.


jmartinesp:
After some more testing, it seems like this doesn't completely removes the deadlock, it just makes it less likely to happen... in any case, I think the changes to drop the media timelines early make sense.

ganfra (approved):

PR #6676, opened 2026-04-28
Show reactions in the pinned messages list

bxdxnn:
## Content

Support showing/adding reactions for messages in the pinned messages list

## Motivation and context

They weren't shown before, I'm not sure if that was intentional though.

## Tests

- Test that you can see the reactions for messages in the pinned messages list's timeline
- Test that you can add reactions to messages in the pinned messages list's timeline

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty (commented):
A comment for now: reactions are not shown on pinned event list in purpose, because they may not be accurate. Element X iOS does not show them either.

The problem we observed in the past is related to the reaction aggregation, which is not always working correctly. Reaction are events in the timeline, and for old events (pinned or not, but pinned events could get more late reactions than not pinned ones), that can get reactions from the time they were sent to now, it can render inaccurate reactions list.

I have been told that some improvement has been done SDK side regarding this, so it's maybe better now.

bxdxnn:
@bmarty I've checked on a 2 months old message, and the reactions seems to be accurate and the same if I load all messages in-between.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/pinned/list/PinnedMessagesListView.kt
```diff
@@ -195,11 +199,27 @@ private fun PinnedMessagesListLoaded(
         )
     }
 
+    fun onReactionClick(emoji: String, event: TimelineItem.Event) {
+        event.eventOrTransactionId.let { eventOrTransactionId ->
```
bmarty: This `let` statement can be removed to simplify the code.

bmarty (approved):
I have tested and reactions are correctly displayed on the pinned events timeline.

PR #6682, opened 2026-04-28
Add a way to tweak MAS url.

bmarty:
## Content

Be able to override the url related to MAS

## Motivation and context

Fill a customer request

## Tests

- Check that there is no impact in the login flow, the reset crypto identity flow and the link new device flow on FOSS version.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty:
@jmartinesp I have added a cache support for the .well-known file content.

libraries/wellknown/impl/src/main/kotlin/io/element/android/libraries/wellknown/impl/DefaultSessionWellknownRetriever.kt
```diff
@@ -10,29 +10,70 @@ package io.element.android.libraries.wellknown.impl
...
+    private val sessionCoroutineScope: CoroutineScope,
 ) : SessionWellknownRetriever {
     private val domain by lazy { matrixClient.userIdServerName() }
 
     override suspend fun getElementWellKnown(): WellknownRetrieverResult<ElementWellKnown> {
         val url = "https://$domain/.well-known/element/element.json"
+        val cacheData = cacheStore.getData(url)
+        if (cacheData != null) {
+            Timber.d("Element .well-known data retrieved from cache for $domain")
+            // If the cache is outdated, trigger a refresh in background but still return the cached value
```
jmartinesp, line 44: That's odd, I think. Shouldn't you try to load the new value, and if it fails, return the previous one?

bmarty: The idea was to return a value as fast as possible. There are few reasons for value to change on this file, so I guess this is acceptable?

jmartinesp: But once you have the updated value, wouldn't you be interested in using that one instead? With this approach you'll use the previous one until you check again, right? Maybe something like a `Flow` would be better?

bmarty: Not sure, the url is used as a one shot, to open the Web page, so a new emitted value should be just ignored?

Note: an update of the .well-known file is really an edge case, we should not bother too much about it.

jmartinesp (approved):
Thanks, I think we can merge it like this. 

I was trying to figure out why we needed 2 differently scoped services, and I finally remembered we have 2 differently scoped well-known retrievers, one using OkHttp for that and another one using the SDK because to use the SDK's `get_url` function we need a built `Client` 😅 . I'm sorry that delayed merging the PR.

PR #6683, opened 2026-04-28
Remove legacy `mx-reply` from `toPlainText` formatted event contents

### LINKED ISSUE #6673: Room list displays legacy `in reply to` text instead of the actual reply
due to replies from legacy applications that send `<mx-reply>` tags

Side question: Where is the `in reply to` string coming from? It's not in the formatted body of the reply and Element-X doesn't display that string anywhere else that I'm aware of. Is it being needlessly inserted given that it should never be used?

jmartinesp:
## Content

What the title says, when formatting a text message to display in plain text, ignore the contents of the `mx-reply` elements.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/6673.

## Tests

Send or receive a text message with a formatted body and an `mx-reply` element. Check the contents of this HTML element  are not added to the latest event text in the room list.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

    - [x] No.

### DIFF
--- a/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/ToHtmlDocument.kt
+++ b/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/ToHtmlDocument.kt
@@ -12,8 +12,10 @@ import io.element.android.libraries.matrix.api.permalink.PermalinkData
 import io.element.android.libraries.matrix.api.permalink.PermalinkParser
 import io.element.android.libraries.matrix.api.timeline.item.event.FormattedBody
 import io.element.android.libraries.matrix.api.timeline.item.event.MessageFormat
-import io.element.android.wysiwyg.utils.HtmlToDomParser
+import org.jsoup.Jsoup
 import org.jsoup.nodes.Document
+import org.jsoup.nodes.Document.OutputSettings
+import org.jsoup.safety.Safelist
 
 /**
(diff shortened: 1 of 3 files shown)


libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/ToHtmlDocument.kt
```diff
@@ -60,3 +62,35 @@ private fun fixMentions(
         }
     }
 }
+
+/** Custom Html to DOM parser, based on the one included in the rich text editor library. */
+private object HtmlToDomParser {
```
bmarty: The fact that the name is the same than in the RTE lib confused me at the beginning, because I thought that this class was not used. Then I saw that an import as been removed.

But maybe for clarity rename to `CustomHtmlToDomParser` or something else.

libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/ToPlainText.kt
```diff
@@ -78,6 +78,9 @@ private class PlainTextNodeVisitor : NodeVisitor {
             } else {
                 builder.append("• ")
             }
+        } else if (node is Element && node.tagName() == "mx-reply") {
```
bmarty: Maybe create a const for `"mx-reply"`?

bmarty (approved):
Thanks!

PR #6686, opened 2026-04-29
Rename `OIDC` components and variables to `OAuth`

jmartinesp:
## Content

What the title says. `:libraries:oidc` modules were also renamed.

Note some enums and DB-bound types were left untouched, otherwise they'd create issues when restoring the user's session.

## Motivation and context

This matches the new behavior in the SDK.

## Tests

Run the app with a logged in session, maybe run the unit tests.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
```diff
@@ -299,7 +299,7 @@ class RustMatrixAuthenticationService(
     /**
      * callbackUrl should be the uriRedirect from OidcClientMetadata (with all the parameters).
      */
-    override suspend fun loginWithOidc(callbackUrl: String): Result<SessionId> {
+    override suspend fun loginWithOAuth(callbackUrl: String): Result<SessionId> {
```
bmarty, line 302: Update the comment?

bmarty (approved):
Thanks. I can spot some remaining oidc instance in the code base, maybe rename them all?
- `oidcRedirectSchemeBase`
- `Oidc redirection`
- the file `oidc.md` (name and content)

Also about the renaming, Open Id Connect is a protocol using OAuth2, as far as I understand. Does this renaming make sense? I guess this topic have been discussed SDK side.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
jmartinesp: I couldn't find the referenced `OidcClientMetadata` component, so I left it like this. Maybe it's what preceded `OAuthAction` and the comment was already outdated?

bmarty: Ah yes, probably...

jmartinesp:
> Does this renaming make sense? I guess this topic have been discussed SDK side.

It seems like we're not fully following OIDC's spec (we mostly do, but there are some exceptions) and that's why we don't want to use `OIDC` everywhere and use a more generic `OAuth` term.

bmarty (approved):
Thanks!

PR #6694, opened 2026-04-29
Timeline: jump-to-unread FAB and mark-as-read shortcut

jennaharris7:
# Timeline: jump-to-unread + mark-as-read on the timeline FABs

## Content

Two additions to the room timeline, gated behind a new `JumpToUnread` feature flag (off by default; toggleable in developer options):

- **Jump-to-unread FAB** — chevron-up button that appears stacked above the scroll-to-bottom FAB whenever the read marker is above the visible viewport. Tapping centres the timeline on the marker.
- **Mark as read** — long-pressing either FAB opens a small popup with a "Mark as read" action that marks the room fully read at the latest event.

Both FABs show a small green dot when there are unreads in their direction:
- Top-of-FAB dot on the up-arrow when there are messages above the viewport.
- Bottom-of-FAB dot on the down-arrow when new messages have arrived since you scrolled away.

The dot is binary (visible / not visible), not a count.

## Motivation

When you scroll up in a busy room, the read marker leaves the screen and there's no quick way back. The chevron-up FAB takes you back in one tap. The green dots act as lightweight "there's something here" cues without cluttering the FAB with a number. The long-press shortcut lets you clear unread state without scrolling all the way back to the bottom.

## Screenshots / GIFs

*style updated to match figma:
******
Screenshot 2026-05-05 at 10 22 56 AM
******

screenshot1screenshot2

## Edge case not covered

If the read marker isn't in the loaded timeline window (e.g. focused-event timeline, paginated gaps), the chevron-up FAB won't appear. Closing this requires SDK plumbing to surface the fully-read event id so we can paginate-and-centre on it. Reasonable to land as a follow-up; this PR is gated behind `FeatureFlags.JumpToUnread` so the limited form can soak.

## Tests

> Enable the `Jump to unread messages` feature flag in developer options before exercising the manual cases below.

- [ ] Open a busy room with unreads → scroll up → chevron-up FAB appears stacked above chevron-down at bottom-right with a green dot above it; tap → timeline animates the read marker into view.
- [ ] Scrolled away from the bottom; receive a message from another user → green dot appears below the chevron-down FAB.
- [ ] Tap the chevron-down FAB or scroll back to the bottom → green dot clears.
- [ ] Long-press either FAB → "Mark as read" popup appears → tap → all unreads clear.
- [ ] Disable the feature flag → chevron-up FAB never appears; chevron-down FAB shows no green dot regardless of state. (Long-press menu still works.)
- [ ] Set the system font scale to maximum → "Mark as read" popup text wraps cleanly.
- [ ] Enable TalkBack → both FABs render and remain tappable; jump-to-unread FAB is announced.
- [ ] Unit tests pass: `./gradlew :features:messages:impl:testDebugUnitTest`.
- [ ] Konsist tests pass: `./gradlew :tests:konsist:testDebugUnitTest`.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Samsung Galaxy S24 on Android 16

## Checklist

    - [ ] No.


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
```diff
@@ -133,6 +138,7 @@ class TimelinePresenter(
         val prevMostRecentItemId = rememberSaveable { mutableStateOf<UniqueId?>(null) }
 
         val newEventState = remember { mutableStateOf(NewEventState.None) }
+        val newMessagesCount = remember { mutableIntStateOf(0) }
```
ganfra: Should probably be part of `NewEventState.FromOther` instead

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
```diff
@@ -268,7 +278,35 @@ class TimelinePresenter(
...
+        // [displayJumpToUnread]; the values are always computed so the state stays up to date if the
+        // feature flag flips at runtime.
+        val readMarkerIndex = remember { mutableIntStateOf(-1) }
+        val unreadMessagesCount = remember { mutableIntStateOf(0) }
+        LaunchedEffect(timelineItems) {
+            val items = timelineItems
+            withContext(dispatchers.computation) {
+                var markerIdx = -1
+                var unread = 0
+                for ((i, item) in items.withIndex()) {
```
ganfra: This logic won't work when the marker is not in memory, which can happens pretty easily (gaps, event-cache pagination)

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -358,19 +401,41 @@ private fun BoxScope.TimelineScrollHelper(
...
             .padding(end = 24.dp, bottom = 12.dp),
-        onClick = { jumpToBottom() },
+        onClick = ::jumpToBottom,
+    )
+    val jumpToUnreadDescription = if (unreadMessagesCount > 0) {
+        pluralStringResource(CommonPlurals.a11y_jump_to_unread_messages_count, unreadMessagesCount, unreadMessagesCount)
+    } else {
+        stringResource(id = CommonStrings.a11y_jump_to_unread_messages)
+    }
+    TimelineFab(
```
ganfra: `TimelineFab`  name is not precise enough on the meaning of the component, maybe something like `JumpToPositionButton`

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -380,22 +445,64 @@ private fun JumpToBottomButton(
...
+/**
+ * Small accent badge overlaid on a timeline FAB. Shows the count when it's between 1 and 9, otherwise a dot.
+ * Renders nothing when [count] is zero or negative.
+ */
+@Composable
+private fun TimelineCountBadge(
+    count: Int,
+    modifier: Modifier = Modifier,
+) {
+    if (count <= 0) return
```
ganfra: With a `when` would be better.
Also needs inputs from Design team

libraries/ui-strings/src/main/res/values/temporary.xml
```diff
@@ -7,4 +7,9 @@
 
 <resources>
     <string name="common_black">"Black"</string>
+    <string name="a11y_jump_to_unread_messages">"Jump to unread messages"</string>
```
ganfra: TODO: move to localazy

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -327,6 +363,13 @@ private fun BoxScope.TimelineScrollHelper(
         }
     }
 
+    fun jumpToReadMarker() {
+        if (readMarkerIndex < 0) return
```
ganfra: So like said previously, this won't work if the read marker is not loaded in memory

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -301,6 +324,19 @@ private fun BoxScope.TimelineScrollHelper(
             lazyListState.firstVisibleItemIndex < 3 && isLive
         }
     }
+    val isReadMarkerOffTop by remember {
```
ganfra: Should probably be name `isJumpToUnreadVisibile`

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -301,6 +324,19 @@ private fun BoxScope.TimelineScrollHelper(
...
+        derivedStateOf {
+            if (!displayJumpToUnread || readMarkerIndex < 0) {
+                false
+            } else if (forceJumpToReadMarkerVisibility) {
```
ganfra: Can we simplify this check? Also `forceJumpToReadMarkerVisibility` should win over anything else.

ganfra (commented):
Thanks for the PR, first comments after a quick look, I'll get back on this next week.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
jennaharris7: The iOS sibling PR (element-hq/element-x-ios#5506) hit the same limitation — they landed on the conclusion to ship behind a feature flag and follow up with SDK-supported scrolling. Proposing the same here. The proper fix needs an SDK accessor for the fully-read marker event id (it's in m.fully_read account data but not currently surfaced when the marker is outside the loaded window), plus rewiring the FAB to use Timeline.Mode.FocusedOnEvent on click. I’ve added an edge case not covered in the pr description, apologies for not including it initially!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
jennaharris7: Fixed in commit 78aa500d23eaf00fabfa47c30b902c327848a6a7

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
jennaharris7: Fixed in 7148f2f28f6d012c1eca395233396a4199ae9140

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
jennaharris7: Refactored to use when in 7148f2f28f6d012c1eca395233396a4199ae9140, will wait for design team feedback for other updates

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
jennaharris7: Fixed in 7148f2f28f6d012c1eca395233396a4199ae9140

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
jennaharris7: The forceJumpToReadMarkerVisibility win is fixed in 7148f2f2 and the check refactor is in 200b98b7

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
```diff
@@ -220,6 +225,13 @@ class TimelinePresenter(
                     timelineController.focusOnLive()
                 }
                 TimelineEvent.HideShieldDialog -> messageShieldDialogData.value = null
+                TimelineEvent.MarkAllAsRead -> sessionCoroutineScope.launch {
+                    val latestEventId = room.liveTimeline.getLatestEventId().getOrElse {
+                        Timber.tag(tag).w(it, "Failed to get latest event id to mark as fully read")
+                        return@launch
```
bmarty: I find a bit clearer to avoid `return` statement in a `getOrElse` block, but just return `null`.


```suggestion
                        null
```

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -358,47 +415,158 @@ private fun BoxScope.TimelineScrollHelper(
...
+            icon = CompoundIcons.ChevronUp(),
+            contentDescription = stringResource(id = CommonStrings.a11y_jump_to_unread_messages),
+            modifier = Modifier.padding(bottom = 12.dp),
+            isVisible = isJumpToUnreadVisible,
+            hasUnread = true,
+            onClick = ::jumpToReadMarker,
+            onMarkAsRead = onMarkAllAsRead,
+            testTag = TestTags.jumpToUnreadButton,
+        )
+        Box(modifier = Modifier.size(36.dp)) {
```
bmarty, line 453: What is the goal of this Box? Can it be removed?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -358,47 +415,158 @@ private fun BoxScope.TimelineScrollHelper(
...
+                            ) + fadeIn(animationSpec = tween(180)),
+                            exit = scaleOut(
+                                animationSpec = tween(140),
+                                targetScale = 0.9f,
+                                transformOrigin = transformOrigin,
+                            ) + fadeOut(animationSpec = tween(140)),
+                        ) {
+                            // Hand-rolled instead of DropdownMenuItem: padding here is tighter
+                            // than DropdownMenuItem's 12.dp default to match the Figma spec.
+                            Row(
```
bmarty, line 536: Maybe extract the Popup container to the design system module?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -433,3 +601,80 @@ internal fun TimelineViewPreview(
...
+            forceJumpToReadMarkerVisibility = true,
+        )
+    }
+}
+
+@PreviewsDayNight
+@Composable
+internal fun TimelineViewWithReadMarkerNoIndicatorsPreview() = ElementPreview {
+    TimelineViewWithReadMarker(hasUnreadAbove = false, hasUnreadBelow = false)
+}
```
bmarty, line 663: I am confused because the preview does have indicators. Is there an issue?

Image

bmarty (commented):
Thanks!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
jennaharris7: You are absolutely correct - the preview is incorrectly named, the jump to unreads fab would always have an indicator - I've renamed to reflect reality

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
jennaharris7: This is needed to keep the jump to unread button in the same spot in the ui regardless of the jump to bottom fab visibility (matching figma comment) - I've added a code comment to explain

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
jennaharris7: Before I extract — would you be thinking specifically of the styled menu surface (the shadow + rounded + bgCanvasDefaultLevel1 + borderDisabled Row, with padding) as a new component in libraries/designsystem? The Popup itself plus CenterStartOfAnchorPositionProvider and the right-center TransformOrigin are pretty FAB-shaped, so pulling those out would either bake the FAB anchor into the design system or expose enough knobs that callers re-implement positioning. The surface is easily reusable though and would be straightforward to extract. Which do you prefer?

jennaharris7:
The update here, together with the sdk change here: https://github.com/matrix-org/matrix-rust-sdk/pull/6569 , closes the jump-to-unread edge case:
                                                                                                                                                                        
  - Out-of-window marker — when the m.fully_read event is older than the loaded timeline (gaps, focused-event entry, long pagination), the chevron-up FAB now appears and routes the tap through the focused-event path so you actually land on the marker. Previously it stayed silently hidden.                                           
  - Also added eager hide on mark-as-read — long-pressing → Mark as read now drops the FAB the instant the SDK call returns, rather than waiting for the next RoomInfo push to catch up.                                                                                                                                                             
                  
This brings the Android behavior in line with https://github.com/element-hq/element-x-ios/pull/5506.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
bmarty: OK, if this is only used here for now, so the code can stay like that. We can extract later, with parameters depending on the needs.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
bmarty: Thanks, I did not realized that `JumpToPositionButton` can draw nothing.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
```diff
@@ -358,47 +431,160 @@ private fun BoxScope.TimelineScrollHelper(
...
         }
     }
 }
 
+@Composable
+private fun TimelineUnreadIndicator(
+    isVisible: Boolean,
+    modifier: Modifier = Modifier,
+) {
+    if (!isVisible) return
```
bmarty: I'd remove the `isVisible` parameter and change the call site to

```kotlin
if (hasUnread) {
    TimelineUnreadIndicator(...)
}
```

Also maybe we could use the existing `UnreadIndicatorAtom` ? Only the border seems to be different. If this is not easy to do, let's keep this code.

libraries/ui-strings/src/main/res/values/temporary.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?><!--
+  ~ Copyright (c) 2026 Element Creations Ltd.
+  ~
+  ~ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+  ~ Please see LICENSE files in the repository root for full details.
+  -->
+
+<resources>
+    <string name="action_mark_as_read">"Mark as read"</string>
+    <string name="a11y_jump_to_unread_messages">"Jump to unread messages"</string>
```
bmarty: Maybe change to
```suggestion
    <string name="a11y_jump_to_unread_messages">"Jump to first unread message"</string>
```

bmarty (commented):
Thanks, can you also check why the tests are not passing please?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt
jennaharris7: Fixed in commit #40f8b48256

libraries/ui-strings/src/main/res/values/temporary.xml
jennaharris7: Fixed in commit 2c2d803e58

bmarty (approved):
Thanks for the update. I will merge on my branch so that I can record the screenshots

PR #6698, opened 2026-04-30
Remove SignInWithClassic FeatureFlag to enable the feature.

### LINKED ISSUE #6669: [Task] Enable `SignInWithClassic` by default
once Element Classic is released with https://github.com/element-hq/element-android/pull/9136 and https://github.com/element-hq/element-android/pull/9137.

The feature flag can probably be entirely removed.

bmarty:
Closes #6669

Draft, as we do not want to merge it now. But it can be reviewed, so we are ready to merge whenever we want.


bmarty:
Note: we will include this on EXA 26.05.1

ganfra (approved):
Why not marking as isFinished first?

bmarty:
> Why not marking as isFinished first?

I guess we could do that, but if we want to revert, we will have to release again and we can just revert the PR.

PR #6702, opened 2026-04-30
Notification settings: pick custom sounds for messages and call ringtones

### LINKED ISSUE #6672: Add in-app settings link to system picker for app-wide message sound and call ringtone
### What

Introduce two app-wide settings under **Settings → Notifications**:

1. **Message sound** — opens the system ringtone picker, stores the selected URI, and applies it to `NOISY_NOTIFICATION_CHANNEL_ID`.
2. **Call ringtone** — opens the system ringtone picker, stores the selected URI, and applies it to `RINGING_CALL_NOTIFICATION_CHANNEL_ID`.

### Why

- Lets users distinguish Element X notifications from other apps without changing OS-wide defaults.
- Lets users separate message tone from call ringtone. Today the ringing-call channel inherits `Settings.System.DEFAULT_RINGTONE_URI` (see `libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/channels/NotificationChannels.kt` around line 177) and the noisy-message channel is hardcoded to `R.raw.message` (around lines 121–131), so users have no in-app way to differentiate the two.
- Discoverability — many users don't realize Android channel sound is configurable per-channel via system settings, so an in-app entry point matches user expectations.

### How

- Persist the picked URIs in session preferences (separate keys for the message sound and the call ringtone).
- Use a channel-version counter to force `NotificationChannel` recreation when the URI changes (Android only reads `setSound(...)` at channel-create time).
- Note: this re-introduces dynamic channel IDs that were removed in the noisy-channel cleanup at `NotificationChannels.kt` lines 91–101.

Yes

jennaharris7:
## Content

Notification sound picking moves in-app for both **messages** and **incoming calls**.

For message notifications, tapping the row opens a dialog with four options:
- **Element default** — Element X bundled sound
- **Element fade** — Element X bundled sound
- **System default** — Android's notification tone
- **Choose another sound…** — opens the system ringtone picker

The current bundled "Element default" is always reachable through the dialog, so users can return to it after experimenting with custom tones. Silent is reachable inside the system picker.

_Open to feedback on the dialog's option set, labels, and presentation — happy to iterate on the UX based on review._

For incoming call ringtones, tapping the row opens the system picker directly — there's no bundled call ringtone to surface as a preset.

Custom picks are copied into app-private storage and re-exposed through the app's FileProvider (`DefaultNotificationSoundCopier`), so a chosen sound can't disappear when the source app is uninstalled or the URI grant lapses.

Users on prior builds who never picked a sound keep hearing the bundled tone — a one-shot migration on first launch flips their stored state to Element default and bumps the channel version. Gated on `version == 0`, so it runs at most once.

### Channel recreation

Android forbids mutating an existing channel's sound, so each pick increments a per-slot version counter and creates a new channel id (`{base}_v{n}`); the previous versioned channel is deleted. Persistence and the version bump happen in a single transaction.

### Display name

The row shows the friendly tone name, falling back through `Ringtone.getTitle()` → `OpenableColumns.DISPLAY_NAME` of the source URI → localised "Custom". The live resolver skips our own FileProvider URIs, where `getTitle()` would otherwise surface our internal filename (`call_sound.ogg`).

### Failure path

If the copy fails, an inline dismissible error appears below the row and the persisted state is left untouched.

## Motivation and context

Closes #6672.

Today, changing the notification sound requires leaving Element X to dig through Android's per-channel system settings, which most users never discover. Bringing the picker in-app makes the setting discoverable, gives messages and calls independent controls, and lets users return to the bundled Element default sound at any time.

## Screenshots / GIFs

**Subject to design review
screenshot1screenshot2
screenshot3screenshot4
screenshot5screenshot6

## Tests

- Open **Settings → Notifications**. The **Sounds** section shows two rows: **Message sound** (defaults to "Element default") and **Call ringtone** (defaults to "System default").
- Tap **Message sound** → dialog with three options. Pick **Choose another sound…** → system picker opens. Pick a tone → row reads the tone's friendly name; sending a message plays it.
- Tap **Message sound** → **Element default** → row reads "Element default"; sending a message plays the bundled `message.mp3`. 
- Tap **Message sound** → **Element fade** → row reads "Element fade"; sending a message plays the bundled `element_fade.mp3`. 
- Tap **Message sound** → **System default** → row reads "System default"; sending a message plays Android's notification tone.
- Reach **Silent** via "Choose another sound…" → row reads "Silent"; messages produce no sound.
- Tap **Call ringtone** → system picker opens directly. Verify Default / Silent / Custom all behave the same way; a placed call rings with the chosen tone.
- Force-stop and re-open → all picks persist.
- Upgrade-from-prior-build: install the previous build with the message sound left at its default, install this build over the top → row silently switches to "Element default"; the channel still plays `message.mp3`.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Samsung Galaxy S24, Android 16

## Checklist

    - [ ] No.


bmarty:
Thanks! A question before I review the PR: would it be possible, when the user selects a sound, to copy the selected file in the private application storage, so that the sound file can never be deleted / lost / updated? All the resilience part of this PR could then be removed, that would simplify the PR a lot. Also it will prevent the file to be updated (by a third party app), and so the sound to change without Element app to be aware of, even with the resilience system.

jennaharris7:
We can absolutely do that - how do we want to handle the current message.mp3 in the app?  This will be the starting point for users, do we want to show the label for this to be "Default", or "Custom" (and leave the default label for the true system default?) I've code the code changes in place (barring a change to the message.mp3 label), I just want to run it through some testing before I push

bmarty:
> how do we want to handle the current message.mp3 in the app? This will be the starting point for users, do we want to show the label for this to be "Default", or "Custom" (and leave the default label for the true system default?)

Maybe something like:
- "Default" for the system default
- "Element" or "Element's sound" for the Element default sound (could be actually the app name "$AppName's sound")

@mxandreas any thoughts?

mxandreas:
I have taken a quick look at it, and please see also my comment on the iOS PR for more context: https://github.com/element-hq/element-x-ios/pull/5527#issuecomment-4387283645

1) It looks like we're in agreement to use Element Default for the default tone.
2) iOS also offers the alternative Element tone, I think Android should have it available as well, and I proposed to rename it Element Fade (open for better names).
3) iOS does not currently let you to customize ringtone for calls - does anyone know if that is something that is platform-specific? E.g. not something that is possible or good practice to do on iOS?
4) Regarding the copy, I'd make both platforms to either use "sound" or "tone" - both are used, there seem to be no clear winner here but perhaps sound is a bit more modern and use.

mredig:
> 4. Regarding the copy, I'd make both platforms to either use "sound" or "tone" - both are used, there seem to be no clear winner here but perhaps sound is a bit more modern and use.

iOS PR author here - I used 'tone' because it's the terminology used on iOS. 

Screenshot 2026-05-06 at 9 49 07 AM

I don't know what Android uses, but I would suggest remaining consistent with the platform.

jennaharris7:
> I have taken a quick look at it, and please see also my comment on the iOS PR for more context: [element-hq/element-x-ios#5527 (comment)](https://github.com/element-hq/element-x-ios/pull/5527#issuecomment-4387283645)
> 
> 1. It looks like we're in agreement to use Element Default for the default tone.
> 2. iOS also offers the alternative Element tone, I think Android should have it available as well, and I proposed to rename it Element Fade (open for better names).
> 3. iOS does not currently let you to customize ringtone for calls - does anyone know if that is something that is platform-specific? E.g. not something that is possible or good practice to do on iOS?
> 4. Regarding the copy, I'd make both platforms to either use "sound" or "tone" - both are used, there seem to be no clear winner here but perhaps sound is a bit more modern and use.

I've gone ahead and added the Element fade tone in the selection picker - subject to name or ui changes  (Android does use the term "Sound", which differs from iOS)

features/announcement/impl/src/main/kotlin/io/element/android/features/announcement/impl/DefaultAnnouncementService.kt
```diff
@@ -41,16 +40,14 @@ class DefaultAnnouncementService(
...
-        ) { _, newNotificationSoundStatus ->
-            buildList {
-                if (newNotificationSoundStatus == AnnouncementStatus.Show) {
-                    add(Announcement.NewNotificationSound)
+        return announcementStore.announcementStatusFlow(Announcement.NewNotificationSound)
+            .map { newNotificationSoundStatus ->
+                buildList {
+                    if (newNotificationSoundStatus == AnnouncementStatus.Show) {
+                        add(Announcement.NewNotificationSound)
+                    }
```
bmarty: This change is correct, but the current version will have to be restored if we have several `StatusFlow` to observe in the store. That's why we kept the `combine` and used a `flowOf(Unit)`, because `combine` require at least 2 flows (ofc).

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSoundPicker.kt
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright (c) 2025 Element Creations Ltd.
+ * Copyright 2025 New Vector Ltd.
```
bmarty: New files should only have the current year copyright to only one line with `Copyright (c) 2026 Element Creations Ltd.`

Can you check the other new files please?

features/preferences/impl/src/main/res/values/temporary.xml
```diff
@@ -0,0 +1,23 @@
...
+    <string name="screen_notification_settings_call_ringtone_label">Call ringtone</string>
+    <string name="screen_notification_settings_sound_element_default">Element default</string>
+    <string name="screen_notification_settings_sound_element_fade">Element fade</string>
+    <string name="screen_notification_settings_sound_system_default">System default</string>
+    <string name="screen_notification_settings_sound_silent">Silent</string>
+    <string name="screen_notification_settings_sound_custom_fallback">Custom</string>
+    <string name="screen_notification_settings_sound_copy_error_message">Couldn\'t use this sound. Please try another.</string>
+    <string name="screen_notification_settings_sound_copy_error_dismiss_a11y">Dismiss sound copy error</string>
+    <string name="screen_notification_settings_message_sound_dialog_title">Message sound</string>
+    <string name="screen_notification_settings_message_sound_dialog_choose_other">Choose another sound…</string>
```
bmarty, line 20: For the diff of terminology between iOS and Android, we have a convention to use suffixes key `_ios` or `_android`.

For instance in Localazy we will have two entries with:

|key|value|
|-|-|
|`screen_notification_settings_message_sound_dialog_choose_other_ios`|Choose another tone…|
|`screen_notification_settings_message_sound_dialog_choose_other_android`|Choose another sound…|

Then iOS (resp. Android) importer will ignore keys with suffix `_android` (resp. `_ios`)

CC @mredig

bmarty (commented):
Thanks!
A few remarks, but nothing blocking.

For the UX, I found a bit weird to not have the same behavior for the 2 items in the settings:

Image

Clicking on `Message sound` first open this dialog:

Image

But clicking on `Call ringtone` directly open the system sound picker. Maybe it could open the same dialog with just 2 options: `System default` and `Choose another sound...`.

please wait for @mxandreas to share his POV on this before doing any change.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/NotificationSoundPicker.kt
jennaharris7: Fixed in commit f25e3862a8

features/preferences/impl/src/main/res/values/temporary.xml
jennaharris7: I've updated the screen_notification_settings_message_sound_dialog_choose_other_android - should we also discuss aligning the other string choices for this?  Here is a current breakdown:

  Android key: screen_notification_settings_sounds_section_title                                                                                                        
  Android value: "Sounds"                                                                                                                                            
  iOS counterpart: screen_notification_settings_configuration_message_sound_section_title                                                                               
  iOS value: "Sound"                                                                                                                                                 
  ────────────────────────────────────────                                                                                                                              
  Android key: screen_notification_settings_sound_element_default                                                                                                    
  Android value: "Element default"                                                                                                                                      
  iOS counterpart: message_sound_elementx_default                                                                                                                    
  iOS value: "Element Default"                                                                                                                                          
  ────────────────────────────────────────                                                                                                                              
  Android key: screen_notification_settings_sound_element_fade                                                                                                       
  Android value: "Element fade"                                                                                                                                         
  iOS counterpart: message_sound_elementx_pro_default                                                                                                                
  iOS value: "Element Fade"                                                                                                                                             
  ────────────────────────────────────────                                                                                                                           
  Android key: screen_notification_settings_sound_custom_fallback
  Android value: "Custom"                                                                                                                                               
  iOS counterpart: screen_notification_settings_configuration_message_sound_custom_sound_button_title
  iOS value: "Custom Sound..."                                                                                                                                          
  ────────────────────────────────────────              
  Android key: screen_notification_settings_message_sound_dialog_choose_other_android
  Android value: "Choose another sound…"
  iOS counterpart: (same as above)
  iOS value: "Custom Sound..."

features/announcement/impl/src/main/kotlin/io/element/android/features/announcement/impl/DefaultAnnouncementService.kt
jennaharris7: Fixed in commit f25e3862a8

features/preferences/impl/src/main/res/values/temporary.xml
mxandreas: > 've updated the screen_notification_settings_message_sound_dialog_choose_other_android - should we also discuss aligning the other string choices for this? Here is a current breakdown:

Thanks for the overview. Yes, we should most certainly align, whenever the same thing exists on both sides. Quickly looking through the keys and English - I'd use the Android ones as reference, let me know if it makes sense.

mxandreas:
> please wait for @mxandreas to share his POV on this before doing any change.

I am guessing the diff between message and call ringtone behavior is that for the call ringtone there is no Element default and no Element fade, and so it does not make sense to open the intermediate modal, as there is nothing to select from there? @jennaharris7 please confirm.

jennaharris7:
> > please wait for @mxandreas to share his POV on this before doing any change.
> 
> I am guessing the diff between message and call ringtone behavior is that for the call ringtone there is no Element default and no Element fade, and so it does not make sense to open the intermediate modal, as there is nothing to select from there? @jennaharris7 please confirm.

That is correct, there are no element specific defaults for ringtone - previous to this pr we were just using the system default ringtone - I can add the modal as suggested with System default as the only option (aside from select other) or leave out whichever you guys decide

features/preferences/impl/src/main/res/values/temporary.xml
mredig: Jenna and I can sync up on key naming. As for terminology, we've already discussed. However, I'd like to push back on capitalization and the ellipsis. "Element Default" and "Element Fade" aren't just UI labels -- they're the actual names of the sounds, where title case is unambiguous. "Custom Sound..." is admittedly just a button label, and while title case, having had roots in the HIG, still keeps more in line with Apple platforms, I'll defer to whatever the project's convention is for consistency. The ellipsis though follows a long-standing Apple convention signaling that more steps follow -- these are the kinds of small details iOS users notice, even if they can't articulate why.

features/announcement/impl/src/main/kotlin/io/element/android/features/announcement/impl/DefaultAnnouncementService.kt
bmarty: it's https://github.com/element-hq/element-x-android/pull/6702/commits/8f6a9aead5492dd69b8961f849a1de3229fc72d2 but OK, thanks!

bmarty:
> > > please wait for @mxandreas to share his POV on this before doing any change.
> > 
> > 
> > I am guessing the diff between message and call ringtone behavior is that for the call ringtone there is no Element default and no Element fade, and so it does not make sense to open the intermediate modal, as there is nothing to select from there? @jennaharris7 please confirm.
> 
> That is correct, there are no element specific defaults for ringtone - previous to this pr we were just using the system default ringtone - I can add the modal as suggested with System default as the only option (aside from select other) or leave out whichever you guys decide

The problem with the current behavior is that the user can be lost if they first select another sound, and then want to restore the default sound. They will have to manually select the default sound, and it will not be handled by the application as the default sound. So having a dialog with the 2 options `System default` and `Choose another sound...` could help.

features/preferences/impl/src/main/res/values/temporary.xml
mxandreas: Ok, I leave it for you to figure it out - it is fine as long as it respects the existing patterns in place. I do not think we have many differences between iOS and Android when it comes to copy so please double check all such cases - there must be a very good reason for that.

FWIW - I agree that Element Default and Fade could rather be title case as they are title of the sound. cc @americanrefugee

mxandreas:
> The problem with the current behavior is that the user can be lost if they first select another sound, and then want to restore the default sound. They will have to manually select the default sound, and it will not be handled by the application as the default sound. So having a dialog with the 2 options `System default` and `Choose another sound...` could help.

Right, let's then have this intermediate modal with the 2 options you proposed.

features/preferences/impl/src/main/res/values/temporary.xml
mredig: An iOS, some of the sounds provided are system sounds, so they would be exclusive to iOS. Would you want their keys to have the `_ios` suffix?

mredig: (I'm clarifying compared to keys that would have equivalents on both platforms, but differ)

mredig: Here's where we landed:

- title case for the sound titles
- matched up the similar keys
- realized `screen_notification_settings_sound_custom_sound_button_title` (iOS) and `screen_notification_settings_message_sound_dialog_choose_other_android` are functionally different. On iOS, it allows the user to select their own sound. On Android, it opens the OS notification settings where customization can occur from the os's offerings. These aren't the same function, and therefore shouldn't have the same string or key.

jennaharris7:
> > The problem with the current behavior is that the user can be lost if they first select another sound, and then want to restore the default sound. They will have to manually select the default sound, and it will not be handled by the application as the default sound. So having a dialog with the 2 options `System default` and `Choose another sound...` could help.
> 
> Right, let's then have this intermediate modal with the 2 options you proposed.

A matching modal has been added for ringtone selection, and the strings have been adjusted to match iOS where applicable

bmarty (approved):
Thanks for the update!

bmarty:
@jennaharris7 can you fix the test so that we can merge the PR? Thanks!

bmarty:
Thanks. Merging to my branch so that I can record the screenshots.

PR #6724, opened 2026-05-05
Revert PR #6642

jmartinesp:
This reverts commit 6ba4679908ae1a3c7b0f698c10fda649aff9f8a5.

---

It seems like we didn't test this properly and there are some issues with back press and the currently embedded EC version.


features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenView.kt
```diff
@@ -256,18 +248,15 @@ private fun WebView.setup(
 
 private fun WebView.addBackHandler(onBackPressed: () -> Unit) {
     addJavascriptInterface(
-        JavascriptBackHandler {
-            onBackPressed()
+        object {
+            @Suppress("unused")
+            @JavascriptInterface
+            fun onBackPressed() = onBackPressed()
```
bmarty, line 254: Suspecting a stack overflow risk, but the lambda will be used...

bmarty (approved):
Thanks, both device back and EC pip buttons are working fine.

PR #6732, opened 2026-05-06
Fix back button sometimes not working after exiting a thread

### LINKED ISSUE #6689: Going back from thread breaks back action
When pressing the android navigation back button at the lower button of the screen or the back button in the upper-left corner provided by the app the next back navigations have no actions. 

## Steps to reproduce
1. Go into a chat with threads
2. Open any thread
3. Go back with android back button or app provided back button 
4. App switches to chat
5. Further back actions does not work

## Version
26.04.4

jmartinesp:
## Content

Make sure the `markingAsReadAndExiting` value is reset when the coroutine is either finished successfully or cancelled.

## Motivation and context

When I reproduced the issue, it looked like the coroutine used to mark as read and then exit the room was canceled, leaving the `markingAsReadAndExiting` variable with `true` value and preventing the exit block from running again.

Hopefully fixes https://github.com/element-hq/element-x-android/issues/6689

## Tests

Not easy to test, I just tried entering and leaving a thread several times with different gestures and timings and then trying to leave the room and it worked every time.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
@@ -272,6 +272,8 @@ class MessagesPresenter(
                             }
                         }
                         navigator.close()
+                    }.invokeOnCompletion {
+                        markingAsReadAndExiting.set(false)
                     }
                 }
             }


ganfra:
We had already discussed that, it's not just about thread, any other screen you open from the timeline and close quickly will make this happen. It's again a`BackHandler` issue!

jmartinesp:
Are you sure? I could definitely check with the debugger that it's this coroutine that's not being run again. Like, I got to the `if (!markingAsReadAndExiting.getAndSet(true))` line consistently, but then that was already `true` so the logic in the true case wasn't being executed.

So the back gesture/tap on the back icon was working, it's this presenter logic that got stuck. I'm not saying the back handler can't create similar symptoms in other screens though.

ganfra:
No I mean, the coroutine got cancelled because the `BackHandler` is called when you press back too soon

jmartinesp:
Ok, clarified in chat: using the back handler (gesture/icon) while already animating to any other screen from the chat one will trigger this, that's what @ganfra meant. The changes still solves the issue, the alternative would be disabling the back handler as soon as we start moving to a different screen.

bmarty (approved):
Thanks!

PR #6748, opened 2026-05-07
Adapt to new DM definition changes in the SDK

### LINKED ISSUE #4034: https://github.com/element-hq/element-x-android/blob/b2a9ebb2d25da51a13627cca06ef7e86a4697ab8/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/RoomIsDmCheck.kt#L18

### LINKED ISSUE #6717: [Task] Use `DmRoomDefinition` and `RoomInfo.isDm`

jmartinesp:
## Content

These changes try to centralise every check about whether a room is a DM or not using the new SDK APIs.

- Use `ClientBuilder.dmRoomDefinition(DmRoomDefinition.TwoPeople)`, which makes the SDK consider a DM a room with 1 direct target and at most two non-service members in it.
- Add `RoomMember.isServiceMember` to tell apart normal room member and service/functional ones.
- Use `Room.isDm` as a shortcut to `Room.roomInfoFlow.value.isDm`, remove all other `isDm` implementations so only the SDK one is valid.
- Replace `isDirect` checks with `isDm` where possible for `Room`, `SpaceRoom`, `RoomInfo`, etc.

## Motivation and context

Centralises how the DM checks are done.

Fixes https://github.com/element-hq/element-x-android/issues/6717.
Also fixes https://github.com/element-hq/element-x-android/issues/4034, since the previous issue is a task that implements the needed changes.

## Screenshots / GIFs

No screenshot changes, since just the internal checks changed.

## Tests

- Check a direct room with > 2 users *is not* a DM.
- Check a direct room with <= 2 non service users and the rest of them being service users *is* a DM.
- Check DMs don't display the avatar or display name of the service members (they're not part of the heroes).
- Notifications and incoming calls also tell apart DMs from non-DMs correctly based on the same logic.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.


frebib:
I swapped out my patches for this and it seems to work perfectly.
This PR fixes #4034

frebib:
One thing I've noticed that this doesn't fix is rooms in spaces with service members. It displays `$user and 2 others` where the two others are me and the bot user

jmartinesp:
`SpaceRoom`s (naming is hard) can't calculate their `isDm` value on demand and have to rely on a cached value that's updated when we load the room members with a `/members` request, which happens when you open the room or the room details.

Can you confirm if after opening the room and loading all the members you still don't see it as a DM? I may have to change at least how the FFI struct is created.

frebib:
~Opening the room doesn't change anything in the space room list (?). In the timeline/MessagesView, it the room name is correct. It's only the the `SpaceRoom` object where the room name is calculated incorrectly.~

Edit: I'm wrong, actually. Prior to opening the room I see the correct avatar, but the room's name is `$user and 2 others`, then after opening the room I see the wrong avatar (either that of the bridge bot, or the avatar of the space room if that's the default/fallback) and the room name changes to `$user, $bot`

>  you still don't see it as a DM

The issue isn't the fact that it's a DM/not a DM here. It's the fact that the name is generated from the heroes but incorrectly. I suppose it's only tangentially related to this PR in the fact that changes here could/should in theory affect the way that room name generation works. Does the Rust SDK not use the same "generate a room name from heroes logic" for `SpaceRoom` as it does for the regular `RoomInfo`?

jmartinesp:
@frebib that means we probably have to find a way to compute those on demand too, which is going to be difficult 🫤 .

That said, if you have the time, could you try changing this line:

https://github.com/matrix-org/matrix-rust-sdk/blob/65072dd8a82f378d835480a3683c4cea10371f96/crates/matrix-sdk-ui/src/spaces/room.rs#L148

So it's `heroes: Some(known_room.heroes().to_vec()),` instead? The difference is `RoomInfo::heroes` just returns whatever the server sent as a response, including the service members and `Room::heroes` filters out those. I didn't realise the wrong value was being returned, but let me know if that fixes your issue with avatars/names after loading the members.

Also, you'd have to change the `let display_name = matrix_sdk_base::Room::compute_display_name_with_fields(` call above that to use `known_room.heroes()` instead of `room_info.heroes()`.

frebib:
Tried that but it seems to have not helped. ~It did at least make the behaviour consistent between "not opening a room" and "opening a room" where in both cases it shows the same as the "not opening a room" before: the correct avatar and the room name being `$user, and 2 others`~

Edit: seems also not the case. I don't think it made any difference. Perhaps I did it wrong. Let me try a few more things to make sure my changes are actually making it into the build apk.

frebib:
Either I'm doing something fundamentally wrong or it doesn't work. Tried applying the changes as you suggest
```diff
diff --git crates/matrix-sdk-ui/src/spaces/room.rs crates/matrix-sdk-ui/src/spaces/room.rs
index 70826ac99..67476e3f7 100644
--- crates/matrix-sdk-ui/src/spaces/room.rs
+++ crates/matrix-sdk-ui/src/spaces/room.rs
@@ -118,7 +118,7 @@ impl SpaceRoom {
         let display_name = matrix_sdk_base::Room::compute_display_name_with_fields(
             name.clone(),
             room_info.canonical_alias(),
-            room_info.heroes().to_vec(),
+            known_room.heroes().to_vec(),
             known_room.joined_members_count(),
         )
         .to_string();
@@ -140,7 +140,7 @@ impl SpaceRoom {
             is_direct: Some(known_room.direct_targets_length() != 0),
             children_count,
             state: Some(known_room.state()),
-            heroes: Some(room_info.heroes().to_vec()),
+            heroes: Some(known_room.heroes().to_vec()),
             via: vec![],
             suggested: false,
         }
```
```
crates/matrix-sdk-ui/src/spaces/room.rs
64:    /// A list of room members considered to be heroes.
65:    pub heroes: Option<Vec<RoomHero>>,
87:            known_room.as_ref().map(|r| r.heroes().to_vec()).unwrap_or_default(),
107:            heroes: known_room.map(|r| r.heroes()),
121:            known_room.heroes().to_vec(),
143:            heroes: Some(known_room.heroes().to_vec()),

crates/matrix-sdk-ui/src/spaces/room_list.rs
1070:            heroes: None,
```
but the behaviour remains the same as before. I'll leave this here as to not pollute your PR any more. Thanks

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/notifications/edit/EditDefaultNotificationSettingPresenter.kt
```diff
@@ -154,9 +154,9 @@ class EditDefaultNotificationSettingPresenter(
     private fun CoroutineScope.setDefaultNotificationMode(mode: RoomNotificationMode, action: MutableState<AsyncAction<Unit>>) = launch {
         action.runUpdatingStateNoSuccess {
             // On modern clients, we don't have different settings for encrypted and non-encrypted rooms (Legacy clients did).
-            notificationSettingsService.setDefaultRoomNotificationMode(isEncrypted = true, mode = mode, isOneToOne = isOneToOne)
+            notificationSettingsService.setDefaultRoomNotificationMode(isEncrypted = true, mode = mode, isOneToOne = isDm)
```
bmarty: maybe rename the param `isOneToOne` too?

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/roomlist/RoomSummary.kt
```diff
@@ -21,5 +21,5 @@ data class RoomSummary(
         is LatestEventValue.Remote -> latestEvent.timestamp
         is LatestEventValue.RoomInvite -> latestEvent.timestamp
     }
-    val isOneToOne get() = info.activeMembersCount == 2L
+    val isDm get() = info.isDm
```
bmarty: Maybe we can remove the `get()` now?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomInfoExt.kt
```diff
@@ -18,7 +18,7 @@ import org.matrix.rustcomponents.sdk.RoomInfo
  */
 fun RoomInfo.elementHeroes(): List<MatrixUser> {
     return heroes
-        .takeIf { isDirect && activeMembersCount.toLong() == 2L }
+        .takeIf { isDm }
```
bmarty, line 21: We should probably filter out service members here, or the SDK is doing that for us (i.e. `heroes` does not contain service members)?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -169,6 +170,7 @@ class RustMatrixClientFactory(
             )
             .enableShareHistoryOnInvite(true)
             .threadsEnabled(featureFlagService.isFeatureEnabled(FeatureFlags.Threads), threadSubscriptions = false)
+            .dmRoomDefinition(DmRoomDefinition.TWO_MEMBERS)
```
bmarty, line 173: Does this change requests a clear of the cache?

bmarty (approved):
Thanks! I got only minor remarks.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomInfoExt.kt
frebib: https://github.com/matrix-org/matrix-rust-sdk/pull/6535/changes#diff-40ec73160d057ed5bbe28cf81af28f56e27394369779ead9676055b6206cc860R482
The SDK filters them out
(although now I look at this again, I notice that the above code is for `Room` and not `RoomInfo`)

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
jmartinesp: No, it's not really needed. This changes how the SDK treats the DMs on runtime, but it doesn't change anything in the DBs/caches.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomInfoExt.kt
jmartinesp: > (although now I look at this again, I notice that the above code is for Room and not RoomInfo)

The `heroes` in `ffi::RoomInfo` come from the `sdk::Room::heroes` function that already filters out service members.

PR #6752, opened 2026-05-07
Fix wakelock not being released when EC returns an error

bxdxnn:
## Content

Release wakelock when Element Call exits with an error

## Motivation and context

It wasn't released, and so I had the proximity sensor active even when I've closed Element Call.

## Tests

- Close the Element Call error dialog (e.g. on an unsupported server) then test if the screen doesn't turns off on the proximity sensor

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenEvent.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenEvent.kt
@@ -14,4 +14,5 @@ sealed interface CallScreenEvent {
     data object Hangup : CallScreenEvent
     data class SetupMessageChannels(val widgetMessageInterceptor: WidgetMessageInterceptor) : CallScreenEvent
     data class OnWebViewError(val description: String?) : CallScreenEvent
+    data object CallEnded : CallScreenEvent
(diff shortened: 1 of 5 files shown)


jmartinesp:
I just tested this branch and I don't think it's solving the issue. If I got it right, the issue seems to be `WebView.onCallStopped` is called and that should in turn launch a coroutine that releases the wakelock... but the provided coroutine scope is already cancelled at that point, so the code inside the `coroutineScope.launch { ... }` block never runs.

I *believe* since the WebView was destroyed, it would be OK to not use the `proximitySensorMutex` at all there, not needing to use the coroutine scope to launch a coroutine either and just calling:

```kotlin
if (proximitySensorWakeLock?.isHeld == true) {
    proximitySensorWakeLock?.release()
}
```

Directly inside the method.

---

Yep, that seemed to do the trick when I just tested it. I'll leave it up to you to update the PR, otherwise let me know and I can create a new one with the mentioned changes.

bxdxnn:
@jmartinesp thanks for investigating this, I would prefer for you to fix it properly, I'm not very familiar with the logic here.

PR #6757, opened 2026-05-08
Improve `FetchPushForegroundService`'s reliability

jmartinesp:
## Content

- Don't use DI, we can just create the notification channel. This should speed up the creation of the service and reduce the number of `ForegroundServiceDidNotStartInTimeException` received. Also use `MainScope` instead of the app's coroutine scope.
- Move the wakelock releasing mechanism to `onDestroy` so it's always used. Previously, this would only happen when `stopService` was called, which would only happen when `stopSelf()` is called, but not when the OS or the service manager stops the service.
 
## Motivation and context

Fix the `ForegroundServiceDidNotStartInTimeException` we're still seeing in Sentry.

## Tests

Just use notifications as usual.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] Yes. I asked Claude what could be causing the issues, which confirmed a suspicion of mine and also added a possible cause I hadn't identified.
    - [ ] No.


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/FetchPushForegroundService.kt
```diff
@@ -42,26 +45,33 @@ class FetchPushForegroundService : Service() {
...
 
     private var isOnForeground = false
 
-    override fun onCreate() {
-        Timber.d("Creating FetchPushForegroundService")
+    private fun ensureNotificationChannelExists() {
+        NotificationManagerCompat.from(this).createNotificationChannelsCompat(
+            listOf(
+                NotificationChannelCompat.Builder(CHANNEL_ID, NotificationManagerCompat.IMPORTANCE_LOW)
+                    .setName(getString(CommonStrings.common_android_fetching_notifications_title))
```
bmarty: Setting an empty name will crash the application, and this can be the case if the translation is empty.


```suggestion
                    .setName(getString(CommonStrings.common_android_fetching_notifications_title).ifEmpty { "Syncing notifications…" })
```

bmarty (approved):
Thanks!

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/push/FetchPushForegroundService.kt
bmarty: By the way the key `common_android_fetching_notifications_title` should be `common_fetching_notifications_title_android` to be ignored by the iOS string importer

PR #6765, opened 2026-05-11
Stop removing the `logs` dir when clearing cache

jmartinesp:
## Content

When we manually clear the cache from developer settings, don't remove the `logs` directory.

## Motivation and context

When we remove this directory, we also remove the current log file the SDK is using and usually this means it keeps trying to write new log lines to this deleted file while silently failing. So we end up with no old logs and no *current* logs until the log rotation logic creates a new log file.

To be honest, I can't remember if we removed this directory on purpose so we had a way to free some space taken by the logs or if it was a side-effect of it being in the cache directory.

It also matches the Element X iOS behavior.

## Tests

Clear the cache, check the log files are still there.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.


bmarty (approved):
OK, I would prefer the SDK issue to be fixed, by creating a new file if the current one get deleted, but I am not blocking this change.

Issue #6764, opened 2026-05-11
[Task] EX LLS: reply and thread


bmarty:
Please see and discuss the details in the meta issue.

manuroe:
Done in https://github.com/element-hq/element-x-android/pull/6767 and https://github.com/matrix-org/matrix-rust-sdk/pull/6563

PR #6767, opened 2026-05-12
Prevent user from starting LLS in thread

bmarty:
## Content

Prevent user from starting LLS in thread. We only allow sharing the current location in a thread. 

## Motivation and context

Part of #6764.

Other points of #6764 has to be handled by the SDK, this is done in https://github.com/matrix-org/matrix-rust-sdk/pull/6563

## Tests

- enable LLS flag
- open a room
- try to share location
- see that you can start a LLS
- open a thread
- try to share location
- see that you cannot start a LLS

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/location/impl/src/main/kotlin/io/element/android/features/location/impl/share/ShareLocationPresenter.kt
+++ b/features/location/impl/src/main/kotlin/io/element/android/features/location/impl/share/ShareLocationPresenter.kt
@@ -171,7 +171,7 @@ class ShareLocationPresenter(
             dialogState = dialogState,
             trackUserLocation = trackUserPosition,
             hasLocationPermission = permissionsState.isAnyGranted,
-            canShareLiveLocation = isLiveLocationSharingEnabled,
+            canShareLiveLocation = isLiveLocationSharingEnabled && timelineMode.canShareLiveLocation(),
             appName = appName,
             startLiveLocationAction = startLiveLocationAction.value,
             eventSink = ::handleEvent,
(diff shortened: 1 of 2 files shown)


features/location/impl/src/main/kotlin/io/element/android/features/location/impl/share/ShareLocationPresenter.kt
```diff
@@ -210,4 +210,9 @@ class ShareLocationPresenter(
     }
 }
 
+private fun Timeline.Mode.canShareLiveLocation() = when (this) {
+    is Timeline.Mode.Thread -> false
```
jmartinesp, line 214: Should this also exclude pinned events timeline? The current UI won't allow you to start a live location share in that screen, but it may be worth it to explicitly ensure that here.

jmartinesp (approved):
Thanks!

features/location/impl/src/main/kotlin/io/element/android/features/location/impl/share/ShareLocationPresenter.kt
bmarty: I thought about having all the case in the `when` statement, but I preferred to keep the code simple. Since there is no composer in Pinned event timeline, I guess it's fine. Maybe `Timeline.Mode` could have a read-only / writable concept.

PR #6772, opened 2026-05-12
Add missing strings `theme.black`

bmarty:
Add missing string in Localazy and rename other related to theme.


jmartinesp (approved):
Thanks!

PR #6774, opened 2026-05-12
Avoid SQLCipher key derivation

jmartinesp:
## Content

Use a random generated key instead of a passphrase with SQLCipher for new sessions.

This needs a series of DB re-key operations, performed as part of a 'dummy' DB migration. Note this is not reversible, and downgrading to a previous app version will result in a broken app.

## Motivation and context

This means ~600ms faster startup time on my test device. And since we aren't using a user-provided value as a passphrase, there is no real need for performing key derivation - even less so every time we want to open a DB instead of when we generate the key.

## Tests

1. In a previous version of the app, have a signed in user.
2. Install this version and start the app.
3. If everything works, it went well. The app boot time should also be faster.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.


jmartinesp:
I already asked about how this should be implemented, but I'm tagging @element-hq/security anyway to double check the implementation matches what they expect.

libraries/encrypted-db/src/main/kotlin/io/element/encrypteddb/passphrase/RandomSecretPassphraseProvider.kt
```diff
@@ -35,9 +36,15 @@ class RandomSecretPassphraseProvider(
...
+        return file.delete()
+    }
+
     private fun generateSecret(): ByteArray {
         val buffer = ByteArray(size = secretSize)
         SecureRandom().nextBytes(buffer)
-        return buffer
+        // Derive a key from the random secret to ensure it has the right properties for SQLCipher
+        // 256 bits (32 bytes) is the required length for raw SQLCipher keys
+        return buffer.doKeyDerivation(bitCount = 256)
```
ganfra: This class is used by other db than the CacheStore, is it ok? (just checking)

libraries/encrypted-db/src/main/kotlin/io/element/encrypteddb/SqlCipherDriverFactory.kt
```diff
@@ -23,16 +24,40 @@ import net.zetetic.database.sqlcipher.SupportOpenHelperFactory
...
-        val factory = SupportOpenHelperFactory(passphrase)
-        return AndroidSqliteDriver(schema = schema, context = context, name = name, factory = factory)
+        val factory = if (passphrase.size == 32) {
+            // Raw SQLCipher keys are 32 bytes long
+            val key = "x'${passphrase.toHexString()}'".toByteArray()
+            SupportOpenHelperFactory(key)
+        } else {
+            // Usual passphrase, which will be processed by SQLCipher's internal key derivation function
+            SupportOpenHelperFactory(passphrase)
+        }
```
ganfra: I'm not sure to follow this check? When are we going to have the 32 bytes long key? And there is no "constant value" we can check?

ganfra (commented):
Some questions

libraries/encrypted-db/src/main/kotlin/io/element/encrypteddb/passphrase/RandomSecretPassphraseProvider.kt
jmartinesp: Yes, it's supposed to be used by all client-side DBs using SQLCipher.

libraries/encrypted-db/src/main/kotlin/io/element/encrypteddb/SqlCipherDriverFactory.kt
jmartinesp: For the migration we need both the old key and a new one. The steps will be:

1. Open the DB with the old 'passphrase' (which is actually a random key of size != 32, in fact it should be 256 bytes IIRC), used as an actual passphrase as we did so far.
2. When migrating the DBs, reset the stored key and re-key the DB using this new key, which will have 32 bytes.
3. After that, every time we restore the session we use the 32 byte key as a raw key, providing it using the `x'<hex-string>'` format [SQLCipher expects](https://www.zetetic.net/sqlcipher/sqlcipher-api/#PRAGMA_key).

ganfra: There is no formatting method provided by SQLCipher for the key?

dkasak:
Some questions:

- The PR mentions that it switches to using a "random generated key" rather than a passphrase. What kind of passphrase were we using before? Where was it coming from?
- Internally, we discussed using a CSPRNG for the generation of the encryption key. There were some concerns with the currently used RNG. Can you remind me of that and tell me whether anything had changed in that respect? This PR doesn't appear to contain changes in this direction.
- Also internally I believe I advised using HKDF as the KDF rather than PBKDF2, which makes little sense when the supplied secret is not a "password" or "passphrase" but completely randomly generated. Yet, the PR seems to be using PBKDF2 again. Did we miscommunicate?

jmartinesp:
> The PR mentions that it switches to using a "random generated key" rather than a passphrase. What kind of passphrase were we using before? Where was it coming from?

We were also using an auto-generated key using [`SecureRandom`](https://developer.android.com/reference/java/security/SecureRandom), but providing it to SQLCipher as a passphrase, which triggered the key derivation.

> Internally, we discussed using a CSPRNG for the generation of the encryption key. There were some concerns with the currently used RNG. Can you remind me of that and tell me whether anything had changed in that respect? This PR doesn't appear to contain changes in this direction.

`SecureRandom()`, instantiated without an algorithm, will use the best fit the device is capable of. In my test and personal devices, this meant `SHA1PRNG`, but in other devices it could use `NativePRNG` or `DBRG`. AFAICT we don't have any other available, unless we implement something maybe in the Rust SDK or include a separate library for this.

> Also internally I believe I advised using HKDF as the KDF rather than PBKDF2, which makes little sense when the supplied secret is not a "password" or "passphrase" but completely randomly generated. Yet, the PR seems to be using PBKDF2 again. Did we miscommunicate?

Ah, sorry, I misunderstood that! Let me check if I can use `HKDF` instead.

libraries/encrypted-db/src/main/kotlin/io/element/encrypteddb/SqlCipherDriverFactory.kt
jmartinesp: No mentions of blob formatting or `x'` in https://github.com/sqlcipher/sqlcipher-android AFAICT 🫤 .

jmartinesp:
@dkasak [these](https://developer.android.com/reference/javax/crypto/SecretKeyFactory) are the available algorithms for `SecretKeyFactory` on Android, would any of those work? Alternatively, I think Tink's [Hkdf](https://javadoc.io/doc/com.google.crypto.tink/tink/1.2.1/com/google/crypto/tink/subtle/Hkdf.html) class could be used instead.

dkasak:
First of all, to resolve the `SHA1PRNG` mystery. [Once upon a time](https://android-developers.googleblog.com/2016/06/security-crypto-provider-deprecated-in.html), this was the only algorithm provided by the old `Crypto` provider and was not a CSPRNG. The `Crypto` provider was deprecated in Android N (API 24) and removed in Android O (API 26), but given many codebases assumed the existence of `SHA1PRNG` and asked for it explicitly, this name was retained in the modern implementation even though asking for it *will no longer get you an actual `SHA1PRNG` implementation*. The modern Conscrypt implementation [contains a comment](https://github.com/google/conscrypt/blob/ef2f4f0ec049fe2b4a97a0e683a88334341e5d6d/common/src/main/java/org/conscrypt/OpenSSLProvider.java#L411) explaining the same phenomenon.

The modern approach is not to specify the provider/algorithm but simply to construct `SecureRandom` using the default no-argument constructor, as we are doing. This should result in a CSPRNG backed by OpenSSL and `/dev/urandom`.

The modern approach is not to specify the provider or algorithm and to construct `SecureRandom` using the no-argument constructor. On Android API 26 and later this resolves to Conscrypt's `OpenSSLRandom`, which delegates to BoringSSL's CSPRNG (seeded from the Linux kernel via `getrandom(2)` / `/dev/urandom`).

We should be able to verify that this is so by running:

```kotlin
val sr = SecureRandom()
println("Algorithm: ${sr.algorithm}")      // SHA1PRNG
println("Provider:  ${sr.provider.name}")  // AndroidOpenSSL
println("Class:     ${sr.javaClass.name}")
```

Note: I haven't run this myself.

So just constructing a bare `SecureRandom()` and using `nextBytes()` is the way to go. Given that this is already a CSPRNG (see first sentence of <https://developer.android.com/reference/java/security/SecureRandom>), we don't need to do any special treatment of the resulting output, and therefore I would simply remove any KDF call for this purpose.

The scenario where we would still reach for a KDF would be one where we had some kind of root secret available and we wanted the secret we are discussing in this PR to be derived from the root secret, such that we don't have N separate secrets to manage but only a single root secret from which all the others are derived. I don't think we are in that situation, but you'll know better than me.

So, to recap:

- Just construct an argumentless `SecureRandom()` and generate the SQLCipher key from there directly
- Pass to SQLCipher using the `PRAGMA key = "x'...'";` format to avoid a key derivation there
- Avoid doing any kind of KDF on the Android side either
- Consider whether we have multiple secrets running around we could rationalise away by deriving them from a root secret

jmartinesp:
Just to confirm, I ran the code you mentioned:

```
val sr = SecureRandom()
println("Algorithm: ${sr.algorithm}")      // SHA1PRNG
println("Provider:  ${sr.provider.name}")  // AndroidOpenSSL
println("Class:     ${sr.javaClass.name}")
```

And the results were the expected ones:

```
Algorithm: SHA1PRNG
Provider:  AndroidOpenSSL
Class:     java.security.SecureRandom
```

jmartinesp:
I'm thinking about reusing the `ClientSecret` util added in https://github.com/element-hq/element-x-android/pull/6835, so maybe we should review and merge that one first.

ganfra (approved):
Thanks for the changes, makes sense!

PR #6775, opened 2026-05-12
Make timeline sender avatars and display names update reactively

### LINKED ISSUE #5176: Users don't see updated avatars on EX-EX, frequently
### Steps to reproduce

Customer reports:

If two users are using EXA, and one changes their avatar, that will not be seen by the other user.

Possibly refreshing the app or cache helps but I did not press them to test

Have we seen this before?

#### What did you expect?

Second user should see the update

### Your phone model

_No response_

### Operating system version

_No response_

25.08.2

### Homeserver

customer

### Will you send logs?

No

No

jennaharris7:
## Content

Let live `RoomMember` updates override the event-level sender snapshot in `TimelineItemEventFactory`, so timeline rows already in the diff cache reflect avatar / display name changes that arrived via sync. Banned members keep the existing hide-identity behavior.

## Motivation and context

Closes #5176. Customers report that when another user changes their avatar or display name, the update doesn't appear until the app is restarted or the cache is cleared.

The issue was tagged `X-Needs-Rust`, but no `matrix-rust-sdk` change turned out to be necessary: the SDK already emits profile changes as `ProfileChangeContent` timeline events. The only SDK-layer tweak is in `MatrixTimelineDiffProcessor`, which now treats `ProfileChangeContent` (alongside the already-handled `RoomMembershipContent`) as a trigger for `membershipChangeEventReceived` so subscribers refresh the cached members list.

## Open question
Do we want this behind a feature flag for users who may want to keep the behavior of seeing historical snapshots of timeline events?

## Tests

- New unit tests for `ProfileDetails.withLiveMemberOverride`: live member wins, null falls back, banned falls back, ambiguous name propagates, `Unavailable` snapshot still uses live data.
- Manual: change a peer's avatar from another client, confirm the timeline reflects it without restarting EXA.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Samsung Galaxy S24, Android 16

## Checklist

- [ ] Accessibility has been taken into account.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/TimelineItemsFactory.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/TimelineItemsFactory.kt
@@ -19,6 +19,7 @@ import io.element.android.features.messages.impl.timeline.model.TimelineItem
 import io.element.android.libraries.androidutils.diff.DiffCacheUpdater
 import io.element.android.libraries.androidutils.diff.MutableListDiffCache
 import io.element.android.libraries.core.coroutine.CoroutineDispatchers
+import io.element.android.libraries.matrix.api.core.UserId
 import io.element.android.libraries.matrix.api.room.RoomMember
 import io.element.android.libraries.matrix.api.timeline.MatrixTimelineItem
 import kotlinx.collections.immutable.ImmutableList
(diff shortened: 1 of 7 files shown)


jmartinesp:
Sorry for the long review time, the code looks good to me overall. About this question:

> Do we want this behind a feature flag for users who may want to keep the behavior of seeing historical snapshots of timeline events?

I'd say we do, I think when we implemented this in the classic apps we did so behind a developer option/advanced settings toggle. @mxandreas WDYT?

jmartinesp:
Actually, it seems like this should be possible with a pure SDK implementation: https://github.com/matrix-org/matrix-rust-sdk/issues/6575

I'm looking at ways to implement this, it seems like following the example from `AmbiguityMap` would do the trick.

jmartinesp:
Since https://github.com/matrix-org/matrix-rust-sdk/pull/6576 was merged, this should now be fixed in the SDK and we can probably close this PR. Thanks again for the client-side fix, but it's better to have it centralised in the SDK.

PR #6006, opened 2026-01-14
When a duplicate room list entry is found, report it and remove it

jmartinesp:
## Content

- When a duplicate room list entry is found, instead of just adding some logs, report them to Sentry if the user has enabled analytics, with some extra context info.
- Besides reporting them, remove the duplicate from the room list entries to display in either the main room list or a space.
- Modify tests so they still work after the remove duplicate post-processing.

## Motivation and context

Avoid crashes related to duplicate room id keys being used and try to figure out where this issue is coming from: is it us or the SDK?

Related rageshake: https://github.com/element-hq/element-x-android-rageshakes/issues/9275

## Tests

Unless you modify the code to force a duplicate room list entry it's quite difficult to test 😓 .

An example Sentry report: https://sentry.tools.element.io/organizations/element/issues/11222277/events/ce9d71ad9d7e4d548f8e354c00955f21/

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly define what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

Issue #4182, opened 2025-01-22
Duplicated room around midnight


bmarty:
### Steps to reproduce

- First reported internally, creating an issue to track the problem.
- Sometimes I can observe that a room is duplicated in the room list.
- It seems to happen when messages are sent after midnight.
- Maybe due to the computation of dates that are triggered at midnight to update relative dates rendering?
- Rust team claims that this cannot occur, so the bug may be app side.

To reproduce the issue (steps to be confirmed)

1. let the app display the room list
2. from another session, and just after midnight, send 2 messages.

Note: I tried to write unit test the repro the issue but without success

## Phone screenshot

<img width="409" alt="Image" src="https://github.com/user-attachments/assets/fe0fa476-df85-4869-a2a2-c461413d64fa" />

## Web screenshot

<img width="869" alt="Image" src="https://github.com/user-attachments/assets/d8855e18-74f3-4fa7-aca1-9f15ffcd7416" />

#### What did you expect?

No room duplication

#### What happened instead?

Duplicated room

Sending a new message do not fix the issue:

<img width="402" alt="Image" src="https://github.com/user-attachments/assets/62ed05e5-0396-4939-8037-04349c5fe974" />

Restarting the app fixes the issue

<img width="398" alt="Image" src="https://github.com/user-attachments/assets/14f9379f-6790-49aa-b3b9-c0376dc979d3" />

### Your phone model

na

### Operating system version

na

### Application version and app store

latest develop

### Homeserver

na

### Will you send logs?

Yes

### Are you willing to provide a PR?

Yes

xundeenergie:
Yes, can confirm.

manuroe:
@90vgt: thanks for the report. Did that "Arcadia ..." room get upgraded?

We are working on room upgrade support in EX right now. It will be available in few weekls.

90vgt:
@manuroe I've checked the room and it's on version 9. Does upgrading the room resolve this issue? 
That sounds great. The issue with room duplication is very minor and rare, so I'll wait until room upgrading in EX is ready. 
Thanks for the excellent work!

PR #6791, opened 2026-05-14
Fix room list duplicate-detection telemetry crashing before it can report

jennaharris7:
## Content
Capture `updates.description()` at the top of `RoomSummaryListProcessor.updateRoomSummaries` instead of inside the dedupe branch, so the trackError payload is built while the Rust `Room` objects are still valid. Also adds regression tests covering both the UI and SDK dedupe safety nets.

## Motivation and context
The room list has had a duplicate-detection safety net since #5364 and #6006, with `analyticsService.trackError` calls meant to surface offending updates to Sentry so the team can root-cause #4182 and #5031. The reporting path is broken: `applyUpdate` consumes each `Room` via `entry.use { ... }` (which destroys it), and `describe()` reads from the same `Room`. By the time the dedupe branch builds the trackError message, every `Room` payload is destroyed, so the call throws and the trackError never reaches Sentry — silently masking exactly the diagnostic the team needs.

Switched describe() from Room.displayName() to Room.id(). The roomId is the dedupe key, so it's the more useful identifier in the diagnostic message; id() is also a cheaper accessor than displayName() (which can compute names from membership for DMs), avoiding the per-update FFI cost now that describe() runs on every update;  and it keeps room names — which can be user-set — out of analytics payloads.    

**This is not a fix for the duplicate-room bug itself.** Both #4182 and #5031 remain open. This PR makes the telemetry actually fire so we can start collecting the data we need to diagnose them.

## Tests
- New regression test `Insert that triggers dedupe is reported via trackError without crashing` proves the trackError fires end-to-end with a meaningful description.
- New regression tests for the existing dedupe safety nets at both the SDK processor layer and the UI mapper layer (`RoomSummaryListProcessorTest`, `RoomListDataSourceTest`).

## Tested devices
- [x] Unit tests pass locally (`./gradlew :libraries:matrix:impl:testDebugUnitTest :features:home:impl:testDebugUnitTest`)

### DIFF
--- a/features/home/impl/src/test/kotlin/io/element/android/features/home/impl/datasource/RoomListDataSourceTest.kt
+++ b/features/home/impl/src/test/kotlin/io/element/android/features/home/impl/datasource/RoomListDataSourceTest.kt
@@ -14,6 +14,8 @@ import io.element.android.features.home.impl.FakeDateTimeObserver
 import io.element.android.libraries.androidutils.system.DateTimeObserver
 import io.element.android.libraries.dateformatter.test.FakeDateFormatter
 import io.element.android.libraries.matrix.api.roomlist.RoomListService
+import io.element.android.libraries.matrix.test.A_ROOM_ID
+import io.element.android.libraries.matrix.test.A_ROOM_ID_2
 import io.element.android.libraries.matrix.test.notificationsettings.FakeNotificationSettingsService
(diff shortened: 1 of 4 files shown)


jmartinesp (approved):
Thanks! Having this back would be quite helpful to understand if a fix for the root issue works.

PR #6793, opened 2026-05-15
Attempt to fix room list item duplicates at midnight

### LINKED ISSUE #4182: Duplicated room around midnight

jmartinesp:
## Content

This seems to happen because of a race condition between `RoomListDataSource.observeDateTimeChanges` and `RoomListDataSource.replaceWith` being called at almost the same time and the first one using the newly received items from observing the timeline items but not updating the cache which will be later reused by `replaceWith`, containing incorrect indices.

Updating `diffCacheUpdater` inside `rebuildAllRoomSummaries` should fix this. The changes introduced in https://github.com/element-hq/element-x-android/pull/6791 should also make sure the issue is gone.

## Motivation and context

*Should* fix https://github.com/element-hq/element-x-android/issues/4182 and https://github.com/element-hq/element-x-android/issues/5031.

## Tests

Remove the line added in this PR and run the regression test. It should fail, proving the issue.

## Checklist

    - [x] Yes (used Claude to confirm where this issue was coming from). In this case, please request a review by Copilot.
    - [ ] No.

### DIFF
--- a/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/datasource/RoomListDataSource.kt
+++ b/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/datasource/RoomListDataSource.kt
@@ -83,8 +83,8 @@ class RoomListDataSource(
 
     val loadingState = roomList.loadingState
 
-    fun launchIn(coroutineScope: CoroutineScope) {
-        roomList
+    fun launchIn(coroutineScope: CoroutineScope): Job {
+        return roomList
             .summaries
             .onEach { roomSummaries ->
                 replaceWith(roomSummaries)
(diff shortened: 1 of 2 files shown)


ganfra (approved):
Seems fine!

PR #6715, opened 2026-05-03
Fix media viewer flickering and crashing

bxdxnn:
## Content

1. Reverts `6a4fed2`
2. Reimplements it in a much simpler way (just adds `reverseLayout = true`)

## Motivation and context

Fixes #6695
Fixes #6709

I haven't tried to fix the previous (pre-`6a4fed2`) media viewer header flicker, but at least this fixes the regressions introduced by `6a4fed2`.

## Tests

- Open a media

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

- This PR was made with the help of AI:
    - [x] Yes. In this case, please request a review by Copilot.
    - [ ] No.
- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly defines what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

Issue #6695, opened 2026-04-30
26.04.4 was installed by Google Play although pre-release Media Viewer "jumping"


chrissi55:
### Steps to reproduce

1. Where are you starting? What can you see?
2. What do you click?
3. More steps…

1. I send a photo via Element X (made with photo app or element doesn't matter)
2. I see the photo in the chat and open it
3. I see the header of the photo with description "pumping and springing"
4. another receipient opened the photo on his / her device and see the same 
5. more another person see the photo switching to another (older photo from the same chat)

Seems as if the media viewer in 26.04.4 has a severe bug.
Not happening on iOS or Element Web (Desktop)

The Media viewer is no longer usable that way - because photos doesn't stay calm so buttons for details or download media disappear when photo switches or closes on itself.

I wouldn't have installed 26.04.4 version if google haven't done so. 

#### What did you expect?
That media viewer shows exactly one photo without jumpinmg or switching in media viewer

#### What happened instead?
media switches unmotivated by its own or crash / closes media viewer

### Your phone model

Pixel

### Operating system version

16

### Application version and app store

26.04.4

### Homeserver

1.152.0

### Will you send logs?

No

### Are you willing to provide a PR?

No

chrissi55:
downgraded to 26.04.3 manually by apk file from github direrctly solves the problem, so definitely a prob of the latest (pre-release)
26.04.**4**

jmartinesp:
I'm sorry, but I can't really reproduce the issue, I've tried sending images several times and the media viewer behaves as expected. Could you send some logs with the URL of this issue in the message so we can check what's happening in your case?

chrissi55:
How can i produce logs in android?

The error occurs not by sending media!
It occurs when re-open sended media in timeline.
Open it twice or more than twice a time and you should see that the opened photo autmatically switches to another photo from the timeline unwanted by user

jmartinesp:
> How can i produce logs in android?

You can go to settings -> send bug report after reproducing the issue.

> The error occurs not by sending media! It occurs when re-open sended media in timeline. Open it twice or more than twice a time and you should see that the opened photo autmatically switches to another photo from the timeline unwanted by user

Oh, that's what I tried too, but you mentioned you sent some media before, so I was trying to reproduce the whole flow in case something went wrong with matching the local event in the send queue with the remote event once it's uploaded.

chrissi55:
OK if done that and send a log report after reproducing the error.
[...]

PR #6800, opened 2026-05-18
Fix media viewer flickering

jmartinesp:
## Content

Make sure the index and the underlying data are updated in sync.

This was caused by the data being loaded causing an index update that got out of sync with the displayed items, and for an instant, the pager index pointed to the wrong data until it was refreshed

- Turn `MediaViewerDataSource.collectAsState` into `MediaViewerDataSource.produceState` so we can modify both the index and the pager data at the same time without an intermediate recomposition.
- When calculating when to display the snackbars for 'no more items', take into account the previous state so we only react to changes to the underlying data source while keeping the same index (so you were at index 0 and there aren't newer items so the 'loading' item is removed, or you were at N-1 and the 'loading' item for older items is removed).

## Motivation and context

There were quite a few complaints (but no issues opened that I could find?) about the media viewer 'flickering' its media info in the top app bar from several users, and @dkasak  also complained about the wrong item being displayed when opening it (I couldn't reproduce this one, but hope these changes fix it).

## Tests

Open the media viewer from either the gallery or a timeline item:

- Check there is no flickering in the top app bar.
- Check the right item is displayed.
- Check when one of the loaders (backwards/forwards) disappears while you were viewing it, a snackbar appears with some warning.
- Check the media viewer still works as expected, loading new items in both directions.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.

### DIFF
--- a/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
+++ b/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
@@ -168,7 +168,7 @@ class MediaViewerDataSource(
                 is MediaItem.LoadingIndicator -> add(
                     MediaViewerPageData.Loading(
                         direction = mediaItem.direction,
-                        timestamp = systemClock.epochMillis(),
+                        timestamp = mediaItem.timestamp,
                         pagerKey = pagerKeysHandler.getKey(mediaItem),
                     )
                 )
--- a/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
+++ b/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
@@ -181,7 +181,9 @@ class MediaViewerDataSource(
     }
 
     suspend fun loadMore(direction: Timeline.PaginationDirection) {
-        galleryDataSource.loadMore(direction)
+        if (galleryDataSource.isReady) {
+            galleryDataSource.loadMore(direction)
+        }
     }
 
     suspend fun loadMedia(data: MediaViewerPageData.MediaViewerData) {
--- a/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
+++ b/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
@@ -11,12 +11,13 @@ package io.element.android.libraries.mediaviewer.impl.viewer
 import androidx.annotation.VisibleForTesting
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.ProduceStateScope
 import androidx.compose.runtime.State
-import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.remember
+import androidx.compose.runtime.produceState
 import io.element.android.libraries.architecture.AsyncData
 import io.element.android.libraries.core.extensions.mapCatchingExceptions
+import io.element.android.libraries.matrix.api.core.EventId
 import io.element.android.libraries.matrix.api.media.MatrixMediaLoader
 import io.element.android.libraries.matrix.api.media.MediaFile
 import io.element.android.libraries.matrix.api.timeline.Timeline
@@ -36,13 +37,16 @@ import kotlinx.collections.immutable.persistentListOf
 import kotlinx.collections.immutable.toImmutableList
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.withContext
 import timber.log.Timber
 
 class MediaViewerDataSource(
     mode: MediaViewerMode,
+    private val coroutineScope: CoroutineScope,
     private val dispatcher: CoroutineDispatcher,
     private val galleryDataSource: MediaGalleryDataSource,
     private val mediaLoader: MatrixMediaLoader,
@@ -74,13 +78,28 @@ class MediaViewerDataSource(
         localMediaStates.clear()
     }
 
+    /**
+     * Helper function to translate the [dataFlow] result to a Compose [State] that can be observed in the UI.
+     */
     @Composable
-    fun collectAsState(): State<ImmutableList<MediaViewerPageData>> {
-        return remember { dataFlow() }.collectAsState(initialData())
+    fun produceState(
+        producer: suspend ProduceStateScope<ImmutableList<MediaViewerPageData>>.(StateFlow<ImmutableList<MediaViewerPageData>>) -> Unit
+    ): State<ImmutableList<MediaViewerPageData>> {
+        return produceState(initialValue = initialData()) {
+            producer(dataFlow())
+        }
+    }
+
+    /**
+     * Find the index of the page corresponding to the given eventId, or null if not found.
+     */
+    fun findEventIndex(eventId: EventId?): Int? {
+        if (eventId == null) return null
+        return dataFlow().value.indexOfFirst { (it as? MediaViewerPageData.MediaViewerData)?.eventId == eventId }.takeIf { it >= 0 }
     }
 
     @VisibleForTesting
-    internal fun dataFlow(): Flow<ImmutableList<MediaViewerPageData>> {
+    internal fun dataFlow(): StateFlow<ImmutableList<MediaViewerPageData>> {
         return galleryDataSource.groupedMediaItemsFlow()
             .map { groupedItems ->
                 when (groupedItems) {
@@ -107,6 +126,11 @@ class MediaViewerDataSource(
                     }
                 }
             }
+            .stateIn(
+                scope = CoroutineScope(coroutineScope.coroutineContext + dispatcher),
+                started = SharingStarted.Lazily,
+                initialValue = initialData(),
+            )
     }
 
     private fun initialData(): ImmutableList<MediaViewerPageData> {
--- a/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
+++ b/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
@@ -46,7 +46,7 @@ import timber.log.Timber
 
 class MediaViewerDataSource(
     mode: MediaViewerMode,
-    private val coroutineScope: CoroutineScope,
+    coroutineScope: CoroutineScope,
     private val dispatcher: CoroutineDispatcher,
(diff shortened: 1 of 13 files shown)


libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerNode.kt
```diff
@@ -113,19 +114,22 @@ class MediaViewerNode(
...
-        navigator = this,
-        dataSource = MediaViewerDataSource(
-            mode = inputs.mode,
-            dispatcher = coroutineDispatchers.computation,
-            galleryDataSource = mediaGallerySource,
-            mediaLoader = mediaLoader,
-            localMediaFactory = localMediaFactory,
-            systemClock = systemClock,
-            pagerKeysHandler = pagerKeysHandler,
+    private val presenter by lazy {
```
ganfra: Do we really need the lazy here?

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
```diff
@@ -71,18 +73,33 @@ class MediaViewerDataSource(
...
+            producer(dataFlow())
+        }
+    }
+
+    /**
+     * Find the index of the page corresponding to the given eventId, or null if not found.
+     */
+    fun findEventIndex(eventId: EventId?): Int? {
+        if (eventId == null) return null
+        return dataFlow().value.indexOfFirst { (it as? MediaViewerPageData.MediaViewerData)?.eventId == eventId }.takeIf { it >= 0 }
```
ganfra: We are recreating the flow each time here.
The `StateFlow` should probably be kept as a variable instead.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
```diff
@@ -71,18 +73,33 @@ class MediaViewerDataSource(
...
         localMediaStates.clear()
     }
 
+    /**
+     * Helper function to translate the [dataFlow] result to a Compose [State] that can be observed in the UI.
+     */
     @Composable
-    fun collectAsState(): State<ImmutableList<MediaViewerPageData>> {
-        return remember { dataFlow() }.collectAsState(initialData())
+    fun produceState(
```
ganfra, line 85: Maybe we don't need this if we just expose the `StateFlow` instead?

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerPresenter.kt
```diff
@@ -205,23 +208,19 @@ class MediaViewerPresenter(
         currentIndex: IntState,
         data: State<ImmutableList<MediaViewerPageData>>,
     ) {
-        // With newest-first ordering, forward loading indicator is at the first index
-        val isRenderingLoadingForward by remember {
-            derivedStateOf {
-                currentIndex.intValue == 0 &&
-                    data.value.size > 1 &&
-                    data.value.firstOrNull() is MediaViewerPageData.Loading
+        var previousIndex by remember { mutableIntStateOf(currentIndex.intValue) }
```
ganfra: The code is almost the same between the 2 methods, can you create only one with a lambda to compute the `isLoading` value?

ganfra (commented):
Some quick remarks

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerNode.kt
jmartinesp: It seems like we do: passing the `lifecycleScope` breaks the `DefaultMediaViewerEntryPointTest` tests with some error about the main dispatcher. There may be some other solution though.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
jmartinesp: Ah, good call. I didn't want to change underlying logic that much, but since I already added the `stateIn`, it makes sense to only do it once.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerPresenter.kt
jmartinesp: Sure, let me try.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
jmartinesp: I added this so I wouldn't have to expose the `initialData()` function, it would just be included here.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerPresenter.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/6800/commits/c3438593fbe819addb5c934550a861691146f50f

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerNode.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/6800/commits/69a7407157fdf3b50f35960e5d92fad2316d8ff2

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/6800/commits/17d90dfd45d732aeff9dcf5c8eca0d962a89545f

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
jmartinesp: I've tried different variations of these and if we don't want to expose the `initialData()` function, the other alternative I see (since we also need the comparator) seems to be:

```kotlin
val initialData = remember { dataSource.dataFlow.value }
val data = produceState(initialData) {
    dataSource.dataFlow
        .distinctUntilChanged(::pageDataComparator)
        .collectLatest { new ->
            if (value.isEmpty() && new.isNotEmpty()) {
                currentIndex.intValue = dataSource.findEventIndex(inputs.eventId) ?: 0
            } else if (currentIndex.intValue > 0 && value.firstOrNull() is MediaViewerPageData.Loading &&
                new.firstOrNull() !is MediaViewerPageData.Loading) {
                // Restore index based on the eventId after the initial items have been loaded
                currentIndex.intValue = dataSource.findEventIndex(inputs.eventId) ?: 0
            }
            value = new
        }
}
```

I don't really like this solution though, I think the existing one is probably cleaner.

ganfra (approved):
Thanks for the changes, LGTM (but not tested)

jmartinesp:
Agh, the pager comparator breaks the remote pagination since it doesn't allow new loaders to be displayed, and that prevents new paginate actions to be triggered.

jmartinesp:
The latest changes should have fixed it:

- I tested several scenarios and I saw no flickering.
- We're loading items as usual.
- I didn't see any issues with indices.

But the new changes probably require a new review since they weren't trivial.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
```diff
@@ -168,7 +168,7 @@ class MediaViewerDataSource(
                 is MediaItem.LoadingIndicator -> add(
                     MediaViewerPageData.Loading(
                         direction = mediaItem.direction,
-                        timestamp = systemClock.epochMillis(),
+                        timestamp = mediaItem.timestamp,
```
ganfra, line 171: Are we not using systemClock on purpose in case of error, so the timestamp is used to retrigger a new pagination? 
```
 LaunchedEffect(dataForPage.timestamp) {
                        state.eventSink(MediaViewerEvent.LoadMore(dataForPage.direction))
                    }
```

ganfra (commented):
One more remark after the changes

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
jmartinesp: Hmm... If there was an error I was expecting the associated `LaunchedEffect` to be removed and then re-added, which should trigger the action again. Let me check if I can add a UI test for this case to verify it.

jmartinesp: I think this covers it: https://github.com/element-hq/element-x-android/pull/6800/commits/d1911a03ae3e745d47e1b5246e59651af86f1d8f

I believe this matches the real scenario, since I'm not sure how to simulate that in a real device.

jmartinesp: I actually hacked `TimelineMediaGalleryDataSource` so it would emit 2 loading indicator items with exactly the same values with an error between them and I could see `LoadMore` being triggered once per indicator.

bxdxnn:
> There were quite a few complaints (but no issues opened that I could find?) about the media viewer 'flickering' its media info in the top app bar from several users, and @dkasak also complained about the wrong item being displayed when opening it (I couldn't reproduce this one, but hope these changes fix it).

Some more info:
* There are two separate "flickering" issues:
  * wrong item being displayed, media switches, and a long flickering in the top bar: post-`6a4fed2` regression, tracked in #6695 which got fixed by #6715 already and is included in v26.05.1, but I guess that release is still in some partial rollout stage because there are still a lot of duplicated reports about this issue
  * top bar flickers a bit (icons flicker their disabled state, sender/date might disappear for a moment) when opening the media viewer: pre-`6a4fed2` existing bug, wasn't tracked by anything. This PR probably fixes this problem, as I was trying to do a similar solution

jmartinesp:
Oh, I had completely forgotten about that previous PR of yours. Thanks @bxdxnn !

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
ganfra: Ok, lets merge then!

ganfra:
@jmartinesp feel free to merge when you want :)

PR #6817, opened 2026-05-19
[Link new device] Rotate QrCode instead of showing an error

### LINKED ISSUE #6806: [Task] Replace the "sign in was not completed in time" QR code error with an automatic retry.
Android alter ego for https://github.com/element-hq/element-x-ios/issues/5483

When the user want to link a new device by showing a QR code, the application generate a new QR code each time an expiration error occurs.

bmarty:
## Content

When a QrCode is displayed, the user has about 1 minute to scan the QrCode. After the rendezvous session is closed and cannot be used anymore. To mitigate this, the application is now automatically rotating the QrCode 10 times, to give enough time for the user to scan the code. After 10 times, the existing error screen is shown, to avoid having infinite retry.

There is still a doubt on how the timeout is detected (I asked in the internal room), but the test are working fine here.

The application is showing a spinner instead of the QrCode only if the QrCode (re)generation is taking more than 1 second, to have a smooth transition between 2 consecutive QrCodes.

## Motivation and context

Closes #6806 

## Screenshots / GIFs

https://github.com/user-attachments/assets/3b1ec537-2ed5-4c19-83eb-a548ce78c3ba

## Tests

- Login to a server that have the Link new device feature (for instance element.io)
- On the setting click on "Link new device"
- Click on "Mobile device"
- A first QrCode should be displayed
- Every minutes, during 10 minutes, a new QrCode should be displayed
- After 10 minutes an error is displayed

It should be possible to scan any QrCode to link a new device
All back press should be correclty handled (for st QrCode and next ones).

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/linknewdevice/impl/src/main/kotlin/io/element/android/features/linknewdevice/impl/screens/qrcode/ShowQrCodePresenter.kt
+++ b/features/linknewdevice/impl/src/main/kotlin/io/element/android/features/linknewdevice/impl/screens/qrcode/ShowQrCodePresenter.kt
@@ -22,6 +22,9 @@ import io.element.android.libraries.architecture.Presenter
 import io.element.android.libraries.core.log.logger.LoggerTag
 import io.element.android.libraries.matrix.api.linknewdevice.LinkMobileStep
 import io.element.android.libraries.matrix.api.logs.LoggerTags
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
 import timber.log.Timber
 
 private val tag = LoggerTag("ShowQrCodePresenter", LoggerTags.linkNewDevice)
@@ -36,20 +39,54 @@ class ShowQrCodePresenter(
         fun create(initialData: String): ShowQrCodePresenter
     }
 
+    private var loadingJob: Job? = null
+
     @Composable
     override fun present(): ShowQrCodeState {
         var qrCodeRotationCounter by remember { mutableIntStateOf(MAX_QR_CODE_ROTATION) }
-        val data by produceState<AsyncData<String>>(AsyncData.Success(initialData)) {
+        val state by produceState(
+            initialValue = ShowQrCodeState(
+                data1 = AsyncData.Success(initialData),
+                data2 = AsyncData.Uninitialized,
+                dataToRender = 1,
+            )
+        ) {
             linkNewMobileHandler.stepFlow.collect { step ->
+                val currentValue = value
                 when (step) {
                     is LinkMobileStep.QrReady -> {
-                        value = AsyncData.Success(step.data)
+                        loadingJob?.cancel()
+                        if (currentValue.dataToRender == 1) {
+                            value = currentValue.copy(
+                                data2 = AsyncData.Success(step.data),
+                                dataToRender = 2,
+                            )
+                        } else {
+                            value = currentValue.copy(
+                                data1 = AsyncData.Success(step.data),
+                                dataToRender = 1,
+                            )
+                        }
                     }
                     is LinkMobileStep.QrRotating -> {
                         if (qrCodeRotationCounter-- > 0) {
                             Timber.tag(tag.value).d("Rotating QrCode")
                             linkNewMobileHandler.rotateQrCode()
-                            value = AsyncData.Loading()
+                            // Ensure that outdated data is not rendered too long while rotating QR code
+                            loadingJob = launch {
+                                delay(1000)
+                                if (currentValue.dataToRender == 1) {
+                                    value = currentValue.copy(
+                                        data2 = AsyncData.Loading(),
+                                        dataToRender = 2,
+                                    )
+                                } else {
+                                    value = currentValue.copy(
+                                        data1 = AsyncData.Loading(),
+                                        dataToRender = 1,
+                                    )
+                                }
+                            }
                         } else {
                             Timber.tag(tag.value).w("Max QR code rotation reached, not rotating anymore")
                             linkNewMobileHandler.onTooManyRotation()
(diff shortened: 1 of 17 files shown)


features/linknewdevice/impl/src/main/kotlin/io/element/android/features/linknewdevice/impl/screens/qrcode/ShowQrCodeView.kt
```diff
@@ -55,11 +58,24 @@ fun ShowQrCodeView(
...
+                        contentAlignment = Alignment.Center,
+                    ) {
+                        CircularProgressIndicator()
+                    }
+                }
+                else -> {
+                    QrCodeImage(
+                        data = str,
+                        modifier = Modifier
+                            .size(220.dp)
```
jmartinesp: Super nit: maybe use a shared modifier?

features/linknewdevice/impl/src/main/kotlin/io/element/android/features/linknewdevice/impl/screens/qrcode/ShowQrCodePresenter.kt
```diff
@@ -0,0 +1,58 @@
...
+        val data by produceState<AsyncData<String>>(AsyncData.Success(initialData)) {
+            linkNewMobileHandler.stepFlow.collect { step ->
+                when (step) {
+                    is LinkMobileStep.QrReady -> {
+                        value = AsyncData.Success(step.data)
+                    }
+                    is LinkMobileStep.QrRotating -> {
+                        Timber.tag(tag.value).d("Rotating QrCode")
+                        linkNewMobileHandler.rotateQrCode()
+                        value = AsyncData.Loading()
```
jmartinesp: It's probably fine, but should the order here be reversed, first change the state then rotate?

features/linknewdevice/impl/src/main/kotlin/io/element/android/features/linknewdevice/impl/screens/qrcode/ShowQrCodeState.kt
```diff
@@ -0,0 +1,16 @@
...
+
+package io.element.android.features.linknewdevice.impl.screens.qrcode
+
+import io.element.android.libraries.architecture.AsyncData
+
+data class ShowQrCodeState(
+    val data1: AsyncData<String>,
+    val data2: AsyncData<String>,
+    val dataToRender: Int,
+)
```
jmartinesp, line 14: Could we just pass a single data value here and then use `AnimatedContent` instead of 2 nodes with `AnimatedVisibility`?

```kotlin
@Composable
private fun QrCodeOrLoading(
    data: String?,
) {
    AnimatedContent(
        modifier = modifier,
        targetState = data,
        transitionSpec = {
            ContentTransform(fadeIn(), fadeOut())
        },
    ) { state ->
        val qrCodeModifier = Modifier.size(220.dp)
        if (state == null) {
            Box(
                modifier = qrCodeModifier,
                contentAlignment = Alignment.Center,
            ) {
                CircularProgressIndicator()
            }
        } else {
            QrCodeImage(
                data = state,
                modifier = qrCodeModifier,
            )
        }
    }
}
```

Seems to work here:

https://github.com/user-attachments/assets/3256726c-aa4e-4a4e-974f-30a0e5288cef

bmarty: It was to animate between 2 QrCode but yes I was reinvented the wheel, thanks for the hint about `AnimatedContent`. Changed in 5dd897dae81adfdf55601c7b16541180b89c9106

features/linknewdevice/impl/src/main/kotlin/io/element/android/features/linknewdevice/impl/screens/qrcode/ShowQrCodeView.kt
```diff
@@ -55,11 +64,17 @@ fun ShowQrCodeView(
...
-                    .size(220.dp)
-            )
+            AnimatedContent(
+                targetState = state.data.dataOrNull(),
+                transitionSpec = {
+                    fadeIn().togetherWith(fadeOut())
+                }
+            ) { data ->
+                QrCodeOrLoading(
+                    modifier = modifier.size(220.dp),
```
jmartinesp: The modifier could be used for `AnimatedContent` instead. I don't think it's important though.

jmartinesp (approved):
Thanks for the latest changes, it looks way cleaner now!

features/linknewdevice/impl/src/main/kotlin/io/element/android/features/linknewdevice/impl/screens/qrcode/ShowQrCodeView.kt
bmarty: It is actually wrong to use `modifier` parameter again here. I will fix, good catch! (I thought that the detekt rules were catching this mistakes, maybe it's broken?)

jmartinesp: I think since the migration to detekt 2 lots of checks have been broken. I tested a revert to 1.x the other day (to [this version](https://github.com/mrmans0n/compose-rules/releases/tag/v0.4.28) of the compose-rules) and it found > 50 issues.

bmarty: OK, fixed in b4f23d9d9e1d994de465d0ec91d506cf9c372917, let's revert the Detekt tool and rules separately.

PR #6823, opened 2026-05-20
Fix 'Conversation label cannot be empty' error

jmartinesp:
## Content

Make sure we provide a non-blank room name or id to the shortcut info builder.

## Motivation and context

This happens when building a `ShortcutInfoCompat` in `DefaultNotificationConversationService.onSendMessage` when the provided room name is not null but it's empty.

Fixes a rageshake: https://github.com/element-hq/element-x-android-rageshakes/issues/13220

## Tests

<del>To be honest, I'm not entirely sure how to reproduce this because an empty room name state event translates to a room with an `Empty room` name according to my tests, but I guess some users somehow bypassed this. Assuming you'd be able to do the same, the flow to test this would be:</del>

- Rename a room to a blank value like a single whitespace character.
- Send a message to that room.
- Check the app doesn't crash and the shortcut is created using the room id.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.


bmarty (approved):
Thanks!

PR #6835, opened 2026-05-21
Use a raw key for the SDK stores for new sessions

jmartinesp:
## Content

- Add `ClientSecret` to tell apart the legacy passphrases from the raw keys and centralise parsing/formatting.
- Use the `ClientSecret` to set the passphrase/key in the `SqliteStoreBuilder`.

## Motivation and context

Faster session restoration times, since we skip the key derivation from the passphrase (which isn't needed since we already provide a random key). Related to https://github.com/element-hq/element-x-android/pull/6774, which implements a similar change for client-side DBs.

## Tests

- With this branch installed, log into a new session.
- Then restart the app.
- The session restoration should be faster than before.

Remember to remove your session after the test, otherwise you'd be stuck with a non-working session if you go back to other branches while this one isn't merged.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.


ganfra (approved):
Looking great, thank you!

PR #6855, opened 2026-05-24
Fix formatting inconsistencies in latest event summaries

bxdxnn:
## Content

|Before|After|
|-|-|
|Poll: poll body|**Poll**: poll body|
|Sticker (sticker body)|**Sticker**: sticker body|

## Motivation and context

Make it consistent with the other message types

## Tests

- Send a sticker/poll message
- Go back to the room list

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/DefaultRoomLatestEventFormatter.kt
+++ b/libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/DefaultRoomLatestEventFormatter.kt
@@ -93,8 +93,8 @@ class DefaultRoomLatestEventFormatter(
                 message.prefixIfNeeded(senderDisambiguatedDisplayName, isDmRoom, isOutgoing)
             }
             is StickerContent -> {
-                val message = sp.getString(CommonStrings.common_sticker) + " (" + content.bestDescription + ")"
-                message.prefixIfNeeded(senderDisambiguatedDisplayName, isDmRoom, isOutgoing)
+                content.bestDescription.prefixWith(sp.getString(CommonStrings.common_sticker))
+                    .prefixIfNeeded(senderDisambiguatedDisplayName, isDmRoom, isOutgoing)
             }
             is UnableToDecryptContent -> {
                 val message = sp.getString(CommonStrings.common_waiting_for_decryption_key)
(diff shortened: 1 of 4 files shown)


libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/DefaultRoomLatestEventFormatter.kt
```diff
@@ -110,8 +110,8 @@ class DefaultRoomLatestEventFormatter(
                 stateContentFormatter.format(content, senderDisambiguatedDisplayName, isOutgoing, RenderingMode.RoomList)
             }
             is PollContent -> {
-                val message = sp.getString(CommonStrings.common_poll_summary, content.question)
-                message.prefixIfNeeded(senderDisambiguatedDisplayName, isDmRoom, isOutgoing)
+                content.question.prefixWith(sp.getString(CommonStrings.a11y_poll))
```
jmartinesp: Was this change needed? It seems like the result would be the same, and if needed I'd rather create a new value in localazy with the needed format instead of reusing the accessibility text here instead.

bxdxnn: Yes, this change is needed, otherwise the "Poll" wouldn't be bold.

jmartinesp: Ah, I see, thanks! Let me create some new string in localazy we can use here.

libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/PrefixWith.kt
```diff
@@ -20,6 +20,10 @@ internal fun CharSequence.prefixWith(prefix: String): AnnotatedString {
             append(prefix)
         }
         append(": ")
-        append(this@prefixWith)
+        if (this@prefixWith is AnnotatedString) {
+            append(this@prefixWith)
+        } else {
+            append(this@prefixWith.toString())
+        }
```
jmartinesp, line 27: I don't think this is needed, `AnnotatedString.append(CharSequence?)` already handles this case internally if I'm not mistaken. Since I'm doing some other changes I'll revert it too.

jmartinesp (approved):
Thanks, I'm merging it with a couple of tweaks.

Issue #6656, opened 2026-04-25
Spontaneous logout


bxdxnn:
### Steps to reproduce

1. Open the app

#### What happened instead?

Got logged out

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

jmartinesp:
In the logs I see:

```
2026-04-25T10:58:37.572921Z TRACE matrix_sdk::client: Token refresh: Using OAuth 2.0. | crates/matrix-sdk/src/client/mod.rs:1479
2026-04-25T10:58:37.580153Z  WARN matrix_sdk::authentication::oauth: error refreshing an OAuth 2.0 token: failed to refresh token: Server returned error response: invalid_grant: The provided access grant is invalid, expired, or revoked. | crates/matrix-sdk/src/authentication/oauth/mod.rs:1265
2026-04-25T10:58:37.580793Z ERROR matrix_sdk::client::futures: Token refresh: OAuth 2.0 refresh_token rejected with invalid grant | crates/matrix-sdk/src/client/futures.rs:128
2026-04-25T10:58:37.580863Z ERROR matrix_sdk::client::futures: Token refresh: OAuth 2.0 refresh_token rejected with invalid grant | crates/matrix-sdk/src/client/futures.rs:128
2026-04-25T10:58:37.580906Z ERROR matrix_sdk::encryption: Couldn't setup and resume recovery Sdk(Http(RefreshToken(OAuth(RefreshToken(ServerResponse(StandardErrorResponse { error: invalid_grant, error_description: Some("The provided access grant is invalid, expired, or revoked."), error_uri: None })))))) | crates/matrix-sdk/src/encryption/mod.rs:1865
2026-04-25T10:58:37.580960Z ERROR matrix_sdk::encryption::recovery: Couldn't update the recovery state: Sdk(Http(RefreshToken(OAuth(RefreshToken(ServerResponse(StandardErrorResponse { error: invalid_grant, error_description: Some("The provided access grant is invalid, expired, or revoked."), error_uri: None })))))) | crates/matrix-sdk/src/encryption/recovery/mod.rs:631
2026-04-25T10:58:37.581222Z ERROR matrix_sdk::client::futures: Token refresh: OAuth 2.0 refresh_token rejected with invalid grant | crates/matrix-sdk/src/client/futures.rs:128 | spans: next_sync_with_lock > sync_once{conn_id="encryption"  timeout=30000}
[...]

PR #6859, opened 2026-05-25
Add better logs to track token update failures

jmartinesp:
## Content

1. Make some logs use `info` log level instead of `debug`, so they appear in most user's bug reports.
2. Make the anonymized tokens even harder to reverse.
3. Detect when the tokens we should be saving match the current ones, as that's an error.

## Motivation and context

Logs related to session restoration issues don't seem to contain enough info, e.g. (https://github.com/element-hq/element-x-android/issues/6656).

## Tests

Check the new logs have the right log level and no PII is disclosed.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
@@ -131,7 +131,7 @@ class RustMatrixClientFactory(
             analyticsService = analyticsService,
             workManagerScheduler = workManagerScheduler,
         ).also {
-            Timber.tag(it.toString()).d("Creating Client with access token '$anonymizedAccessToken' and refresh token '$anonymizedRefreshToken'")
+            Timber.tag("RustMatrixClient").i("Creating Client with access token '$anonymizedAccessToken' and refresh token '$anonymizedRefreshToken'")
         }
     }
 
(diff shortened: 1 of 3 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/util/Token.kt
```diff
@@ -16,7 +16,8 @@ private val sha256 by lazy { MessageDigest.getInstance("SHA-256") }
 
 @OptIn(ExperimentalStdlibApi::class)
 private fun anonymizeToken(token: String): String {
-    return sha256.digest(token.toByteArray()).toHexString()
+    // Only keep the first 32 bytes of the token to avoid displaying too much information.
```
bmarty: Maybe more accurate with:

```suggestion
    // Only keep the first 16 bytes of the hashed token to avoid displaying too much information.
```

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -132,7 +132,7 @@ class RustMatrixClientFactory(
             analyticsService = analyticsService,
             workManagerScheduler = workManagerScheduler,
         ).also {
-            Timber.tag(it.toString()).d("Creating Client with access token '$anonymizedAccessToken' and refresh token '$anonymizedRefreshToken'")
+            Timber.tag("RustMatrixClient").i("Creating Client with access token '$anonymizedAccessToken' and refresh token '$anonymizedRefreshToken'")
```
bmarty, line 134: The previous tag was crazy :)

bmarty (approved):
Thanks!

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
jmartinesp: I think it was done to track the client instance, but I can't remember why 😅 .

Issue #6501, opened 2026-03-31
startup slow


thigg:
### Your use case

If I start signal for android after restart the app is available after <<1s. 
Element X takes about 3s.

I have contacts in signal and element. I noticed I am avoiding using element for those because signal is faster to use.

### Are you willing to provide a PR?

No

charlag:
Just installed X, it takes even more time to start than Classic

charlag:
Probably needs it's own issue but also just opening a chat loses a bunch of frames where Classic didn't struggle

bxdxnn:
Most of the time is spent on decrypting the app's DB.
I wonder if it's possible to improve it by e.g. running it asynchronously alongside other startup tasks, or to improve the decryption process somehow...

jmartinesp:
I ran some tests the other day using [perfetto](https://perfetto.dev/) to try to understand what's causing the app startup to be so slow and I found:

- 600ms initializing [JNA](https://github.com/java-native-access) (aka FFI bridge with the Rust SDK). There's no way to improve this until [uniffi](https://github.com/mozilla/uniffi-rs) supports JNI instead.
- 800ms building the client - this is after a patch in the SDK to load the client's cache stores in parallel. There's probably something else that can be done to speed this up, but I didn't have the time to check what's making it so slow.
- 400ms loading user certificates. Now this is a fun one: loading the user CAs from the Android `KeyStore` is super slow, and most users won't need this feature. However, to know if you have to add some custom user certificates to the SDK you first need to load such certificates. The only way to fix this would be to add some UI choice at some point for 'use custom certificates' that can be changed both after and before logging in.
- 900ms of SQLCipher initializing for the session DBs on the client side: that's probably what you were referring to @bxdxnn . SQLCipher is slow to initialize on purpose, it uses [key derivation](https://www.zetetic.net/sqlcipher/performance/#adjust-key-derivation) to create the actual keys it'll use to encrypt/decrypt the data. Reducing the key derivation rounds does improve the time it takes to open the DB and using a provided key manually sets this almost to 0... but it makes the app less secure and it wasn't backwards-compatible, from what I saw in my tests. So first, we'd have to decide if we want to make using the DB faster at the cost of it being less secure, then, how should we do it (less iterations? we provide the key?), and then figure out how to do it without breaking existing sessions.

[...]

PR #6861, opened 2026-05-25
Ask for user confirmation before trusting user CA certificates

bxdxnn:
## Motivation and context

1. Faster app startup (https://github.com/element-hq/element-x-android/issues/6501#issuecomment-4427942057)
2. It also adds more security that we don't trust random user CAs (that might be not intended to use with Element, e.g. to just access some private website) by default.

## Tests

- Have a CA certificate installed
  - The dialog should appear on launch
- Don't have a CA certificate installed
  - The dialog shouldn't appear on launch
- Run an MITM attack on the app (the MITM cert should be installed on the device):
  - Certificates not trusted: it should show the "Offline" badge on top
  - Certificates trusted: it should work normally

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


jmartinesp:
Thanks, I was planning on doing some work in this area in the near future.

While I agree not loading the CA certs if possible is better, hiding this behaviour behind a feature flag can be problematic: users who already logged in and need them will suddenly find they can't connect to their HS without any explanations, and users who haven't logged in yet have no easy way to enable this behaviour in release builds (enabling developer options is not intuitive, on purpose).

Maybe for new logins what we could do is: try detecting if there are any custom CAs installed in the device in the authentication flow and at that point display a dialog confirming whether these custom CAs should be used or not, while keeping them enabled by default for existing users to ensure backwards-compatibility.

Note this is just an idea and it should be confirmed by product (@mxandreas , I'm pinging you so you can check this once you're back from your holidays).

bxdxnn:
> who haven't logged in yet have no easy way to enable this behaviour in release builds (enabling developer options is not intuitive, on purpose).

1. You can tap on the app version on the initial launch screen to enable the developer settings
2. IMO, enabling custom CAs is dangerous, so it also shouldn't be intuitive.

> Maybe for new logins what we could do is: try detecting if there are any custom CAs installed in the device in the authentication flow and at that point display a dialog confirming whether these custom CAs should be used or not, while keeping them enabled by default for existing users to ensure backwards-compatibility.

That makes sense, but I'm not sure why we shouldn't also have this confirmation for the existing users too? The logic also applies to old logins, disabling/enabling custom CAs wouldn't break anything if they don't have them.

jmartinesp:
> IMO, enabling custom CAs is dangerous, so it also shouldn't be intuitive.

Well, this is true for most users, but there could be some customers in air-gapped environments who would probably depend on custom certs. I think displaying a dialog/intermediate screen with a detailed explanation about the risks is both user-friendly and informative.

> but I'm not sure why we shouldn't also have this confirmation for the existing users too?

I think having this confirmation means either displaying it unconditionally, which could be alarming for users who don't know this option exists at all, or only restoring the session data after having loaded the CAs to check they're there (which is what we want to avoid) and letting the user confirm if they want to use them if they're present. This could be a bit difficult since we'd have to add an intermediate state between not logged in and logged in to display the UI for this.

This 2nd option would make more sense and be easier to implement as part of the login flow, I think: you select a HS, we load the CAs there if they weren't loaded yet, which is ok because it doesn't really matter if it takes some extra 200-400ms at that point, and only after you confirm we create the SDK's `HttpClient` using the certs or not. We cache this decision in disk so when we restore the session in the future we can check this value to decide whether to load the certs or not.

bxdxnn:
Reimplemented as per the discussion

bxdxnn:
To make it clear for the product team, the current flow is roughly this:
* Load user CAs at the app startup
  * If there are CAs present
    * If we haven't displayed this dialog before, display this dialog:
      * Title: Trust user CA certificates?
      * Content: Your device has user-installed CA certificates. Do you want Element to trust them? User CA certificates can let others intercept or decrypt your secure network traffic. If you don't recognize a certificate or its owner, remove it in the device settings.
      * Buttons:
        * Confirm: Trust certificates
          * In this case, we always load and trust user CAs until the user goes to developer options and disables loading user CAs.
        * Decline: No
          * In this case, we never load CAs again until the user goes to developer options and enables loading user CAs.
    * If we already have displayed this dialog and user has consented to trusting user CAs, just load and trust the CAs
  * If there are no CAs present and the dialog was displayed before and the user has consented to trusting user CAs, do nothing.
  * If there are no CAs **and** the dialog wasn't displayed before, we never load CAs again until the user goes to developer options and enables loading user CAs.

mxandreas:
I am afraid this would need to be reviewed by security first (cc @dkasak ).

From a user perspective - I'd really like to keep things as simple as possible, as there is enough complexity - thus it must be of high importance and impact majority of users. There must be an elaboration/reasoning for this in the PR which can be understood by everybody who has not been hands-on this topic and have all the necessary context.

jmartinesp:
> I am afraid this would need to be reviewed by security first (cc @dkasak ).

We're already adding user certs by default to those used in the SDK, this is just limiting that behaviour because it's not performant to do so if there's no need to.

Adding a custom user certificate on Android is quite difficult on purpose, you'd have to download it, go to settings -> security -> more security -> encryption and credentials -> user credentials -> install, and you'd be asked to confirm whether you want to install the custom cert and some explanations why you shouldn't if you don't know what you're doing (it can't be done directly from the file), so while a 'phishing' attack of sorts is possible, it's the equivalent of a user downloading a random executable file from the internet, the antivirus saying it might be a virus and the user ignoring it and running it with admin privileges anyway 😓 .

> From a user perspective - I'd really like to keep things as simple as possible, as there is enough complexity - thus it must be of high importance and impact majority of users. There must be an elaboration/reasoning for this in the PR which can be understood by everybody who has not been hands-on this topic and have all the necessary context.

Using a certificate authority (CA) is necessary to validate the certificates the homeservers will provide in an HTTPS connection, and in some cases custom CAs are needed for those since public CAs won't be able to validate this certificate: an important case are air-gapped environments, where the client won't be able to access the internet and won't be able to reach the public CAs the OS trusts, so a custom (user-provided) CA is needed.

However, to detect *if* we need to use the user-added certs we first have to load **all** certs from the Android KeyStore into memory and this can take almost 0.5s, making the startup of the app quite slow, in most cases for no reason at all.

What I proposed is, when you try to log into a HS, load the user certs, check if one is needed to reach the HS, and in that case, either ask the user to confirm whether they want to use it or directly save a session-bound flag that we can check when we restore the session to understand whether we need to add the user certs or not. However, as we're already including the certs by default, this could be a potential problem for existing users that rely on this behaviour, so we'd have to either:

- When we don't know if the certs should be used (old logins), assume we need them and keep the current behaviour.
- Do a migration where we check if the user has custom certs and set this flag in the session data that we store in disk based on the result.

mxandreas:
> However, to detect if we need to use the user-added certs we first have to load all certs from the Android KeyStore into memory and this can take almost 0.5s, making the startup of the app quite slow, in most cases for no reason at all.

I am probably still missing some pieces in terms of why we need to determine if user-added certs are needed. I assumed the system to take care of TLS and trust whatever (CA) certificates the system has, e.g. if a CA/certificate is used that does not ship with Android, then one is using MDM or whatever to install an additional certificate and then the TLS works again.

> and in that case, either ask the user to confirm whether they want to use it or directly save a session-bound flag that we can check when we restore the session to understand whether we need to add the user certs or not.

This is clearly a no go, a regular user does not understand anything about certs and it only increases their effort to use the app.

jmartinesp:
> I am probably still missing some pieces in terms of why we need to determine if user-added certs are needed. I assumed the system to take care of TLS and trust whatever (CA) certificates the system has, e.g. if a CA/certificate is used that does not ship with Android, then one is using MDM or whatever to install an additional certificate and then the TLS works again.

You'd think so, but it turns out it doesn't. The JVM stack on Android can handle that automatically, but the Rust SDK, which is what performs the network requests, can't. We need to manually provide them if they're needed, and providing them is a very slow operation that needs to happen when the app starts. However, if you don't need those certs, your app is almost 0.5s slower to start for no good reason.

> This is clearly a no go, a regular user does not understand anything about certs and it only increases their effort to use the app.

As I explained above, if the user has a user-provided CA cert, either some admin installed it for them or they had to go through the whole flow I mentioned above, which means they either got some understanding of what they're doing or they're following some guide, given its complexity:

> Adding a custom user certificate on Android is quite difficult on purpose, you'd have to download it, go to settings -> security -> more security -> encryption and credentials -> user credentials -> install, and you'd be asked to confirm whether you want to install the custom cert and some explanations why you shouldn't if you don't know what you're doing (it can't be done directly from the file)

We can automatically provide the certs without asking the users, but I thought confirming their usage could be better security-wise. Another option is having some other build configuration option that forces the usage of some user CAs for custom-built apps, so we'd skip the confirmation in that case. 

I'm fine with skipping the confirmation if you think it's simpler and just saving whether custom CAs are needed or not on login/migration, or we can close this and keep the current behaviour even though it hurts performance.

mxandreas:
> As I explained above, if the user has a user-provided CA cert, either some admin installed it for them or they had to go through the whole flow I mentioned above, which means they either got some understanding of what they're doing or they're following some guide, given its complexity:

Even if the user were involved in this process earlier, the less we ask them, the better. I think a default expectation is that the certs are installed for the user so they don't need to be involved and also there are no further questions. This is especially important as there are already some things in the end to end encryption area we fully can't hide. The less we "leak" those "technical questions" to the end user, the better.

> or we can close this and keep the current behaviour even though it hurts performance.

If the performance hurts everybody, including those who have 0 user-provided certs, I'd configure this with some kind of explicit flag/configuration.

jmartinesp:
> Even if the user were involved in this process earlier, the less we ask them, the better.

Ok, that should also make the flow easier for us.

> If the performance hurts everybody, including those who have 0 user-provided certs, I'd configure this with some kind of explicit flag/configuration.

I believe this means going with the 'set the flag on login/app migration' plan then, and not asking for any confirmation, unless we want to set that up using some MDM value, but the question is how do we do that if accessing the server with the info might need the CA to be active, we might have a chicken-and-egg situation 😅 .

With the MDM option, we could set this flag per-session, which would allow us to add the custom CA certs only when needed, and keep the rest of the sessions fast to load. Adding an app migration to check if we're currently using user CA certs and setting the flag to true/false depending on that result, then doing the same for new logins. Maybe this could be added to the `SessionData` we store in a local DB. However, I'm not entirely sure how feasible this would be, given the problem mentioned above.

Otherwise, we'd set that for all users using an app-wide flag (as it is in the PR now, which can be accessed through the developer settings), but that means:
- If the user currently has user certs, we enable the flag (app migration/auto detection?).
- If a new user logs in and doesn't use the certs we can't just set the flag to false, because the previous user needs the custom CA, so their start-up time would be affected.
- Maybe, every time we load the certs (so, having the flag enabled or not set at all yet), detect if there are 0 certs and disable the flag automatically, which will only be enabled again either by manually doing so in the developer options, or when a new user logs in and we check for user CA certs again.

That seems a bit harder to handle, but it's also more centralised and can be toggled by the user if something goes wrong.

bxdxnn:
@jmartinesp I didn't get it, what was decided? If there are certs at the first startup, we should always load them, if there aren't, we never load them, no confirmation dialogs, and there's a dev option to toggle cert loading?

jmartinesp:
I wanted to add some info about this, but I was quite busy today: if all goes well we'll use [this](https://github.com/matrix-org/matrix-rust-sdk/pull/6645) instead.

We'll remove the special case for Android added to avoid issues with the revocation of certificates and OCSP not being available, only content revocation lists (CRL, which is the case for Let's Encrypt), which will force us to add some HTTP traffic exceptions for some CRL domains, but it *should* be fine overall. 

When tested again, the simplified platform verifier code was only checking a single cert against both the system and user certs you have on the Android device and it took microseconds to do so, so providing the user certs when building the client is no longer needed.

bmarty:
We decided in an internal meeting that we will not integrate this work for now.

PR #6887, opened 2026-05-26
Add unread count to the room unread indicator

### LINKED ISSUE #6867: Show unread message count instead of green dot in ElementX rooms
### Steps to reproduce

**Description:**
In the ElementX app, when receiving new messages in a room, only a green dot is shown as an indicator. There is no way to see how many unread messages have been received.

**Expected behavior:**
The number of unread messages should be displayed on the room list (similar to how it works in Element Classic), instead of just a green dot.

**Actual behavior:**
Only a green dot is shown, without any message count. This makes it difficult to prioritize which rooms need attention first.

**Affects:**
    ElementX (latest version)
    Works correctly in Element Classic (shows unread count)

Additional context:
This issue has been reported multiple times before by other users, but unfortunately it has not been fixed or properly addressed yet.

**Suggested fix:**
Please replace the green dot indicator with the actual number of unread messages per room, as it was in Element Classic.

**Priority:**
Critical — greatly impacts usability and room management.

Thank you for looking into this urgently.

Image

bxdxnn:
## Content

Add unread count to the room unread indicator in the room list.

The indicator will expand a bit depending on the number of characters. If the count is more than 99, "99+" will be shown instead.

## Motivation and context

Fixes #6867 and the duplicate issues of it

Feature parity with Element Classic and it has been frequently requested.

## Tests

- Go to the room list

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/atoms/UnreadIndicatorAtom.kt
+++ b/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/atoms/UnreadIndicatorAtom.kt
@@ -10,10 +10,10 @@ package io.element.android.libraries.designsystem.atomic.atoms
 
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.defaultMinSize
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.runtime.Composable
-import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
@@ -35,24 +35,26 @@ fun UnreadIndicatorAtom(
     isVisible: Boolean = true,
     contentDescription: String? = null,
 ) {
-    Box(
-        modifier = modifier
-            .semantics {
+    when {
+        !isVisible -> Spacer(modifier = modifier.size(size))
+        count != null && count >= 1 -> CounterAtom(
+            count = count.toInt(),
+            modifier = modifier.semantics {
                 contentDescription?.let { this.contentDescription = it }
-            }
-            .defaultMinSize(size, size)
-            .clip(CircleShape)
-            .background(if (isVisible) color else Color.Transparent),
-        contentAlignment = Alignment.Center,
-    ) {
-        if (count != null && count > 1) {
-            CounterAtom(
-                count = count.toInt(),
-                containerColor = Color.Transparent,
-                contentColor = ElementTheme.colors.bgCanvasDefault,
-                textStyle = ElementTheme.typography.fontBodySmMedium,
-            )
-        }
+            },
+            containerColor = color,
+            contentColor = ElementTheme.colors.bgCanvasDefault,
+            textStyle = ElementTheme.typography.fontBodySmMedium,
+        )
+        else -> Box(
+            modifier = modifier
+                .semantics {
+                    contentDescription?.let { this.contentDescription = it }
+                }
+                .size(size)
+                .clip(CircleShape)
+                .background(color),
+        )
     }
 }
 
--- a/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/atoms/UnreadIndicatorAtom.kt
+++ b/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/atoms/UnreadIndicatorAtom.kt
@@ -10,9 +10,10 @@ package io.element.android.libraries.designsystem.atomic.atoms
 
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.layout.defaultMinSize
 import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
@@ -29,6 +30,7 @@ import io.element.android.libraries.designsystem.theme.unreadIndicator
 fun UnreadIndicatorAtom(
     modifier: Modifier = Modifier,
     size: Dp = 12.dp,
+    count: Long? = null,
     color: Color = ElementTheme.colors.unreadIndicator,
     isVisible: Boolean = true,
     contentDescription: String? = null,
@@ -38,10 +40,20 @@ fun UnreadIndicatorAtom(
             .semantics {
                 contentDescription?.let { this.contentDescription = it }
             }
-            .size(size)
+            .defaultMinSize(size, size)
             .clip(CircleShape)
-            .background(if (isVisible) color else Color.Transparent)
-    )
+            .background(if (isVisible) color else Color.Transparent),
+        contentAlignment = Alignment.Center,
+    ) {
+        if (count != null && count > 1) {
+            CounterAtom(
+                count = count.toInt(),
+                containerColor = Color.Transparent,
+                contentColor = ElementTheme.colors.bgCanvasDefault,
+                textStyle = ElementTheme.typography.fontBodySmMedium,
+            )
+        }
+    }
 }
 
 @PreviewsDayNight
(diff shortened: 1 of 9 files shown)


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/atoms/UnreadIndicatorAtom.kt
```diff
@@ -38,10 +42,20 @@ fun UnreadIndicatorAtom(
...
             .clip(CircleShape)
-            .background(if (isVisible) color else Color.Transparent)
-    )
+            .background(if (isVisible) color else Color.Transparent),
+        contentAlignment = Alignment.Center,
+    ) {
+        if (count != null && count > 1) {
+            Text(
+                modifier = Modifier.padding(horizontal = 4.dp),
+                text = if (count > 99) "+99" else count.toString(),
```
Susurrus: Wouldn't ">99" or "99+" make more sense instead?

bxdxnn: This PR would require a product/design review AFAIK, so it's up to them to decide anyway. For example, we might as well remove this overflow logic as there's plenty of space, even if you'd have millions of unread messages.

But I'll change it to "99+" as you've suggested, it indicates an open range.

jmartinesp:
I'm not the reviewer of this PR, but I don't think we want to include this change at the moment: 

With the sync logic used in Element Android classic, you loaded everything that was pending since the last time you synced, and that was quite slow, but it resulted in accurate up to date data. 

In Element X on the other hand, we have sliding sync, that only returns the latest N events in a room depending on how it's set up (events, not messages), which is way faster... but can be problematic if we want to display accurate values like the unread count. So it could be easy that you see a room with maybe `5` unread notifications and you open it and there were 50 or 100, because the previous one happened a long time ago and we didn't load the data needed. 

We're working on back paginating the rooms in background to make the unread counts more accurate, but that's not ready yet.

By the way, if you wanted to display a number in the indicators you can't just use `numberOfUnreadMessages`: that's what we'd want to display (maybe?) if the room didn't have notifications enabled. If it had, we'd want to use `numberOfUnreadNotifications` if I'm not mistaken.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/atoms/UnreadIndicatorAtom.kt
```diff
@@ -38,10 +42,20 @@ fun UnreadIndicatorAtom(
...
+    ) {
+        if (count != null && count > 1) {
+            Text(
+                modifier = Modifier.padding(horizontal = 4.dp),
+                text = if (count > 99) "99+" else count.toString(),
+                style = ElementTheme.typography.fontBodySmMedium,
+                color = ElementTheme.colors.bgCanvasDefault,
+            )
+        }
+    }
```
bmarty, line 58: Can the existing `CounterAtom` be used here?

bmarty (commented):
Thanks. As you said, this change will require a product decision.

CC @mxandreas

bxdxnn:
@jmartinesp I agree with you, this change felt "good enough" but of course it would not work correctly ATM, it seems like we can just set it behind a feature flag for now?

sasan7071-maker:
As the person who reported this issue, I think you should enable the unread count even if it's not 100% accurate. It would still be very useful and would solve most of the problem for us. We can always improve the accuracy later.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/atoms/UnreadIndicatorAtom.kt
```diff
@@ -38,10 +40,20 @@ fun UnreadIndicatorAtom(
...
             }
-            .size(size)
+            .defaultMinSize(size, size)
             .clip(CircleShape)
-            .background(if (isVisible) color else Color.Transparent)
-    )
+            .background(if (isVisible) color else Color.Transparent),
+        contentAlignment = Alignment.Center,
+    ) {
+        if (count != null && count > 1) {
```
bmarty: Why not "count >= 1"?

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/atoms/UnreadIndicatorAtom.kt
```diff
@@ -38,10 +40,20 @@ fun UnreadIndicatorAtom(
...
+            CounterAtom(
+                count = count.toInt(),
+                containerColor = Color.Transparent,
+                contentColor = ElementTheme.colors.bgCanvasDefault,
+                textStyle = ElementTheme.typography.fontBodySmMedium,
+            )
+        }
+    }
```
bmarty, line 58: It's weird to have the CounterAtom inside the Box that draws the existing drawing indicator.
I would write:
```
@Composable
fun UnreadIndicatorAtom(
    size: Dp = 12.dp,
    count: Long? = null,
    color: Color = ElementTheme.colors.unreadIndicator,
    isVisible: Boolean = true,
    contentDescription: String? = null,
) {
    when {
        !isVisible -> Spacer(modifier = modifier.size(size))
        count != null && count >= 1 -> CounterAtom(
            count = count.toInt(),
            ...
        )
        else -> Box(
            modifier = modifier
                .semantics {
                    contentDescription?.let { this.contentDescription = it }
                }
                .size(size)
                .clip(CircleShape)
                .background(color)
        )
    }
}
```

bmarty (changes requested):
Thanks for the update! I guess it's OK now with the feature flag.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/atoms/UnreadIndicatorAtom.kt
sasan7071-maker: I'm just a user, not a developer. Can you share an APK for Samsung A55 with this feature enabled temporarily, so I can use it until the full fix is ready?

sasan7071-maker:
I'm just a user, not a developer. Can you share an APK for Samsung A55 with this feature enabled temporarily, so I can use it until the full fix is ready?

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/atoms/UnreadIndicatorAtom.kt
bxdxnn: @sasan7071-maker please don't spam reviews with unrelated comments

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/atoms/UnreadIndicatorAtom.kt
bxdxnn: That's up to product/design decision, but IMO, to make it less noisy, I think having the indicator without a number would make it clear that there's only one unread message, otherwise there will be a lot of numbers in the room list.

But if this is not wanted, we can make it `count > 0` instead.

bxdxnn:
@sasan7071-maker this kind of behavior (duplicating issues, flooding PRs with unrelated comments) is not very appropriate and may result in you being blocked from engaging in this repo.

If you need customer support and want your issues to be prioritized, please buy the [Element Enterprise](https://try.element.io/get-started) solution to get proper support for your problems in the app. I'm just a volunteer doing this work for free.

The APK build is [here](https://github.com/bxdxnn/element-x-android/actions/runs/26630492575), you would need to enable the feature flag in the developer options (tap 7 times on the app version) afterwards.

sasan7071-maker:
I have downloaded and used this, and it worked very well — very useful.
If I were you, this should be moved to the next pre-release and main release.
It was excellent for my needs.
Please consider including it in the next version.
Thank you.

sasan7071-maker:
I downloaded the APK build is [here](https://github.com/bxdxnn/element-x-android/actions/runs/26630492575). it worked very well — very useful.
I think you should enable the "unread count" feature flag in the developer options for the next pre-release and release. I'm waiting for this useful feature.
This feature is not just useful, it is vital for us. We look forward to seeing it enabled in the developer options.

mxandreas:
Indeed the eventual goal is to show the number of messages (so it is also compatible with Web) however, we have postponed this until we get a reliable number. The work to build the foundation for getting a reliable number is currently in progress.

bmarty (approved):
Thanks for the update!
@mxandreas since the counter is behind a disabled by default feature flag, I guess we can merge the PR?

mxandreas:
> Thanks for the update! @mxandreas since the counter is behind a disabled by default feature flag, I guess we can merge the PR?

Yes, that's fine.

PR #6902, opened 2026-05-27
Hide message notification content when the app is locked

bxdxnn:
## Content

New setting (enabled by default, not shown when the app PIN code is not setup) that makes the app use the fallback notification (the "You have (count) new messages" notification) when the app is closed (i.e. locked).

## Motivation and context

ATM you can still read and reply from the notifications on new messages even when the app is locked, which almost defeats the entire purpose of having an app PIN lock.

## Tests

Setup the "Screen lock" PIN code

- Have the option enabled
  - Have the app opened
    - You should receive notifications normally
  - Have the app closed
    -  You should receive the fallback notification
- Have the option disabled
  - You should receive notifications normally

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

    - [ ] No.


bmarty:
FTR we have this settings in Element Classic (but not defaulting to hiding message content in notification):

image

jmartinesp:
@mxandreas another PR to check once you're back 😅 .

We were also discussing this behaviour earlier and a valid point was whether we want to link the 2 options and only have this hiding of notification content option available if there's a PIN code, or if it should always be present (disabled by default) and the user can change it at any time.

mxandreas:
> ATM you can still read and reply from the notifications on new messages even when the app is locked, which almost defeats the entire purpose of having an app PIN lock.

I am curious about the use cases or real life situations for this one. I understand the phone is unlocked (otherwise one could hide content in notifications, right), then a new message arrives and then you would not want its content to be shown? A few examples would be appreciated.

jmartinesp:
I think when this was asked before it was for the case where you have notifications as heads up (so, displaying above everything else even if you don't open the notification drawer) and someone is looking at your phone/recording it somehow.

So you want to know there is new content in the app, you just don't want to display the content unless you unlock the app.

mxandreas:
> and someone is looking at your phone/recording it somehow

I understand that this can theoretically happen - you show your phone to someone and because of bad timing some sensitive message appears but for me this is an edge case. So, unless there is a very clear scenario in which it happens regularly and there is no workaround, I'd not spend time on it.

bmarty:
We also said that it can be a user wish to hide the content of the notification from the OS, i.e. the notification content will not be leaked in the system notification history, or even possibly sent to third parties. It will please paranoid/privacy-focused users/customers to have such settings. But we could implement it later.

Domimueller85:
das ist schön. 👍

bxdxnn:
@mxandreas besides the content hiding, it also fixes the issue that you can use the quick notification actions for replying to bypass the PIN lock of the app.
I personally don't know what's the use-case of the whole PIN lock feature, but since it exists, I think the app should be really fully locked if so, including notifications.

mxandreas:
Could we avoid another new setting and couple this tightly to the PIN lock. So, whenever you have activated PIN, the notifications won't show the content?

bmarty (approved):
LGTM, thanks!

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotificationDrawerManager.kt
```diff
@@ -197,6 +200,8 @@ class DefaultNotificationDrawerManager(
             it.sessionId
         }
 
+        val isAppLocked = lockScreenService.lockState.value == LockScreenLockState.Locked
```
bmarty: This check is strange to me, because when the app is unlocked and in foreground, the message content will be displayed in the notification. Then the app can be backgrounded (or killed), and the user will be able to send a message from the notification without having to unlock the application. 

I think it is better/simpler to use `lockScreenService.isPinSetup()`

bmarty (changes requested):
Thanks for the update. Can you check my remarks and see why the tests are failing?

bmarty (approved):

PR #6821, opened 2026-05-20
Disable cron jobs and private SSH key jobs in forks

jmartinesp:
## Content

What the title says.

## Motivation and context

I recently tested some changes related to forks in my fork of the repo and I started receiving periodic emails about the scheduled jobs/flows failing every day. On top of that, there were some other failures when the jobs required some credentials they didn't have (tokens, ssh keys, etc.).

With this, I stopped receiving emails for the cron jobs and I saw the same flows that previously failed because they lacked the required secrets succeed.

## Checklist

- This PR was made with the help of AI:
    - [ ] Yes. In this case, please request a review by Copilot.
    - [x] No.
- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly defines what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #6905, opened 2026-05-28
Revert "Disable cron jobs and private SSH key jobs in forks"

jmartinesp:
Reverts element-hq/element-x-android#6821

It seems like the implementation used for the checks is wrong: it should be `if: !github.event.repository.fork` or something similar, otherwise it'll actually prevent PRs from forks from running correctly, like: https://github.com/element-hq/element-x-android/actions/runs/26561159423/job/78244408046?pr=6898

Let's revert this for now.


bmarty (approved):

bmarty:
Thanks for the follow-up!

Issue #6737, opened 2026-05-06
Telegram style message selection


Kommynct:
#### What would you like to do?

I want to be able to easily select many messages with a long press.

#### Why would you like to do it?

To easily copy-paste or forward multiple messages without the selection flow interrupting the conversation view or requiring many extra taps.

#### How would you like to achieve it?

Long-pressing a message should immediately begin selection mode and create tickboxes. You should then be able to:

* tap additional messages to toggle selection
* swipe up/down to quickly select many messages
* continue scrolling normally while selecting

Actions like copy, forward, and delete should appear in the top bar.

The important part is keeping the UI visually stable and lightweight while selecting, instead of switching into a heavy modal workflow.

### Have you considered any alternatives?

The current approach works functionally, but feels slower and more disruptive for selecting many or non-adjacent messages. Checkbox-heavy or modal approaches used by many apps also tend to break conversational context.

### Additional context

I recommend looking at and copying Telegram’s UX for this. Its selection flow feels extremely natural because:

* selection is message-centric
* scrolling continues normally
* the layout barely changes
* multi-select remains fast even across long histories

It preserves spatial continuity and minimizes mode-switching, which makes bulk actions much more comfortable to use.

### Are you willing to provide a PR?

No

PR #6918, opened 2026-05-30
Add message multi-select for bulk copy, forward and delete

manfrommedan:
### Content

Adds an opt-in multi-select mode to the timeline, behind a new Labs flag
(Multi-select messages, off by default).

- Long-press a text message to start selecting, then tap rows to add or remove
  them.
- Or pick "Select" from the message action menu. This is the entry point for
  media, where a long-press already opens the menu.
- While selecting, a top bar shows the count and offers bulk Copy, Forward and
  Delete. The selection is capped at 30 and survives rotation and process death.
- When every selected message is a file, the top bar also offers Save, which
  writes them all to the Downloads folder. A gallery counts as all of its items.

Delete only appears when you have redaction rights and only redacts the messages
you are allowed to. Bulk redact and forward run through the active timeline, so
they work in threads and focused timelines too. With the flag off, nothing
changes.

Save is hidden rather than disabled for a selection which holds anything that is
not a file, since there would be nothing for it to save. Files are saved one at a
time so a batch of videos does not start several downloads at once, and progress
is shown in a banner rather than blocking the room. The banner can be cancelled;
files already written are kept. Saving goes through a new `MediaFileSaver` in the
media viewer api, as `LocalMediaActions` is internal to its impl module.

Two points where I would welcome a second opinion:

- The progress banner is a new surface. A snackbar at the end would be less code,
  but a batch of videos takes long enough that silence reads as nothing happening.
- Hidden versus disabled for Save. Copy stays disabled because it is the action
  people look for first, while a Save action that cannot apply seemed better left
  out of the bar.

### Motivation and context

Relates to #6737. Selecting messages one at a time through the action sheet is
slow and breaks the reading flow; this keeps the timeline stable and lets you
act on many or non-adjacent messages at once.

### Screenshots / GIFs

`@PreviewsDayNight` previews cover the selection top bar, the selection
indicator and a timeline in selection mode, so the screenshot tests pick them
up in light and dark.

### Tests

Unit tests cover enter/toggle/clear, the 30 cap, bulk copy and forward ordering,
bulk delete (success and partial failure), the "Select" menu intercept, the
selectable-content filter, and the forward presenter with multiple events.

### Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): Android 12, Android 16

### Checklist

- [x] Changes are based on the `develop` branch.
- [x] The PR title describes the user-facing change (for the release notes).
- [x] I added unit tests.
- [x] Tested on a device in light and dark themes.
- [x] Accessibility considered (Checkbox role and state on selectable rows).
- [x] Self-reviewed the PR.

Note for the reviewer: the new English strings are in
`features/messages/impl/src/main/res/values/temporary.xml`. Please import them
into Localazy and I will rebase onto the generated keys.


manfrommedan:
Small follow-up: while testing this on a device I found a bug in the selection handling and folded the fix into this PR.

Repro: enter selection on a media message, scroll it slightly so it leaves the screen and comes back, cancel the selection, then tap the message. It would toggle selection again instead of opening the viewer. It was intermittent, and scrolling far enough that the row gets disposed and recreated made it go away.

Cause: the tap/long-press handlers read `selectionState.isActive` from the state snapshot captured when the row last composed. A `LazyColumn` row that is retained off-screen across the clear is not recomposed, so it kept a stale `isActive=true` and toggled on the next tap. The fix reads the flag through a single `rememberUpdatedState` value, so a handler captured by a stale row still observes the current state.

I did not add a unit test for it: the handler logic in isolation is correct, and the bug only appears from `LazyColumn` retention combined with a recomposition and input timing window. A Compose UI test recomposes deterministically (`waitForIdle`), so it would pass on both the broken and the fixed code and give false confidence rather than catch the regression. I verified it by hand on-device. Better to have caught it here than after a release.

bmarty:
Hello @manfrommedan , thanks for the PR! In order to keep consistency between Element clients (Android/Web/iOS) can I ask you to create an issue on https://github.com/element-hq/element-meta/issues describing the behavior of the feature? It must include all the requirements (ex limit selection to 30 messages, etc.). This way we can involve the product and the design team, and make this PR mergeable.

Another remark, the code which manage the multiple selection using press and drag seems a bit complex, I'd prefer that it's removed from this PR. We may introduce this gesture later. Let's keep things simple for now.

manfrommedan:
Thanks @bmarty.

I removed the press-and-drag selection from this PR. Selection is now started from the "Select" action and a long-press on a text message, and toggled by tapping rows. The DragToSelect registry and gesture (and their test) are gone, which removes the most complex part. The drag can come back later as a separate change if we want it.

I also opened a behaviour issue so product and design can have a look: element-hq/element-meta#3255. It covers the entry points, the bulk actions, the cap and the redaction rules, and keeps the drag gesture out of scope.

The new strings are still in temporary.xml. I'll rebase onto the generated keys once they are imported into Localazy.

mxandreas:
It would help to speed up the review if there was a build that I can install.

manfrommedan:
Thanks! I just pushed a fix for the Konsist failures (state defaults, `toImmutableSet`, preview name), so the checks should pass once the workflows are approved to run again. The "Build APKs (debug)" run will then have an installable debug APK under its Artifacts.

Note that multi-select is behind a Labs flag and is off by default. To try it: Settings > Advanced settings > Labs > turn on "Multi-select messages". Then long-press a text message (or use "Select" from the message menu) to start, tap rows to add/remove, and use the top bar for bulk copy/forward/delete. Selection is capped at 30 and survives rotation.

mxandreas:
I tried to run this in a simulator and the app version number indicates the right branch, but I do not see this feature, e.g. the ability to select multiple messages. Is there an additional feature flag for this or?

mxandreas:
> I tried to run this in a simulator and the app version number indicates the right branch, but I do not see this feature, e.g. the ability to select multiple messages. Is there an additional feature flag for this or?

Nvm, found it in labs.

mxandreas:
I've taken a quick look. We should not change how long-press and short-press work. We should stick to the approach that short-press does nothing and long-press opens bottom sheet. This is the case now and this is also the case in WA and Signal. I confirmed this with @amshakal as well.

So, for now we would enable users to select multiple messages only by:
1. Long-press on message.
2. Choose Select from the context menu.

In terms of specific UI/design, I have recorder a video. Hope this helps the designers to see if there are specific issues to fix or not. cc @americanrefugee .

Btw, I do not actually see a need to put this feature into Labs. Either we're happy and we release it or not. 


https://github.com/user-attachments/assets/16b76440-1ccf-427a-a837-687e531fdc82

americanrefugee:
I have a few thoughts...

1. The video provided changes the entire logic. Previously, you selected "Forward" and then were taken to a menu of potential recipients to select. We should keep that behavior.

2. On WhatsApp, if you long-press a message then there is only "Forward". Tap that and then the multi-select buttons appear at once with the original message selected - no need to separately choose "Select". From there, the composer changes to an action bar with either Forward or Share. I think we copy this patter. Below are screenshots for reference (iOS).

3. For Android specifically, the header can just say "X selected" instead of "X messages selected"


1WA-long press message
2WA-select messages
3WA-select recipients

mxandreas:
> The video provided changes the entire logic. Previously, you selected "Forward" and then were taken to a menu of potential recipients to select. We should keep that behavior.

@americanrefugee I think you're mixing up 2 different things. In here, we're talking about selecting multiple *messages* and then either copying, deleting or forwarding these messages. In case you choose to forward, selecting the recipients is the next step and is completely independent of if you choose 1 or more messages.

> On WhatsApp, if you long-press a message then there is only "Forward". Tap that and then the multi-select buttons appear at once with the original message selected - no need to separately choose "Select"

This may be case on iOS. On Android, a long-press activates both multi-selection mode, and also shows the "toolbar" at the top, which then has all the various actions like copy, forward, delete, etc. which we have in the bottom sheet. This is how it looks:



However, in our case we can't use that model, because it is weird to go both into multi-select and open bottom sheet. And at the same time we do not want to re-define what short-press and long-press do, hence the Select option in the bottom sheet (which is exactly what Signal also does).

> For Android specifically, the header can just say "X selected" instead of "X messages selected"

Yes, we can keep it shorter as I do not think the "messages" adds any further clarification.

manfrommedan:
Just my two cents on the gesture model — I think users would genuinely appreciate this version.

The idea is simple: a tap always does the most useful thing for whatever you tapped — media opens, and text (which has nothing to open) shows its actions, so a tap is never wasted. Long-press then starts the multi-select, which keeps that quick. It's close to how Telegram feels on Android, so it should come across as familiar rather than surprising.

Happy to go with whatever works best for everyone in the end — just wanted to share why I think this one's a nice little improvement on the usability side. 🙂

americanrefugee:
> I think you're mixing up 2 different things.

You're right. I got confused. Understanding the video better now :)

I'm ok with the proposal, but want to get an opinion from @amshakal 

Then the question becomes what exactly we do on iOS since a few UI patters are different.

> The idea is simple: a tap always does the most useful thing for whatever you tapped — media opens, and text (which has nothing to open) shows its actions, so a tap is never wasted. Long-press then starts the multi-select, which keeps that quick. It's close to how Telegram feels on Android, so it should come across as familiar rather than surprising.

For Telegram or WhatsApp on iOS there is only long press for actions. I can see the usability benefit, but this would break our existing convention which I don't think we want to entertain right now. It's important to keep as much as we can the exact same on both platforms.

manfrommedan:
One thing worth weighing: strict iOS/Android parity can leave Android users without patterns that feel natural on their own platform. A lot of apps deliberately diverge so each side feels native rather than identical, and I'd argue the tap-aware version is the better Android experience. Totally respect it if parity is the priority for now though.

americanrefugee:
> One thing worth weighing: strict iOS/Android parity can leave Android users without patterns that feel natural on their own platform. A lot of apps deliberately diverge so each side feels native rather than identical, and I'd argue the tap-aware version is the better Android experience. Totally respect it if parity is the priority for now though.

Interesting... I ran the scenario through Claude and this is the concluding statement I got back after some more detailed analysis:

> For your chat app, the safe cross-platform choice is long-press for message actions on both platforms — it's what users expect everywhere. If you want to surface common actions faster on Android, you could additionally show a faint ⋮ tap target on messages, but that's an enhancement, not a requirement.

I'm not suggesting that Claude is "right" or that you are "wrong". But it gives me more confidence that long-press on Android at least isn't totally alien. So for now I think we should just keep long-press for both and revaluate if we get more feedback in the future. It's (usually) easier / more robust to maintain parity rather than diverge anyway.

But thank you @manfrommedan for your feedback! It definitely gives us more perspective to consider when developing features for Android in the future :)

manfrommedan:
Sounds good, and thanks for digging into it properly. Glad the perspective was useful. I'll leave the final call and the conventional-model tweak to the team, and I'm happy to see the feature land whichever way works best across both platforms. 🙂

mxandreas:
> The idea is simple: a tap always does the most useful thing for whatever you tapped — media opens, and text (which has nothing to open) shows its actions, so a tap is never wasted. Long-press then starts the multi-select, which keeps that quick. It's close to how Telegram feels on Android, so it should come across as familiar rather than surprising.

Similarly to what Aaron said above - this sounds quite practical to me, if we started from a clean slate. But given we already have a pattern, and this patter is also established in several other apps, we do not want to change it lightly.

Thus, the conclusion for now is that:
* Short-press continues to do nothing.
* Long-press continues to open the bottom sheet menu.
* Multi-select is done via the Select in the bottom sheet menu.

Agreed?

manfrommedan:
Agreed, that works for me. 🙂

amshakal:
Agreed. Aaron and I will will go through the designs propose a few changes. A quick one might be changing the button copy from 'select' to 'select messages'.

Kommynct:
> > The idea is simple: a tap always does the most useful thing for whatever you tapped — media opens, and text (which has nothing to open) shows its actions, so a tap is never wasted. Long-press then starts the multi-select, which keeps that quick. It's close to how Telegram feels on Android, so it should come across as familiar rather than surprising.
> 
> Similarly to what Aaron said above - this sounds quite practical to me, if we started from a clean slate. But given we already have a pattern, and this patter is also established in several other apps, we do not want to change it lightly.
> 
> Thus, the conclusion for now is that:
> 
>     * Short-press continues to do nothing.
> 
>     * Long-press continues to open the bottom sheet menu.
> 
>     * Multi-select is done via the Select in the bottom sheet menu.
> 
> 
> Agreed?

I want the telegram design, if you use telegram for a while it feels incredibly nice. I have given it to multiple tech illiterate elderly people and faced no issues with this. I think at the very least this should be toggleable, and honestly telegram is way more popular than element x, I'm not sure what the value is in sticking to an older, frankly clunkier design. Especially considering element x isn't even properly released/stable yet.

I think the menu should be the short press which currently does nothing anyway? long press should be relegated to selection, which is also how it works for text fields!

manfrommedan:
Yeah, honestly it didn't fully land for me either, I'd have gone with the Telegram flow too if it were up to me. But I ended up taking the compromise.

It's a cross-client app at the end of the day, so the team wants the gesture to line up across iOS, web and desktop, and one client running off with its own pattern is a tough sell. Not much I can do about that one. I did leave room for a toggle later if they ever warm up to the idea.

Kommynct:
> Yeah, honestly it didn't fully land for me either, I'd have gone with the Telegram flow too if it were up to me. But I ended up taking the compromise.
> 
> It's a cross-client app at the end of the day, so the team wants the gesture to line up across iOS, web and desktop, and one client running off with its own pattern is a tough sell. Not much I can do about that one. I did leave room for a toggle later if they ever warm up to the idea.

also as for feature parity this seems like something where a toggle in the config should be used on both platforms no? Is there any reason ios users wouldn't also want this?

manfrommedan:
Rebased on develop, and added bulk save: when every selected message is a file, the top bar offers Save and writes them all to Downloads. It stays hidden for a mixed selection.

Happy to swap the progress banner for a snackbar at the end, or make Save disabled rather than hidden, if either fits better.

PR #6923, opened 2026-06-01
Create log messages from `WebView` just once

jmartinesp:
## Content

What the title says.

Also, check the retrieved message for passwords before formatting the log line instead of after: this way we don't create a new formatted log line to just discard it.

## Motivation and context

We were creating the same message twice by mistake. We saw some crashes related to `OutOfMemoryError`s where this code was mentioned: while it's probably not the root cause, it's probably not helping either 😅 .

## Tests

Not much to test, other than the logs still being displayed and the log lines with passwords in them being discarded.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.


bmarty (approved):
Good catch, thanks!

PR #6944, opened 2026-06-02
Add well-known parsing and extension seam for custom recovery passphrase

jennaharris7:
Adds support for a user-chosen recovery passphrase. Base story: element-hq/element-meta#3228 (ER-233).

The feature itself (the custom passphrase UI and the strength estimator) lives entirely in the enterprise repo — see element-hq/element-android-enterprise#24. 

This PR contains only the FOSS-side pieces that enterprise builds on
- Extension seam
- Well-known parsing
- Design system addition for PasswordVisibilityToggle

 ## Tested devices

  - [x] Physical
  - [ ] Emulator
  - OS version(s): Android 16

  ## Checklist

      - [ ] No.


bmarty:
> Submodule pointer is intentionally not bumped here

What is the reason for this? The CI cannot build the project if the ref is not updated.

jennaharris7:
> > Submodule pointer is intentionally not bumped here
> 
> What is the reason for this? The CI cannot build the project if the ref is not updated.

You're right — sorry for the confusion. The pointer was left unbumped because the two new EnterpriseService methods are implemented in element-android-enterprise#24, and I wanted to avoid landing develop on a non-main enterprise commit before that PR merged. But leaving the gitlink at the old commit means the Enterprise variant can't compile — so the workable path would be to point to the enterprise pr, merge that one, then bump the pointer again here to the newly created pointer on develop. 

It seems we’re changing directions here though, moving this feature to live entirely in enterprise. So instead of bumping the submodule here, the FOSS-side pieces of this PR will be moving into the enterprise repo and this PR will shrink accordingly / likely be superseded.

I'll hold off on changes here until the split has been reworked — will update both PRs once that's sorted.

jennaharris7:
> > Submodule pointer is intentionally not bumped here
> 
> What is the reason for this? The CI cannot build the project if the ref is not updated.

The feature has been moved to enterprise - one open question: the strings are currently also housed in enterprise, whereas the iOS team were forced to move their strings to the FOSS due to technical limitations.  The enterprise submodule has no translation pipeline today — no localazy.xml, no values-* translations, and it isn't referenced in tools/localazy/config.json.  Do we want to:
 
1. Keep in the enterprise submodule but no translations and no shared source-of-truth with iOS.
2. Keep in the enterprise submodule + extend Localazy - this reuses the existing service (no new infra), but puts proprietary pro_ strings into the public, community-translate Localazy project 
3. Mirror iOS and keep the strings in FOSS even though the code lives in enterprise. This keeps both platforms on the same Localazy keys + translations.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/encryption/RustEncryptionService.kt
```diff
@@ -131,19 +131,28 @@ class RustEncryptionService(
 
     override suspend fun enableRecovery(
         waitForBackupsToUpload: Boolean,
-    ): Result<Unit> = withContext(dispatchers.io) {
+        passphrase: String?,
+    ): Result<String> = withContext(dispatchers.io) {
```
jmartinesp, line 135: Shouldn't this better be `String?` and return `null` if there isn't a passphrase? Seems easier to follow than later checking if the returned value is empty.

features/securebackup/impl/src/main/kotlin/io/element/android/features/securebackup/impl/setup/CustomPassphraseDerivations.kt
```diff
@@ -0,0 +1,51 @@
...
+    val meetsMinLength = requirements?.isSatisfiedBy(passphrase) ?: true
+    val mismatch = confirm.isNotEmpty() && passphrase != confirm
+    val canContinueFromEntry = requirements != null &&
+        passphrase.isNotEmpty() &&
+        meetsMinLength &&
+        setupState is SetupState.Init
+    val canSubmit = requirements != null &&
+        passphrase.isNotEmpty() &&
+        passphrase == confirm &&
+        meetsMinLength &&
```
jmartinesp:
```
requirements != null &&
        passphrase.isNotEmpty() &&
        meetsMinLength
```

This part seems common to both variables, maybe it could be extracted to its own variable so it's easier to read? I'm not sure what I'd call it though since I'm not sure what it's supposed to mean 😅 . `isValidPassphrase` maybe?

features/securebackup/impl/src/main/kotlin/io/element/android/features/securebackup/impl/setup/SecureBackupSetupPresenter.kt
```diff
@@ -53,10 +65,81 @@ class SecureBackupSetupPresenter(
...
+        // anyway, so we just clear it on success and accept loss on process death.
+        var customPassphrase by remember { mutableStateOf("") }
+        var customPassphraseConfirm by remember { mutableStateOf("") }
+        var customEntryStep by remember { mutableStateOf(CustomEntryStep.Entry) }
+        // Single-flight guard: a button tap and an IME-Done in the same frame both see
+        // canSubmit=true and would each launch an enableRecovery coroutine. The state machine
+        // dedupes UserCreatesKey, but only this flag stops the duplicate SDK call.
+        var customSubmitInFlight by remember { mutableStateOf(false) }
+        // Handle to the in-flight custom submit so a back press can abort it (see CancelCustomPassphraseSubmit).
+        var customSubmitJob by remember { mutableStateOf<Job?>(null) }
```
jmartinesp: I may be wrong, but couldn't `customSubmitJob?.isActive == true` act the same as `customSubmitInFlight == true`? It's not wrong, but I think using the `Job` to check this should be enough.

features/securebackup/impl/src/main/kotlin/io/element/android/features/securebackup/impl/setup/SecureBackupSetupPresenter.kt
```diff
@@ -53,10 +65,81 @@ class SecureBackupSetupPresenter(
...
+                )
+            }
+        }
+
+        // Strength is an enterprise-only estimation; null while the field is empty or in FOSS builds.
+        // Only computed on the Entry step, the only place the indicator is rendered.
+        val customPassphraseStrength by remember {
+            derivedStateOf {
+                customPassphrase
+                    .takeIf { customEntryStep == CustomEntryStep.Entry && it.isNotEmpty() }
```
jmartinesp: Everything in the presenter seems to use `remember { derivedStateOf { ... } }`, which again isn't wrong, but it's not performant either if I'm not mistaken. In this case in example, I think `remember(customPassphrase, customEntryStep) { ... }` would also work and be more efficient.

features/securebackup/impl/src/main/kotlin/io/element/android/features/securebackup/impl/setup/SecureBackupSetupView.kt
```diff
@@ -144,10 +241,269 @@ private fun Content(
...
+}
+
+// Stops mirror iOS PasswordStrengthBar.gradient: Compound red900 / lime700 with the same vivid
+// orange & yellow midpoints. Intentionally fixed (not light/dark semantic) so the gradient reads
+// identically to iOS in both themes.
+private val passphraseStrengthGradientStops = listOf(
+    0.25f to Color(0xFFD51928), // Compound red900
+    0.5f to Color(0xFFFF9500), // orange
+    0.75f to Color(0xFFFFCC00), // yellow
+    1.0f to Color(0xFF54C424), // Compound lime700
```
jmartinesp: Why not use the actual color tokens? You can reference the raw color tokens using `LightColorTokens/DarkColorTokens.color___`. You'd have to add an annotation with `@OptIn(CoreColorToken::class)` to remove a warning, but that should work.

jmartinesp (commented):
I'm halfway through the review and I added some comments about style. I think @bmarty will take over though.

features/securebackup/impl/src/main/kotlin/io/element/android/features/securebackup/impl/setup/CustomPassphraseDerivations.kt
jmartinesp: Oh, I see this was removed later.

features/securebackup/impl/src/main/kotlin/io/element/android/features/securebackup/impl/setup/SecureBackupSetupView.kt
jmartinesp: It seems like this is gone too. If it's added in some other repo, I think the suggestion still stands.

features/securebackup/api/src/main/kotlin/io/element/android/features/securebackup/api/SecureBackupSetupEntryPoint.kt
```diff
@@ -0,0 +1,30 @@
...
+import io.element.android.libraries.architecture.FeatureEntryPoint
+import io.element.android.libraries.architecture.NodeInputs
+
+/**
+ * Entry point for the recovery-key setup screen.
+ *
+ * The default (FOSS) implementation builds the standard auto-generated-key setup node. Enterprise
+ * builds replace this binding to vend a richer setup node (e.g. a homeserver-advertised custom
+ * recovery passphrase) while falling back to the standard node when no custom spec applies.
+ */
```
bmarty: I am not sure if we want to give details about what's in Pro build in the open. Maybe just remove this comment?

libraries/wellknown/impl/src/main/kotlin/io/element/android/libraries/wellknown/impl/Mapper.kt
```diff
@@ -17,4 +22,17 @@ internal fun InternalElementWellKnown.map() = ElementWellKnown(
...
+
+private fun InternalCustomRecoveryPassphraseRequirements.toPublic(): CustomRecoveryPassphraseRequirements? {
+    val min = minCharacterCount ?: run {
+        Timber.tag(loggerTag.value).w("custom_recovery_passphrase_settings missing min_character_count; ignoring spec")
+        return null
+    }
+    if (min <= 0) {
+        Timber.tag(loggerTag.value).w("custom_recovery_passphrase_settings.min_character_count must be > 0; ignoring spec")
+        return null
+    }
```
bmarty, line 46: Instead of ignoring spec, maybe we could have a default min number of chars hard-coded in the app?

bmarty (approved):
Thanks, the feature (in pro) is working well. Only small comments.

libraries/wellknown/impl/src/main/kotlin/io/element/android/libraries/wellknown/impl/Mapper.kt
jennaharris7: Added a min floor of 1 for this in a205d03b43

features/securebackup/api/src/main/kotlin/io/element/android/features/securebackup/api/SecureBackupSetupEntryPoint.kt
jennaharris7: Good call - comment removed in 1df91ca8d5

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/encryption/RustEncryptionService.kt
jennaharris7: The downstream empty-check this was guarding against was removed with the FOSS custom flow in a050716123; the "" is now just a documented internal placeholder for the passphrase path (the key is intentionally kept out of the session-scoped flow). Happy to switch Done.recoveryKey to String? if you'd still prefer the idiomatic null, but it would require the auto-gen path to handle a null which it currently doesn’t need to do? (This was converted to String in an earlier commit)

features/securebackup/impl/src/main/kotlin/io/element/android/features/securebackup/impl/setup/SecureBackupSetupPresenter.kt
jennaharris7: This was resolved as a part of this refactor: 5bf1246eea

features/securebackup/impl/src/main/kotlin/io/element/android/features/securebackup/impl/setup/SecureBackupSetupView.kt
jennaharris7: This was fixed in the enterprise repo: d4947948d2

jennaharris7:
one open question: the strings are currently also housed in enterprise, whereas the iOS team were forced to move their strings to the FOSS due to technical limitations. The enterprise submodule has no translation pipeline today — no localazy.xml, no values-* translations, and it isn't referenced in tools/localazy/config.json. Do we want to:
> 
> 1. Keep in the enterprise submodule but no translations and no shared source-of-truth with iOS.
> 2. Keep in the enterprise submodule + extend Localazy - this reuses the existing service (no new infra), but puts proprietary pro_ strings into the public, community-translate Localazy project
> 3. Mirror iOS and keep the strings in FOSS even though the code lives in enterprise. This keeps both platforms on the same Localazy keys + translations.

Per discussion we have chosen to go with option 3, the strings have been moved to FOSS here: 4f2c2b773a

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/encryption/RustEncryptionService.kt
jmartinesp: It's not big deal, we can keep the check.

features/securebackup/impl/src/main/kotlin/io/element/android/features/securebackup/impl/setup/SecureBackupSetupPresenter.kt
jennaharris7: Fixed in 22d08bc

PR #6961, opened 2026-06-04
Make Vulkan a *not required* feature

jmartinesp:
## Content

- Override the `required=true` value from MapLibre.
- Instead, check this on runtime using `DeviceHasVulkanSupport`.
- If the device doesn't support Vulkan, which is needed to render the maps, prevent them from opening the map-based screens and display a dialog instead.

## Motivation and context

Support devices that have no/unknown Vulkan support.

## Screenshots / GIFs

Should be in the PR.

## Tests

Not much to test unless you can force your device to report not supporting Vulkan. Maybe you could reverse the check in `DeviceHasVulkanSupport` and try opening a shared location or sharing a new location.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.

### DIFF
--- a/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/system/DeviceHasVulkanSupport.kt
+++ b/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/system/DeviceHasVulkanSupport.kt
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2026 Element Creations Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.androidutils.system
+
+import android.content.Context
+import dev.zacsweers.metro.Inject
+import io.element.android.libraries.di.annotations.ApplicationContext
+
+private const val VULKAN_VERSION_1_0 = 0x400003
+
+/**
+ * Checks if the device supports Vulkan 1.0.
+ *
+ * This is needed for the location screens that contain maps using MapLibre UI components.
+ *
+ * Needed until https://github.com/maplibre/maplibre-native/issues/3079 is resolved and we can automatically choose between OpenGL and Vulkan renderers,
+ * or no devices support OpenGL anymore.
+ */
+@Inject
+class DeviceHasVulkanSupport(
+    @ApplicationContext private val context: Context,
+) {
+    operator fun invoke(): Boolean {
+        return context.packageManager.hasSystemFeature("android.hardware.vulkan.version", VULKAN_VERSION_1_0)
+    }
+}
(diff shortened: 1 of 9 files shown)


libraries/ui-strings/src/main/res/values/localazy.xml
```diff
@@ -546,4 +546,6 @@ Are you sure you want to continue?"</string>
   <string name="timeline_decryption_failure_historical_event_user_not_joined">"You don\'t have access to this message"</string>
   <string name="timeline_decryption_failure_unable_to_decrypt">"Unable to decrypt message"</string>
   <string name="timeline_decryption_failure_withheld_unverified">"This message was blocked either because you did not verify your device or because the sender needs to verify your digital identity."</string>
+  <string name="vulkan_not_supported_dialog_content_android">"Your device is too old, a device with Android 8 or newer is required."</string>
```
bxdxnn, line 549: Maybe be explicit that the Vulkan support is needed instead of the Android version?
e.g. some people might have custom ROMs with the latest Android, but the device still doesn't support Vulkan.

bxdxnn (approved):
Thanks!

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/system/DeviceHasVulkanSupport.kt
```diff
@@ -0,0 +1,31 @@
...
+ * Needed until https://github.com/maplibre/maplibre-native/issues/3079 is resolved and we can automatically choose between OpenGL and Vulkan renderers,
+ * or no devices support OpenGL anymore.
+ */
+@Inject
+class DeviceHasVulkanSupport(
+    @ApplicationContext private val context: Context,
+) {
+    operator fun invoke(): Boolean {
+        return context.packageManager.hasSystemFeature("android.hardware.vulkan.version", VULKAN_VERSION_1_0)
+    }
```
bmarty, line 30: Do we really need to provide the version here?

libraries/ui-strings/src/main/res/values/localazy.xml
jmartinesp: That was our first thought, but it seemed like something quite technical that most users probably wouldn't understand.

bmarty (approved):
Thanks, tested OK on my device.

libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/system/DeviceHasVulkanSupport.kt
jmartinesp: I guess not, but since it's already stated in the manifest, we might as well add it too.

bmarty: I do not get it, the version is not stated in the manifest. Do I miss something?

bmarty: My remark is that we could write:
```kotlin
context.packageManager.hasSystemFeature("android.hardware.vulkan.version")
```

jmartinesp: I meant in the merged one (or the one that takes the value from MapLibre [here](https://github.com/maplibre/maplibre-native/blob/35126d6a51d040de349c6c269b5fa744e0022378/platform/android/MapLibreAndroid/src/vulkan/AndroidManifest.xml#L5)).

But yeah, it could have been used without the version since it's `1.0` anyway.

PR #6984, opened 2026-06-06
Fix message drafts not being saved in Threads

bxdxnn:
## Content

Save thread roots when saving drafts

## Motivation and context

Wasn't working properly before, probably also fixes some other weird bugs regarding saving drafts from threads.

## Tests

- Write a draft message in a thread then go back to the room list and back to that thread
  - It should be still there

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
@@ -49,6 +49,7 @@ import io.element.android.libraries.designsystem.utils.snackbar.SnackbarDispatch
 import io.element.android.libraries.designsystem.utils.snackbar.SnackbarMessage
 import io.element.android.libraries.di.annotations.SessionCoroutineScope
 import io.element.android.libraries.matrix.api.core.EventId
+import io.element.android.libraries.matrix.api.core.ThreadId
 import io.element.android.libraries.matrix.api.core.UserId
 import io.element.android.libraries.matrix.api.permalink.PermalinkBuilder
 import io.element.android.libraries.matrix.api.permalink.PermalinkParser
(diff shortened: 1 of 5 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
```diff
@@ -110,6 +111,7 @@ class MessageComposerPresenter(
     @Assisted private val navigator: MessagesNavigator,
     @Assisted private val timelineController: TimelineController,
     @Assisted private val isInThread: Boolean,
+    @Assisted private val threadRoot: ThreadId?,
```
bmarty, line 113: it looks like `isInThread` is always equal to `threadRoot != null`, so I guess `isInThread` could be removed from the constructor and moved in the class body.

bmarty (commented):
Thanks, one comment, and I trigger the CI.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
```diff
@@ -138,10 +139,11 @@ class MessageComposerPresenter(
         fun create(
             timelineController: TimelineController,
             navigator: MessagesNavigator,
-            isInThread: Boolean,
+            threadRoot: ThreadId?,
         ): MessageComposerPresenter
     }
 
+    private val isInThread: Boolean get() = threadRoot != null
```
bmarty: `get()` is not necessary here, since the val is immutable.


```suggestion
    private val isInThread = threadRoot != null
```

bmarty (changes requested):
Thanks for the update. One final remark.

bmarty (approved):
Thanks!

PR #6985, opened 2026-06-07
Fix "Sent" checkmark disappearing when the message is read and "Share presence" is disabled

bxdxnn:
## Content

Don't compute read receipts if the presence sharing is disabled, and remove the (now) unnecessary checks in timeline-related views for rendering read receipts.

## Motivation and context

1. Fixes the bug in the title
2. Makes the code simpler and more robust due to not having to check for the setting everytime now

## Tests

- Have the last message read while "Share presence" is disabled

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


jmartinesp:
Thanks for fixing the issue, I think ideally we could even move preventing the read receipt computation to `TimelineItemsFactoryConfig.computeReadReceipts = preference` when we create timelines, but that makes it harder to work with as it would have to be instantiated asynchronously in `TimelinePresenter`, so I'm not sure it's worth it.

I just fixed the preview you changed so it behaves how it used to, so we can see all the different states (it was now missing a few of them since the read receipts were taking some space). I think this means new screenshot tests need to be recorded (or the previous commit changing them removed) and only you can do it 😅 .

jmartinesp:
<del>Can you also upload an empty commit so I can trigger the CI? I don't think mine would unlock it.</del>

---

A rebase made it work.

jmartinesp (approved):
There are some issues with codecov at the moment, but the tests pass. Merging this!

PR #6994, opened 2026-06-08
Detect RTL text in events and adjust the text rendering

### LINKED ISSUE #6974: Persian (RTL) messages overlap with the timestamp
### Steps to reproduce

Issue Description:
When sending a message in Persian (or any RTL language), the last line of the message text overlaps with the timestamp (time label) shown on the same side. This makes the last few characters of the message unreadable.

Steps to reproduce:

Change the software language or keyboard to Persian (RTL)

Type a multi-line Persian message

Send the message

Look at the last line of the message – it crashes into the timestamp

Expected behavior:
The message text should be properly wrapped and padded so that the last line ends before reaching the timestamp area. The timestamp should remain visible and not overlap with the message content.

Suggested fix:

Add proper RTL padding/margin for message bubbles in RTL mode

Ensure text wrapping accounts for the timestamp's width

Apply different layout rules for RTL languages (Persian, Arabic, Hebrew)

Additional context:
This issue does not occur in LTR languages like English. It only happens when the message is written in a right‑to‑left script. This is a critical accessibility and usability issue for Persian (Farsi), Arabic, and Hebrew users.

### Outcome

 What did you expect?
Add proper RTL padding/margin for message bubbles in RTL mode

 What happened instead?
When sending a message in Persian (or any RTL language), the last line of the message text overlaps with the timestamp (time label) shown on the same side.

Image

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

Yes

Yes

jmartinesp:
## Content

Adjust the timeline event rendering when the text/layout direction of the content doesn't match the one from the device (e.g. a message with RTL text in a device with LTR locale).

This mainly changes the position of the timestamp.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/6974.

cc @mxandreas / @americanrefugee  maybe?

## Screenshots / GIFs

| Before | After |
| --- | --- |
| image | image |

## Tests

In a device with LTR locale (most languages) open a room with messages written in RTL (Arabic, Hebrew, Persian...). Check they look fine. Optionally, change the locale to one of those and open rooms with messages written in both LTR and RTL languages.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.


sasan7071-maker:
I just installed the debug apk, but the problem was solved 
Here is the screenshot
 
Screenshot_20260608_221213_Element X dbg

jmartinesp:
Oh, that's weird. It's working on the UI previews and it seemed to work fine when I tested it on a real device, using an English locale and reading messages in Arabic. Can I ask you which language your device is using? Is there some public room I could use for testing?

sasan7071-maker:
I'm using Persian/Farsi (an RTL language). The main issue is not whether the text is left-aligned or right-aligned.
The real problem is the spacing between the message text and the timestamp.
Currently, the timestamp and the text overlap each other. They render on top of each other instead of having proper separation.
Your debug version actually made the problem worse and is not useful for me.
Please focus on fixing the overlapping issue, not just changing LTR/RTL alignment. The text and timestamp should not be placed on top of one another.
Thank you.
Screenshot_20260609_135611_Element

jmartinesp:
I added some changes that I hope fix the issue 🤞 .

sasan7071-maker:
let me check it out....

sasan7071-maker:
While the original overlap issue is resolved , a new issue has appeared specifically for Persian (Farsi) text.
After submitting a message in Persian (Farsi) , the message text exhibits a visible shaking or jittering effect. but it works correctly for the message in English.

Persian (Farsi) - an RTL works shaking.
English (LTR) - works correctly. 

Please watch the attached clip to see the problem

https://github.com/user-attachments/assets/d4dc16bc-922d-4a17-84fe-dd71400a93c7

jmartinesp:
I've tried simplifying how we detect and apply the text direction to the layout, which I think has fixed the issues. Could you please double check @sasan7071-maker ?

sasan7071-maker:
I think you solved the problem, but it would be better if the timestamp were on the right side of the text messages.
timestamp is on the right side in all the massanger like whatapp and telegram 

Screenshot_20260610_221200_Element X dbg

sasan7071-maker:
I think you solved the problem, but it would be better if the timestamp were on the right side of the text messages.
Like all the other messenger WhatsApp and telegram the timestamp is on the right side of the text messages.
Thanks.

Screenshot_20260610_221200_Element X dbg

jmartinesp:
Could you share a screenshot of what that looks like on WhatsApp or Telegram? I'm looking at what my Telegram client does (it might not match yours) and it's just placing the timestamp below the text no matter what when we have RTL text in an LTR context. Is this what you expect?

sasan7071-maker:
As you can see in the screenshots, the time of the messages is shown at the bottom right:
WhatsApp"
Screenshot_20260611_111922_WhatsApp
Telegram:
Screenshot_20260611_112044_Telegram

jmartinesp:
I have a commit with changes that would make this work as WhastApp/Telegram, but I'd rather double check that's what we want to do. @americanrefugee what do you think we should do when the text content direction doesn't match the layout direction (e.g. content is RTL but layout is LTR)? 

Should we do like WhatsApp and Telegram and put the timestamp next to the text if it has enough space **and it's a single line**, otherwise below it? Or should we flip the direction of the layout instead so the timestamp and the content are placed in reverse but finding if there is free space in the last line is consistent?

*Removed screenshots*

I think doing what the other IMs do makes sense, even if it's a bit more complex. It'll also mean this matches what users have seen in other apps, which is usually good.

americanrefugee:
Agreed: Let's do what the other IMs do.

After all L/R placement is not the issue, so let's do what's already expected.

btw @jmartinesp the screenshots you attached above are broken for me

sasan7071-maker:
Thank you for working on this fix. The main issue I’m seeing is the overlap between message text and timestamp, which I highlighted in the attached screenshot.

In my opinion, placing the timestamp at the bottom right works best – and that’s how the current stable version of the app behaves.

Please just work on the overlap between message text and timestamp.
It is a screenshot from last release 26.06.02:
Screenshot_20260612_182633_Element X

jmartinesp:
Sorry for the delay @bmarty I think the PR should be ready to review now. Since lots of things changed, maybe it's better if you review all the code at once. Let me know if you'd rather have me rebase the code.

sasan7071-maker:
I downloaded and installed the debug APK from the link. The original issue with timestamp and text overlapping has been fixed.

However, there is another issue:
1. For single-line messages, the spacing between the text and the timestamp (let's call this 'Space A') and the top/bottom padding ('Space B') are correct.
2. For multi-line messages (2+ lines) , 'Space A' (between text and timestamp) is missing, and 'Space B' (the vertical padding at the top/bottom of the bubble) is too large. This makes the message bubble height much taller than it should be.
3. Compared to the last release, the height of multi-line messages has increased too much

Element X dbg Screenshot:
Screenshot_20260613_085721_Element X dbg

jmartinesp:
But that's how it works on WhatsApp? RTL text in a new line in a LTR context will always be to the right in an LTR context, which is also where the timestamp will be. To avoid this overlap, what they seem to do is push the timestamp to the next line unconditionally if I'm not mistaken.

In this other screenshot I found from WhatsApp, both in the 1st and 3rd messages there should be enough space to add the timestamp in the last line *if it was on the left*, but as we want to keep it to the right, as you confirmed and we iterated on before, there's no available space there and it gets moved to the next line.

image

sasan7071-maker:
Let me explain the problem in a different way. As you can see in the screenshot, the height of Persian messages is significantly larger compared to English messages.

1. Number 1 points to the height difference between a two-line Persian message and a two-line English message.
2. Number 2 points to the height difference between a three-line Persian message and a three-line English message.
3. Number 3 points to the height difference between a four-line Persian message and a four-line English message.

The height of multi-line messages has increased too much

image

jmartinesp:
Ok, I just tested it again with WhastApp and I saw it's not really unconditionally adding a new line if there's enough free space in a multiline message to put the timestamp next to the message, my bad. I'll see if I can achieve something similar without breaking our current rendering or affecting the performance by having to measure everything several times 🫤 .

jmartinesp:
I iterated on this a bit and I think it's working as you expect now @sasan7071-maker , if there is enough space next to the text even if it's multiline it'll place the timestamp there. A new APK will be generated soon if you want to test it.

sasan7071-maker:
Unfortunately the issue hasn't been resolved yet. 
Screenshot:
Screenshot_20260615_180555_Element X dbg

jmartinesp:
Let's give it another try. I think I reproduced the problem and fixed it, APKs should be re-built soon 🤞 .

sasan7071-maker:
You've fixed the message and timestamp overlap issue, but the padding and spacing is now excessive and doesn't look perfect. Please see Screenshot 1.
In your previous version, having the timestamp on the left side for Persian and Arabic text was much better and it was perfect – as shown in Screenshot 2.
Let's go with the approach from Screenshot 2. the timestamp on the left side for Persian and Arabic text is so perfect. I believe it's a much better solution and It solves all problems.

**Please apply code that the timestamp is on the left side for Persian and Arabic text as shown in Screenshot 2:**

image

jmartinesp:
I'm sorry, but we already agreed with design to do what other IMs do, so I think we'll continue either with the latest changes, which match what WhatsApp does or the previous one which matches what Telegram does:

| WhatsApp | Telegram |
| --- | --- |
| Screenshot_2026-06-16-08-19-14-27_6012fa4d4ddec268fc5c7112cbb265e7 | Screenshot_2026-06-16-08-19-44-49_948cd9899890cbd5c2798760b2b95377 |

@americanrefugee sorry for the ping again, but which one do you think works best?

sasan7071-maker:
I know you are right, but the timestamp on the left side for Persian solves all problems and it is perfect.
actually the timestamp on the right side is useless.
Please place the timestamp on the left side.

sasan7071-maker:
the timestamp on the left side for Persian makes it so beautiful. Please be sure to place the timestamp on the left side as shown in Screenshot. in my opinion it is perfect and there is no issue with this code.

**Please apply code that the timestamp is on the left side for Persian and Arabic text as shown in screenshot :**

image

americanrefugee:
Ok, after further consideration: **let's go with the timestamp on the left for RTL languages.**

We acknowledge that this might look weird when there's a mix of RTL / LTR messages in a room since the timestamp won't be in the same place. However, this is an edge case and I'd rather optimize for the majority.

If after releasing this we get pushback from RTL users that they would rather have the timestamp on the right, then we already know what to do (go with the Telegram approach).

Sorry for all the back-and-forth! I just want our RTL language users to be happy, not forced to live with a compromise simply because it's what other apps are doing.

jmartinesp:
Ok, I've uploaded the latest changes with what @americanrefugee suggested, matching @sasan7071-maker 's expectations. @bmarty the PR should now be ready to review, sorry for all the unexpected changes.

americanrefugee (approved):

jmartinesp:
@bmarty I rebased the PR so it now only contains a few incremental commits instead of ~15 contradicting ones 😅 . Ready to review.

tests/uitests/src/test/snapshots/images/features.messages.impl.timeline.components_TimelineItemEventRowRtlContent_en.png
bmarty, line 1: This screenshot looks OK, I do not see any diff in the other screenshots about timeline touched by this PR.

bmarty (approved):
Thanks!

tests/uitests/src/test/snapshots/images/features.messages.impl.timeline.components_TimelineItemEventRowRtlContent_en.png
jmartinesp: Yeah, me neither. It must be something super subtle 🫤 .

jmartinesp: Ah, it looks like it's an issue with rounded corners somehow 🫤 

image

PR #6996, opened 2026-06-08
change(location): ensure permissions are always requested at least once

### LINKED ISSUE #6850: Sharing location: Weird behavior when Location permission is `Ask every time`
### Steps to reproduce

1. Set the Location permission of the app to `Ask every time`
2. Tap on "Share location"

#### What did you expect?
I would expect to be asked to accept that the app acceses the location once.

Once accepted, the app should focus on my actual location so that I can share it.
It is normal that LLS is not allowed in this case. iOS displays this:

Image

#### What happened instead?
Android keeps displaying:

Image

### Your phone model

_No response_

### Operating system version

_No response_

Galaxy A50

### Homeserver

12

### Will you send logs?

No

No

ganfra:
## Content

As the title says, we should at least try to request the permissions once (if permissions are not yet granted of course).

## Motivation and context
Closes https://github.com/element-hq/element-x-android/issues/6850

## Tests

- Reset permissions of the app 
- Open the Share location screen 
- Play with the permission state (kill app, go back, etc etc)

## Tested devices

- [X] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,5 +1,39 @@
 <component name="ProjectCodeStyleConfiguration">
   <code_scheme name="Project" version="173">
+    <JavaCodeStyleSettings>
+      <option name="IMPORT_LAYOUT_TABLE">
+        <value>
+          <package name="android" withSubpackages="true" static="true" />
+          <package name="androidx" withSubpackages="true" static="true" />
+          <package name="com" withSubpackages="true" static="true" />
+          <package name="junit" withSubpackages="true" static="true" />
+          <package name="net" withSubpackages="true" static="true" />
+          <package name="org" withSubpackages="true" static="true" />
+          <package name="java" withSubpackages="true" static="true" />
+          <package name="javax" withSubpackages="true" static="true" />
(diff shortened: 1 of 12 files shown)


.idea/codeStyles/Project.xml
```diff
@@ -1,5 +1,39 @@
...
+          <emptyLine />
+          <package name="java" withSubpackages="true" static="false" />
+          <emptyLine />
+          <package name="javax" withSubpackages="true" static="false" />
+          <emptyLine />
+          <package name="" withSubpackages="true" static="false" />
+          <emptyLine />
+        </value>
+      </option>
+    </JavaCodeStyleSettings>
```
bmarty: I have the same problem, this file is updated on my side. But not on @jmartinesp env, so we should not update it, I guess...

bmarty (approved):
Tested OK, the permissions is kept even when set to "Ask every time" and the app is killed, but it may depend on OS I guess.

.idea/codeStyles/Project.xml
ganfra: Yeah...reverted

PR #7012, opened 2026-06-11
build-rust-sdk: allow skipping git update for kotlin components

richvdh:
Updating matrix-rust-components-kotlin from git on every build is time-consuming, and requires network connectivity. It's useful to be able to skip it, particularly when working offline.

Add an option to the build script to skip this step.


jmartinesp (approved):
Thanks!

PR #7017, opened 2026-06-12
Fix ANRs when calculating DB sizes for Sentry

jmartinesp:
## Content

Instead of using the SDK's `Client.getStoreSizes()` method, manually check the sizes using the file system.

For this, the `SessionPaths` need to be synchronously accessible from a `MatrixClient`. I removed `SessionPathsProvider` too since it's not needed anymore.

## Motivation and context

What the title says: fix most of the ANRs in the latest version of the app, caused by using `runBlocking` for fetching the DB sizes from the Rust SDK to be uploaded as extra data for Sentry so we can check if some performance issues can be related to large DBs.

## Tests

It's difficult to test, unless you have access to the Sentry analytics 😅 .

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.


jmartinesp:
Alternatively, we could have `RustMatrixClient` load the session paths on `init` on a coroutine and have `sessionPaths` be optional in `MatrixClient`, or behind some (sync) function.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/paths/SessionPaths.kt
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (c) 2026 Element Creations Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.matrix.api.paths
```
bmarty, line 8: I do not understand why this class is now in the `api` module. It is a mistake?

jmartinesp: No, it's the fastest way to expose the paths to the DB sizes use case.

jmartinesp: We need some way to access those paths synchronously from at least that use case. So either I expose them using this data class or I have to create some other one that contains them. And I also need to access them in a component that's not bound to the session, so just creating some annotation for them and adding them separately for `SessionScope` won't work for this either 🫤 .

bmarty: OK, thanks for clarifying.

bmarty (approved):

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/paths/SessionPaths.kt
jmartinesp: No worries, I know this code isn't great, but given our current architecture I didn't see an easy way to improve it.

PR #6068, opened 2026-01-22
Draft: Remove white background from stickers/images

jmartinesp:
## Content

- Remove unwanted white/blur background behind stickers and images in timeline items once media is loaded.
- Keep blurhash placeholder behavior only while loading.
- Resolve merge conflicts by merging `origin/develop` into this branch and reconciling `TimelineItemImageView` plus related UI snapshot files.

## Motivation and context

This PR fixes visual artifacts (white background) on stickers/images and updates the branch to remain compatible with the latest `develop` changes after conflict resolution feedback.

## Screenshots / GIFs

UI snapshot files were updated as part of the change and conflict resolution.

## Tests

- Attempted to run module checks for `:features:messages:impl`.
- In this environment, Gradle could not resolve Android plugin `com.android.application:8.13.2`, so local verification could not be completed end-to-end.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

- [ ] I am aware of the [etiquette](https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#etiquette).
- This PR was made with the help of AI:
    - [x] Yes. In this case, please request a review by Copilot.
    - [ ] No.
- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly defines what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [ ] You've made a self review of your PR

PR #4595, opened 2025-04-15
Send SVG images as files

jmartinesp:
## Content

Android doesn't have support for SVG files so we can't create a thumbnail or get the dimensions and other metadata needed for the `m.image` message type, so we need to send them as plain files.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/4593 (partially).

## Tests

Download any SVG file, send it in some room.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 14

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly define what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #7020, opened 2026-06-14
Add full SVG support

bxdxnn:
## Content

Support SVG throughout the app (avatars, timeline, media, etc.)

Needs https://github.com/matrix-org/matrix-rust-sdk/pull/6662

1. Revert #4595
￼2. Wire Coil's `SvgDecoder` to media loader
3. Fallback to full content fetching in case a thumbnail doesn't exist
4. Fetch a bit more metadata from the SVG when sending it (thumbnail still won't be set because SVGs are commonly light in size like normal messages)

## Motivation and context

There was poor to no support before.

## Tests

- Send SVG (needs https://github.com/matrix-org/matrix-rust-sdk/pull/6662)
- See SVG

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

    - [ ] No.
- [x] UI change has been tested on both light and dark themes (related: #6068)

### DIFF
--- a/libraries/matrixmedia/impl/src/main/kotlin/io/element/android/libraries/matrix/ui/media/CoilMediaFetcher.kt
+++ b/libraries/matrixmedia/impl/src/main/kotlin/io/element/android/libraries/matrix/ui/media/CoilMediaFetcher.kt
@@ -36,17 +36,16 @@ internal class CoilMediaFetcher(
         return when (val kind = mediaData.kind) {
             is MediaRequestData.Kind.Content -> fetchContent(mediaSource)
             is MediaRequestData.Kind.Thumbnail -> {
-                val thumbnailResult = fetchThumbnail(mediaSource, kind)
-                if (thumbnailResult.isSuccess) {
-                    thumbnailResult.getOrThrow()
-                } else {
-                    val error = thumbnailResult.exceptionOrNull()
-                    if (error?.isNetworkError() == true) {
-                        null
-                    } else {
-                        fetchContent(mediaSource)
+                fetchThumbnail(mediaSource, kind).fold(
+                    onSuccess = { it },
+                    onFailure = { error ->
+                        if (error.isNetworkError()) {
+                            null
+                        } else {
+                            fetchContent(mediaSource)
+                        }
                     }
-                }
+                )
             }
             is MediaRequestData.Kind.File -> fetchFile(mediaSource, kind)
         }
--- a/libraries/matrixmedia/impl/src/main/kotlin/io/element/android/libraries/matrix/ui/media/CoilMediaFetcher.kt
+++ b/libraries/matrixmedia/impl/src/main/kotlin/io/element/android/libraries/matrix/ui/media/CoilMediaFetcher.kt
@@ -13,6 +13,7 @@ import coil3.decode.ImageSource
 import coil3.fetch.FetchResult
 import coil3.fetch.Fetcher
 import coil3.fetch.SourceFetchResult
+import io.element.android.libraries.matrix.api.exception.isNetworkError
 import io.element.android.libraries.matrix.api.media.MatrixMediaLoader
 import io.element.android.libraries.matrix.api.media.MediaSource
 import io.element.android.libraries.matrix.api.media.toFile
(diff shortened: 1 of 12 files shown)


libraries/matrixmedia/impl/src/main/kotlin/io/element/android/libraries/matrix/ui/media/CoilMediaFetcher.kt
```diff
@@ -34,7 +35,19 @@ internal class CoilMediaFetcher(
...
+                    thumbnailResult.getOrThrow()
+                } else {
+                    val error = thumbnailResult.exceptionOrNull()
+                    if (error?.isNetworkError() == true) {
+                        null
+                    } else {
+                        fetchContent(mediaSource)
+                    }
+                }
+            }
```
bmarty, line 49: A better way to write this is:

```kotlin
            is MediaRequestData.Kind.Thumbnail -> {
                fetchThumbnail(mediaSource, kind).fold(
                    onSuccess = { it },
                    onFailure = { error ->
                        if (error.isNetworkError()) {
                            null
                        } else {
                            fetchContent(mediaSource)
                        }
                    }
                )
            }
```

libraries/matrixmedia/impl/src/main/kotlin/io/element/android/libraries/matrix/ui/media/ImageLoaderFactories.kt
```diff
@@ -41,6 +42,7 @@ class DefaultImageLoaderFactory(
         return ImageLoader.Builder(context)
             .components {
                 add(okHttpNetworkFetcherFactory)
+                add(SvgDecoder.Factory())
```
bmarty: I do not think that we need the Svg decoder for the `NotLoggedInFlowNode`, this can be removed.

libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/AndroidMediaPreProcessor.kt
```diff
@@ -83,10 +88,6 @@ class AndroidMediaPreProcessor(
     ): Result<MediaUploadInfo> = withContext(coroutineDispatchers.computation) {
         runCatchingExceptions {
             val result = when {
-                // Special case for SVG, since Android can't read its metadata or create a thumbnail, it must be sent as a file
-                mimeType == MimeTypes.Svg -> {
-                    processFile(uri, mimeType)
```
bmarty, line 88: I would keep this first check and call `processSvgImage(uri, mimeType)` here for code simplicity. Also the change on `notCompressibleImageTypes` is not strictly necessary.

libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/AndroidMediaPreProcessor.kt
```diff
@@ -220,6 +226,76 @@ class AndroidMediaPreProcessor(
...
+            blurhash = null,
+        )
+        return MediaUploadInfo.Image(
+            file = file,
+            imageInfo = imageInfo,
+            thumbnailFile = null,
+        )
+    }
+
+    private fun extractSvgDimensions(file: File): Pair<Long, Long> {
```
bmarty: Maybe extract this function to a dedicated class?

bmarty (changes requested):
Thanks, a few remarks.

libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/SvgDimensionExtractor.kt
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (c) 2025 Element Creations Ltd.
+ * Copyright 2023-2025 New Vector Ltd.
```
bmarty: One line with 2026 is enough for new files.

libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/SvgDimensionExtractor.kt
```diff
@@ -0,0 +1,68 @@
...
+
+import android.util.Xml
+import org.xmlpull.v1.XmlPullParser
+import java.io.File
+
+class SvgDimensionExtractor(
+    private val defaultWidth: Long = 640L,
+    private val defaultHeight: Long = 480L,
+) {
+    fun extractDimensions(file: File): Pair<Long, Long> {
```
bmarty: maybe cleaner to return `android.util.Size` instead of `Pair<Long, Long>`.

bmarty (commented):
Thanks for the update. 2 more remarks then I think that we're good to merge this change.

libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/SvgDimensionExtractor.kt
bxdxnn: Seems like more code just for the type name IMO, but OK.

bmarty (approved):
Thanks!

PR #7034, opened 2026-06-16
Fix thread list not loading older threads

bxdxnn:
## Content

More simple condition to trigger pagination

## Motivation and context

Fix thread list pagination not working sometimes.
With the previous behavior, I had to scroll back sometimes to make it load further.

## Tests

- Scroll thread list and check if it loads to the oldest thread

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/threads/list/ThreadsListView.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/threads/list/ThreadsListView.kt
@@ -162,14 +162,13 @@ private fun ScrollHelper(
     val lastVisibleItemIndex by remember {
         derivedStateOf { listState.firstVisibleItemIndex + listState.layoutInfo.visibleItemsInfo.size - 1 }
     }
-    val needsPagination by remember {
+    val shouldPaginate by remember {
         derivedStateOf {
-            val canLoadNewItems = listState.isScrollInProgress || listState.firstVisibleItemScrollOffset == 0
-            canLoadNewItems && lastVisibleItemIndex == listState.layoutInfo.totalItemsCount - 1
+            lastVisibleItemIndex >= listState.layoutInfo.totalItemsCount - 1
         }
     }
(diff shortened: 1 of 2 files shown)


bmarty (approved):
LGTM, @jmartinesp can you have a second look please?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/threads/list/ThreadsListView.kt
```diff
@@ -162,14 +162,13 @@ private fun ScrollHelper(
     val lastVisibleItemIndex by remember {
         derivedStateOf { listState.firstVisibleItemIndex + listState.layoutInfo.visibleItemsInfo.size - 1 }
     }
-    val needsPagination by remember {
+    val shouldPaginate by remember {
         derivedStateOf {
-            val canLoadNewItems = listState.isScrollInProgress || listState.firstVisibleItemScrollOffset == 0
-            canLoadNewItems && lastVisibleItemIndex == listState.layoutInfo.totalItemsCount - 1
```
jmartinesp, line 168: Maybe this could be:

```kotlin
val canLoadNewItems = listState.isScrollInProgress || listState.layoutInfo.totalItemsCount == 0
            canLoadNewItems && lastVisibleItemIndex >= listState.layoutInfo.totalItemsCount - 1
```

jmartinesp (approved):
I think the check was kind of useful even if wrong to avoid trying to trigger more paginations than needed, but since I think the SDK has a check to avoid performing multiple paginations maybe we can simplify it as you did.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/threads/list/ThreadsListView.kt
bxdxnn: Also works I guess

PR #7070, opened 2026-06-21
Fix Threads list stuttering

bxdxnn:
## Content

Optimize Threads list:
* Paginate using the diffs from SDK so that we won't call `timelineItemContentFactory.create()` on every row again when paginating
* Move item processing out of the main thread to fix animation stuttering
* Minor fix in `ThreadsListView.kt` to fix the effect restarting on every scroll frame
* ~Check the Threads availability in the view so that we can show the top bar icon instantly by just checking the feature flag, then show a hint text in the view if there are no threads available. Also improves UX/discoverability (when we roll-out this and the whole Threads feature) IMO, though maybe needs a product decision, but since this is behind a feature flag I guess it's OK.~

## Motivation and context

Fix the stuttering in the Threads list on opening/scrolling, along with other delays (huge especially on slow network/device).

## Tests

- Open Threads list and scroll it, should be much smoother now.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/room/threads/ThreadsListService.kt
```diff
@@ -11,6 +11,7 @@ import kotlinx.coroutines.flow.Flow
 
 interface ThreadsListService {
     fun subscribeToItemUpdates(): Flow<List<ThreadListItem>>
+    fun subscribeToItemDiffs(): Flow<List<ThreadListDiff>>
```
bmarty, line 13: `subscribeToItemUpdates` is not used anymore, we could remove it.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/threads/list/ThreadsListPresenter.kt
```diff
@@ -34,53 +41,53 @@ class ThreadsListPresenter(
...
-                        val latestTimelineEvent = item.latestEvent?.content?.let {
-                            timelineItemContentFactory.create(
-                                itemContent = it,
-                                eventId = item.latestEvent!!.eventId,
-                                isEditable = false,
-                                sender = item.latestEvent!!.senderId,
-                                senderProfile = item.latestEvent!!.senderProfile,
-                            )
+                .collect { diffs ->
+                    Timber.d("Received thread list diffs: $diffs")
```
bmarty: This log will contain the last messages content, we should not do that. Please remove the log, or log only technical data.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/threads/list/ThreadsListView.kt
```diff
@@ -325,6 +371,27 @@ internal fun ThreadsListViewPreview() {
...
+            ),
+            onThreadClick = {},
+            onBackClick = {},
+        )
+    }
+}
+
+@PreviewsDayNight
+@Composable
+internal fun ThreadsListViewEmptyPreview() {
```
bmarty: You should rather introduce a new class `ThreadsListStateProvider : PreviewParameterProvider<ThreadsListState>` and use it in the existing preview `ThreadsListViewPreview`.

Additionally, `ThreadListItemRowPreview` can be removed (can be in a separate PR as it's not related to this change).

bmarty (changes requested):
Thanks for the PR!

I have tested the feature and it's working well.

For the new empty view, I let @mxandreas / @americanrefugee  confirm that this is what we want, or provide an alternative design / wording if necessary. The threads icon will be always displayed and in case there are no threads in the room, this view will be displayed:

Image

Thanks!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/threads/list/ThreadsListPresenter.kt
```diff
@@ -34,53 +41,53 @@ class ThreadsListPresenter(
...
     @Composable
     override fun present(): ThreadsListState {
         val coroutineScope = rememberCoroutineScope()
         val threadsListService = room.threadsListService
 
+        var hasLoadedFirstPage by remember { mutableStateOf(false) }
+
         val threads by produceState(initialValue = persistentListOf(), key1 = threadsListService) {
-            threadsListService.subscribeToItemUpdates()
+            var isFirstDiff = true
```
jmartinesp: You probably don't need this and can always set `hasLoadedFirstPage` to true instead: the `MutableState` will do an equality check before propagating changes, so `true -> true` does nothing.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
```diff
@@ -323,7 +313,7 @@ class MessagesPresenter(
             topBarSharedHistoryIcon = topBarSharedHistoryIcon,
             successorRoom = roomInfo.successorRoom,
             threads = Threads(
-                hasThreads = canOpenThreadList && threadsList.isNotEmpty(),
+                hasThreads = canOpenThreadList,
```
jmartinesp: Do we really want to display a button for an empty screen? I think we agreed with design to only display the thread list button when it's useful.

jmartinesp (commented):
A couple of comments about the changes. Also, if we want to have an empty screen, it might make sense to use an `AsyncData` wrapper for the thread list, and check whether its `dataOrNull()` value contains something instead of creating a new variable and exposing it in the state. There are several examples of this in the app.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/threads/list/ThreadsListPresenter.kt
jmartinesp: Note you can remove this if you use the `AsyncData` wrapper as suggested above.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesPresenter.kt
ganfra: Yeah I had the same in mind

bxdxnn: If that's really not wanted, I can try making a "threads exist in the room" check in the SDK with cache. But I'll wait for the product decision.

americanrefugee: Agreed - only display the tread list button if there are threads.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/threads/list/ThreadsListPresenter.kt
bxdxnn: Will address after the product decision.

bxdxnn:
I've reverted the delay optimization as it's too complicated to do it properly ATM, something like https://github.com/matrix-org/matrix-spec-proposals/pull/4360 is needed for it to work fast and correct but it seems to be WIP (?) and doesn't have an implementation yet.

bxdxnn:
@bmarty FYI product/deisgn decision isn't required anymore as I've reverted the new screen/UX thing.

bxdxnn:
Test failures doesn't seems to be related...

jmartinesp:
@bmarty should we try to unblock this PR by taking another look?

PR #7090, opened 2026-06-23
Fix the message edit being lost when attaching media

### LINKED ISSUE #7083: If you need to edit (add a photo) to your previous message
### Steps to reproduce

If you need to edit (add a photo) to your previous message, it is not edited, but a new one is created

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

Yes

Yes

manfrommedan:
## Content

Starting an edit and then attaching media (photo, video, file) used to drop the edit. The attachment is sent as a new message, which is expected since a text message can't be replaced by media, but the edit itself was lost, so sending afterwards posted the leftover text as another new message instead of editing the original.

`handlePickedMedia` now leaves the composer mode untouched while you're editing, so the edit stays active and the text still edits the original message on the next send. Reply mode keeps being cleared, since the attachment becomes the reply.

## Motivation and context

Fixes #7083.

## Screenshots / GIFs

No UI change, this only affects composer state handling.

## Tests

- Edit a message, attach a photo, send the photo, then send the text: the original message is edited instead of a new one being created.
- Unit tests in `MessageComposerPresenterTest` cover the edit being preserved across a media pick, and a reply still being cleared.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 16

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
@@ -288,8 +288,7 @@ class MessageComposerPresenter(
                         inReplyToEventId = inReplyToEventId,
                     )
 
-                    // Reset composer since the attachment has been sent
-                    messageComposerContext.composerMode = MessageComposerMode.Normal
+                    resetComposerModeAfterAttaching()
                 }
                 is MessageComposerEvent.SetMode -> {
                     localCoroutineScope.setMode(event.composerMode, markdownTextEditorState, richTextEditorState)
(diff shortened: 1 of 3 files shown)


bmarty (commented):
Thanks. This look good to me, but I think that it will have to be updated when #6519 will be merged. I am working on this one now, and I'd prefer to merge #6519 first as it is quite big.

manfrommedan:
Sounds good. One problem worth flagging, and the fix is about 2 minutes.

**Problem:** #6519's new `handlePickedMediaList` resets `composerMode` to `Normal` unconditionally (it only reads the `Reply` mode), exactly like the single-pick `handlePickedMedia` this PR guards. So an in-progress edit is still lost when picking from the gallery, i.e. the #7083 behaviour via multi-pick.

**Proposal:** the same one-line guard, `if (!messageComposerContext.composerMode.isEditing)`. I'll apply it to both methods when I rebase #7090 on top of #6519, or it's trivial to fold straight into #6519 while you're already there.

bmarty:
Hello @manfrommedan , now that #6519 has been merged, can you revisit this PR please?

manfrommedan:
Rebased on #6519. Attaching media no longer drops an in-progress edit: all paths (SendUri, single and multi picker) now go through one reset that clears reply/normal but keeps the edit. Added tests for both, all green.

bmarty (approved):
Thanks!

PR #7092, opened 2026-06-24
Bump enterprise submodule: custom recovery key finishes via snackbar

jennaharris7:
Bump the `enterprise` submodule to pick up the custom recovery key snackbar-finish change (element-hq/element-android-enterprise#28).


jennaharris7:
Superseded — #28 is merged and develop's submodule already includes it (pin `7bbf7db`), so this backward pointer bump is no longer needed.

PR #7112, opened 2026-06-26
Support gallery messages

bmarty:
Follow up for #6519 

Support sending and rendering gallery messages (a message with multiple photos)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
```diff
@@ -123,52 +136,51 @@ class AttachmentsPreviewPresenter(
...
-        LaunchedEffect(
-            mediaOptimizationSelectorState.displayMediaSelectorViews,
-            mediaOptimizationSelectorState.videoSizeEstimations,
-            currentAttachment,
-            imageEditorState,
-            isApplyingImageEdits,
-        ) {
-            // If the media optimization selector is not displayed, we can pre-process the media
-            // to prepare it for sending. This is done to avoid blocking the UI thread when the
-            // user clicks on the send button.
```
jmartinesp, line 135: This comment still seems valuable.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
```diff
@@ -123,52 +136,51 @@ class AttachmentsPreviewPresenter(
...
-            @Suppress("ComplexCondition")
+        LaunchedEffect(mediaOptimizationSelectorState.displayMediaSelectorViews, mediaOptimizationSelectorState.selectedVideoPreset) {
             if (mediaOptimizationSelectorState.displayMediaSelectorViews == false &&
-                preprocessMediaJob == null &&
                 imageEditorState == null &&
-                !isApplyingImageEdits) {
-                if (mediaAttachment.localMedia.info.mimeType.isMimeTypeVideo() && mediaOptimizationSelectorState.videoSizeEstimations.dataOrNull() == null) {
-                    Timber.d("Waiting for video size estimations to be able to select the best video compression preset before pre-processing the media")
-                    return@LaunchedEffect
-                }
```
jmartinesp, line 144: I think removing this may be breaking selecting the highest preset available automatically in the 'per-media' media quality feature. e.g. your video is short, so it fits in the HS max upload size in high quality (1080p) but instead standard (720p) is selected.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
```diff
@@ -192,12 +204,29 @@ class AttachmentsPreviewPresenter(
...
+                                    ?: mediaOptimizationConfigProvider.get().videoCompressionPreset,
                             )
+                            preprocessMediaJob = coroutineScope.launch(dispatchers.io) {
+                                preProcessAttachments(
+                                    attachments = attachmentsAndEdits.map { it.attachment },
+                                    mediaOptimizationConfig = config,
+                                    displayProgress = true,
+                                    sendActionState = sendActionState,
+                                )
+                            }
```
jmartinesp: Why do we need to re-run the pre-processing? Shouldn't this already have happened or be happening and we should just wait for it to finish?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
```diff
@@ -206,35 +235,27 @@ class AttachmentsPreviewPresenter(
...
-                                    dismissAfterSend = false,
-                                    inReplyToEventId = inReplyToEventId,
-                                )
-                            } finally {
-                                editedTempFileToDelete?.safeDelete()
-                                // Clean up the pre-processed media after it's been sent
-                                mediaSender.cleanUp()
-                            }
+                            sendGalleryPreProcessed(
+                                mediaUploadInfos = allMediaUploadInfos,
```
jmartinesp, line 269: I see this is calling a method that can send a single file or a gallery. Should we rename this method?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
```diff
@@ -328,16 +359,21 @@ class AttachmentsPreviewPresenter(
...
-                                appliedImageEdits = pendingState.edits
+                                editedTempFiles[currentIndex]?.safeDelete()
+                                editedTempFiles = editedTempFiles + (currentIndex to editedMedia.file)
                                 currentAttachment = Attachment.Media(editedMedia.localMedia)
+                                attachmentsAndEdits = attachmentsAndEdits.toMutableList().also {
+                                    it[currentIndex] = AttachmentAndEdits(
+                                        currentAttachment,
+                                        pendingState.edits,
+                                    )
+                                }.toImmutableList()
```
jmartinesp, line 395: I think `ImmutableList.replacingAt(index, value)` exists.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
```diff
@@ -352,11 +388,14 @@ class AttachmentsPreviewPresenter(
...
                 }
+                is AttachmentsPreviewEvent.SetCurrentCarouselIndex -> {
+                    currentIndex = event.index
+                }
             }
         }
 
         return AttachmentsPreviewState(
-            attachment = currentAttachment,
+            attachments = attachmentsAndEdits.map { it.attachment }.toImmutableList(),
```
jmartinesp: Do we want to create new lists in every recomposition? Although I'm not sure if having a separate variable we update at the same time is better 🫤 .

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
```diff
@@ -365,92 +404,62 @@ class AttachmentsPreviewPresenter(
...
-
-            MediaOptimizationConfig(
-                compressImages = false,
-                videoCompressionPreset = videoCompressionPreset,
-            )
-        } else {
-            // Otherwise, we just rely on the user preferences for media optimization
-            mediaOptimizationConfigProvider.get()
-        }
-    }
```
jmartinesp, line 392: Oh, so I guess here's my answer to the question above, we're removing the code to find the best suited quality for the media... Can't we make it work with multiple items?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewView.kt
```diff
@@ -77,6 +89,20 @@ import io.element.android.libraries.ui.utils.formatter.rememberFileSizeFormatter
 import io.element.android.wysiwyg.display.TextDisplay
 import kotlinx.collections.immutable.ImmutableList
 import kotlinx.collections.immutable.persistentListOf
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.collectLatest
+import kotlin.time.Duration.Companion.milliseconds
+
+private val SingleItemPreviewRenderer = object : LocalMediaRenderer {
```
jmartinesp: I wasn't sure why this was referencing a sample resource, then I realised it's used only for previews (since the file also contains 'Preview' in its name it wasn't that obvious). Should we add some comments to it, or put it closer to the previews at least?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewView.kt
```diff
@@ -266,16 +293,75 @@ private fun AttachmentPreviewContent(
     ) {
         Box(
             modifier = Modifier
-                .weight(1f),
-            contentAlignment = Alignment.Center
+                .fillMaxSize()
+                .weight(1f)
```
jmartinesp: Aren't these 2 modifiers contradicting?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewView.kt
```diff
@@ -266,16 +293,75 @@ private fun AttachmentPreviewContent(
...
+                            if (isScrolling) {
+                                isPillVisible = true
+                            } else {
+                                delay(2000.milliseconds)
+                                isPillVisible = false
+                            }
+                        }
+                }
+
+                androidx.compose.animation.AnimatedVisibility(
```
jmartinesp, line 323: Import?

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewView.kt
```diff
@@ -266,16 +293,75 @@ private fun AttachmentPreviewContent(
...
+                val firstAttachment = state.attachments.first()
+                when (firstAttachment) {
+                    is Attachment.Media -> {
+                        localMediaRenderer.Render(firstAttachment.localMedia)
+                    }
                 }
             }
         }
-        val mediaInfo = (state.attachment as? Attachment.Media)?.localMedia?.info
+        val mediaInfo = (state.attachments.first() as? Attachment.Media)?.localMedia?.info
```
jmartinesp: We only add the selector UI for the first item?

jmartinesp (commented):
I've reviewed the first few classes and I've found some issues, mainly with the 'per-item media quality selector' feature, it disables some optimizations and they don't seem to work well together overall.

I'll continue testing and reviewing tomorrow.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/event/TimelineItemGalleryContent.kt
```diff
@@ -0,0 +1,66 @@
...
+    val blurhash: String?,
+    val duration: Duration = Duration.ZERO,
+) {
+    enum class Type {
+        Image,
+        Video,
+        Audio,
+        File;
+
+        fun isMedia() = this in setOf(Image, Video)
```
jmartinesp: `isVisualMedia`? All of the types are media.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/messagesummary/DefaultMessageSummaryFormatter.kt
```diff
@@ -56,6 +58,16 @@ class DefaultMessageSummaryFormatter(
             is TimelineItemVideoContent -> context.getString(CommonStrings.common_video)
             is TimelineItemFileContent -> context.getString(CommonStrings.common_file)
             is TimelineItemAudioContent -> context.getString(CommonStrings.common_audio)
+            is TimelineItemGalleryContent -> context.getString(CommonStrings.common_gallery)
+            is TimelineItemAttachmentsContent -> {
+                val count = content.attachments.size
+                val extensions = content.attachments.take(3).joinToString { it.fileExtension }
```
jmartinesp, line 64: Should this use `.distinct()` or do we want to display something like `jpg, jpg, png`?.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/media/FormattedBody.kt
```diff
@@ -0,0 +1,24 @@
...
+
+fun FormattedBody.map(): RustFormattedBody = RustFormattedBody(
+    format = format.map(),
+    body = body,
+)
+
+private fun io.element.android.libraries.matrix.api.timeline.item.event.MessageFormat.map(): RustMessageFormat {
+    return when (this) {
+        io.element.android.libraries.matrix.api.timeline.item.event.MessageFormat.HTML -> RustMessageFormat.Html
+        io.element.android.libraries.matrix.api.timeline.item.event.MessageFormat.UNKNOWN -> RustMessageFormat.Unknown("")
```
jmartinesp: Imports?

libraries/matrix/impl/src/test/kotlin/io/element/android/libraries/matrix/impl/timeline/item/event/EventMessageMapperTest.kt
```diff
@@ -0,0 +1,318 @@
...
+import org.junit.Test
+import org.matrix.rustcomponents.sdk.EmoteMessageContent
+import org.matrix.rustcomponents.sdk.LocationContent
+import org.matrix.rustcomponents.sdk.MessageType
+import org.matrix.rustcomponents.sdk.NoticeMessageContent
+import org.matrix.rustcomponents.sdk.TextMessageContent
+import org.matrix.rustcomponents.sdk.UnstableVoiceContent
+import org.matrix.rustcomponents.sdk.AssetType as RustAssetType
+
+class EventMessageMapperTest {
```
jmartinesp, line 45: Oh, I didn't realise we didn't have a test for this mapping. Thanks for adding it!

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/datasource/TimelineMediaItemsFactory.kt
```diff
@@ -32,58 +30,51 @@ class TimelineMediaItemsFactory(
...
-        timelineItems: List<MatrixTimelineItem>,
-    ) {
-        val newTimelineItemStates = ArrayList<MediaItem>()
-        for (index in diffCache.indices().reversed()) {
-            val cacheItem = diffCache.get(index)
-            if (cacheItem == null) {
-                buildAndCacheItem(timelineItems, index)?.also { timelineItemState ->
-                    newTimelineItemStates.add(timelineItemState)
+            val newTimelineItemStates = ArrayList<MediaItem>()
+            for (index in timelineItems.indices.reversed()) {
```
jmartinesp, line 43: Why reversed? Is this done to keep the pagination direction?

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/datasource/TimelineMediaItemsFactory.kt
```diff
@@ -32,58 +30,51 @@ class TimelineMediaItemsFactory(
...
-        diffCache[index] = timelineItem
-        return timelineItem
+    private fun isItemUnchanged(
+        currentTimelineItem: MatrixTimelineItem.Event,
+        previousItems: List<MatrixTimelineItem>,
+    ): Boolean {
+        if (previousItems.isEmpty()) return false
+        val previousIndex = previousItems.indexOfFirst { it is MatrixTimelineItem.Event && it.uniqueId == currentTimelineItem.uniqueId }
+        if (previousIndex < 0) return false
+        val previousItem = previousItems[previousIndex] as? MatrixTimelineItem.Event ?: return false
```
jmartinesp: Could this just use `find`? Do we need the index?

jmartinesp (commented):
Ok, here's another round of comments. I may have missed some stuff since the PR is huge, but I think once the issues with the per-item media quality feature are fixed, it should be mergeable.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
bmarty: Renamed to `sendMedia`

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
bmarty: I have introduced a derived state `editedAttachments`

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewView.kt
bmarty: Yes, it was not clear/ideal. What I have done: extract SingleItemPreviewRenderer, make it a class and rename it to SampleMediaRenderer

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewView.kt
bmarty: I have restored the previous code

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewView.kt
bmarty: Not sure, since we are in a ColumnScope, it will use the `ColumnScope.AnimatedVisibility` function.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
bmarty: I cannot find such method

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
bmarty: It was not happening and the application was stuck when the user clicked the send buton. I will check the code again

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/messagesummary/DefaultMessageSummaryFormatter.kt
bmarty: Good remark, I guess we can revisit this later.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/media/FormattedBody.kt
bmarty: I have removed this dead code.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
jmartinesp: Ah, it's in `PersistentList` not `ImmutableList`, so you need to make `attachmentsAndEdits` have a `.toPersistentList()` at the end for this to work (having it as immutable is recommended anyway).

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
jmartinesp: Maybe some of the changes broke something related to this, or maybe there's some case we didn't realise could cause this pre-processing to be cancelled 🤔 .

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewView.kt
jmartinesp: Ah, I see, there are 2 contradicting methods. I think prefixing `this@Column.AnimatedVibility` fixes it, but we can also keep the FQN.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
bmarty: Restored.

bmarty: Restored

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
bmarty: Restored

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewView.kt
bmarty: I am making a selector for each media

bmarty:
@jmartinesp I have pushed a few commits, but I'll do more user tests on my side.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/SampleMediaRenderer.kt
```diff
@@ -0,0 +1,32 @@
...
+import androidx.compose.ui.res.painterResource
+import io.element.android.libraries.designsystem.utils.CommonDrawables
+import io.element.android.libraries.mediaviewer.api.local.LocalMedia
+import io.element.android.libraries.mediaviewer.api.local.LocalMediaRenderer
+
+/**
+ * An implementation of [LocalMediaRenderer] that displays a sample background image.
+ * To be used for Previews only.
+ */
+class SampleMediaRenderer : LocalMediaRenderer {
```
jmartinesp: Nit: `internal class`?

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/datasource/TimelineMediaItemsFactory.kt
```diff
@@ -65,16 +65,15 @@ class TimelineMediaItemsFactory(
...
-        val previousItem = previousItems[previousIndex] as? MatrixTimelineItem.Event ?: return false
-        return previousItem.event.eventId == currentTimelineItem.event.eventId &&
-               previousItem.event.timestamp == currentTimelineItem.event.timestamp
-    }
+private fun MatrixTimelineItem.Event.isUnchanged(
+    previousItems: List<MatrixTimelineItem>,
+): Boolean {
+    val previousItem = previousItems
+        .filterIsInstance<MatrixTimelineItem.Event>()
+        .find { it.uniqueId == uniqueId }
```
jmartinesp: I *think* directly using `.find { it is MatrixTimelineItem.Event && it.uniqueId == uniqueId}` would be slightly more performant, since the `filterIsInstance` above will probably filter all the items first, then use `find` in the results.

jmartinesp (approved):
It seems to work pretty well now, thanks! Great job making it work with the per-item media quality. I added a couple of nits, but nothing important.

My main concern is that some other clients (like Element Web) don't seem to support this event type and there is no fallback version. I guess that's why it's behind a FF.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/datasource/TimelineMediaItemsFactory.kt
bmarty: Updated. I use the first approach to avoid casting the result, but OK.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/SampleMediaRenderer.kt
bmarty: updated (this class is in the `impl` module, so it's already sort of internal by architecture).

bmarty:
Thanks for the review @jmartinesp . I have pushed a fix and take your remark into account.

I am wondering when `sendAsFile` is true if you should display the media quality selection. This is a bit out of scope of this PR, just sharing this point.

PR #7134, opened 2026-07-01
Add arrow buttons to filter carousel for single-pointer navigation

### LINKED ISSUE #6402: A11y: [Chats] Filter carousel operable only via swipe gestures....
- **Checkpoint ID:** 11.2.5.1/4.1
- **Component:** Chats

### Issue Description
Filter carousel operable only via swipe gestures.

### Recommendation
Provide single-pointer (clickable) arrows for carousel navigation.

--- 
*Report Date: 25.02.2026*

bmarty:
## Summary

- Wraps the `LazyRow` of room list filters in an outer `Row` with left/right `IconButton` arrow controls (fixes #6402)
- Arrow buttons appear when scroll is possible (`canScrollBackward` / `canScrollForward`) and scroll by 300dp with `animateScrollBy`
- New strings `a11y_filter_carousel_scroll_left` / `a11y_filter_carousel_scroll_right` added to `temporary.xml`

## Test plan

- [ ] With filters shown in the Chats screen, verify left/right arrow buttons appear when the carousel can be scrolled
- [ ] Verify the arrows work with a single pointer (touch or mouse) and with TalkBack
- [ ] Verify the arrows disappear when the carousel is fully scrolled to each end

### DIFF
--- a/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/filters/RoomListFiltersView.kt
+++ b/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/filters/RoomListFiltersView.kt
@@ -17,6 +17,7 @@ import androidx.compose.foundation.gestures.animateScrollBy
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
@@ -32,6 +33,7 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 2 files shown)


bmarty:
In discussion with @gaelledel , let's revisit when we will have a decision about this.

gaelledel:
This is an issue for both of our apps Android + iOS

This suggestion adds arrow buttons on the chat filters.  This is what the assessor is recommending in order to make the app more accessible for the case of users who can't use swipe gesture.
image (8)

Whilst I understand the assessor's recommendation, this is not something I am keen on including in our apps as this is a visually invasive solution that also deteriorates typical/gesture based consumption of the area for the majority of users. The area for gesture and selection of filters becomes very constrained and this is going to be amplified on smaller width screens. By design, our filter chips auto scroll when selected which reduces the need for swipe gesture.
Additionally, this is not a pattern that I have observed in competitor apps whether direct or non direct. i.e: Slack, teams, Whatsapp, Youtube (to name a few)

So rather than adding anything on the UI, I would first check that any Assistive technology functions properly:

1/VoiceOver/TalkBack 
Can we check we can reach all chips via native scroll-into-view behaviour — this might already close most of the gap.
i.e: Does swiping right with one finger (the standard screen-reader navigation gesture, distinct from a normal two-finger-free swipe) move focus through each filter chip one at a time? 

2/ External hardware
Fix/verify switch-control and keyboard tab order includes off-screen chips.

3/ Voice control: can a user say "tap People" and have it work, even if "People" isn't currently visible?

Then when these checks are done I can go back to the assessor.

For further consideration:
It may be possible to display an alternate control when apps detect that an assistive tech is Active (reduced motion, VoiceOver running, switch control on). 
This would need mobile engineers to research @bmarty @Velin92 
This way the default gesture-first experience stays untouched for most users, and we can think of an alternative visual affordance for the users who need it (not necessarily the arrow system above). If we see it is possible, I can do more design research on how to do it.

FYI @daniellekirkwood  @amshakal @americanrefugee @manuroe

PR #7139, opened 2026-07-01
Initial content scanner integration

### LINKED ISSUE #7122: [Task] Enable the content scanner in the SDK based on the well-known contents

jmartinesp:
## Content

Create the base APIs to be able to use an optional content scanner for the media in the app:

- Create `:features:contentscanner` modules.
- Create a `ContentScanner` wrapper for the `RustContentScanner` one.
- Create a `ContentScannerUrlProvider` interface and implement a noop version for the FOSS app so we don't instantiate a content scanner in that case.
- Add the shared code to use a provided content scanner (coming from the enterprise code) to set it up in the SDK and expose it in the `@SessionScope`.
- Split the code for fetching the Element well-known and caching it so the caching part can be used by other components - it will be used by enterprise.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/7122.

## Tests

Check with FOSS that everything works as before.

With Pro, use the counterpart branch to check the element-well known file is fetched. Sadly, we have no example HS with this value, so the most you can probably do is using `Proxyman` or some other MITM proxy to override the response for that endpoint and point to a content scanner instance you have running somewhere. Details on how to run a content scanner server are detailed in the parent story.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.

### DIFF
--- a/features/contentscanner/api/build.gradle.kts
+++ b/features/contentscanner/api/build.gradle.kts
@@ -1,0 +1,23 @@
+plugins {
+    id("io.element.android-compose-library")
+    id("kotlin-parcelize")
+}
+
+android {
+    namespace = "io.element.android.features.contentscanner.impl"
+}
+
+dependencies {
+    implementation(projects.libraries.core)
+    implementation(projects.libraries.architecture)
+    implementation(projects.libraries.matrix.api)
+    implementation(projects.libraries.matrixui)
+    implementation(projects.libraries.designsystem)
+
+    testImplementation(libs.test.junit)
+    testImplementation(libs.coroutines.test)
+    testImplementation(libs.molecule.runtime)
+    testImplementation(libs.test.truth)
+    testImplementation(libs.test.turbine)
+    testImplementation(projects.libraries.matrix.test)
+}
--- a/features/contentscanner/api/build.gradle.kts
+++ b/features/contentscanner/api/build.gradle.kts
@@ -4,7 +4,7 @@ plugins {
 }
 
 android {
-    namespace = "io.element.android.features.contentscanner.impl"
+    namespace = "io.element.android.features.contentscanner.api"
 }
 
 dependencies {
--- a/features/contentscanner/impl/build.gradle.kts
+++ b/features/contentscanner/impl/build.gradle.kts
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2026 Element Creations Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+import extension.setupDependencyInjection
+
+plugins {
+    id("io.element.android-compose-library")
+    id("kotlin-parcelize")
+}
+
+android {
+    namespace = "io.element.android.features.contentscanner.api"
+}
+
+setupDependencyInjection()
+
+dependencies {
+    api(projects.features.contentscanner.api)
+    implementation(projects.libraries.core)
+    implementation(projects.libraries.architecture)
+    implementation(projects.libraries.matrix.api)
+    implementation(projects.libraries.matrixui)
+    implementation(projects.libraries.designsystem)
+
+    testImplementation(libs.test.junit)
+    testImplementation(libs.coroutines.test)
+    testImplementation(libs.molecule.runtime)
+    testImplementation(libs.test.truth)
+    testImplementation(libs.test.turbine)
+    testImplementation(projects.libraries.matrix.test)
+}
--- a/features/contentscanner/impl/build.gradle.kts
+++ b/features/contentscanner/impl/build.gradle.kts
@@ -13,7 +13,7 @@ plugins {
 }
 
 android {
-    namespace = "io.element.android.features.contentscanner.api"
+    namespace = "io.element.android.features.contentscanner.impl"
 }
 
 setupDependencyInjection()
--- a/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/scanner/ContentScanner.kt
+++ b/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/scanner/ContentScanner.kt
@@ -1,0 +1,24 @@
+/*
+ * Copyright (c) 2026 Element Creations Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.matrix.api.scanner
+
+import io.element.android.libraries.matrix.api.media.MediaSource
+
+/**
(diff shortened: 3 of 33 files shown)


features/contentscanner/api/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
+plugins {
+    id("io.element.android-compose-library")
+    id("kotlin-parcelize")
+}
+
+android {
+    namespace = "io.element.android.features.contentscanner.impl"
```
bmarty:
```suggestion
    namespace = "io.element.android.features.contentscanner.api"
```

Also this module seems to be empty?

features/contentscanner/impl/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+
+import extension.setupDependencyInjection
+
+plugins {
+    id("io.element.android-compose-library")
+    id("kotlin-parcelize")
+}
+
+android {
+    namespace = "io.element.android.features.contentscanner.api"
```
bmarty:
```suggestion
    namespace = "io.element.android.features.contentscanner.impl"
```

Also this module seems to be empty?

libraries/wellknown/impl/src/test/kotlin/io/element/android/libraries/wellknown/impl/DefaultWellknownRetrieverTest.kt
```diff
@@ -0,0 +1,291 @@
...
+        }
+    } else if (response != null) {
+        every { enqueue(any()) } answers {
+            val callback = firstArg<okhttp3.Callback>()
+            callback.onResponse(this@mockk, response)
+        }
+    } else {
+        error("Either response or error must be provided")
+    }
+}
```
bmarty, line 291: We could use `mockwebserver` but it's also OK.

bmarty (commented):
Only small remarks, I let @ganfra do the proper review

libraries/wellknown/impl/src/test/kotlin/io/element/android/libraries/wellknown/impl/DefaultWellknownRetrieverTest.kt
jmartinesp: I just tried using `MockWebServer` and while it works, it's also harder to set up and in some cases I end up having to use a mix of mocking (for spies) and the web server, so I'm not sure it's worth it. If @ganfra thinks we should change this too I can upload those changes.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -113,6 +118,23 @@ class RustMatrixClientFactory(
...
+        // If no content scanner URL is available, the contentScanner will be null.
+        val contentScanner = domainName?.let {
+            contentScannerUrlProvider.getContentScannerUrl(domainName)
+                .getOrNull()
+                ?.let { contentScannerUrl ->
+                    val contentScanner = ContentScanner(contentScannerUrl)
+                    client.setContentScanner(contentScanner)
+                    RustContentScanner(
+                        client = client,
+                        rustScanner = ContentScanner(contentScannerUrl),
```
ganfra: Why are we creating two `ContentScanner` instances?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/scanner/RustContentScanner.kt
```diff
@@ -0,0 +1,45 @@
...
+    val json = this.json
+    return try {
+        if (json != null) {
+            RustMediaSource.fromJson(json)
+        } else {
+            RustMediaSource.fromUrl(safeUrl)
+        }
+    } catch (e: LinkageError) {
+        // Used for tests, since we can't instantiate an actual `RustMediaSource` because the native library can't be loaded
+        if (Class.forName("org.junit.Test") != null) {
```
ganfra: I don't think it can be null, it'll probably throw `ClassNotFoundException` instead?

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/scanner/ContentScanner.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+interface ContentScanner {
+    /**
+     * Scans the given [mediaSource] for potential security risks.
+     *
+     * @param mediaSource The media source to scan.
+     * @return A [Result] containing a [Boolean] indicating whether the content is safe (true) or potentially unsafe (false).
+     * If the scan fails, the [Result] will contain an exception.
+     */
+    suspend fun scan(mediaSource: MediaSource): Result<Boolean>
```
ganfra, line 26: What is the purpose of this over passing the scanner to the `RustClient`? If there is some diff, maybe we should document?

ganfra (commented):
Some remarks

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/scanner/RustContentScanner.kt
jmartinesp: Good catch, thanks! https://github.com/element-hq/element-x-android/pull/7139/commits/0271e0cb98eef1979c54e95bbabbec1193736b3f

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/scanner/ContentScanner.kt
jmartinesp: Done in https://github.com/element-hq/element-x-android/pull/7139/commits/2a1fc35cd4f08d4f8f9bf3b639d5d9b56ea2c200

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
jmartinesp: I rewrote this so many times I didn't realise this was there, sorry. Fixed in https://github.com/element-hq/element-x-android/pull/7139/commits/271e9b18b5bfe1f46128b4b224fdc6a61887deeb

libraries/wellknown/impl/src/test/kotlin/io/element/android/libraries/wellknown/impl/DefaultWellknownRetrieverTest.kt
ganfra: Ok for me like this

ganfra (approved):
Thanks for the changes, LGTM now!

PR #7145, opened 2026-07-02
Feature/per room notification channels

PolynomialDivision:
## Content

Automatically creates a per-room notification channel the first time a room notifies you, grouped under new "Private chats"/"Rooms" categories, so any room can appear as an Android Conversation and be marked Priority.

## Motivation and context

You can prioritize certain chats and Element-X is closer integrated in the Android Conversation system.

## Screenshots / GIFs

image

## Tests

- Install the updated apk
- Go to Settings->Apps->Element-X->Notifications
- Select a Conversation and select it as Priority or similar

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): GrapheneOS

## Checklist

    - [ ] No.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/attachments/preview/AttachmentsPreviewPresenter.kt
@@ -41,6 +41,8 @@ import io.element.android.libraries.core.mimetype.MimeTypes.isMimeTypeVideo
 import io.element.android.libraries.di.annotations.SessionCoroutineScope
 import io.element.android.libraries.matrix.api.core.EventId
 import io.element.android.libraries.matrix.api.permalink.PermalinkBuilder
+import io.element.android.libraries.matrix.api.room.JoinedRoom
+import io.element.android.libraries.matrix.api.room.getDirectRoomMember
 import io.element.android.libraries.matrix.api.timeline.Timeline
 import io.element.android.libraries.mediaupload.api.MediaOptimizationConfig
 import io.element.android.libraries.mediaupload.api.MediaOptimizationConfigProvider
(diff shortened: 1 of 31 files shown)


PolynomialDivision:
can someone may approve the workflow runs? maybe @ganfra ?

jmartinesp:
Thanks for the contribution. I do personally like the the way you've grouped rooms and DMs and added options to create and customise conversations for rooms, but I'm not sure if we want the whole 'everything that receives a notification is now a conversation, but up to a max of X', which also adds extra complexity since we now have to track which channels are new/old or were modified. Also, the fact that this seems heavily vibe-coded makes me a bit uneasy.

---

@mxandreas this PR seems promising, although I'm not sure if we want to keep the whole logic.

To explain this a bit, this has the following pros:

1. Allows us to create notification channel 'groups' for both rooms and DMs separately, so the users can customise in the OS settings the default behaviour they want for those (e.g. I want DMs to always have this sound, or rooms to always be silent or only have vibration but no sound, etc., or even fully disable notifications just for rooms or DMs).
2. Adds an 'Android notification settings' shortcut item in the room notification settings so you can quickly access the notification settings for the room/DM in the OS and perform these changes instead of having to long press on an existing notification and modifying the behaviour or manually navigate to these settings in the OS settings app.
3. Adds an easy way to treat a room as a 'conversation' on Android, allowing the customisations explained above and giving it preferential screen state when a notification is received.

And, at least in my opinion, a big con:

For each notification you receive for a room you haven't been notified in before, a new notification channel is created, and it's marked as a conversation. 

This could flood the OS notification settings of the app with lots of conversations (at least on my Android 15 device I have now 10 conversations basically filling the whole screen space until I scroll down), and having them being added so easily feels a bit wrong. Also, it makes us need to have lots of logic to track which conversations/channels are outdated and should be replaced, which ones have been modified and we shouldn't touch, etc. I also think it's using the wrong storage location for that (a preferences text file instead of a proper database).

I think we'd be interested in these changes, but instead of automatically adding the rooms as conversations as soon as you receive something, maybe we could have some toggle on the room notification settings screen to manually mark the room as a conversation on Android, and allow this customisation and prioritisation. I think it should be fairly easy with the code added in this PR. What do you think?

This is an example of what could be customised, and how it works in this PR:

https://github.com/user-attachments/assets/c13460ba-7931-4e74-9e24-3960879d7907

mxandreas:
Unfortunately, I currently do not have time to do a deep dive into this and give a more specific answer. However, at the conceptual level, the issue is that notifications overall aren't yet stable enough - we have seen challenges in various areas and levels - issues related to Android OS, issues related to the app notification settings, issues related to push rules complexity, issues specific to air-gapped networks, etc. Therefore, I think we should not yet introduce more options and flexibility into this area.

PR #7162, opened 2026-07-07
Default to dark theme for EC

toger5:
This PR configures the Call presenter and element call webview (via url param) to use the dark theme.
(this is the intended design on light and dark system configuration)

 
Screenshot 2026-07-10 at 18 45 26

## Content

This change is related (and required) for: https://github.com/element-hq/element-call/pull/4067

## Motivation and context
https://github.com/element-hq/element-call/pull/4067

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
+++ b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/ui/CallScreenPresenter.kt
@@ -84,7 +84,7 @@ class CallScreenPresenter(
         var ignoreWebViewError by rememberSaveable { mutableStateOf(false) }
         var webViewError by remember { mutableStateOf<String?>(null) }
         val languageTag = languageTagProvider.provideLanguageTag()
-        val theme = if (ElementTheme.isLightTheme) "light" else "dark"
+        val theme = "dark"
 
         DisposableEffect(Unit) {
             coroutineScope.launch {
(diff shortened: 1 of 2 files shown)


bmarty (commented):
I think a better way to do this is to force the Dark theme at the Activity level:
Add the lines:
```kotlin
                ForcedDarkElementTheme(
                    colors = colors,
                ) {
```

at the beginning of the block `ElementThemeApp` in the class `ElementCallActivity`.

That should do it and also take care of the status bar and other side effects of forcing the theme to dark.

Here is a patch:
[elementCallDarkTheme.patch](https://github.com/user-attachments/files/29748227/elementCallDarkTheme.patch)

toger5:
> I think a better way to do this is to force the Dark theme at the Activity level:

This indeed sounds like a smarter way. I like it

toger5:
We might still need the dark theme url param (this is also needed as we benefit from the header colouring on the OS) but EC might not use the browser theme configuration but only listen on the url param config.

toger5:
@bmarty The errors I am getting 
```
> uses-permission: name='android.permission.REQUEST_INSTALL_PACKAGES'
6d6
< uses-permission: name='android.permission.REQUEST_INSTALL_PACKAGES'
12d11
< uses-permission: name='android.permission.INTERNET'
18a18
> uses-permission: name='android.permission.INTERNET'
20d19
< uses-permission: name='android.permission.ACCESS_LOCAL_NETWORK'
Error: Process completed with exit code 1.

```
and
```
Run ./scripts/checkEnv.js...maestro.js.JsEvaluationException: Fatal: missing env variable MAESTRO_PASSWORD
```
are not clear to me how to solve.

This might be an obvious issue I would just waste time on. Maybe you can help out?

jmartinesp (approved):

PR #7077, opened 2026-06-22
Fix local servers not working on Android 17

bxdxnn:
## Content

Ask for `ACCESS_LOCAL_NETWORK` permission if changing the server to a local server

## Motivation and context

Fixes #7073

## Tests

- Connect to a local server

## Tested devices

I don't have such device to check it properly, sorry. @Eskuero can you build this branch and see it yourself? I can also provide an APK from CI if you want.

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

- [x] I am aware of the [etiquette](https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#etiquette).
- This PR was made with the help of AI:
    - [x] Yes. In this case, please request a review by Copilot.
    - [ ] No.
- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly defines what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #7164, opened 2026-07-07
Declare ACCESS_LOCAL_NETWORK permission so local homeservers work on Android 17

### LINKED ISSUE #7073: Can't login to personal server anymore (on Android 17?)
### Steps to reproduce

So I have a local and not federated synapse server, that I access via a personal VPN on my Pixel 9 running GrapheneOS. Until now this worked with no issues but since I yesterday the apps seemed unable to connect anymore. Yesterday I did upgrade my Pixel 9 Pro to the Android 17 based build.

The chats stopped refreshing. I though it got in some buggy state and wiped it to relogin. But now it doesn't even seem to be able to connect with the server and permanently stays looping the circle until it says it couldn't connect.

However is not a connectivity issue. I can actually login and use my server just fine with Element Classic on the very same device. I can also still login perfectly using Element X on a Xiaomi device with Android 14.

I tried enabling some compatibility mode to see if it's caused by the GrapheneOS extra security features but it still fails. It never crashed anyway.

[Element X log 5ff1a45e6a83.txt](https://github.com/user-attachments/files/29197950/Element.X.log.5ff1a45e6a83.txt)

### Outcome

Being able to move into the next screen writing my credentials

### Your phone model

Pixel 9 Pro

### Operating system version

Android 17

26.06.4

### Homeserver

Synapse 

### Will you send logs?

Yes

No

mallafix:
## Content

Declare the `ACCESS_LOCAL_NETWORK` permission in the matrix api library manifest (next to `INTERNET`).

## Motivation and context

Since the targetSdk was raised to 37, Android 17's local network protections are enforced. Element X currently does not declare `ACCESS_LOCAL_NETWORK`, so:

- all traffic to homeservers on private addresses (192.168.x.x, 10.x.x.x, mDNS) is silently dropped — sync and push notifications stop working whenever the device is on the same LAN as a self-hosted homeserver,
- users cannot even grant the permission manually, because without the manifest declaration Android shows no toggle and `appops` rejects mode changes.

Declaring the permission makes the "Nearby devices" permission toggle available in the app settings, which is the minimal fix suggested in the discussion of #7077 ("devs can just add the permission to the manifest"). A follow-up could request the permission at runtime during server selection, as attempted in #7077, but this one-liner already unblocks affected self-hosters.

Fixes #7073

## Tests

- Reproduced on a Pixel 9a (Android 17, v26.06.x from the store) with a self-hosted Synapse on a private address: connection attempts are rejected by the system (visible via `appops get ... ACCESS_LOCAL_NETWORK` showing a fresh `rejectTime` on every app start), no permission toggle available.
- A build with the declared permission was already tested by a user in #7077 (comment by @pdc1): after manually granting "Nearby devices", connecting to a local homeserver works again. This PR applies exactly that declaration.

## Checklist

- [x] Changes has been tested on an Android device or Android emulator with API 24 (see Tests above — reproduction verified on device, fixed build verified by community testing in #7077)
- [x] UI change has been tested on both light and dark themes (no UI change)
- [x] Accessibility has been taken into account. (no UI change)

### DIFF
--- a/libraries/matrix/api/src/main/AndroidManifest.xml
+++ b/libraries/matrix/api/src/main/AndroidManifest.xml
@@ -8,5 +8,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <!-- Required since targetSdk 37 (Android 17) to reach homeservers on the local network. -->
+    <uses-permission android:name="android.permission.ACCESS_LOCAL_NETWORK" />
 
 </manifest>


bxdxnn (changes requested):
We need to ask for this permission too if the user changes or already changed his server to a local one.

bxdxnn:
Anyway, this PR should be closed as there's an Element-submitted one there: #7165

PR #7198, opened 2026-07-13
Allow homeserver to disable encryption in Element Pro

jmartinesp:
## Content

Add `force_disable_e2ee` support in the Element well known retriever and cache.

Use it to disable creating rooms with e2ee enabled or enabling encryption in existing ones.

## Motivation and context

Allowing a homeserver to disable encrypted room creation and disable enabling encryption in an existing room in Element Pro clients.

## Tests

- Hardcode `DefaultSessionEnterpriseService.isEncryptionDisabledByHomeserver` to return true.
- Open an existing non-encrypted room, check the security settings screen doesn't contain the 'enable encryption' toggle.
- Try creating private rooms and DMs, both should be unencrypted (without this, they would be encrypted by default).

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.

### DIFF
--- a/features/createroom/impl/build.gradle.kts
+++ b/features/createroom/impl/build.gradle.kts
@@ -43,6 +43,7 @@ dependencies {
     implementation(libs.coil.compose)
     implementation(projects.libraries.featureflag.api)
     implementation(projects.features.invitepeople.api)
+    implementation(projects.features.enterprise.api)
     api(projects.features.createroom.api)
 
     testCommonDependencies(libs, true)
@@ -53,4 +54,5 @@ dependencies {
     testImplementation(projects.libraries.permissions.test)
     testImplementation(projects.features.startchat.test)
     testImplementation(projects.libraries.featureflag.test)
+    testImplementation(projects.features.enterprise.test)
 }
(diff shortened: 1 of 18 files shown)


bxdxnn:
I think basically any deployment that trust its own server would want this feature as well? Why is this Element Pro?

jmartinesp:
> I think basically any deployment that trust its own server would want this feature as well? Why is this Element Pro?

I think Element's strategy is using E2EE chats as much as possible, this is mostly a workaround to allow some customers to upgrade their Element Classic users to Element X, and we made it this way mainly because we didn't think community users would be interested in it and because it's way faster than adding an MSC for it, but I don't think we'd be against implementing a proper MSC with a similar goal if it existed (cc @mxandreas maybe?).

mxandreas:
Yes, our goal is to generally optimize the clients for E2EE (Pro or not) as optimizing for "everything" tends to lead poor experience for everybody or a more complex product that is harder to understand, configure, maintain and further enhance. Having said that, we don't of course do anything intentionally to just block non-E2EE usage - this only applies when we have to make compromises.

Therefore, yes, we do not want to put the disabling of E2EE into our clients generally - it is rather keeping backward compatibility with what existed before for those who have paid for that in the past.

bmarty (approved):
LGTM, thanks!

PR #7202, opened 2026-07-14
Fix timelines not loading after marking a room as read from notification when app was killed

jmartinesp:
## Content

- When creating a new `SessionGraph`, make sure any `SessionPreferencesStore` that would be used in it is freshly created by removing existing ones.
- Try reusing an active room if it's present instead of instantiating a new one and its timeline, which is quite slow (this is just an optimization I added on top of the fix).

## Motivation and context

This can only happen when the app was killed by the OS and then you received some notification.

This 'mark as read' notification action creates a new `DefaultSessionPreferencesStore` when trying to fetch `isSendPublicReadReceiptsEnabled`, which would contain a different session coroutine scope than the one created when instantiating the navigation graph, resulting in flows that never emit any items and close automatically as 'completed'.

This in turn prevented the timeline items subscription in `TimelinePresenter` from progressing, since we're also checking `isRenderReadReceiptsEnabled` there, which is closed and in turn closes the timeline items flow.

## Tests

This is tricky to reproduce. With an unpatched version:

- Open the app, then send it to background.
- Simulate the OS killing the app with `adb shell am kill io.element.android.x.debug` and make sure it's not running anymore.
- With another account, send a message that will trigger a notification. This will wake the app, but without any UI or navigation graph (meaning no `SessionScope` or DI graph).
- Tap on 'mark as read' in the notification. This will trigger the problematic method call that caches a session prefs store.
- Then open the app and try opening any room. Check their timeline won't display any items, but the logs mention those items being received.

Then try the same with this branch, it should work.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.

### DIFF
--- a/app/src/main/kotlin/io/element/android/x/di/DefaultSessionGraphFactory.kt
+++ b/app/src/main/kotlin/io/element/android/x/di/DefaultSessionGraphFactory.kt
@@ -15,7 +15,7 @@ import io.element.android.libraries.matrix.api.MatrixClient
 
 @ContributesBinding(AppScope::class)
 class DefaultSessionGraphFactory(
-    private val appGraph: AppGraph
+    private val appGraph: AppGraph,
 ) : SessionGraphFactory {
     override fun create(client: MatrixClient): Any {
         return appGraph.sessionGraphFactory.create(client)
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationBroadcastReceiverHandler.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationBroadcastReceiverHandler.kt
@@ -105,13 +105,21 @@ class NotificationBroadcastReceiverHandler(
     @Suppress("unused")
     private fun handleMarkAsRead(sessionId: SessionId, roomId: RoomId, threadId: ThreadId?) = appCoroutineScope.launch {
         val client = matrixClientProvider.getOrRestore(sessionId).getOrNull() ?: return@launch
-        val isSendPublicReadReceiptsEnabled = sessionPreferencesStore.get(sessionId, this).isSendPublicReadReceiptsEnabled().first()
+        val isSendPublicReadReceiptsEnabled = sessionPreferencesStore.get(
+            sessionId = sessionId,
+            // Make sure `sessionCoroutineScope` is used here, otherwise we'll have several instances using different coroutines,
+            // and it will not work as expected.
+            sessionCoroutineScope = client.sessionCoroutineScope,
+        ).isSendPublicReadReceiptsEnabled().first()
         val receiptType = if (isSendPublicReadReceiptsEnabled) {
             ReceiptType.READ
         } else {
             ReceiptType.READ_PRIVATE
         }
-        val room = client.getJoinedRoom(roomId) ?: return@launch
+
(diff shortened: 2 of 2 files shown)


app/src/main/kotlin/io/element/android/x/di/DefaultSessionGraphFactory.kt
```diff
@@ -12,12 +12,18 @@ import dev.zacsweers.metro.AppScope
...
 class DefaultSessionGraphFactory(
-    private val appGraph: AppGraph
+    private val appGraph: AppGraph,
+    private val sessionPreferencesStoreFactory: SessionPreferencesStoreFactory,
 ) : SessionGraphFactory {
     override fun create(client: MatrixClient): Any {
+        // Remove the session preferences store for the given session ID when creating a new session graph.
+        // There might be an outdated one already stored using an outdated session coroutine scope.
+        sessionPreferencesStoreFactory.remove(client.sessionId)
+
```
bmarty: Good catch, but this looks like a workaround.

I wonder if a better fix would be to remove the cache in `DefaultSessionPreferencesStoreFactory` and add `@SingleIn(SessionScope::class)` annotation on `SessionPreferencesModule.providesSessionPreferencesStore`.

I haven't fully checked that it works, though

app/src/main/kotlin/io/element/android/x/di/DefaultSessionGraphFactory.kt
```diff
@@ -12,12 +12,18 @@ import dev.zacsweers.metro.AppScope
...
 
 @ContributesBinding(AppScope::class)
 class DefaultSessionGraphFactory(
-    private val appGraph: AppGraph
+    private val appGraph: AppGraph,
+    private val sessionPreferencesStoreFactory: SessionPreferencesStoreFactory,
 ) : SessionGraphFactory {
     override fun create(client: MatrixClient): Any {
+        // Remove the session preferences store for the given session ID when creating a new session graph.
+        // There might be an outdated one already stored using an outdated session coroutine scope.
```
bmarty:
```suggestion
        // There might be an outdated one already stored using the application coroutine scope.
```

bmarty (approved):
Thanks. A remark about the fix (good catch!). Maybe alternatives are nicer, but more complex, so I let you decide what you want to do. Keeping your solution also works for me.

app/src/main/kotlin/io/element/android/x/di/DefaultSessionGraphFactory.kt
jmartinesp: I believe we added this cache because we had some issues with several `DefaultSessionPreferencesStore` instances being instantiated at the same time with different coroutine scopes. This broke some stuff like app migrations or the `NotificationBroadcastReceiverHandler`.

These 2 components need to access/create one of these stores from the `AppScope`, so we need a way to reach it. That said, I think I might have found an easier solution: the problem here is we're instantiating the session store using `appCoroutineScope` when it should be using `SessionCoroutineScope`, and I can take that from the `matrixClient` we just got/restored. Using that seems to fix the issue too.

jmartinesp:
Simplified this in https://github.com/element-hq/element-x-android/pull/7202/commits/f4110e2532ae1e9070c5bb648960ed93048bac14, replacing the usage of the app coroutine scope with the proper session coroutine scope.

bmarty (approved):
The new alternative is even better. Thanks!

PR #7204, opened 2026-07-14
Try fixing DM creation in Maestro

jmartinesp:
## Content

Several fixes to the Maestro flow (again).

Also added a room cleanup step to try to minimise issues with duplicate rooms or DMs.

This PR is better reviewed by checking the full changes instead of commit-by-commit.

## Motivation and context

Fix Maestro again.

## Tests

[CI passed](https://github.com/element-hq/element-x-android/actions/runs/29343363715).

## Checklist

    - [ ] No.

### DIFF
--- a/.maestro/tests/roomList/cleanUpRoomList.yaml
+++ b/.maestro/tests/roomList/cleanUpRoomList.yaml
@@ -1,0 +1,21 @@
+appId: ${MAESTRO_APP_ID}
+---
+# Clean up the room list if it was left in a bad state by previous tests
+- runFlow:
+    when:
+      visible: 'aRoomName'
+    commands:
+      - longPressOn: 'aRoomName'
+      - tapOn:
+          text: "Leave room"
+          index: 0
+      - tapOn: "Leave"
+- runFlow:
+    when:
+      visible: ${MAESTRO_INVITEE1_MXID}
+    commands:
+      - longPressOn: ${MAESTRO_INVITEE1_MXID}
+      - tapOn:
+          text: "Leave room"
+          index: 0
+      - tapOn: "Leave"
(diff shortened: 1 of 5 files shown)


.maestro/tests/roomList/cleanUpRoomList.yaml
```diff
@@ -0,0 +1,21 @@
+appId: ${MAESTRO_APP_ID}
+---
+# Clean up the room list if it was left in a bad state by previous tests
```
bmarty, line 3: Good idea, thanks. I am wondering what could happen if we have 2 maestro test runs in parallel and the room is deleted when it should not. But the risk is very very low, so we should not bother.

features/invitepeople/impl/src/main/kotlin/io/element/android/features/invitepeople/impl/InvitePeopleView.kt
```diff
@@ -302,13 +302,13 @@ private fun InvitePeopleConfirmModal(
...
                 leadingIcon = IconSource.Vector(CompoundIcons.Close()),
-                modifier = Modifier.weight(1f).testTag(TestTags.confirmInviteUnknown),
+                modifier = Modifier.weight(1f),
             )
             Button(
                 text = stringResource(CommonStrings.action_invite),
                 onClick = onInvite,
                 leadingIcon = IconSource.Vector(CompoundIcons.Check()),
-                modifier = Modifier.weight(1f)
+                modifier = Modifier.weight(1f).testTag(TestTags.confirmInviteUnknown),
```
bmarty, line 311: Good catch!

bmarty (approved):
Thanks a lot!

.maestro/tests/roomList/cleanUpRoomList.yaml
jmartinesp: I think I actually made sure some time ago Maestro tests won't run in parallel. I'd have to double check though 😅 .

---

Yup, seems like that's already considered:

```
# Allow only one to run at a time, since they use the same environment.
# Otherwise, tests running in parallel can break each other.
concurrency:
  group: maestro-test
```

bmarty: wow, I was not aware of this. Thanks!

PR #7228, opened 2026-07-16
Fix logging via QR code when using MDM-configured homeserver

mxandreas:
## Content

When logging in via QR Code, it is checked that the homeserver the user is logging in to, is the same as configured by MDM. This check has a bug (e.g. comparing https://example.com vs example.com). The same check is applied when the homeserver is pre-configured via custom app.

For now, we remove this check because of the following:
* Even if in some cases the check can be easily fixed by ensuring there is always protocol there, it could be that we end up comparing server/domain name (e.g. https://example.com with the homeserver URL https://element.example.com/) and the check fails again for another reason.
* The primary purpose of the feature MDM-configured homeserver was to simplify sign in (not necessarily to block users to sign in any other homeserver).
* However, since in some cases there is interest that no other homeserver than configured by MDM is used (even when using QR code login), then there will be a project soon to add the check back and work robustly.

On iOS the check was already removed before.

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [x] No.


bmarty (approved):
Thanks!

PR #7236, opened 2026-07-17
Group Element Pro changes into a dedicated release notes section

manuroe:
Pro-only PRs (labelled PR-Element-Pro) leaked into Element X release notes, causing confusion. They now group into a single Element Pro section instead of scattering through Features/Bugfixes. Granularity is dropped on purpose for these changes.

Android equivalent of https://github.com/element-hq/element-x-ios/pull/5899.
The new tag is available in the project: https://github.com/element-hq/element-x-android/labels?q=pr-element-pro

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -28,6 +28,10 @@ changelog:
       labels:
         - PR-Dependencies
 
+    - title: 💼 Element Pro
+      labels:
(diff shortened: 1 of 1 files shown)


bmarty (approved):
Thanks!

Issue #7072, opened 2026-06-22
[Task] Add the ability to search for messages within a room #5620


YvesDutch:
[Please see and discuss the details in the meta issue.](https://github.com/element-hq/element-x-ios/pull/5620)

What is the status of However, message search is a project the core team is going to attack in the coming weeks. We are currently gathering the product and design insights. Sadly, I have to close your PR.?

Thanks.

bmarty:
You can track https://github.com/element-hq/element-meta/issues/3252

Issue #6557, opened 2026-04-08
Feature request: Add search functionality inside rooms (in-chat message search)


sasan7071-maker:
### Your use case

**Problem / Current behavior**
In Element X, there is no way to search for messages inside a room/chat.
Once you are inside a room, there is no search bar or option to find specific messages, keywords, or older conversations within that room.

**Expected behavior**
I would like to have a search feature inside each room that allows users to:

Search for keywords or phrases within that specific room
See results with timestamps and quick jump to the message
Filter results if possible (e.g. by sender, date, media, etc.)

This is a very common and useful feature in modern messaging apps (Telegram, WhatsApp, Discord, Slack, and even the old Element client had basic search).

**Why this matters**
Makes it much easier to find important messages in long or busy rooms
Especially helpful in group chats with hundreds or thousands of messages
Improves productivity and user experience significantly

**Steps to reproduce / Current limitation**
Open any room in Element X
Try to find a specific message or keyword from earlier
There is no search option available

**Additional context**
Global search (searching across all rooms) might already exist or not, but in-room search is missing.
Please consider adding a search icon at the top of the chat screen.

**Platform:**
 Element X Android
 Element X iOS

### Are you willing to provide a PR?

Yes

sepsol:
iOS version could benefit from this as well

nohkumado:
That's definitely a showstopper for a lot of the people I brought to matrix.... And their reaction is reliably: wtf?? Why can't your crappy thing do something as basic as searching, let's go back to WhatsApp.... :'(

GalaxyTimeMachine:
This is the main reason I hate using element, and wouldn't if I didn't have to.

feroom:
The request to add a search function is indeed highly relevant for a vast number of servers and users.
I join in the request.

Trancep0rt:
> That's definitely a showstopper for a lot of the people I brought to matrix.... And their reaction is reliably: wtf?? Why can't your crappy thing do something as basic as searching, let's go back to WhatsApp.... :'(

Same reaction Here. I join in,too!

Issue #4149, opened 2025-01-14
Search option missing


tutle:
### Steps to reproduce

Hello, happy new year for everybody. 
on my android device, there is no "search" option in private room in a community but there is in public, really missing (also missing search medias and links like in common IM).
It's a bug or it's normal ?
Sincerely
Yvan
Public room menu
![Image](https://github.com/user-attachments/assets/8829d76d-85b9-41da-b0fb-e19770913d5e)
Crypted room menu
![Image](https://github.com/user-attachments/assets/2a55f9d8-eafe-4e78-8788-0d3f4e8eda01)

### Outcome

Correction

### Your phone model

Doogee V20 pro

### Operating system version

Android 12

### Will you send logs?

Yes

### Are you willing to provide a PR?

Yes

sunjam:
Duplicate of #3709

Issue #3709, opened 2024-10-19
search messages in room


KiaraGrouwstra:
#### What would you like to do?

search the messages in a room

#### Why would you like to do it?

faster than reading all messages

#### How would you like to achieve it?

have a search bar in the room menu

### Have you considered any alternatives?

use a different client

### Are you willing to provide a PR?

No

jdauphant:
If search on encrypted all messages are technically complicated : in that case, search index could be limited to X last days (30 days for example and this setting could be change in settings).

frandavid100:
> If search on encrypted all messages are technically complicated : in that case, search index could be limited to X last days (30 days for example and this setting could be change in settings).

Maybe search whatever messages are stored locally, then have a "load previous results" that stores more messages and searches those?

jdauphant:
> > If search on encrypted all messages are technically complicated : in that case, search index could be limited to X last days (30 days for example and this setting could be change in settings).
> 
> Maybe search whatever messages are stored locally, then have a "load previous results" that stores more messages and searches those?

"Load previous results" could work if there is no performance issues.
Also, to be sure the user get how the search work and appreciate it, I think we should search in all messages for a period (with no gap) and we should tell him this period ("search on the last 15 days").

frandavid100:
> "Load previous results" could work if there is no performance issues.

Is there any reason to think there would?

> Also, to be sure the user get how the search work and appreciate it, I think we should search in all messages for a period (with no gap) and we should tell him this period ("search on the last 15 days").

Not sure about that, since it might complicate the UI and the devs seem to be going in the opposite direction; but I'm not against it either.

lionking102:
I also encountered this issue. A search function would be amazing and, I believe, quite important for many users.

How do other major messaging apps handle searching through the entire chat history?

maurerle:
This issue is very important for me, as I would like to search the chat history from time to time.
Having this feature on the desktop element client is a good alternative, but somehow not really sufficient :/

saihaze:
When implementing this, CJK should be considered. Tokenising these languages is quite different because they are not space-seperated.

nathanael-h:
Related issue for iOS: https://github.com/element-hq/element-x-ios/issues/3766

vanyasem:
Meta Issue: https://github.com/element-hq/element-meta/issues/1044
Related Meta Issue: https://github.com/element-hq/element-meta/issues/2004

[...]

PR #7249, opened 2026-07-21
Index messages locally in an encrypted on-device store so encrypted rooms can be searched at all

hayaksi1:
### Problem

Element X cannot search messages in encrypted rooms. The server-side `/search` endpoint only ever
sees ciphertext, so the one search path the app has is structurally unable to answer for E2EE
rooms — it returns nothing rather than reporting that it cannot help. The visible symptom is
element-hq/element-x-android#4149: the in-room search option is present in unencrypted rooms and
absent in encrypted ones. The workaround users describe ends in searching from another client
instead (element-hq/element-x-android#7072).

The Rust SDK now ships a local, encrypted, tantivy-backed message index — the `matrix-sdk-search`
crate tracked by matrix-org/matrix-rust-sdk#5350 — but **nothing in this app has ever wired it
up**. This adds the matrix-layer plumbing to do so.

### Fix

- `MatrixClient` gains `isMessageSearchAvailable` and `messageSearchService`.
  `RustMatrixClientFactory` decides availability once, from
  `featureFlagService.isFeatureEnabled(MessageSearch) && clientSecret != null`, and calls
  `withSearchIndexStore(path = <fileDir>/search-index, password = clientSecret)` on the
  `ClientBuilder`. A session with no client secret gets no index — see **Scope**.
- A new `libraries/matrix/api/.../search/` package keeps the SDK out of the UI, per the wrapping
  convention: `MessageSearch` is one stateful cursor exposing
  `results: StateFlow<ImmutableList<MessageSearchResult>>` and
  `paginationState: StateFlow<MessageSearchPaginationState>`, with `setQuery` and `paginate`
  returning `Result`. Its lifetime is the `CoroutineScope` that created it, so there is no
  `close()` to forget.
- `RustMessageSearch` funnels the SDK's non-suspend listener batches through a single
  `Channel` consumer so diffs stay ordered, and disposes its `TaskHandle`s via
  `scope.coroutineContext.job.invokeOnCompletion`. `MessageSearchResultsProcessor` applies
  `SearchServiceResultsUpdate` diffs under a mutex against a list kept index-parallel with the
  SDK's, because the diffs are positional. Both it and `MessageSearchResultMapper` use an
  exhaustive `when` with no `else`, so an SDK bump fails compilation rather than silently
  dropping a variant.
- `TantivyQueryEscaper` escapes query syntax so what the user types is searched as text. Today
  a query containing `:)`, a stray `"`, or the word `AND` is parsed as tantivy query syntax and
  can fail the whole search. Escaping runs in a single forward pass so added backslashes are not
  re-escaped, and the `+` Must prefix is applied *after* escaping — prefixing first would emit a
  literal `\+` that parses cleanly and silently restores OR semantics.
- `getCacheSize()` now counts the index directory, and disabling the flag deletes it.

Everything is behind `FeatureFlags.MessageSearch` (`feature.message_search`,
`defaultValue = { false }`, `isFinished = false`). With the flag off — the default — no index
store is attached, no index directory is created, and no other code path changes.

## Motivation and context

Relates to element-hq/element-x-android#3709 (search messages in room) — the canonical tracker,
and the thread that argues for this design.
Relates to element-hq/element-x-android#6557 (in-room keyword search with timestamps and
jump-to-message).
Relates to element-hq/element-x-android#4149 (closed as a duplicate of #3709; a defect report,
with screenshots).
Relates to element-hq/element-x-android#7072.

None of these is closed here — the change ships behind a disabled feature flag, so they are cited
as context rather than as fixes.

## Screenshots / GIFs

There is no UI change in this pull request.

## Tests

Covered by unit tests rather than by manual steps.

`TantivyQueryEscaperTest` — expectations checked against a real tantivy 0.26.1 parser, covering
each escaped character in leading and non-leading position, the bare-word operators
(`AND`/`OR`/`NOT`/`IN`/`TO`), the no-re-escaping property, and the deliberate case where a token
with no letter or digit is left optional because the default tokenizer indexes no terms for it.

`MessageSearchResultsProcessorTest` — every `SearchServiceResultsUpdate` variant:
Append, PushBack, PushFront, Set, Insert, Remove, Truncate, Reset, PopBack, PopFront and Clear,
asserting the local list stays index-parallel with the SDK's.

`RustMessageSearchTest` — lazy subscription on first `setQuery`; that `setQuery` resets
`paginationState` to `Idle(endReached = false)` before results arrive, since a fresh SDK cursor
reports `endReached = true`; ordered delivery of batched listener callbacks; and handle disposal
when the scope completes.

`RustMatrixClientFactoryTest` — the index store is attached when the flag is on and a client
secret exists, and not attached when either is missing.

New fakes for downstream modules: `FakeMessageSearch`, `FakeMessageSearchService`,
`FakeFfiSearchService`.

### Scope

Deliberately not attempted here, so it is not mistaken for an oversight:

- **Search is entirely local.** The on-device index only; there is no server-side `/search`
  fallback. Anything the index never saw is unfindable.
- **Every query word is mandatory.** Each escaped token becomes a `+` Must clause, so one typo
  returns zero results where an OR query would have degraded to partial matches. The trade-off
  is deliberate — OR over a local index returned mostly noise — but it is a trade-off.
- **Matching is whole-word and unstemmed.** The `*` prefix operator is escaped, so partial words
  do not match, and `photo` does not match `photos`. Tokenisation is fixed when the index is
  created and is effectively English-first, so a script the default tokenizer does not segment is
  served poorly; both are worth their own follow-ups rather than a fix buried here.
- **Deliberate query syntax no longer works** — `"phrase"`, `-exclude`, ranges, regex and field
  lookups are all escaped to literal text. None of it was advertised in the UI, and malformed
  use of it used to kill the whole search.
- **What the SDK indexes is opaque from here.** Indexing is a side effect of events passing
  through the SDK's event cache. Media and sticker indexing, edit semantics, and
  removal-on-redaction — currently broken upstream on matrix-org/matrix-rust-sdk#5350 — are all
  SDK-side.
- **Sessions with no client secret never get an index** and search is silently unavailable to
  them: legacy pre-schema-v5 sessions, and sessions created by a release build between
  2024-01-18 and 2024-04-15.
- **The flag is read at client-build time only**, so toggling it mid-session does nothing until
  restart.
- Sender filtering and space scoping are out of scope; happy to file linked tracking issues if
  useful.

### Notes for reviewers

First of four. The rest, each depending only on this one: the search screen and its top-bar
entry point; the background history backfill; and the event-cache coverage bootstrap. Since a
pull request here must be based on a branch in this repository, they follow serially rather than
as a stack.

This is ~1,430 insertions, above the 500 guideline. Roughly half is tests and fakes, and
splitting the tantivy escaping or the API definition away from the implementation they cover
would make review harder rather than easier — but if you would prefer `TantivyQueryEscaper` and
its test as a separate ~200-line change, say so and I will pull it out.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Android 16 (API 36), Pixel 10 Pro emulator (arm64-v8a)

## Checklist

    - [ ] No.

### DIFF
--- a/app/src/main/kotlin/io/element/android/x/initializer/PlatformInitializer.kt
+++ b/app/src/main/kotlin/io/element/android/x/initializer/PlatformInitializer.kt
@@ -22,6 +22,14 @@ import timber.log.Timber
 
 private const val ELEMENT_X_TARGET = "elementx"
 
+// The SDK sets no global log level, and `matrix_sdk_search` is absent from its default target list,
+// so without this directive every log line from the message search index is silently dropped.
+// The indexing task itself lives under `matrix_sdk::event_cache`, so enable the Event Cache trace
+// log pack as well to follow a message from sync into the index.
+// Debuggable builds only: the SDK logs the full message body at debug level, and release logs can be
+// uploaded via rageshake.
+private const val MATRIX_SDK_SEARCH_TARGET = "matrix_sdk_search"
+
 class PlatformInitializer : Initializer<Unit> {
     override fun create(context: Context) {
         val appBindings = context.bindings<AppBindings>()
@@ -36,7 +44,12 @@ class PlatformInitializer : Initializer<Unit> {
             writesToLogcat = runBlocking { featureFlagService.isFeatureEnabled(FeatureFlags.PrintLogsToLogcat) },
             writesToFilesConfiguration = bugReporter.createWriteToFilesConfiguration(),
             logLevel = logLevel,
-            extraTargets = listOf(ELEMENT_X_TARGET),
+            extraTargets = buildList {
+                add(ELEMENT_X_TARGET)
+                if (appBindings.buildMeta().isDebuggable) {
+                    add(MATRIX_SDK_SEARCH_TARGET)
+                }
+            },
             traceLogPacks = runBlocking { preferencesStore.getTracingLogPacksFlow().first() },
             sdkSentryDsn = appBindings.sentrySdkDsn()?.value?.takeIf { it.isNotBlank() },
         )
--- a/app/src/main/kotlin/io/element/android/x/initializer/PlatformInitializer.kt
+++ b/app/src/main/kotlin/io/element/android/x/initializer/PlatformInitializer.kt
@@ -22,14 +22,6 @@ import timber.log.Timber
 
 private const val ELEMENT_X_TARGET = "elementx"
 
-// The SDK sets no global log level, and `matrix_sdk_search` is absent from its default target list,
-// so without this directive every log line from the message search index is silently dropped.
-// The indexing task itself lives under `matrix_sdk::event_cache`, so enable the Event Cache trace
-// log pack as well to follow a message from sync into the index.
-// Debuggable builds only: the SDK logs the full message body at debug level, and release logs can be
-// uploaded via rageshake.
-private const val MATRIX_SDK_SEARCH_TARGET = "matrix_sdk_search"
-
 class PlatformInitializer : Initializer<Unit> {
     override fun create(context: Context) {
         val appBindings = context.bindings<AppBindings>()
@@ -44,12 +36,7 @@ class PlatformInitializer : Initializer<Unit> {
             writesToLogcat = runBlocking { featureFlagService.isFeatureEnabled(FeatureFlags.PrintLogsToLogcat) },
             writesToFilesConfiguration = bugReporter.createWriteToFilesConfiguration(),
             logLevel = logLevel,
-            extraTargets = buildList {
-                add(ELEMENT_X_TARGET)
-                if (appBindings.buildMeta().isDebuggable) {
-                    add(MATRIX_SDK_SEARCH_TARGET)
-                }
-            },
+            extraTargets = listOf(ELEMENT_X_TARGET),
             traceLogPacks = runBlocking { preferencesStore.getTracingLogPacksFlow().first() },
             sdkSentryDsn = appBindings.sentrySdkDsn()?.value?.takeIf { it.isNotBlank() },
         )
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
@@ -82,33 +82,23 @@ class RustMatrixClientFactory(
         // This secret is called 'passphrase' for historical reasons, but it can be a raw key or an actual passphrase
         val clientSecret = sessionData.passphrase?.let(ClientSecret::fromString)
         val sessionPaths = sessionData.getSessionPaths()
-        val isMessageSearchAvailable = isMessageSearchAvailable(clientSecret)
+        val isMessageSearchAvailable = featureFlagService.isFeatureEnabled(FeatureFlags.MessageSearch)
         val indexDirectory = File(sessionPaths.fileDirectory, SEARCH_INDEX_DIRECTORY)
 
         if (!isMessageSearchAvailable && indexDirectory.exists()) {
             // With search off, events reach the event cache unindexed and the index can never
             // catch up on them. Delete it so the next enable rebuilds it from a state where
             // coverage can actually be guaranteed, instead of resuming a silently stale index.
-            Timber.tag("RustMatrixClient").i("Message search is unavailable, deleting the stale search index")
+            Timber.tag("RustMatrixClient").i("Message search is disabled, deleting the stale search index")
             indexDirectory.deleteRecursively()
         }
 
-        val client = buildAndRestoreClient(sessionData, clientSecret, isMessageSearchAvailable)
-        create(client, sessionData, isMessageSearchAvailable)
-    }
-
-    private suspend fun buildAndRestoreClient(
-        sessionData: SessionData,
-        clientSecret: ClientSecret?,
-        isMessageSearchAvailable: Boolean,
-    ): Client {
-        val baseClientBuilder = getBaseClientBuilder(
-            sessionPaths = sessionData.getSessionPaths(),
+        val client = getBaseClientBuilder(
+            sessionPaths = sessionPaths,
             clientSecret = clientSecret,
             slidingSyncType = ClientBuilderSlidingSync.Restored,
             isMessageSearchAvailable = isMessageSearchAvailable,
         )
-        val client = baseClientBuilder.clientBuilder
             .homeserverUrl(sessionData.homeserverUrl)
             .username(sessionData.userId)
             .use { it.build() }
@@ -127,7 +117,8 @@ class RustMatrixClientFactory(
         )
 
         client.restoreSession(sessionData.toSession())
-        return client
+
+        create(client, sessionData, isMessageSearchAvailable)
     }
 
     suspend fun create(
@@ -187,20 +178,13 @@ class RustMatrixClientFactory(
         }
     }
 
-    private suspend fun isMessageSearchAvailable(clientSecret: ClientSecret?): Boolean =
-        featureFlagService.isFeatureEnabled(FeatureFlags.MessageSearch) && clientSecret != null
-
     internal suspend fun getBaseClientBuilder(
         sessionPaths: SessionPaths,
         clientSecret: ClientSecret?,
         slidingSyncType: ClientBuilderSlidingSync,
-        // Null reads the live flag; the restore path passes its own snapshot instead, so a flag
-        // flipped mid-restore cannot make the coverage bookkeeping and the built client disagree
-        // about whether an index exists.
-        isMessageSearchAvailable: Boolean? = null,
-    ): BaseClientBuilder {
-        val messageSearchAvailable = isMessageSearchAvailable ?: isMessageSearchAvailable(clientSecret)
-        val clientBuilder = clientBuilderProvider.provide()
+        isMessageSearchAvailable: Boolean,
+    ): ClientBuilder {
+        return clientBuilderProvider.provide()
             .run {
                 sqliteStoreBuilderProvider.provide(sessionPaths)
                     .secret(clientSecret)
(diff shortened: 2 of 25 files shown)


sanyamseac:
@jmartinesp @ganfra Is there any plan to merge this pr anytime soon, or will it take time to review and test?

jmartinesp:
> @jmartinesp @ganfra Is there any plan to merge this pr anytime soon, or will it take time to review and test?

We're about to start working in this feature on Android, and while we appreciate the conrtibutions, the problem is some of the PRs clash with our strategy (like adding the logic on Rust instead of Android) or designs, so we'd rather build those ourselves. Also, merging huge AI-generated PRs has come back to bite us in the past, so we'd rather proceed with caution with this feature.

That said, using this PR as a base seems fine: it looks like it provides some nice scaffolding for the feature. Let's see if we can make it mergeable.

jmartinesp:
Odd, the failing CI flows pass locally, and a rebase didn't seem to help. It also happened in a different PR.

app/src/main/kotlin/io/element/android/x/initializer/PlatformInitializer.kt
```diff
@@ -36,7 +44,12 @@ class PlatformInitializer : Initializer<Unit> {
             writesToLogcat = runBlocking { featureFlagService.isFeatureEnabled(FeatureFlags.PrintLogsToLogcat) },
             writesToFilesConfiguration = bugReporter.createWriteToFilesConfiguration(),
             logLevel = logLevel,
-            extraTargets = listOf(ELEMENT_X_TARGET),
+            extraTargets = buildList {
+                add(ELEMENT_X_TARGET)
+                if (appBindings.buildMeta().isDebuggable) {
+                    add(MATRIX_SDK_SEARCH_TARGET)
+                }
```
jmartinesp: This is not needed. If we add an extra target, it should be done in the SDK.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -81,11 +81,34 @@ class RustMatrixClientFactory(
...
+        val isMessageSearchAvailable = isMessageSearchAvailable(clientSecret)
+        val indexDirectory = File(sessionPaths.fileDirectory, SEARCH_INDEX_DIRECTORY)
+
+        if (!isMessageSearchAvailable && indexDirectory.exists()) {
+            // With search off, events reach the event cache unindexed and the index can never
+            // catch up on them. Delete it so the next enable rebuilds it from a state where
+            // coverage can actually be guaranteed, instead of resuming a silently stale index.
+            Timber.tag("RustMatrixClient").i("Message search is unavailable, deleting the stale search index")
+            indexDirectory.deleteRecursively()
+        }
```
jmartinesp, line 94: Wouldn't it make more sense to delete it if the feature flag is disabled?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -81,11 +81,34 @@ class RustMatrixClientFactory(
...
+
+        val client = buildAndRestoreClient(sessionData, clientSecret, isMessageSearchAvailable)
+        create(client, sessionData, isMessageSearchAvailable)
+    }
+
+    private suspend fun buildAndRestoreClient(
+        sessionData: SessionData,
+        clientSecret: ClientSecret?,
+        isMessageSearchAvailable: Boolean,
+    ): Client {
```
jmartinesp: Why split this? AFAICT it's only used here, which would be the same as inlining both methods.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -81,11 +81,34 @@ class RustMatrixClientFactory(
...
+        val baseClientBuilder = getBaseClientBuilder(
             sessionPaths = sessionData.getSessionPaths(),
             clientSecret = clientSecret,
             slidingSyncType = ClientBuilderSlidingSync.Restored,
+            isMessageSearchAvailable = isMessageSearchAvailable,
         )
+        val client = baseClientBuilder.clientBuilder
```
jmartinesp: Again, what's the reason for not keeping the previous format? It's more verbose now and I don't see why we need to extract the builder to a variable.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -155,17 +181,26 @@ class RustMatrixClientFactory(
...
         slidingSyncType: ClientBuilderSlidingSync,
-    ): ClientBuilder {
-        return clientBuilderProvider.provide()
+        // Null reads the live flag; the restore path passes its own snapshot instead, so a flag
+        // flipped mid-restore cannot make the coverage bookkeeping and the built client disagree
+        // about whether an index exists.
+        isMessageSearchAvailable: Boolean? = null,
+    ): BaseClientBuilder {
+        val messageSearchAvailable = isMessageSearchAvailable ?: isMessageSearchAvailable(clientSecret)
+        val clientBuilder = clientBuilderProvider.provide()
```
jmartinesp: I don't think this makes sense, the flag can't be flipped mid-restore AFAICT.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -193,6 +228,22 @@ class RustMatrixClientFactory(
...
+                // Note: every ClientBuilder call returns a NEW reference, so this must stay in
+                // expression position — using `if (flag) withSearchIndexStore(...)` as a statement
+                // would silently drop the result and leave the index disabled.
+                if (messageSearchAvailable) {
+                    // The index is encrypted at rest with the session secret, reusing the exact
+                    // string the SDK's SQLite stores already use. When there is no secret we skip
+                    // indexing entirely rather than writing message bodies to disk in plaintext.
+                    withSearchIndexStore(
+                        path = File(sessionPaths.fileDirectory, SEARCH_INDEX_DIRECTORY).absolutePath,
+                        password = checkNotNull(clientSecret).formattedAsString(),
```
jmartinesp: This method accepts a nullable `password`, so I think we should allow passing a null `clientSecret?.formattedAsString()`. I'm not sure why we'd avoid indexing messages if there is no secret, even if they're not encrypted the index would still be used for search if I'm not mistaken.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -218,9 +269,24 @@ class RustMatrixClientFactory(
...
+            clientBuilder = clientBuilder,
+            isMessageSearchAvailable = messageSearchAvailable,
+        )
     }
 }
 
+internal data class BaseClientBuilder(
+    val clientBuilder: ClientBuilder,
+    val isMessageSearchAvailable: Boolean,
+)
```
jmartinesp: I think this class wouldn't be needed if we enforced passing `isMessageSearchAvailable` in the `getBaseClientBuilder` method? There's no need to return something you just passed.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
```diff
@@ -81,6 +81,7 @@ class RustMatrixAuthenticationService(
     // Ideally it would be possible to get the sessionPath from the Client to avoid doing this.
     private var sessionPaths: SessionPaths? = null
     private var currentClient: Client? = null
+    private var currentClientIsMessageSearchAvailable = false
```
jmartinesp: Instead of using this, we should probably check the feature flag using `FeatureFlagService`?

jmartinesp (changes requested):
After a quick look, I have a few questions and possible improvements to the code.

app/src/main/kotlin/io/element/android/x/initializer/PlatformInitializer.kt
hayaksi1: Removed, the file is back to the develop version. I'll propose adding `matrix_sdk_search` to the SDK's default targets instead.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
hayaksi1: Changed: availability is now just the feature flag, so the index is deleted exactly when the flag is off.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
hayaksi1: Inlined back into `create()`.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
hayaksi1: Restored the chained format. The variable was only there for the `BaseClientBuilder` wrapper, which is gone now.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
hayaksi1: Removed; `getBaseClientBuilder` now takes a plain `isMessageSearchAvailable: Boolean`.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
hayaksi1: Changed to `clientSecret?.formattedAsString()`. Indexing now depends only on the feature flag; sessions without a secret get an unencrypted index, same as their SQLite stores.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
hayaksi1: Removed. The parameter is required now and the method returns `ClientBuilder` again.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
hayaksi1: Done, the field is gone; the service now injects `FeatureFlagService` and reads the flag at the call sites.

jmartinesp (approved):
Thanks for the changes, I think the code looks good now! Don't mind the Konsist tests, I think I know where they're coming from and there's no way they could work for forks, so I'll just force merge this.

jmartinesp:
Actually, the PR that should fix the issues with Konsist was just merged, so let's use this PR to test it works as expected for fork PRs.

PR #7252, opened 2026-07-21
Rebuild the search index when it cannot cover what the event cache already holds

hayaksi1:
### Problem

`withSearchIndexStore` hands the SDK a directory, and from then on the indexer sees exactly one
thing: events as they pass through the event cache. Everything already sitting in the
**event cache store** — `matrix-sdk-event-cache.sqlite3` — predates that indexer, and back-paginating
it re-hydrates those events from disk, which — as far as can be observed from this side of the FFI —
never reaches the indexer. The flag is read once, at `ClientBuilder` time, so the index is always created at least
one restart *after* the cache has started filling — which means the initial sync window of every
room, and therefore the entirety of a small room, lands in the hole and stays there for the life
of the session. The result is a search that returns nothing for a conversation the user can
plainly see on screen, with no error.

### Fix

- `RustMatrixClientFactory.create` now deletes the event cache store — the database and its
  `-wal` and `-shm` sidecars, sidecars first so an interrupted deletion leaves a merely stale
  database rather than a fresh one next to a replayable WAL — when message search is available
  and the index cannot prove it already covers what is cached.
- The deletion happens before the client is built, so no SQLite handle is open, and that history
  then re-enters through network pagination — the one path the indexer observes. This relies on the
  SDK deduplicating re-fetched events by event id; that is SDK-side behaviour which cannot be
  asserted from this diff, so it is worth a maintainer confirming rather than taking on trust.
- A marker file, `search-index.covered`, records that the index has seen everything since the
  event cache was last empty; it is a sibling of the `search-index` directory rather than a file
  inside it, because the SDK owns that directory's layout.
- `create(client, sessionData, isMessageSearchAvailable)` writes the marker straight away when
  there is no event cache yet — a fresh login, or a session that has been healed — since the
  index covers everything by construction, which spares those sessions a pointless clear on their
  first restart.
- Failing to write the marker is logged and never fails the restore: without it the bootstrap
  runs again on the next start.
- When message search is unavailable the index directory and the marker are deleted, marker
  first, so an interrupted deletion cannot leave a later restore trusting a partial index.
- `getBaseClientBuilder` takes an optional `isMessageSearchAvailable` snapshot so that a flag
  flipped mid-restore cannot make the coverage bookkeeping and the built client disagree about
  whether an index exists.
- Nothing outside the search index and the event cache store is touched — the state store, the
  sliding-sync position, the crypto store and the room list are all left alone, sync still resumes
  incrementally, and with `FeatureFlags.MessageSearch` off no session reaches the bootstrap at
  all.

## Motivation and context

The gap is visible from the outside as a defect rather than as a missing feature: in-room search
is present in unencrypted rooms and absent in encrypted ones, reported with screenshots in
element-hq/element-x-android#4149 and closed as a duplicate of the canonical tracker.

Relates to element-hq/element-x-android#3709
Relates to element-hq/element-x-android#6557
Relates to element-hq/element-x-android#7072

## Screenshots / GIFs

No UI change in this pull request.

## Tests

`RustMatrixClientFactoryTest.kt` gains seven cases on plain JUnit with a `TemporaryFolder` rule
over real files (no Turbine, no Robolectric, no Paparazzi — this is filesystem bookkeeping, not
UI): the store and both sidecars are deleted and the marker written when the index cannot cover
the cache; the client is built exactly once and the marker left in place when index directory *and*
marker are both present; the same when there is no event cache file yet, with the marker written
anyway; a healed session is not cleared a second time once the SDK has recreated the store; the
marker is *not* written when the store file cannot be deleted, so the next start retries; a stale
index and its marker are both removed when the flag is off; and `create` with an already-built
client records coverage for a fresh session. `FakeFfiClient.clearCaches()` fails the test if it is
ever reached, and `FakeFfiClientBuilder.withSearchIndexStore` now creates the directory so the
bookkeeping assertions see the same side effect the real SDK produces.

To watch the effect by hand: enable `feature.message_search`, restart once, and history already cached before
the index existed becomes searchable instead of silently missing.

### Scope

Deliberately not done here, and the parts most worth arguing about:

- This is a one-time heal that *destroys* data: deleting the event cache store costs bandwidth
  and makes the first timeline load after the flag is enabled noticeably slower. It is bounded —
  once per session, not per launch — but it is a real cost paid by every user who turns the flag
  on.
- Detection is file presence plus a marker, not a stored schema version. If the marker cannot be
  written, the bootstrap repeats on the next start, which is the safe direction but is still a
  repeated clear.
- It deliberately avoids the SDK's own `clearCaches()`. That call spares the crypto store, where
  the sliding-sync position lives, so the client is left on a server connection that never
  re-sends the wiped room data; and discarding the then contractually-unstable client afterwards
  aborts in a native finalizer. Both failure modes were reproduced on a device before settling on
  file deletion.
- The obvious alternative is that none of this should live in the app. Would the maintainers
  rather this became an SDK-side capability — the index able to backfill from the event cache
  store, or an explicit "reset the event cache" entry point — than an app deleting SDK files by
  name? If so this PR should be dropped rather than reworked, and the coverage hole tracked
  instead; happy to file a linked tracking issue if useful.
- Sessions with no client secret never get an index at all, so they never bootstrap: that covers
  legacy pre-schema-v5 sessions, whose `passphrase` column was added without a backfill, and
  sessions created by a release build between 2024-01-18 and 2024-04-15, when database encryption
  shipped on nightly and debug only.
- Toggling the flag off deletes the index and the marker, so re-enabling forces a full rebuild
  plus another bootstrap. That is intentional — an index that missed events while search was off
  can never catch up on them — but it does mean the flag is not a cheap toggle.
- History comes back through ordinary bounded network pagination rather than an unbounded load
  of a room's data into memory, which is what OOM-crashed the earlier Android proof of concept —
  that is the reason this heals by deleting a file and letting the normal paths refill it, rather
  than by bulk-reading the cache into the indexer.
- Forcing history back over the network also increases exposure to matrix-org/matrix-rust-sdk#6772,
  where `matrix-sdk-search` writes a raw `origin_server_ts` into the tantivy schema and panics on
  a malformed value. Nothing here works around that; it is an upstream fix.

Sender filtering, space scoping, stemming, non-Western tokenisation, partial-word matching and
edits all remain follow-ups: the index tokenises English-first, matches whole words only, and
carries no sender or space dimension.

### Notes for reviewers

This is PR 4 of a four-PR stack. It depends on PR 1, which adds the index and the feature flag,
and is independent of PRs 2 and 3 — it can be reviewed or dropped without touching them. Roughly
340 changed lines, about 230 of them tests, so the reviewable delta is a single commit
(`Rebuild the search index when it cannot cover the event cache`) and around 110 lines of
production code in `RustMatrixClientFactory.kt`; the split exists to keep every part under the
500-addition guidance. Storage reporting was already handled in PR 1, where `getCacheSize()`
counts the `search-index` directory, so the history this bootstrap pulls back into the index stays
visible in the settings figures rather than silently inflating them. Targets `develop`, labelled `PR-Wip`.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Android 16 (API 36), Pixel 10 Pro emulator (arm64-v8a)

## Checklist

    - [ ] No.


hayaksi1:
With all four PRs applied, searching a room end-to-end looks like this — top bar entry point, search screen, and results from the local index.

ss1
ss2
ss3

jmartinesp:
Thanks for the PR. We appreciate the contribution, but this will be handled by the Rust SDK soon, AFAICT. Having it there is a better solution than having 2 different implementations for each platform.

Issue #6878, opened 2026-05-26
[Task] Manually set a custom status


pixlwave:
Please see and discuss the details in the meta issue.

PR #7257, opened 2026-07-22
Plug emoji picker for custom user status

ganfra:
## Content
- Wires the shared emoji picker into the user-status custom-input flow: tapping the emoji circle opens a bottom sheet with the picker (no Recent tab), and picking an emoji updates
  the pending custom status.
  - Along the way, reshapes the emoji picker infrastructure so it can be reused:
    - Migrates libraries/recentemojis into a new libraries/emoji module pair (api + impl) exposing EmojiPickerPresenter / EmojiPickerRenderer.
    - Converts EmojiPickerPresenter to an assisted-factory (EmojiPickerPresenter.Factory { create(getRecentEmojis) }) so callers pick their own GetRecentEmojis 
    - Adds libraries/emoji/test with FakeEmojiPickerState and fakeEmojiPickerPresenter[Factory]() helpers for consumer tests.

## Motivation and context
Part of https://github.com/element-hq/element-x-android/issues/6878

## Tests

  - [ ] Settings → tap "Set a status" → Custom → tap the emoji circle → sheet opens with all categories, no Recent tab → pick an emoji → the circle updates.
  - [ ] Reactions long-press picker in a room still works (categories + Recent tab from Matrix history).

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/customreaction/EmojiItem.kt
```diff
@@ -89,18 +84,13 @@ fun EmojiItem(
...
-    )
+    val description = contentDescription(item, isSelected)
     val hasSkinTones = !item.skins.isNullOrEmpty()
     Box(
         modifier = modifier
             .onPlaced { coordinates ->
                 val position = coordinates.positionInParent()
                 val parentBounds = coordinates.parentLayoutCoordinates?.size ?: return@onPlaced
-                // Calculate the offset of the item inside its parent as a percentage, to be able to set the initial offset
-                // for the selection inside the skin tone picker relative to its size.
```
jmartinesp, line 103: Do we really want to remove these comments? I thought they made the complex long press emoji tone selector logic easier to understand.

libraries/emoji/impl/src/main/kotlin/io/element/android/libraries/emoji/impl/DefaultEmojibaseProvider.kt
```diff
@@ -11,16 +11,28 @@ package io.element.android.libraries.emoji.impl
...
+    @ApplicationContext private val context: Context,
+    dispatchers: CoroutineDispatchers,
+    @AppCoroutineScope appScope: CoroutineScope,
+) : EmojibaseProvider {
+    private val deferred = appScope.async(dispatchers.io, start = CoroutineStart.LAZY) {
         EmojibaseDatasource().load(context)
     }
+
+    override suspend fun getStore(): EmojibaseStore = deferred.await()
 }
```
jmartinesp, line 36: 👌

libraries/emoji/impl/src/main/kotlin/io/element/android/libraries/emoji/impl/picker/DefaultEmojiPickerPresenter.kt
```diff
@@ -86,7 +93,9 @@ class DefaultEmojiPickerPresenter(
         val searchQuery = queryState.text.toString()
         LaunchedEffect(searchQuery, data) {
             if (searchQuery.isEmpty() || data.allEmojis.isEmpty()) {
-                emojiResults = SearchBarResultState.Initial()
+                if (emojiResults !is SearchBarResultState.Initial) {
+                    emojiResults = SearchBarResultState.Initial()
+                }
```
jmartinesp, line 100: Not that important, but I thought this would work the same without the extra check? `emojiResults` is backed by the `MutableState` and that performs an equality check before assigning a new value.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/userstatus/UserStatusPresenter.kt
```diff
@@ -19,57 +19,90 @@ import androidx.compose.runtime.rememberCoroutineScope
...
                     val current = pickerState as? UserStatusPickerState.CustomInput ?: return
                     pickerState = current.copy(emoji = event.emoji)
                 }
+                // Emoji-picker visibility (only meaningful within CustomInput mode)
+                UserStatusEvent.OpenEmojiPicker -> isEmojiPickerVisible = true
+                UserStatusEvent.DismissEmojiPicker -> isEmojiPickerVisible = false
             }
         }
 
+        val effectivePickerState = when (val current = pickerState) {
```
jmartinesp, line 84: Maybe add a comment here? This part is a bit confusing until you realise we're overriding the computed value in some cases.

jmartinesp (approved):
LGTM, thanks! Just a note about the deleted comments in `EmojiItem`.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/customreaction/EmojiItem.kt
ganfra: Restored the comments in EmojiItem.kt.

libraries/emoji/impl/src/main/kotlin/io/element/android/libraries/emoji/impl/picker/DefaultEmojiPickerPresenter.kt
ganfra: Good spot — I checked and the guard is actually required today: `SearchBarResultState.Initial` is a plain class with no `equals` override, so `structuralEqualityPolicy` sees each new `Initial()` as a distinct value and emits it. Without the guard, tests here (and callers relying on emission counts) fail with extra emissions. I've kept the guard and added a comment explaining why. The proper fix is to make `Initial`/`NoResultsFound` `data object`s on `SearchBarResultState<out T>` (same pattern as `AsyncData<out T>` + `Uninitialized : AsyncData<Nothing>`) — will do that in a follow-up since it also requires updating a bunch of `skipItems(N)` counts in the invitepeople/startchat tests.

features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/userstatus/UserStatusPresenter.kt
ganfra: Added a comment above the block explaining the override.

PR #7260, opened 2026-07-23
Remove no longer temporary strings

jmartinesp:
## Content

- Remove some strings in `temporary.xml` files that are now in Localazy.
- Fix string containing 'Element' instead of the app name placeholder.

## Motivation and context

The sync strings CI flow was failing because we have 2 conflicting values in `temporary.xml` and `localazy.xml`.

## Tests

The app should build.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

- [x] I am aware of the [etiquette](https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#etiquette).
- This PR was made with the help of AI:
    - [ ] Yes. In this case, please request a review by Copilot.
    - [x] No.
- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly defines what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #7261, opened 2026-07-23
Localazy: improve search, parallelize checks

jmartinesp:
## Content

- Make searching for strings and translation files quicker by fine-tuning the search command.
- Do most of the processes (check, format, delete, etc.) in parallel. Download seems to be the only one we can't paralellize.

Built on top of #7260 to apply the fixes added there.

## Motivation and context

Syncing strings is way slower than it should be, since everything is sequential, over > 100 files.

## Tests

Run `./tools/localazy/downloadStrings.sh --verbose` and check the output.

## Checklist

    - [x] No.


features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/chooseaccountprovider/ChooseAccountProviderView.kt
ganfra: We are not using the same logic everywhere to grab the appName, is it expected?

tools/localazy/downloadStrings.sh
```diff
@@ -33,16 +59,56 @@ localazy download --config ./tools/localazy/localazy.json
...
+  fi
+  ./tools/localazy/formatXmlResourcesFile.py "$xml"
+}
+
+checkForbiddenTerms() {
+  local xml="$1"
+  if [[ $verbose == 1 ]]; then
+    echo "Checking forbidden terms in $xml"
+  fi
+  ./tools/localazy/checkForbiddenTerms.py "$xml" || true
```
ganfra: We are not aborting the process here if there is a forbidenTerm I think?

ganfra (approved):
Small remarks, otherwise LGTM

features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/chooseaccountprovider/ChooseAccountProviderView.kt
jmartinesp: I think that's alright, in the end everything is using the injected `BuildMeta` values either from a presenter or through `LocalBuildMeta`. I can change it so it always uses `LocalBuildMeta`, I guess.

tools/localazy/downloadStrings.sh
jmartinesp: I changed this on purpose, because otherwise you'd sync every file from localazy, format them all, then check them and find a single forbidden term, fix it, sync again, format again, then a different one is found, etc.

That said, even if we find all forbidden terms, we should fail the sync in that case... I'll see if there's an easy way to achieve this.

jmartinesp:
@ganfra could you take a look at https://github.com/element-hq/element-android-enterprise/pull/36 too so we can merge this PR?

bmarty:
Thanks for the improvement! It was annoying the the check was just logging warnings.

PR #7267, opened 2026-07-23
Try making `HeaderFooterView` more robust

jmartinesp:
## Content

Create a new custom layout `HeaderFooterLayout` and use it for `HeaderFooterView` when `isScrollable` is enabled. It ensures the content inside the scrollable column is at least as large as the available screen space.

When not using `isScrollable`, a similar strategy to what we had before is used, so nothing really changed there.

## Motivation and context

What we had before (creating a scrollable column with `height(IntrinsictSize.Max)` was needed to make the contents as large as possible, but it had weird side effects in some screens, where it allowed you to scroll up to twice the size of the content and you could just make it disappear and have a blank screen instead.

## Screenshots / GIFs

Not many, which is a good sign! The UI still looks the same, but it should work better.

## Tests

Try using different screens where this component is used, both scrollable and not.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.
- [ x Pull request title will be used in the release note, it clearly defines what will change for the user

### DIFF
--- a/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/pages/HeaderFooterLayout.kt
+++ b/libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/pages/HeaderFooterLayout.kt
@@ -63,33 +63,27 @@ fun HeaderFooterLayout(
             ) {
                 Layout(
                     content = {
-                        movableHeader()
-                        movableContent()
+                        Box { movableHeader() }
+                        Box { movableContent() }
                     },
                     modifier = Modifier.verticalScroll(rememberScrollState()),
                     measurePolicy = { measurables, constraints ->
                         val actualConstraints = constraints.copy(minWidth = 0, minHeight = 0, maxHeight = newHeight)
-                        val headerPlaceable = measurables.firstOrNull()?.measure(actualConstraints)
+                        val headerPlaceable = measurables[0].measure(actualConstraints)
 
-                        val contentPlaceable = if (headerPlaceable != null && measurables.size > 1) {
-                            val availableContentHeight = max(1, actualConstraints.maxHeight - headerPlaceable.height)
-                            val contentConstraints = actualConstraints.copy(minHeight = availableContentHeight, maxHeight = Constraints.Infinity)
-                            measurables[1].measure(contentConstraints)
-                        } else {
-                            null
-                        }
+                        val availableContentHeight = max(1, actualConstraints.maxHeight - headerPlaceable.height)
+                        val contentConstraints = actualConstraints.copy(minHeight = availableContentHeight, maxHeight = Constraints.Infinity)
+                        val contentPlaceable = measurables[1].measure(contentConstraints)
 
-                        val headerHeight = headerPlaceable?.height ?: 0
-                        val contentHeight = contentPlaceable?.height ?: 0
+                        val headerHeight = headerPlaceable.height
+                        val contentHeight = contentPlaceable.height
                         layout(actualConstraints.maxWidth, headerHeight + contentHeight) {
                             var yPosition = 0
 
-                            headerPlaceable?.let {
-                                it.placeRelative(0, yPosition)
-                                yPosition += it.height
-                            }
+                            headerPlaceable.placeRelative(0, yPosition)
+                            yPosition += headerPlaceable.height
 
-                            contentPlaceable?.placeRelative(0, yPosition)
+                            contentPlaceable.placeRelative(0, yPosition)
                         }
                     }
                 )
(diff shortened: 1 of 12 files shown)


libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/pages/HeaderFooterLayout.kt
```diff
@@ -0,0 +1,108 @@
...
+fun HeaderFooterLayout(
+    header: @Composable () -> Unit,
+    footer: @Composable () -> Unit,
+    isScrollable: Boolean,
+    contentInsetsPadding: PaddingValues,
+    modifier: Modifier = Modifier,
+    content: @Composable () -> Unit,
+) {
+    val movableHeader = remember { movableContentOf(header) }
+    val movableContent = remember { movableContentOf(content) }
```
ganfra: Do we really need `movableContent` here? I'm not sure `isScollable` is going to change dynamically

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/pages/HeaderFooterLayout.kt
```diff
@@ -0,0 +1,108 @@
...
+                    .padding(contentInsetsPadding)
+                    .weight(1f, fill = true)
+                    .onSizeChanged {
+                        newHeight = it.height
+                    }
+            ) {
+                Layout(
+                    content = {
+                        movableHeader()
+                        movableContent()
```
ganfra: Maybe use explicit slot IDs, or wrap slot invocation in a Box:
  ```
content = {
      Box { movableHeader() }
      Box { movableContent() }
  } 
```
to prevent bad usage (like content emitting multiple components)?

ganfra (approved):
Small remarks otherwise LGTM

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/pages/HeaderFooterLayout.kt
jmartinesp: Yeah, it's not, but the linter was being a pain. I can suppress the warning instead if you prefer that.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/pages/HeaderFooterLayout.kt
jmartinesp: > or wrap slot invocation in a Box:

Oh, good idea. That should help simplifying the logic.

> Maybe use explicit slot IDs

Doesn't that require `SubcomposeLayout`? I was trying to avoid using those.

jmartinesp: https://github.com/element-hq/element-x-android/pull/7267/commits/7e1dc0ce6e861bfe58ba5dffa2a8cf784427a570

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/pages/HeaderFooterLayout.kt
jmartinesp: https://github.com/element-hq/element-x-android/pull/7267/commits/884583feed965592efb5ed824a66474d9f919fce

jmartinesp:
The new screenshots are weird: the changes I made shouldn't have caused them, and it's exactly what I wanted to avoid 🫤 . I'll try to check what's causing this side-effect.

ganfra:
> The new screenshots are weird: the changes I made shouldn't have caused them, and it's exactly what I wanted to avoid 🫤 . I'll try to check what's causing this side-effect.

Maybe removing `moveableContent`?  but not sure why it'd

jmartinesp:
Looks like it's using the `Box` wrappers that's causing it. I'll have to revert that, although I'm not sure why that's a problem 🫤 .

jmartinesp:
Ok, no issues with screenshots tests this time. It's a shame we can't just use `Box` to simplify the layout calculations.

libraries/designsystem/src/main/kotlin/io/element/android/libraries/designsystem/atomic/pages/HeaderFooterLayout.kt
jmartinesp: This had to be reverted in the end: https://github.com/element-hq/element-x-android/pull/7267/commits/5e12f389a82d34e69f914049b820fdb3a7ce81d5

PR #7280, opened 2026-07-27
Fix media viewer for medias not related to events displaying invalid state

jmartinesp:
## Content

- Remove the `eventId` param for the `ContentScannerService.scan` method, since it wasn't really used and it prevented some medias from being scanned.
- Make sure we validate both the actual media and the thumbnail in the media viewer before deciding if a media is valid or invalid.
- Add some tests, fix others.

## Motivation and context

@ganfra mentioned he ran into this issue.

## Tests

- Go into a room details/user profile screen and tap on the avatars.
- Check they are scanned if content scanner is enabled and display the right value (valid/invalid depending on the scanner config, or always valid if content scanner is disabled).
- Check the media viewer works fine in other cases.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.


libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerDataSource.kt
```diff
@@ -242,18 +242,15 @@ class MediaViewerDataSource(
...
     }
 
-    fun validateMedia(
-        eventId: EventId,
-        mediaSource: MediaSource,
-    ) {
+    fun validateMedia(mediaSource: MediaSource, thumbnailMediaSource: MediaSource?) {
         val validationState = mediaValidationState[mediaSource.safeUrl] ?: return
-        val currentState = validationState.getCurrentMediaState(mediaSource.safeUrl)
+        val currentState = validationState.getCurrentOverallState()
```
ganfra, line 247: `getCurrentOverallState` is not just about the current mediaSource, right? Meaning for gallery it's going to check unrelated media?

jmartinesp: In this case I think it's fine because the way the overall logic works and the `if` below work well together: we don't want to scan again if we're already scanning any of the provided medias or if they all are validated, and that matches overall vs per-media logic.

ganfra (approved):

Issue #7199, opened 2026-07-13
[Task] See call participants in the People list


robintown:
Please see and discuss the details in the meta issue.

thirumani-vihaan:
Hey @robintown! Would it be alright if I worked on this issue? I read through the meta issue, and looking at the codebase architecture, it seems we'll need to observe the active call state (mapping active call participants to the RoomMemberListState) so the RoomMemberListPresenter can pass an isOnCall flag down to the Compose UI for rendering the visual indicator.

If no one else is currently working on this, I'd be happy to put together a PR for this! Let me know if that approach sounds right to you.

Eklavya47:
Hey @robintown, I would like to work on this task if it is not assigned to anyone else. Please let me know

Atul-Khandekar:
hey @robintown , if no one's working on this issue, i would like to pick it up. already looking at the RoomMemberListPresenter state flow. can i grab this?

PR #7300, opened 2026-07-29
[Task] Show active call participants in the people list #7199

Eklavya47:
## Content

I added a video call icon next to call participants in the People List and they are sorted to be appeared at the top of list

## Motivation and context

Implements task #7199 : See call participants in the People list. 

## Screenshots / GIFs

A Composable Preview has been updated to cover the UI change

## Tests

- Added a unit test to verify that active call participants are displayed first in the joined members list.
- Verified the updated Composable Preview displays the call icon correctly.
- Ran project quality checks

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

    - [ ] No.


features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListPresenter.kt
```diff
@@ -93,20 +94,32 @@ class RoomMemberListPresenter(
                 val info = room.info()
                 if (members.getOrDefault(RoomMembershipState.JOIN, emptyList()).size < info.joinedMembersCount / 2) {
                     // Don't display initial room member list if we have less than half of the joined members:
-                    // This result will come from the timeline loading membership events and it'll be wrong.
+                    // This result will come from the timeline loading membership events, and it'll be wrong.
                     return@withContext
                 }
+                val activeParticipants = info.activeRoomCallParticipants
```
jmartinesp: Maybe rename this to `activeCallParticipants` since otherwise `activeParticipants` could mean 'participants of the room'. I'd do the same on `RoomMember.withIdentityState`.

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListPresenter.kt
```diff
@@ -93,20 +94,32 @@ class RoomMemberListPresenter(
...
+                val comparator = Comparator<RoomMember> { a, b ->
```
jmartinesp: Maybe extract the comparator to a private function in the presenter to make the `present` method a bit easier to read? You could then wrap the call in a remember block that changes when the call participants change and pass the the participant list to this functino to generate a new comparator when it changes.

Something like:

```kotlin
private fun memberListComparator(activeCallParticipants: List<UserId>): Comparator<RoomMember> { ... }

@Composable
fun present() {
    ...
    val activeCallParticipants = info.activeRoomCallParticipants
    val comparator = remember(activeCallParticipants) { roomMemberListComparator(activeCallParticipants) }
    ...
}
```

jmartinesp (changes requested):
Thanks, looks good in general, but I think we can improve the code a bit.

Also, after these are fixed you'll need to record new screenshot tests for the CI to pass.

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListPresenter.kt
Eklavya47: I have renamed them

features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListPresenter.kt
Eklavya47: Extracted the comparator to a private function and added the remember block for it as well. Also I recorded new screenshot tests as well.

Eklavya47:
Hi, can you guide me with the codecov check which is unsuccessful. I am not sure what to do here.

jmartinesp:
Sorry for the late reply, I wasn't around last week. 

> Hi, can you guide me with the codecov check which is unsuccessful. I am not sure what to do here.

That's just because the coverage in the file is now a bit smaller than it used to, it's ok, we can merge it like this.

jmartinesp (approved):

PR #7302, opened 2026-07-30
Add support for checking the ESS config in Element Pro

jmartinesp:
## Content

- Create `ElementWellKnownSource` enum to allow deciding where to get the remote enterprise config from.
- Add `RemoteEnterpriseConfigService` to make this decision (this is implemented mostly in the enterprise code).
- Add `UrlContentFetcher` which is a component to fetch data from any URL using a GET request and implement 2 versions: one with a temporary and unauthenticated SDK client for unauthenticated requests and another one with `RustMatrixClient` so we perform the request using the authenticated client in the SDK.
- Remove `SessionElementWelknownRetriever` and move the logic to `WellknownRetriever`, allowing us to use any of the 2 `UrlContentFetcher` instances to actually fetch the data without having to create 2 variations of this class for the 2 DI scopes.
- Use `RemoteEnterpriseConfigService` as much as possible. The only exception is in `DefaultAccountProviderAccessControl` because we want to always check the legacy element well known URL there to check for the `requires_element_pro` config there.

The PR in enterprise repo is https://github.com/element-hq/element-android-enterprise/pull/38.

## Motivation and context

https://github.com/element-hq/element-enterprise/issues/297

## Tests

With an Element Pro version, check the new endpoint is checked, then the existing well known one is.

You can also check unauthenticated ones go through OkHttp and authenticated ones from the SDK, based on the logs.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.


jmartinesp:
There is an implementation without OkHttp (creating temporary Rust clients like iOS does) in https://github.com/element-hq/element-x-android/tree/feat/add-ess-config-endpoint-support-with-assistedinject.

jmartinesp:
I've merge the branch mentioned above, rebased the code and split it into new commits so it's hopefully easier to understand. I think you can review it now @ganfra .

FWIW the temporary client approach is what iOS does and should help keep the network stack consistent in the authenticating and already authenticated scopes.

libraries/wellknown/impl/src/main/kotlin/io/element/android/libraries/wellknown/impl/DefaultWellknownRetriever.kt
```diff
@@ -9,65 +9,130 @@
...
+        // The ESS_CONFIG source is cached with a prefix to differentiate it from the WELLKNOWN_ENDPOINT.
+        val prefix = when (source) {
+            EnterpriseRemoteConfigSource.ESS_CONFIG -> "ess_config"
+            EnterpriseRemoteConfigSource.WELLKNOWN_ENDPOINT -> null
+        }
+        val store = elementWellknownStoreFactory.create(prefix)
+        return when (val cacheData = store.get(checkedHost)) {
+            is WellknownRetrieverResult.Success -> {
+                Timber.d("Using cached well-known for domain $checkedHost")
+                fetchElementWellKnown(checkedHost, source, store)
```
ganfra: Is the fetch here necessary? And if yes, should we launch the fetch on a separate coroutine?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/di/SessionMatrixModule.kt
```diff
@@ -31,7 +32,7 @@ import io.element.android.libraries.matrix.api.verification.SessionVerificationS
 import kotlinx.coroutines.CoroutineScope
 
 @BindingContainer
-@ContributesTo(SessionScope::class)
+@ContributesTo(SessionScope::class, replaces = [GetUrlResolver::class])
```
ganfra: Do we need this `replaces`? This sounds a bit weird

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/GetUrlResolver.kt
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2026 Element Creations Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.matrix.api
+
+fun interface GetUrlResolver {
```
ganfra: `GetUrl` is a bit misleading? On first hand I though it was just returning an `url `  I know it comes from the rust sdk name, but could be something like `UrlContentFetcher` or `HttpGetExecutor`?

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/TemporaryMatrixClient.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import java.lang.AutoCloseable
+
+/**
+ * A Matrix client in an 'unauthenticated' state that can still be used to resolve URLs.
+ *
+ * Note: this client implements an [AutoCloseable] interface, so using [AutoCloseable.close] or [AutoCloseable.use] will clean up any resources associated
+ * with it, such as temporary files.
+ */
+interface TemporaryMatrixClient : GetUrlResolver, AutoCloseable
```
ganfra: Maybe it should be called `UnauthenticatedMatrixClient`? not sure... 
Also, does it need to be part of the `api` module?

libraries/wellknown/impl/src/main/kotlin/io/element/android/libraries/wellknown/impl/DefaultWellknownRetriever.kt
```diff
@@ -9,65 +9,130 @@
...
-                    elementWellknownStore.update(domain, json.encodeToString(result))
-                    WellknownRetrieverResult.Success(result.map())
-                } catch (e: Exception) {
-                    // Is it a 404?
-                    Timber.e(e, "Failed to retrieve Element well-known data for $baseUrl")
-                    if ((e as? HttpException)?.code() == HttpURLConnection.HTTP_NOT_FOUND) {
-                        WellknownRetrieverResult.NotFound
-                    } else {
-                        WellknownRetrieverResult.Error(e)
+        val checkedHost = URL(host.ensureProtocol()).host ?: host
```
ganfra: I think `host`  never returns null, so the fallback doesn't work

ganfra (approved):
Thanks, looking solid. Just small remarks!

libraries/wellknown/impl/src/main/kotlin/io/element/android/libraries/wellknown/impl/DefaultWellknownRetriever.kt
jmartinesp: Oops, you found some test code I forgot to remove. Thanks!

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/TemporaryMatrixClient.kt
jmartinesp: > Maybe it should be called UnauthenticatedMatrixClient? not sure...

That's what I had at first, but I thought the `temporary` part was more important... now I'm not so sure.

> Also, does it need to be part of the api module?

I think so, otherwise how would you use the `TemporaryMatrixClientFactory` to return one of these?

jmartinesp:
@ganfra I uploaded some changes to fix the issues you mentioned, in case you want to double check. Also, remember the [enterprise PR](https://github.com/element-hq/element-android-enterprise/pull/38) should be reviewed too before this one can be merged.

PR #7316, opened 2026-08-03
Add a copy button to code blocks in the timeline

hayaksi1:
## Content

Code blocks in a message now render their own chrome: the language label and a separator at the
top, and a full-width copy row at the bottom, both drawn inside the block's own box. Tapping the
copy row puts that block's code on the clipboard — just that block, not the whole message.

A message is rendered by a single `TextView`, so a code block is a `CodeBlockSpan` rather than a
composable and there is no layout slot to reserve for the chrome. A `LineHeightSpan` across the
block grows its first line upward and its last line downward instead, which leaves the message's
character offsets and the copied text untouched. Two `StaticLayout` behaviours shape how: it
consults every span in a paragraph for every line of that paragraph, so the span picks its lines
inside `chooseHeight` by the character they hold rather than by its attachment range; and it seeds
a wrapped line from the previous line's chosen metrics, so the span recognises the exact metrics it
produced and takes the growth back off a line they leaked onto. The rows are sized to the text
layout's width — the width the block's background is actually drawn at.

The language is read from the formatted body's DOM (`<pre><code class="language-x">`), because it
no longer exists by the time the DOM has become spans, and is matched back to each block by
position. The HTML sanitizer now keeps `class` on `code` — the attribute the spec allows for
`language-*` — so the class survives into that DOM. Blocks that declare no language reserve no
header.

On Android 12 and below a "copied to clipboard" snackbar is shown; on 13+ the system shows its own
confirmation, so we stay quiet.

No new strings — this reuses `CommonStrings.action_copy` and
`CommonStrings.common_copied_to_clipboard`.

## Motivation and context

Copying a code snippet from a message currently means copying the whole message and trimming it by
hand. Element Web has had a copy button on code blocks for a long time
(element-hq/element-web#1974), and Telegram, Discord, Slack and GitHub all do the same — the
interaction is what people expect a code block to offer.

## Tests

- `CodeBlockOverlayTest` covers language extraction from the HTML document, chrome insertion, block
  detection, overlay geometry, and the per-line heights of wrapped, multi-paragraph, single-line
  and unlabelled blocks — under native-graphics Robolectric, since legacy graphics never wraps a
  line and would leave the wrap assertions vacuous.
- `ToHtmlDocumentTest` covers the sanitizer keeping the `language-*` class on a code block.
- Manually: sent messages containing a fenced block with a language, a fenced block without one,
  two blocks in one message, a block mixed with surrounding prose, and a long one-line block that
  wraps over several lines. Checked that the copied text is the block's code exactly, with no
  chrome and no surrounding message text, and that long-press "Copy" on the message still copies
  the whole message.

## Tested devices

- [ ] Physical
- [x] Emulator

## Checklist

- [x] I am aware of the etiquette.
    - [ ] No.
- [x] Accessibility has been taken into account.


jmartinesp:
Thanks for the contribution. This approach is interesting, but IMO it feels too much like hacking the UI and doesn't seem to be the right solution.

Also, I can consistently see some flickering when the layout is animating:

https://github.com/user-attachments/assets/2a694b26-48e3-4006-8504-c6dec8ce61e1

What we should do instead is create a new way to render UI components for the timeline items: right now we're just using a `TextView` under the hood and using spans parsed using the rich text editor's HTML parser to draw all the rich text, but that's not really compatible with interactive UI like we'd have to add here. 

That said, while we've been talking about it for a while, it's not something we're planning to work on soon unless we find some extra time for it: getting basic rendering working would probably be quite easy, but there are quite a few corner cases with measuring texts and placing the other message bubble UI components and paddings that can be quite tricky.

@ganfra I'll let you decide whether to keep the PR open or close it.

hayaksi1:
@jmartinesp 

The concern about extending the current `TextView`-and-spans renderer for interactive UI is understood.

Code blocks represent content users commonly need to reuse. When someone sends a fenced block, it is usually code, a command, or structured text that should be copied independently from the rest of the message. Element Desktop already supports this, and for a collaboration client, the same interaction is valuable on Android/iOS.

Element also offers an alternative to services where users give up control of their data to a large platform. It gives people the freedom to run and control their own communication. Contributions like this aim to make that experience more complete by adding useful native features where possible.

An update has been pushed to address the issue shown in the video. The code-block chrome now reads its geometry during layout rather than composition, so it follows the `TextView` in the same frame while the timeline item is animating. The overlay state is also retained when an item is rebuilt, which removes the blink caused by resetting it between layouts.

The implementation keeps the interactive UI outside the text content itself: message text, character offsets, and the existing long-press copy behaviour remain unchanged. The copy target is positioned from the `TextView` layout and copies only the relevant code block.

I understand there may still be concerns about the design or implementation. I am open to changes and would appreciate specific feedback on what should be adjusted. Feedback from previous Element contributions has helped improve the work, and I am happy to continue refining this PR if the overall approach is acceptable.

Issue #7289, opened 2026-07-28
Custom map_style_url from .well-known is broken: MapTiler API key is appended unconditionally


Norrodar:
### Steps to reproduce

1. Serve a `m.tile_server` block from `/.well-known/matrix/client`, as read since
   #6886:

```json
   "m.tile_server": {
     "map_style_url": "https://api.maptiler.com/maps/streets-v2/style.json?key=YOUR_KEY"
   }
```

2. Sign in to Element X Android with an account on that homeserver
3. Open a shared location from the timeline (tap the preview to go fullscreen)

### Outcome

**Expected:** the fullscreen map renders using the configured style.

**What happened:** the map area is black. Only the MapLibre logo, the (i) button
and the locate button render; the attribution dialog lists nothing but
"MapLibre Android", i.e. no style was loaded at all.

The static preview in the timeline still renders correctly — see below for why
those two differ.

### Cause

`MapTilerTileServerStyleUriBuilder.build()` appends the API key outside the
`if/else`, so it is added to the custom URL as well:

`features/location/api/src/main/kotlin/io/element/android/features/location/api/internal/MapTilerTileServerStyleUriBuilder.kt`

```kotlin
return buildString {
    if (customMapStyleUrl.isNullOrBlank()) {
        val mapId = if (darkMode) darkMapId else lightMapId
        append("$baseUrl/$mapId/style.json")
    } else {
        append(customMapStyleUrl)
    }
    append("?key=$apiKey")   // <-- also applied to customMapStyleUrl
}
```

A custom URL that already carries its own key therefore becomes:

```
https://api.maptiler.com/maps/streets-v2/style.json?key=YOUR_KEY?key=<build-key>
```

The second `?` does not start a new query string, so the value of `key` is
literally `YOUR_KEY?key=<build-key>` — an invalid key. MapTiler rejects it,
MapLibre loads no style, and the result is a black surface.

The existing unit tests pin this behaviour in
`MapTilerTileServerStyleUriBuilderTest.kt`:
`"https://custom.url/style.json"` is expected to produce
`"https://custom.url/style.json?key=anApiKey"`.

### Suggested fix

Move `append("?key=$apiKey")` into the `isNullOrBlank()` branch, so the key is
only added to URLs the app builds itself, and update the two custom-URL tests
accordingly.

Note this also means that today a custom style is always requested with the
app's built-in key rather than the operator's — so a style private to the
operator's own MapTiler account cannot work even without the double-`?` issue.

[...]

PR #7340, opened 2026-08-09
Do not append the application MapTiler key to a custom map style url

hayaksi1:
### Problem

When a homeserver publishes a `map_style_url` in its `.well-known` file, the app appends its own
MapTiler API key to it before loading the style. The resulting URL has two query strings —
`https://tiles.example.org/style.json?key=operatorKey?key=applicationKey` — which the tile server
rejects, so the map fails to load for every user of that homeserver. There is no way for an operator
to opt out, because the key is appended unconditionally.

### Fix

- `MapTilerTileServerStyleUriBuilder.build()` now appends `?key=<application key>` only to the URL
  it composes itself from `MAPTILER_BASE_URL` and the light/dark map ids. A non-blank
  `customMapStyleUrl` is returned verbatim, so whatever credentials the operator put in it survive.
- The blank/`null` path is unchanged, so the bundled MapTiler configuration keeps working exactly as
  before.

## Motivation and context

A custom style URL is the mechanism an operator uses to point the app at their own tile server, and
it is expected to be self-sufficient — this is how Element Web treats it, and how the self-hosted
tile server documentation describes it. Injecting the application's key into a URL the app did not
build both breaks the request and leaks a credential to a third-party host.

Relates to #7289.

## Tests

- `MapTilerTileServerStyleUriBuilderTest` — `custom map uri light` / `custom map uri dark` now
  assert the custom URL is returned unchanged.
- New: `custom map uri with its own api key is not altered` reproduces the report — a MapTiler URL
  that already carries `?key=` must not gain a second one.
- New: `custom map uri with an existing query is not altered` covers a self-hosted URL with an
  unrelated query string.
- New: `blank custom map uri falls back to the built-in style`, and the untouched `light map uri` /
  `dark map uri` cases, guard that the built-in path still receives the application key.

What these do not prove: no test loads a real style document, so this verifies the URL that is
built, not that any particular tile server accepts it.

### Scope

**This is not a change to the static-map URL builder.** `MapTilerStaticMapUrlBuilder`, mentioned in
the same issue thread, has a different composition problem and is deliberately left alone.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/location/api/src/main/kotlin/io/element/android/features/location/api/internal/MapTilerTileServerStyleUriBuilder.kt
+++ b/features/location/api/src/main/kotlin/io/element/android/features/location/api/internal/MapTilerTileServerStyleUriBuilder.kt
@@ -33,10 +33,6 @@ internal class MapTilerTileServerStyleUriBuilder(
             val mapId = if (darkMode) darkMapId else lightMapId
             "$baseUrl/$mapId/style.json?key=$apiKey"
         } else {
-            // A custom style url comes from the homeserver .well-known and is expected to be complete:
-            // it already carries the operator's own credentials, if any. Appending our api key to it
-            // adds a second query string and breaks the request.
-            // See https://github.com/element-hq/element-x-android/issues/7289
             customMapStyleUrl
         }
     }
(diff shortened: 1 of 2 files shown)


bmarty (approved):
Thanks. This seems to match what's Element X iOS is doing, see https://github.com/element-hq/element-x-ios/pull/5633/changes

Issue #4328, opened 2025-02-27
Strikethrough with <s> instead of <del> will be rendered as normal text


SteffoSpieler:
### Steps to reproduce

1. Send a message that uses [strikethrough via \<s\>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/s) (e.g. with Cinny)
2. See the message being rendered like normal text.

#### What did you expect?
I'd expect the text to be stroked through.

#### What happened instead?
The text renders without any visual change.

### Your phone model

Google Pixel 7 Pro

### Operating system version

Android 15

### Application version and app store

Element X version 25.02.0

### Homeserver

Synapse 1.124.0

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

PR #7341, opened 2026-08-09
Render <s> and <strike> as strikethrough instead of dropping the formatting

hayaksi1:
### Problem

A message whose `formatted_body` uses `<s>` or `<strike>` for strikethrough renders as ordinary
text: the strikethrough is gone with no indication that the sender applied any formatting. Only
`<del>` works today. `<strike>` is part of the allowed tag list in the Matrix specification, and
several clients — Cinny among them — send `<s>`, so this affects real messages arriving from other
clients.

### Fix

- `CustomHtmlToDomParser` now allows `s` and `strike` through its Jsoup `Safelist`, so the sanitizer
  stops unwrapping them before the message is ever parsed.
- Immediately after sanitizing, both tags are renamed to `del` in the parsed document. This is what
  makes the fix safe: the span parser recognises `del` only, and its fallback branch does not
  recurse into the children of an unknown tag — so allowing the tags through *without* normalising
  them would drop the text itself, not just its styling.

## Motivation and context

`<strike>` is in the Matrix spec's list of permitted tags in `org.matrix.custom.html` bodies, so
discarding it loses information the sender deliberately included. `<s>` is not in that list, but
supporting it costs nothing here and fixes interoperability with clients that emit it.

Relates to #4328.

## Tests

- `ToHtmlDocumentTest` — new `a s tag is kept and normalised to del` and `a strike tag is kept and
  normalised to del`: each asserts the document contains `<del>gone</del>` **and** that
  `document.text()` is still `"gone"`. The second assertion is the regression guard against the span
  parser's fallback branch silently discarding children.
- New `a del tag is left as is` — proves the normalisation does not disturb the tag that already
  worked.
- New `a s tag nested in another tag is normalised too` — covers markup inside a `<p>`, and asserts
  no `s`/`strike` element survives anywhere in the document.

What these do not prove: they assert on the parsed `Document`, not on the rendered spans, so they
show the tag reaches the renderer in a form it understands — not that the pixels show a line through
the text.

### Scope

**This is not a change to the rich text editor or to what the composer sends.** Element X still
sends `<del>`; this only affects how incoming markup from other clients is interpreted.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/ToHtmlDocument.kt
+++ b/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/ToHtmlDocument.kt
@@ -69,9 +69,6 @@ private object CustomHtmlToDomParser {
         val outputSettings = OutputSettings().prettyPrint(false).indentAmount(0)
         val cleanHtml = Jsoup.clean(html, "", safeList, outputSettings)
         return Jsoup.parse(cleanHtml).also { document ->
-            // `del` is the only strikethrough tag the renderer knows about, so normalise the two
-            // others to it. Without this the content of the tag is dropped, not just its styling.
-            // See https://github.com/element-hq/element-x-android/issues/4328
             document.select("s, strike").forEach { it.tagName("del") }
         }
     }
@@ -85,7 +82,6 @@ private object CustomHtmlToDomParser {
             "em",
             "u",
             "del",
-            // Strikethrough aliases, normalised to `del` in [document]
             "s",
             "strike",
             "code",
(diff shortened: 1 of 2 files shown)


libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/ToHtmlDocument.kt
```diff
@@ -80,6 +85,9 @@ private object CustomHtmlToDomParser {
             "em",
             "u",
             "del",
+            // Strikethrough aliases, normalised to `del` in [document]
```
frebib: Drop this comment

frebib:
Can you provide some before/after screenshots for this change and the other PRs that make visual fixes?
I'd also recommend manually toning back some of the excessive commentary left by your LLM. It makes the code hard to read. Nice fixes though!

bxdxnn (commented):
This should be done in https://github.com/element-hq/matrix-rich-text-editor I believe.

bmarty:
@hayaksi1 can you try to implement the fix at https://github.com/element-hq/matrix-rich-text-editor ?

hayaksi1:
@bmarty I’ve opened [matrix-rich-text-editor#146](https://github.com/element-hq/matrix-rich-text-editor/pull/146) to add support for these tags in the shared editor/parser as well.

If this change should be handled there instead, please feel free to close this PR.

Issue #5383, opened 2025-09-20
exclusive categories simultaneous click


Destinyg133:
### Steps to reproduce

1. Where are you starting? What can you see?
when clicking fsvorites and invites or people and room simultaneously 

![Image](https://github.com/user-attachments/assets/62cb1e19-eb99-4838-8533-c1105f105a62)
![Image](https://github.com/user-attachments/assets/8d059043-3c22-439d-8e07-400d9d1f7eda)

#### What did you expect?

Not to be triggered since they are not compatibile with each other

#### What happened instead?

Both are pressed

### Your phone model

xiaomi mi10t

### Operating system version

android 12

### Application version and app store

EX 25.09.1

### Will you send logs?

No

### Are you willing to provide a PR?

No

PR #7342, opened 2026-08-09
Ignore a room list filter that is incompatible with an already selected one

hayaksi1:
### Problem

Some room list filters are mutually exclusive — *People* against *Rooms*, and *Invites* against
everything else. The UI expresses this by hiding the incompatible chips once a filter is selected,
but nothing prevents the selection itself. If two incompatible chips are tapped in the same frame,
before the list has had a chance to recompose, both filters are applied. The result is a room list
that can never match anything, and the offending chip is hidden, so the user cannot see what is
filtering their rooms or tap it again to undo it.

### Fix

- `DefaultFilterSelectionStrategy.select()` now ignores a filter that is incompatible with one
  already selected, so the write path enforces the same rule `buildFilters()` already applies when
  it decides which chips to show. The currently applied filters win; the late tap is dropped.
- `toggle()`, `deselect()` and `clear()` are unchanged, as is the `FilterSelectionStrategy`
  interface.

## Motivation and context

`RoomListFilter.incompatibleFilters` was only ever consulted when *rendering* the unselected chips.
Enforcing it where the selection is actually mutated closes the window between a tap and the
recomposition that hides the chip.

Relates to #5383.

## Tests

- `RoomListFiltersPresenterTest` — new `toggling two incompatible filters in the same frame keeps
  the first one` sends `ToggleFilter(Favourites)` and `ToggleFilter(Invites)` on the *same* state
  instance, which reproduces two taps landing before a recomposition, and asserts only *Favourites*
  survives and that the chip list is `Favourites` (selected), `Unread`, `People`, `Rooms`.
- New `toggling two incompatible categories in the same frame keeps the first one` covers the
  *People* / *Rooms* pair from the report.
- The existing `initial state`, `toggle rooms filter` and `clear filters event` cases are unchanged
  and guard that ordinary selection, deselection and clearing still behave.

What these do not prove: they exercise the presenter and the real selection strategy, not the
Compose chip row, so they do not prove two physical taps can actually be delivered in one frame on
a device — only that the selection layer is safe if they are.

### Scope

**This is not a change to which filters are mutually exclusive.** The `incompatibleFilters` table is
untouched; this only makes the existing table authoritative when a filter is selected.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/filters/selection/DefaultFilterSelectionStrategy.kt
+++ b/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/filters/selection/DefaultFilterSelectionStrategy.kt
@@ -22,10 +22,6 @@ class DefaultFilterSelectionStrategy : FilterSelectionStrategy {
     override val filterSelectionStates = MutableStateFlow(buildFilters())
 
     override fun select(filter: RoomListFilter) {
-        // Incompatible filters are hidden from the UI rather than disabled, so a filter can still be
-        // selected while its chip is on its way out, typically when two chips are tapped in the same
-        // frame. Ignore it, the currently applied filters win.
-        // See https://github.com/element-hq/element-x-android/issues/5383
         if (selectedFilters.any { it in filter.incompatibleFilters }) return
         selectedFilters.add(filter)
         filterSelectionStates.value = buildFilters()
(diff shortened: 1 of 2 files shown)


bmarty (approved):
Thanks!

Issue #7190, opened 2026-07-12
Forwarding a live location share (LLS) results in an error


Susurrus:
### Steps to reproduce

1. Have a live location share in a room
2. Forward it to a room

#### What did you expect?
Location was shared to another room

#### What happened instead?
Saw an error popup ("Sorry, an error occurred") 

### Application version and app store

26.07.0

### Homeserver

mozilla.org

### Will you send logs?

Yes

### Will you submit a Pull Request?

No

Susurrus:
Uploaded the rageshake when I saw the error popup.

Note that EW doesn't allow forwarding these events either. And the SDK says it's not possible to reply to these events, so I'm inferring that these events aren't supposed to be forwardable and the problem here is that the option exists moreso than it doesn't work.

jmartinesp:
> the problem here is that the option exists moreso than it doesn't work

Makes sense to me. A 'live location event' IIRC is a mix of a state event and timeline events, so I don't they can be forwarded... that said, I'm not the expert on this. Maybe @ganfra can confirm?

PR #7343, opened 2026-08-09
Hide the forward action on live location shares

hayaksi1:
### Problem

The timeline action sheet offers **Forward** on a live location share, but forwarding one always
fails: the SDK rejects that event kind outright, so the user gets an error every time. The action is
therefore visible for exactly the set of events on which it cannot work.

### Fix

- `TimelineItemEventContent.canBeForwarded()` now returns `true` for a location event only when its
  mode is `Static`. Live shares — active or ended — are excluded, so the action is not offered.
- No change is needed in `ActionListPresenter`: its `timelineItem.isRemote &&
  timelineItem.content.canBeForwarded()` check is the function's only call site in the repo, so this
  single gate covers the timeline sheet, the threaded timeline and the pinned-messages list at once.

## Motivation and context

`Mode.Live` corresponds to the SDK's `MsgLikeKind.LiveLocation`, which `RoomContentForwarder`
rejects with `ForwardEventException`. Hiding the action mirrors both Element Web's behaviour and the
existing sticker exclusion in the same function, where an action that cannot succeed is not shown.

Relates to #7190.

## Tests

- `ActionListPresenterTest` — new `compute for a static location item` asserts the full expected
  action list for a static location share and confirms **Forward is still offered** there, so the
  change cannot over-reach.
- New `compute for a live location item does not offer to forward it` asserts the full action list
  for an active live share: Reply, CopyLink, Pin, ViewSource, Redact — with Forward absent, and
  emoji reactions still enabled.
- New `compute for an ended live location item does not offer to forward it` covers the
  `isActive = false` case, which is still `Mode.Live`.

What these do not prove: they assert what the action list contains, not that the SDK would have
rejected the forward — the `RoomContentForwarder` failure path is unchanged and untested here.

### Scope

**This is not an attempt to make live location shares forwardable.** Converting a live share into a
static `m.location` on forward is a product decision, not a defect fix, and is deliberately out of
scope.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/event/TimelineItemEventContent.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/model/event/TimelineItemEventContent.kt
@@ -58,10 +58,12 @@ fun TimelineItemEventContent.canBeForwarded(): Boolean =
         is TimelineItemFileContent,
         is TimelineItemAudioContent,
         is TimelineItemVideoContent,
-        is TimelineItemLocationContent,
         is TimelineItemVoiceContent,
         is TimelineItemGalleryContent,
         is TimelineItemAttachmentsContent -> true
+        // Live location shares can't be forwarded, the SDK rejects them, so we only show the option for static locations
+        // See https://github.com/element-hq/element-x-android/issues/7190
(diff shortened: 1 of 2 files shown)


ganfra (approved):
Thanks, sounds sensible.
@hayaksi1 Can you just add a comment on the statement?

hayaksi1:
@ganfra Done.

Issue #6371, opened 2026-03-17
Unable to save image/file if there are 32 file name conflicts, "Sorry, an error occured" (Unable to save any file named image.png anymore)


ledlamp:
### Steps to reproduce

Pick any image in a chat and save it 32 times, then try saving it again.

#### What did you expect?

I should be able to save an indefinite number of files with the same name

#### What happened instead?

After saving the same filename 32 times I am not able to save any file with the same name (even if it's a different file) because the file name already exists with suffixes going all the way up to (31).

This is huge problem because a lot of images have the same name such as image.png and I am not able to save any of those images anymore because I already have image.png, image (1).png, ..., image (31).png in my Download folder, and I guess the loop that tries different filenames gives up after 32 attempts.

<img width="50%" alt="Image" src="https://github.com/user-attachments/assets/66cbf4e7-3ced-491d-a227-e2990f1d3b02" />

<img width="50%" alt="Image" src="https://github.com/user-attachments/assets/457fe309-180d-4063-ab58-4446e12819ee" />

<img width="50%" alt="Image" src="https://github.com/user-attachments/assets/44c28d2f-fad8-4b02-8c80-08e96a684c85" />

Error from log:

```
2026-03-17T20:01:13.675634Z ERROR elementx: Save on disk failed
java.lang.IllegalStateException: Failed to build unique file: /storage/emulated/0/Download/image.png owner_package_name=io.element.android.x _display_name=image.png mime_type=image/png _data=/storage/emulated/0/Download/image.png is_download=1 relative_path=Download/
	at android.os.Parcel.createExceptionOrNull(Parcel.java:3378)
	at android.os.Parcel.createException(Parcel.java:3354)
	at android.os.Parcel.readException(Parcel.java:3337)
	at android.database.DatabaseUtils.readExceptionFromParcel(DatabaseUtils.java:225)
	at android.database.DatabaseUtils.readExceptionFromParcel(DatabaseUtils.java:177)
	at android.content.ContentProviderProxy.insert(ContentProviderNative.java:591)
	at android.content.ContentResolver.insert(ContentResolver.java:2282)
	at android.content.ContentResolver.insert(ContentResolver.java:2243)
	at io.element.android.libraries.mediaviewer.impl.local.AndroidLocalMediaActions.access$saveOnDiskUsingMediaStore(r8-map-id-61ec08b4fbd12f33364cb071e26815858d814832c3e7576695ecebb2cf1a3240:41)
[...]

PR #7344, opened 2026-08-09
Keep saving files to Downloads once 32 files share the same name

hayaksi1:
### Problem

Saving an image or a file to Downloads stops working once 32 files with the same name are already
there. `MediaStore` de-duplicates the display name itself, but its uniquifier gives up after 32
collisions and throws, so the app shows *"Sorry, an error occurred"* and writes nothing. Because
`image.png` is the name most screenshots and pasted images arrive with, a user who saves media
regularly can reach this state and then never save another file under that name — clearing it
requires deleting files from Downloads by hand.

The same code path also treated a `null` output uri as success: the app reported the file as saved
while writing nothing at all.

### Fix

- New `saveWithUniqueFileName` helper in `libraries/androidutils`: it runs the insert with the
  original name and, if the platform reports that it cannot build a unique name, retries **once**
  with a suffixed name. A second failure propagates, so a genuine error is still surfaced.
- New `String.withFileNameSuffix` helper builds that name by inserting the suffix before the
  extension — `image.png` becomes `image_<timestamp>.png`, and a name without an extension still
  works.
- `AndroidLocalMediaActions.saveOnDiskUsingMediaStore` uses the helper, and now throws `IOException`
  when `ContentResolver.insert` returns `null` instead of silently reporting success.
- `DefaultFileSave.saveOnDiskUsingMediaStore` had byte-identical logic and the identical defect, so
  it uses the same helper.
- The pre-Android-Q path is untouched: it overwrites rather than uniquifying, so it never hits this
  limit, and changing it would be a separate behaviour change.

## Motivation and context

The failure is permanent from the user's point of view and gives no indication of the cause — the
error message says nothing about names or Downloads. Retrying once with a name the platform has not
seen restores the operation without changing anything a user can observe until the 33rd save.

Relates to #6371.

## Tests

- New `FileTest` in `libraries/androidutils` covers `withFileNameSuffix` on a plain name
  (`image.png`), a name with no extension (`README`), a double extension (`archive.tar.gz`) and a
  dotfile (`.hidden`).
- `saveWithUniqueFileName - uses the original name when it works` — no suffixing in the normal case.
- `saveWithUniqueFileName - retries with a suffixed name when the name cannot be made unique` —
  reproduces the report: the first attempt fails the way `MediaStore` fails, the second uses
  `image_1.png`, and both attempts are recorded in order.
- `saveWithUniqueFileName - does not retry more than once` — a persistent failure propagates rather
  than looping.
- The existing `AndroidLocalMediaActionsTest` cases still pass unchanged.
- New `DefaultFileSaveTest`, plus two cases added to `AndroidLocalMediaActionsTest`, drive both call
  sites through a `FakeMediaStoreContentProvider` registered on the `media` authority: the normal
  save inserts under the original name and the bytes land in the provider, and a colliding name is
  retried once with a suffixed one.

What these do not prove: the retry is verified against a stubbed insert, not against a real
`MediaStore`. Nothing here proves that Android throws `IllegalStateException` at exactly 32 files —
that comes from the platform's `FileUtils.buildUniqueFile`, not from this repo, so the 33rd-save
behaviour still deserves a manual check.

### Scope

**This is not the "download into a dedicated Element X folder" request** from the same issue thread;
that is a separate feature. The pre-Android-Q storage path is also deliberately untouched.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


bmarty (approved):
Thanks!

hayaksi1:
Done

bmarty:
`Files.createTempDirectory` is not available on API 24 (see lint reporting this error), can you check this please? Thanks!

hayaksi1:
Done!

bmarty (approved):
LGTM, thanks!

PR #7345, opened 2026-08-09
Show the account a UnifiedPush registration belongs to in the distributor

### LINKED ISSUE #6426: Add Account-Name on UnifiedPush Provider
I like to use the Element-X with multiple-account.
Sometimes i want to check my UnifiedPush-Provider if all works well with PushNotification (e.g. counter of notification).
There i are multiple Element-X instances (one per Account) registert, but i could not distinguish the Accounts.

Please add information during register, which account use which UnifiedPush-Provider.

I believe you just has to fill the field `messageForDistributor` here:

https://github.com/element-hq/element-x-android/blob/908ad8a102cb9b689aa9514871db95c10a76f8a6/libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/RegisterUnifiedPushUseCase.kt#L36

But i am not familiar with Kotlin, how i got the account-name / matrix-id there.

Ping: @p1gp1g is this correct with `messageForDistributor` ?

hayaksi1:
## Content

A UnifiedPush distributor such as ntfy or DAVx5 lists every registration it holds, but Element
registered each one with no message for the distributor, so a second account was indistinguishable
from the first in that list.

`RegisterUnifiedPushUseCase.execute` now takes the session id and passes it as
`messageForDistributor`, which the connector has accepted since 3.x. The value is the raw Matrix ID,
which the project's logging rules treat as non-personal, so no new string resource is needed.

`UnifiedPushProvider` already has the `MatrixClient` in hand at the only call site, so the
`PushProvider` interface is unchanged and nothing outside the module moves.

## Motivation and context

With multiple accounts registered there is no way to tell from the distributor which
registration belongs to which account, so revoking the right one is guesswork.

Fixes #6426

## Tests

- `UnifiedPushProviderTest` — the `register ok` and `register ko` cases now assert the session id is
  forwarded from `registerWith` into the use case, alongside the distributor and client secret.
- `DefaultRegisterUnifiedPushUseCaseTest` follows the new arity; that Robolectric test cannot observe
  the argument handed to the static `UnifiedPush.register`, so it is a no-regression pass only.
- The value actually shown by a distributor's UI is not covered by any test.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/RegisterUnifiedPushUseCase.kt
+++ b/libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/RegisterUnifiedPushUseCase.kt
@@ -12,6 +12,7 @@ import android.content.Context
 import dev.zacsweers.metro.AppScope
 import dev.zacsweers.metro.ContributesBinding
 import io.element.android.libraries.di.annotations.ApplicationContext
+import io.element.android.libraries.matrix.api.core.SessionId
 import io.element.android.libraries.pushproviders.api.Distributor
 import io.element.android.libraries.pushproviders.unifiedpush.registration.EndpointRegistrationHandler
 import kotlinx.coroutines.flow.filter
@@ -21,7 +22,7 @@ import org.unifiedpush.android.connector.UnifiedPush
 import kotlin.time.Duration.Companion.seconds
 
 interface RegisterUnifiedPushUseCase {
-    suspend fun execute(distributor: Distributor, clientSecret: String): Result<Unit>
+    suspend fun execute(distributor: Distributor, clientSecret: String, sessionId: SessionId): Result<Unit>
 }
 
 @ContributesBinding(AppScope::class)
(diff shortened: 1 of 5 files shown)


libraries/pushproviders/unifiedpush/src/main/kotlin/io/element/android/libraries/pushproviders/unifiedpush/RegisterUnifiedPushUseCase.kt
```diff
@@ -21,19 +22,22 @@ import org.unifiedpush.android.connector.UnifiedPush
...
     @ApplicationContext private val context: Context,
     private val endpointRegistrationHandler: EndpointRegistrationHandler,
 ) : RegisterUnifiedPushUseCase {
-    override suspend fun execute(distributor: Distributor, clientSecret: String): Result<Unit> {
+    override suspend fun execute(distributor: Distributor, clientSecret: String, sessionId: SessionId): Result<Unit> {
         UnifiedPush.saveDistributor(context, distributor.value)
         // This will trigger the callback
         // VectorUnifiedPushMessagingReceiver.onNewEndpoint
-        UnifiedPush.register(context = context, instance = clientSecret)
+        // The session id is passed as the message for the distributor, so that its UI can tell one
```
frebib: This comment isn't particularly necessary as the code is self-explanatory and seems to just be LLM copium. Drop it?

bmarty (approved):
Thanks!

bmarty:
This is working for SunUp:

image

But not for Ntfy:

image

image

CC @p1gp1g

p1gp1g:
Showing this value to the user is not required, and ntfy doesn't implement this feature (yet?)

PR #7346, opened 2026-08-09
Ignore a second voice recording start while one is already running

### LINKED ISSUE #5780: In certain situations, recording voice messages cannot be stopped.
### Steps to reproduce

1. Enter any room that allows uploading media and sending messages;
2. **Quickly** tap the "Record" button twice in succession.

#### What did you expect?
n/a

#### What happened instead?
At this point, neither clicking the "Stop" nor "Delete" buttons will stop the recording. You must return to the room list and then re-enter the room.

Additionally, the above operation may cause the app to crash with a small probability, but the app will not prompt you that it crashed previously upon the next launch.
**Additionally, the above actions may cause the app to crash with a small probability**, but the app will not prompt you that it crashed previously upon the next launch.

### Your phone model

OnePlus MT2111

### Operating system version

Oxygen OS 15 (Android 15)

Element X Android 25.11.2

### Homeserver

Synapse 1.142.0

### Will you send logs?

No

No

hayaksi1:
## Content

The composer emits Start whenever the voice message state is Idle, and the state only leaves Idle
once the recorder has produced its first audio buffer. A second tap inside that window started a
second recording, which overwrote the output file, re-initialised the encoder while the first one was
still encoding, and orphaned the first job together with its `AudioReader`.

The orphan kept the microphone open and kept emitting Recording for the rest of the session, so Stop
and Delete appeared to do nothing until the room was left and re-entered, and re-initialising the
encoder under an active encode could crash natively.

`DefaultVoiceRecorder.startRecord` now runs under the mutex it already owns and returns early when a
recording job exists, so the second start is a no-op: no file, no encoder re-init, no second
`AudioReader`. This is not a UX change; the button behaves as it always did.

## Motivation and context

A recording that cannot be stopped holds the microphone for the rest of the session, which
is worse than the visible symptom.

Fixes #5780

## Tests

- `DefaultVoiceRecorderTest` — new `when startRecord is called twice, the second call is ignored`
  calls start twice before the state has left Idle, asserts only one `AudioReader` was ever created,
  then stops and asserts the state settles on Finished with no further emission. An orphaned job fails
  the last assertion. This test fails without the guard.
- `FakeAudioReaderFactory` gained a `createdCount` so the first assertion is possible.
- The native crash path is not covered; no test can re-initialise the real encoder.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/voicerecorder/impl/src/main/kotlin/io/element/android/libraries/voicerecorder/impl/DefaultVoiceRecorder.kt
+++ b/libraries/voicerecorder/impl/src/main/kotlin/io/element/android/libraries/voicerecorder/impl/DefaultVoiceRecorder.kt
@@ -72,11 +72,6 @@ class DefaultVoiceRecorder(
 
     @RequiresPermission(Manifest.permission.RECORD_AUDIO)
     override suspend fun startRecord() = lock.withLock {
-        // The composer can emit a second Start before the first recording has produced its first
-        // buffer, i.e. before the state flips away from Idle. Without this guard the second call
-        // orphans the first job and its AudioReader, which keeps the microphone open and keeps
-        // emitting Recording forever, so Stop and Delete appear to do nothing.
-        // See https://github.com/element-hq/element-x-android/issues/5780
         if (recordingJob != null) {
             Timber.w("Voice recorder is already recording, ignoring this start")
             return@withLock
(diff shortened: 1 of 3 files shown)


bmarty (approved):
Thanks!

bmarty:
FTR: https://github.com/element-hq/element-x-android/issues/5780#issuecomment-5263801700

PR #7347, opened 2026-08-09
Linkify URLs whose host is a bracketed IPv6 address

### LINKED ISSUE #2614: ipv6 URLs are not linkified
### Steps to reproduce

1. Send a HTTP URL with an ipv6 as a host

#### What did you expect?

A linkified Link to the http URL

#### What happened instead?
Parts of the ipv6 address got linkified as a phone number, but the whole URL does not get linkified

Screenshot_20240327-163426

Possibly related to the same issue:
https://github.com/TrianguloY/UrlChecker/issues/337#issuecomment-2020885928

### Your phone model

Pixel 8 Pro

### Operating system version

Android 14

0.4.6

### Homeserver

synapse

### Will you send logs?

No

No

hayaksi1:
## Content

`Patterns.WEB_URL`, which `LinkifyCompat` uses to find links, has no production for a bracketed IPv6
authority, so a URL such as `http://[2001:db8::1]:8008/path` was never turned into a link. Worse, the
phone number matcher claimed the digit runs inside the address, leaving a `tel:` link over part of it.

`LinkifyHelper` now runs its own pass for those URLs. It requires both a scheme and the brackets, so
array syntax like `values[0]:1` is not matched. It runs after `LinkifyCompat.addLinks` — which strips
any span set before it — but before the pre-existing spans are restored, so real anchors and pill
spans keep their precedence. Any span Linkify left inside the address is evicted, and the same
trailing-punctuation trimming as every other link is applied.

## Motivation and context

Self-hosted homeservers and local development servers are commonly reached over IPv6
literals, and those links were not tappable.

Fixes #2614

## Tests

- `LinkifierHelperTest` — new `linkification finds ipv6 URL` asserts a single span covering the whole
  address including port and path.
- New `linkification of an ipv6 URL does not leave a phone number span`, run at `sdk = 30` with a
  German SIM like the existing phone tests, which is the actual regression: no `tel:` span survives.
- New `linkification of an ipv6 URL trims trailing punctuation` and `linkification ignores bracketed
  text which is not a URL` (`values[0]:1` and a bare `[::1]`).
- The existing linkification tests are unchanged and pass, including the phone and parenthesis cases.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/text/LinkifyHelper.kt
```diff
@@ -23,6 +23,12 @@ import timber.log.Timber
  * It also checks the linkified results to make sure URLs spans are not including trailing punctuation.
  */
 object LinkifyHelper {
+    /**
+     * Matches a URL whose authority is a bracketed IPv6 address, i.e. `http://[2001:db8::1]:8008/path`.
+     * A scheme and the brackets are both required, so array syntax such as `values[0]:1` is not matched.
+     */
+    private val IPV6_URL_REGEX = Regex("""(?i)\b(?:https?|ftp)://\[[0-9a-f:.]+(?:%25[0-9a-z._-]+)?](?::\d{1,5})?(?:/\S*)?""")
```
frebib: Using a regex for this isn't a great idea because it'll be inaccurate. For one, you're only capturing 3 protocols here (one of which I doubt is even used now). Perhaps a more fully-featured URI parser might be a better idea.

bmarty (approved):
LGTM, thanks!

PR #7348, opened 2026-08-09
Stop a right to left sender name from reordering a room list preview

### LINKED ISSUE #3338: Left to right text breaks message previews
### Steps to reproduce

1. In the chats views
2. Interaction is not required

#### What did you expect?
Message preview is displayed as normal

#### What happened instead?
A user with a name using left to right characters will cause the rest of the message to Aldi be left to right.

### Your phone model

Pixel 6

### Operating system version

14, 2024082200

version 0.5.0  io.element.android.x versionCode 40005000, insisted from Obtanium

### Homeserver

_No response_

### Will you send logs?

No

No

hayaksi1:
## Content

A room list preview is a single paragraph, and its direction is resolved from the first strong
character. A sender whose display name is Arabic or Hebrew therefore flipped the whole line: the
message body moved to the wrong side of the name and the separating colon jumped to the front.

The sender's display name is now wrapped in Unicode isolate characters before it is used as a prefix,
so its own direction cannot escape it. Both places that interpolate the name are covered: the prefix
path and the emote path, which bypasses it.

`prefixWith` itself is deliberately left alone — the pinned messages banner shares it and only ever
passes localised nouns, so isolating there would be churn. The "You" prefix is localised rather than
user-provided and is not isolated.

## Motivation and context

The preview is the only thing most users read in the room list, and for a room with an
RTL-named member it was rendered in the wrong order.

Fixes #3338

## Tests

- `DefaultRoomLatestEventFormatterTest` — new `a right to left sender name is isolated so it cannot
  reorder the preview` uses an Arabic display name and asserts the exact emitted string, plus that the
  bold span still covers only the name and its isolates.
- The existing expectations across the suite were updated for the two isolate characters, including
  the sticker and poll bold-span offsets, which is the regression guard that the span geometry did not
  drift.
- These assert the emitted string; no test renders the row, so the visual order is not proven here.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/DefaultRoomLatestEventFormatter.kt
+++ b/libraries/eventformatter/impl/src/main/kotlin/io/element/android/libraries/eventformatter/impl/DefaultRoomLatestEventFormatter.kt
@@ -135,7 +135,7 @@ class DefaultRoomLatestEventFormatter(
         val message = when (val messageType: MessageType = type) {
             // Doesn't need a prefix
             is EmoteMessageType -> {
-                return "* $senderDisambiguatedDisplayName ${messageType.body}"
+                return "* ${senderDisambiguatedDisplayName.bidiIsolate()} ${messageType.body}"
             }
             is TextMessageType -> {
                 messageType.toPlainText(permalinkParser)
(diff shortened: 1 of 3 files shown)


bmarty (approved):
LGTM, thanks!

PR #7349, opened 2026-08-09
Do not group timeline messages sent more than five minutes apart

### LINKED ISSUE #4757: [Task] Don't group timeline items if more than 5 minutes has passed

Like on EW and now EXI. The "meta" issue to look at here is @pixlwave's iOS PR in this case: https://github.com/element-hq/element-x-ios/pull/4138

hayaksi1:
## Content

Consecutive messages from the same sender were grouped into one bubble block no matter how far
apart they were sent, so a reply hours later looked like part of the earlier burst and lost its own
timestamp header.

`computeGroupPosition` now folds a five minute limit into the two groupability flags it already
computes. The delta is symmetric between a pair of neighbours, so both items always agree on where the
block splits and every branch of the existing decision table keeps working unchanged. Nothing in the
group-position model or the bubble rendering needed to change.

## Motivation and context

Grouping is meant to show a burst of messages as one unit. Once the gap is long enough the
messages are not a burst, and iOS already draws the line at five minutes.

Fixes #4757

## Tests

- New `TimelineItemsFactoryTest` drives the real factory with events at controlled timestamps and
  asserts the resulting group positions. There was no test anywhere in the module covering
  `TimelineItemGroupPosition` before this.
- Cases: three messages a minute apart group as First/Middle/Last; three messages six minutes apart do
  not group at all; a long gap in the middle splits one block into two; a gap of exactly five minutes
  still groups, which pins the boundary; identical timestamps still group.
- No Paparazzi snapshot changes, since every preview passes `groupPosition` explicitly.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemEventFactory.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemEventFactory.kt
@@ -38,6 +38,9 @@ import io.element.android.libraries.matrix.ui.messages.reply.map
 import kotlinx.collections.immutable.persistentListOf
 import kotlinx.collections.immutable.toImmutableList
 
+/** Events further apart than this are never grouped together, even when sent by the same person. */
+private const val GROUPING_TIMEOUT_MS = 5 * 60 * 1000L
+
 @AssistedInject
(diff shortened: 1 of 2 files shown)


bmarty (approved):
Thanks!

PR #7350, opened 2026-08-09
Do not notify about a redaction whose target event is unknown

### LINKED ISSUE #5766: Notifications are shown for deleted mesages
### Steps to reproduce

1. Check room list
2. Notice notifications
3. Enter room
4. Observe that the event that triggered the notification is actually deletion of a message

#### What did you expect?
Deletion of a message should not notify a user

#### What happened instead?
Deletion of a message notified a user

### Your phone model

Pixel 8a

### Operating system version

Android 16

25.11.2-nightly

### Homeserver

element.io

### Will you send logs?

No

No

hayaksi1:
## Content

A redaction push whose `redacts` field the SDK could not map was reported as an unknown error. The
result processor only suppresses filtered-out and redacted events; every other exception becomes a
visible "you have N notifications" fallback and also raises the battery optimisation banner.

So a deleted message could produce a notification about itself, which is the opposite of what a
redaction means, plus an unrelated banner.

The branch now throws `EventFilteredOut`, which the processor already drops silently, and logs at
warning level so the SDK gap still shows up in push history.

## Motivation and context

A redaction exists to remove content. Turning a redaction the client cannot resolve into
a visible notification is strictly worse than doing nothing.

Fixes #5766

## Tests

- `DefaultNotifiableEventResolverTest` — the null-`redactedEventId` case now asserts the thrown
  exception **is** `EventFilteredOut`. It previously only asserted the result was null, which passed
  for either exception, so it could not have caught this.
- The existing `DefaultNotificationResultProcessor` tests already cover that `EventFilteredOut` is
  dropped and that other exceptions render a fallback.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
@@ -284,10 +284,6 @@ class DefaultNotifiableEventResolver(
                 // Note: this case will be handled below
                 val redactedEventId = content.redactedEventId
                 if (redactedEventId == null) {
-                    // Treat it as filtered out rather than as an error: an unknown error is turned
-                    // into a visible "you have N notifications" fallback plus a battery banner, so a
-                    // redaction we cannot map would notify about a message that was just deleted.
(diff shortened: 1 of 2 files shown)


bmarty (approved):
Thanks!

PR #7351, opened 2026-08-09
Never re-encode a video at a higher bitrate than the source

### LINKED ISSUE #4191: Uploading certain video files is very slow and "compresses" them to a mach larger file size
### Steps to reproduce

When uploading some videos to a room, the "upload" screen sometimes sits with a spinning circle for minutes before finally uploading. Presumably this is when compression of the file is occurring, and if so it may be more user friendly to display some kind of progress to avoid the user believing the process has failed. Uploading a 22MB, 5 minute long video took just over 2 minutes. 

After uploading, certain videos have been "compressed" to a much larger file size. This occurs in both DM and multi-user rooms. 

The video in question was an .mp4 file using H.264 encoding, 25fps, AAC audio and a resolution of 720p x 720p (1:1 aspect ratio). The original file size was 22MB, but it increased to 84MB once uploaded to the room. 

Testing with an .mp4 1080p video with 16:9 aspect ratio seemed to work as intended, reducing the file size from 15.4MB to 11MB. For this video, the time taken to upload it was also significantly less, spending only 20 seconds or so with the spinning upload progress indicator. 

#### What did you expect?
Video is compressed to a smaller size (or not at all if it's already small).

#### What happened instead?
Video ended up being much larger in size than the original, and took about 2 minutes to send. Expecting compression to reduce the size of the video for optimal network performance, or leave it be if no compression required. 

### Your phone model

Samsun S20 Ultra

### Operating system version

Android 13

Element X Android 0.7.6 (40007060)

### Homeserver

Synapse 1.121.1

### Will you send logs?

Yes

No

hayaksi1:
## Content

The compression config computed a target bitrate purely from the output resolution and frame rate,
ignoring the bitrate already extracted from the source. A clip that is small because it was
efficiently encoded, rather than because it is short, was therefore re-encoded upwards: the upload
got slower and the resulting file was larger than the original.

The target is now clamped to the source bitrate when that is known. `METADATA_KEY_BITRATE` is the
container total, video plus audio, so it is a safe upper bound for the video track alone. An unknown
or zero bitrate still falls back to the calculated value, and a high-bitrate source is still
compressed exactly as before.

## Motivation and context

Compression that makes the file bigger is worse than no compression, and the user waits
longer for it.

Fixes #4191

## Tests

- `VideoCompressorConfigFactoryTest` — new case for the reported shape, a 720x720 clip at 615 kbps,
  asserting the target is the source bitrate rather than the much higher calculated optimum.
- New case asserting a high-bitrate 1080p source is still compressed downwards.
- New case asserting an unknown (-1) and a zero bitrate both fall back to the calculated value, so a
  bogus metadata read cannot produce a zero-bitrate encode.
- The four existing resize tests are unchanged.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/VideoCompressorConfig.kt
+++ b/libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/VideoCompressorConfig.kt
@@ -36,11 +36,6 @@ internal object VideoCompressorConfigFactory {
         // If we need to resize the video, we also want to recalculate the bitrate
         val optimalBitrate = resizer.calculateOptimalBitrate(Size(width, height), newFrameRate).toInt()
 
-        // Never ask the encoder for more than the source already carries: re-encoding a video that is
-        // already efficiently encoded at a low bitrate otherwise inflates it, making the upload both
-        // slower and bigger than sending the original. METADATA_KEY_BITRATE is the container total,
-        // video plus audio, so it is a safe upper bound for the video track alone.
-        // See https://github.com/element-hq/element-x-android/issues/4191
         val sourceBitrate = metadata?.bitrate?.takeIf { it > 0 }
         val newBitrate = sourceBitrate?.let { min(optimalBitrate.toLong(), it).toInt() } ?: optimalBitrate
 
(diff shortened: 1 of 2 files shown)


bmarty (approved):
Thanks!

PR #7353, opened 2026-08-09
Send a message with Enter on a physical keyboard, Shift+Enter for a new line

### LINKED ISSUE #860: With physical keyboard, press enter to send message, shift+enter to enter a new line?

When my phone/tablet device is connected to a physical keyboard,

- Press "enter" to send a message
- Press "shift + enter" to enter a newline

Because it's jarring to have to use my finger to press "send" for each message, while I could press "enter" as in most messaging apps I'm using.

With code, sweat and hopefully no tears.

_No response_

### Additional context

Using Nightly from July 11th.

No

hayaksi1:
## Content

With a hardware keyboard attached, Enter inserted a newline and the only way to send was to reach
for the send button, which is the opposite of what every desktop chat client does.

`MarkdownEditText` gains an optional Enter listener and consumes Enter when it comes from a physical
keyboard without Shift. Shift+Enter keeps inserting a newline, and a key event from the on-screen
keyboard is identified by its `VIRTUAL_KEYBOARD` device id and left alone, so nothing changes for
touch users. The key release is consumed as well, so no newline leaks out on key repeat.

`TextComposer` passes its existing send lambda through, guarded by the same `canSendTextMessage`
check the send button already uses, so Enter cannot send an empty message.

## Motivation and context

Element X is usable on tablets, Chromebooks and desktop-mode devices, where reaching for
the send button after every message is a real cost.

Fixes #860

## Tests

- `MarkdownTextInputTest` — new case dispatching a physical-keyboard Enter asserts the send lambda
  runs once and no newline is left in the editor.
- New case for Shift+Enter asserts the send lambda does not run and a newline is inserted.
- New case for an on-screen keyboard Enter (`VIRTUAL_KEYBOARD` device id) asserts the same, which is
  the regression that would break every touch user.
- The existing typing, suggestion and mention cases are unchanged.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/textcomposer/impl/src/main/kotlin/io/element/android/libraries/textcomposer/TextComposer.kt
+++ b/libraries/textcomposer/impl/src/main/kotlin/io/element/android/libraries/textcomposer/TextComposer.kt
@@ -160,6 +160,8 @@ fun TextComposer(
     } else {
         stringResource(id = R.string.rich_text_editor_composer_placeholder)
     }
+    val canSendTextMessage = markdown.isNotBlank() || composerMode is MessageComposerMode.Attachment
+
     val textInput: @Composable () -> Unit = when (state) {
         is TextEditorState.Rich -> {
             val coroutineScope = rememberCoroutineScope()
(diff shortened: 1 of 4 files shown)


bmarty (approved):
Thanks!

PR #7354, opened 2026-08-09
Turn a matrix: URI in a message into a pill

### LINKED ISSUE #5319: Matrix URI is not linkified
### Steps to reproduce

1. Open a chat
2.  Enter text that contains an URI following the [Matrix URI scheme](https://spec.matrix.org/latest/appendices/#matrix-uri-scheme), for example: `matrix:u/someone:example.com`
3. Send the message

#### What did you expect?

URI is linkified

#### What happened instead?

Only the domain part is linkified

---

When I put it in angle brackets instead to explicitly generate a link, the URI even gets pillified.

### Your phone model

_No response_

### Operating system version

_No response_

25.09.1

### Homeserver

_No response_

### Will you send logs?

No

No

hayaksi1:
## Content

A permalink written as a `matrix:` URI stayed plain text, while the same link in matrix.to form
became a pill. The pillifier only scanned `Patterns.WEB_URL`, which cannot match a URI with no web
scheme, even though the permalink parser already understands `matrix:` URIs.

Those URIs are now scanned for as well, before the web pass, since `Patterns.WEB_URL` otherwise
claims the bare domain inside the URI and whichever pass runs first wins the range. Each match also
gets a `URLSpan`: the timeline only reports URLSpans as clicks, so without one the pill would look
right and do nothing.

The composer path is untouched, since it pillifies without permalinks.

## Motivation and context

`matrix:` URIs are the specification's own permalink form, and clients that send them
produced dead plain text here.

Fixes #5319

## Tests

- `DefaultTextPillificationHelperTest` — new cases for a user URI and a room URI, asserting the
  mention span covers the whole URI and that a `URLSpan` carrying the URI is present.
- New case asserting trailing sentence punctuation stays outside the span.
- New case asserting the composer path (`pillifyPermalinks = false`) still produces no pills.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/TextPillificationHelper.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/utils/TextPillificationHelper.kt
@@ -28,9 +28,6 @@ import io.element.android.libraries.textcomposer.mentions.getMentionSpans
 import io.element.android.wysiwyg.view.spans.CodeBlockSpan
 import io.element.android.wysiwyg.view.spans.InlineCodeSpan
 
-/**
- * Matches a `matrix:` URI as defined by MSC2312, stopping before any trailing sentence punctuation.
- */
 private val MATRIX_URI_REGEX = Regex("""matrix:(?:u|user|r|room|roomid|e|event)/\S*[^\s.,;:!?)]""", RegexOption.IGNORE_CASE)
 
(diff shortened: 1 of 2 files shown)


bxdxnn (approved):

bmarty (approved):
LGTM, thanks

PR #7355, opened 2026-08-09
Do not fail the screenshot recording job when there is nothing to commit

### LINKED ISSUE #822: The record screenshot action fails if there is no change to commit.
It should rather log that there is no change to commit and exit gracefully.

See for instance: https://github.com/vector-im/element-x-android/actions/runs/5487107960

hayaksi1:
## Content

`recordScreenshots.sh` runs under `set -e` and ends with an unconditional
`git commit -m "Update screenshots"`. When a recording run produces no diff — the common case once
the snapshots are already up to date — `git commit` exits 1 with "nothing to commit, working tree
clean" and takes the whole workflow down with it, reporting a red job for a successful run.

This stages as before, then returns early when the staged diff is empty. `set -e` is left in place:
it was added deliberately so a failed Gradle run cannot push a commit that deletes every snapshot.
Using `git diff --cached --quiet` as an `if` condition is `set -e`-safe, because its non-zero "there
are changes" status is consumed by the condition. The early return also skips the push, which had
nothing to send.

The unrelated `user.name`/`user.email` mix-up a few lines above is left alone to keep this
single-purpose.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/822

## Tests

- Exercised the commit tail of the script in a scratch repository, against a clean tree and a dirty
  one: the current script exits 1 on a clean tree, the patched one prints "Screenshots are already up
  to date, nothing to commit." and exits 0, and a dirty tree still produces the "Update screenshots"
  commit and reaches the push.
- These do not prove the workflow itself: `recordScreenshots.yml` is `workflow_dispatch`/label gated
  and needs a repository secret, so it cannot run on a fork pull request.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s): n/a, CI script only

## Checklist

    - [ ] No.

### DIFF
--- a/.github/workflows/scripts/recordScreenshots.sh
+++ b/.github/workflows/scripts/recordScreenshots.sh
@@ -78,6 +78,10 @@ else
   git config --local user.name "${INPUT_AUTHOR_EMAIL}"
 fi
 git add -A
+if git diff --cached --quiet; then
+  echo "Screenshots are already up to date, nothing to commit."
+  exit 0
+fi
 git commit -m "Update screenshots"
 
 GITHUB_REPO="https://$GITHUB_ACTOR:$TOKEN@github.com/$REPO.git"


bmarty (approved):
Thanks!

Issue #4008, opened 2024-12-09
FediVerse handles should not be treated as email handles


tigexplorer:
#### What would you like to do?
If I receive a FediVerse handle like @name@server.tld it should be treated as is and not as E-Mail 

#### Why would you like to do it?
Improve handling.

#### How would you like to achieve it?
Tab on the handle to open a dialog for sharing or open with a FediVerse-App 

### Are you willing to provide a PR?

Yes

PR #7356, opened 2026-08-09
Do not linkify the server part of a fediverse handle as an email address

hayaksi1:
## Content

A fediverse handle such as `@name@server.tld` contains a substring that `Linkify.EMAIL_ADDRESSES`
recognises as a valid address, so the timeline renders `name@server.tld` as a tappable `mailto:`
link. Tapping it opens a mail composer addressed to an account that does not accept mail.

This drops a newly added `mailto:` span when the character immediately before it is `@`. It mirrors
what Android already does for web URLs, where `Linkify.sUrlMatchFilter` rejects a match preceded by
`@` — which is why the handle currently produces exactly one stray mailto span and no partial link is
left behind once it is removed.

Author-supplied links are unaffected: spans that were already present before linkification are
excluded from the check and restored as before. One deliberate trade-off: a genuine address written
directly after an `@` — "mail me @support@element.io" — is no longer linkified, which is the same
trade-off Android already makes for web URLs.

The issue also asks for a tap on the handle to offer a share or fediverse-app intent. That is a
separate, undesigned change and is not part of this.

## Motivation and context

Part of https://github.com/element-hq/element-x-android/issues/4008

## Tests

- `libraries/androidutils/src/test/.../text/LinkifierHelperTest.kt` — a handle mid-sentence and a
  handle at the start of the message both produce no spans; a handle next to a real address leaves
  exactly the `mailto:john@doe.com` span; an author-supplied `mailto:` span sitting after an `@` is
  preserved. The existing `linkification finds email` case stays green as the regression guard.
- These cover the span layer only. They do not prove the rendered timeline, which is not
  snapshot-testable here.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/text/LinkifyHelper.kt
+++ b/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/text/LinkifyHelper.kt
@@ -47,6 +47,11 @@ object LinkifyHelper {
                 val start = spannable.getSpanStart(urlSpan)
                 val end = spannable.getSpanEnd(urlSpan)
 
+                if (urlSpan !in oldURLSpans && spannable.isEmailInsideFediverseHandle(urlSpan, start)) {
+                    spannable.removeSpan(urlSpan)
+                    continue
+                }
+
                 // Try to avoid including trailing punctuation in the link.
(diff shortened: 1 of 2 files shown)


bmarty (approved):
LGTM, thanks

PR #7159, opened 2026-07-06
Content scanner (Element Pro): scan timeline events for unsafe content

jmartinesp:
## Content

Second iteration of content scanner: scan timeline events and display errors for dangerous ones.

### Changes:

- Simplify how timeline event contents are displayed in the timeline: instead of having a different implementation of the caption view per event content type, add it in a single place based on the content type.
- Add a `ContentScannerService` that can be used to manually scan medias for an event and update the event content's  validity in a `ContentValidationState`, as long as there is a `ContentScanner` instance available (Element Pro only).
- Add a cache for the results of these scans, living in a room scope, that will be populated while the room is opened and the timeline in use and then discarded when the screen is left.
- Display both invalid content and content not found UI in the timeline items: most items will change the whole message UI based on this, but the gallery and attachment list events will do it per media item.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/7123, https://github.com/element-hq/element-x-android/issues/7125, https://github.com/element-hq/element-x-android/issues/7124.

## Screenshots / GIFs

There are lots in the PR, including some unexpected ones (flaky screenshots and small changes to corners 😓 ).

## Tests

- Use the app as usual, check there are no changes in the timeline.
- Then enable the content scanner (there is a guide [here](https://github.com/element-hq/element-internal/issues/721)), make sure you're using the FOSS app and override the `DefaultContentScannerUrlProvider` so it points to your locally running content scanner, and test the app again.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

- [x] I am aware of the [etiquette](https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#etiquette).
- This PR was made with the help of AI:
    - [ ] Yes. In this case, please request a review by Copilot.
    - [x] No.
- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [x] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly defines what will change for the user
- [x] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #7359, opened 2026-08-09
Show the right image when opening one that has not finished sending

### LINKED ISSUE #6839: Image preview of undelivered images is broken
### Steps to reproduce

1. Be somewhere with spotty internet (low bandwidth, high packet loss, intermittent disconnects)
2. Send a photo to a room
3. Wait for it to turn up (as undelivered to the server) in the timeline
4. Tap on it in the timeline

#### What do you expect?

I expect the tapped image to open in full quality, full screen view.

#### What happens instead?

A random image from the room history opens instead.

### Your phone model

_No response_

### Operating system version

_No response_

Element X v26.04.4 (F-Droid)

### Homeserver

matrix.org

### Will you send logs?

No

No

hayaksi1:
## Content

Tapping an image that has not finished sending does not open it. An item the server has not
acknowledged has no event id, and two separate places treat that as a dead end.

`MessagesView.onContentClick` reads the content validation state by event id and returns when there
is none, so the click never leaves the timeline. That gate arrived with the content scanner in
element-hq/element-x-android#7159; before it, the click went through and opened the media viewer on
the wrong picture, which is how the issue was filed. The check now runs only when there is an event
id to check.

`MediaViewerNode` then falls back to the room-wide media gallery, which is built from the media
timeline and so never contains a local echo. `MediaViewerPresenter` resolves the start page with
`findEventIndex(eventId) ?: 0`, so a null id lands the pager on index 0 — the "random image from the
room history" the report describes. It now builds a one-item gallery from the parameters the viewer
was opened with, the same way an avatar is already handled, and the same fallback Element X iOS uses
in [`TimelineMediaPreviewDataSource`](https://github.com/element-hq/element-x-ios/blob/develop/ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift#L63-L69)
when the tapped item is missing from the timeline it loaded. The item is placed in the bucket the
viewer will actually read: images and videos for the image mode, files for the files and audio mode,
so an unsent file or audio message does not open an empty viewer.

Nothing else moves: the event id was already nullable in the entry point params, and the details
sheet already hides view-in-timeline, forward and delete when it is null. One caveat worth stating —
which item is shown is now correct in every case, but whether it renders at full quality still
depends on the SDK serving a local echo through the media file loader, which is a different path
from the timeline thumbnail.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/6839

## Tests

- `features/messages/impl/.../MessagesViewTest.kt` — clicking an event with no event id invokes the
  content click callback. Reverting the `onContentClick` change fails this test alone, with the
  callback called 0 times; the other 31 cases in the class stay green.
- `libraries/mediaviewer/impl/.../viewer/SingleMediaGalleryDataSourceTest.kt` — building the source
  from params with no event id yields exactly one item carrying the params' media source, and it
  lands in the images bucket for the image mode and in the files bucket for the files and audio mode.
- The existing factory cases for image, video, audio, voice and file stay green.
- These do not cover the node wiring itself, which is a constructor branch with no seam, nor the
  quality of the rendered local echo.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


jmartinesp:
I think we should do like iOS instead, which just opens the media gallery without the event id: for this, we can modify `MessagesView.onContentClick` so it doesn't return early with `eventId = null`, instead just use it for the content validation check, like:

```kotlin
fun onContentClick(event: TimelineItem.Event) {
      Timber.v("onMessageClick= ${event.id}")
      val eventId = event.eventId
      if (event.eventId != null && eventContentValidationState[eventId].getCurrentOverallState() != ContentValidationValue.Valid) return
     ...
```

hayaksi1:
@jmartinesp Done.

libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerNode.kt
```diff
@@ -92,8 +92,7 @@ class MediaViewerNode(
         is MediaViewerEntryPoint.Params.RoomMedia -> {
             val eventId = inputs.eventId
             if (eventId == null) {
-                // Should not happen
-                timelineMediaGalleryDataSource
+                SingleMediaGalleryDataSource.createFrom(inputs, contentValidationCache)
```
jmartinesp: I don't think this change is needed: I just tested it and the existing code opens a media viewer with the timeline data source, where the local image is displayed just fine and you can navigate to the previous ones in a transparent way.

The only issue is that the opened viewer doesn't focus in the item, but I think that's acceptable. It's either that or changing all the underlying logic to handle `EventOrTransactionId` instead, which may be too much to include in a single PR.

jmartinesp (changes requested):
I think this still needs some work for making the behaviour consistent with iOS.

hayaksi1:
@jmartinesp Done.

jmartinesp (approved):
Thanks for the latest changes!

jtrees:
Thanks for the fix!

PR #7360, opened 2026-08-09
Do not notify about an event sent by another account on the same device

### LINKED ISSUE #7253: Multi account: skip notifications from other accounts registered on the same device

If I have bob @alice:example.org and @alice:another.example.com on my phone, I'd like for notifications sent from either account to not ping my phone.

Stands to reason that most of the times I send a message, I know what I have sent already. I suspect there is a vanishingly small minority of people for which they want notifications from other accounts they have registered.

Should be a trivial check to look at the UserIDs connected to the app, and filter out notifications from those senders.

_No response_

### Additional context

I am the worst mobile developer on the planet, but I may give this a tool assisted go at some point in the future if it drives me nuts.

Yes

hayaksi1:
## Content

With several accounts logged in, sending a message from one of them raises a notification on every
other account that shares the room. The device notifies its owner about something the owner has just
done, and the more accounts share a room the more copies arrive.

This resolves the set of logged-in user ids once per push batch and drops an event whose sender is
one of the other accounts, using the filtered-out result the notification pipeline already
understands, so it is recorded in push history rather than turned into a fallback notification.

The receiving session is removed from that set, which keeps the change a strict no-op for a single
account and protects the local echo that the reply action pushes back through the same path.
Filtering happens at the one point that sees every kind of push, so messages, polls, call invites,
ringing calls and room invites are all covered.

Filtering is unconditional and has no setting. Multiple accounts sit behind an experimental flag that
is off by default, so nothing changes for a stock build.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/7253

## Tests

- `libraries/push/impl/src/test/.../notifications/DefaultNotifiableEventResolverTest.kt` — a room
  message and a ringing call sent by a second logged-in session both resolve to the filtered-out
  result; the same message from a user who is not logged in still resolves normally; and a message
  whose sender is the receiving session itself is not filtered, which is the regression guard for the
  smart-reply echo.
- The existing resolver cases stay green, proving the single-account path is untouched.
- These do not prove the on-device behaviour: multi-account is behind an experimental default-off
  flag and needs two accounts on one device to exercise for real.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/DefaultNotifiableEventResolver.kt
@@ -56,6 +56,7 @@ import io.element.android.libraries.push.impl.db.PushRequest
 import io.element.android.libraries.push.impl.notifications.model.InviteNotifiableEvent
 import io.element.android.libraries.push.impl.notifications.model.NotifiableMessageEvent
 import io.element.android.libraries.push.impl.notifications.model.ResolvedPushEvent
+import io.element.android.libraries.sessionstorage.api.SessionStore
 import io.element.android.libraries.ui.strings.CommonStrings
 import io.element.android.services.toolbox.api.strings.StringProvider
 import timber.log.Timber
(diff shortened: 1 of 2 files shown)


bmarty (approved):
Thanks

Half-Shot:
Thank you!

PR #7361, opened 2026-08-09
Only track read receipts on message-like events

### LINKED ISSUE #2311: RRs should only be displayed on non-state events
Will probably require this issue to be solved first to work properly:
https://github.com/matrix-org/matrix-rust-sdk/issues/3065

From a product perspective we would like to not have RRs displayed on state events, but only on messages. So we need to hide them from messages. However this could create issues where RRs would then disappear (like they do on iOS, since we already show them there only for messages), so we need first an API that filters state events for RRs, so that the event is still displayed in the timeline but its RRs are moved to the next non filtered state event (look at the SDK issue for more info)

hayaksi1:
## Content

A read receipt points at whatever event the reader last saw, and that is often a state event: someone
changing the room name or a member joining. The avatar then sits on the membership change rather than
on the last message, which is not what a reader is being told about, and it is the behaviour reported
in the issue.

This asks the SDK to track receipts on message-like events only, for the live timeline and for the
focused and threaded ones, so all three agree. The disabled case for the media and pinned-event
timelines is unchanged.

Nothing in the UI needs to move: the receipt row already renders nothing for an empty receipt list,
and the grouped row already guards on a non-empty list, so state events lose the avatar without
gaining stray space. The call in the state-event row stays because it still drives the sending and
sent indicator for your own state events.

One thing to be explicit about: the SDK attaches the receipt to the preceding event rather than the
next one, which is the placement its API offers.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/2311

The SDK-side support for this is matrix-org/matrix-rust-sdk#5900.

## Tests

- No automated test: there is no test class for the room factory or the joined room, and the timeline
  configuration is passed straight into the Rust SDK with no injectable fake, so the flip cannot be
  asserted from Kotlin.
- The module builds and its existing tests pass, and the enum value used here is present in the
  currently pinned SDK.
- To be explicit about what is not covered: this has not yet been exercised on a device, so the
  rendered placement rests on the SDK's documented behaviour rather than on an observation. Happy to
  add a recording from a room with a trailing state event if that would help the review.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


hayaksi1:
@bmarty @jmartinesp Can you check it?

jmartinesp (approved):
Makes sense. I can't remember why we hadn't done this yet. It probably fell through the cracks.

PR #7362, opened 2026-08-09
Keep the opened image on screen when the media gallery fails to load

### LINKED ISSUE #4973: Technical error displayed when not able to load media in the gallery
### Steps to reproduce

1. Clear application cache
2. Open a room with an image
3. Enter airplane mode so that the application cannot access the homeserver anymore
4. Open the image

#### What did you expect?

The thumbnail is displayed

#### What happened instead?

An error is displayed full screen as below

Image

### Your phone model

Any

### Operating system version

Any

25.06.2

### Homeserver

Any

### Will you send logs?

No

No

hayaksi1:
## Content

Opening an image from the timeline replaces it with a full-screen page reading "error". Two things go
wrong. The viewer already holds the item that was tapped, but building the surrounding media timeline
can fail and the failure overwrites that cached item, so the picture the user asked for disappears.
And the failure page renders the exception message straight from the SDK, which is developer text
with nothing actionable in it.

This emits the failure only when there is nothing cached, so a viewer opened on a specific event
keeps showing that event and stays usable while the rest of the gallery is unavailable. Where there
really is nothing to show, the page now uses the same alert icon, title and message as the other
error states in this screen, built from strings that already exist.

The trade-off is that the timeline error is no longer surfaced in the cache case; it is still logged.
Swapping it for a snackbar would need a new surface in the viewer and is left out, as is a retry
button on the error page.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/4973

## Tests

- `libraries/mediaviewer/impl/src/test/.../datasource/TimelineMediaGalleryDataSourceTest.kt` — a
  focused timeline whose creation fails keeps emitting the cached item and never emits a failure.
- The existing `failing to load timeline should emit an error` case stays green, proving the
  no-cache path still surfaces the error.
- Screenshot coverage for the new error page comes from the existing `@PreviewsDayNight` preview
  (state provider index 15), re-recorded in this branch.
- These do not cover the flick-to-dismiss behaviour of the error page, which is unchanged.

## Screenshots / GIFs

|Before|After|
|-|-|
|||

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty (approved):
Make sense, thanks!

Issue #4588, opened 2025-04-13
some audio messages don't have a length and can't be seeked


bnjbvr:
#### What would you like to do?

Be able to see the length (duration) of an audio message, and be able to seek in all the audio messages I receive.

#### Why would you like to do it?

- To know how long an audio message is, before listening to it.
- To seek backwards and forwards in the audio message, if i want to re-listen to a specific part / skip a part.
- Because ElementX for iOS does it correctly 😎 

#### How would you like to achieve it?

If the audio message's length/waveform is missing from the media, compute it locally (maybe using an audio processing lib).

### Are you willing to provide a PR?

No

bnjbvr:
May be a dup of https://github.com/element-hq/element-x-android/issues/2419?

Issue #2419, opened 2024-02-20
Voice message without waveform doesn't work as intended


UntriexTv:
### Steps to reproduce

Use newest version of meta bridge
Receive a voice message
The pointer doesn't move and user is not able to change position of playback too.
The event Source:
```json
{
  "content": {
    "body": "audio.mp3.ogg",
    "file": {
      "hashes": {
        "sha256": "7petCMnwEXzHNv5znssNelKlneGpW0z9uX7t4FvGocg"
      },
      "iv": "kaYCr7ZctD0AAAAAAAAAAA",
      "key": {
        "alg": "A256CTR",
        "ext": true,
        "k": "Z6FlbdIij6rcLLJFc0dZODuBE3COIE-YOMO5-XJjSGw",
        "key_ops": [
          "encrypt",
          "decrypt"
        ],
        "kty": "oct"
      },
      "url": "mxc://XXX/zSekdasOsQYZFmjmDeaNTGcB",
      "v": "v2"
    },
    "info": {
      "duration": 5000,
      "mimetype": "audio/ogg",
      "size": 24130
    },
    "m.mentions": {},
    "msgtype": "m.audio",
    "org.matrix.msc1767.audio": {
      "duration": 5000
    },
    "org.matrix.msc3245.voice": {}
  },
  "event_id": "$bt277QlOLofsQwBmWipWzfElrxYOr1h-8Ft74_ZbiDE",
  "origin_server_ts": 1708428672000,
  "room_id": "!XPbNhAlKFYbfNVbfir:XXX",
  "sender": "@messenger_100052572662810:XXX",
  "type": "m.room.message",
  "unsigned": {}
}
```

#### What did you expect?
See it working like on element web
![image](https://github.com/element-hq/element-x-android/assets/48880229/6e4ac616-c146-4be8-b97a-776f89548dae)

#### What happened instead?
There is no waveform, or placeholder and skipping is impossible.
(Normal voice message, but no waveform and the green playing line stays at the left after pressing play)

### Your phone model

Pixel 7 and Galaxy Tab s9 FE

### Operating system version

Android 14

### Application version and app store

Element X V. 0.4.4

### Homeserver

matrixdotorg/synapse latest docker

### Will you send logs?

No

### Are you willing to provide a PR?

Yes

PR #7363, opened 2026-08-09
Let a voice message without a waveform be seeked

hayaksi1:
## Content

A voice message that arrives without amplitude data — bridged messages and those sent by clients that
omit it — is unusable. The waveform is empty, so its measured width is zero: the progress overlay
never draws, the cursor is pinned to the left edge, and the touch handler rejects every event because
no x coordinate can fall inside a zero-width range. The message plays, but there is no way to see
where it is or to move it.

This falls back to a flat run of zero amplitudes sized to the canvas when there is no waveform. Width,
progress, cursor and seeking then all work exactly as they do for a real waveform, and the existing
minimum amplitude renders those zeros as the same flat track a genuinely silent recording already
draws, so no new drawing code is needed.

Nothing changes for a message that has a waveform, and the model still carries the empty list, which
is the honest value. Computing a real waveform from the audio is a separate piece of work.

## Motivation and context

Part of https://github.com/element-hq/element-x-android/issues/2419
Relates to https://github.com/element-hq/element-x-android/issues/4588

## Tests

- `libraries/designsystem/src/test/.../components/media/WaveformNormalisationTest.kt` — an empty
  waveform is padded to the requested sample count, stays empty when there is no room to display it,
  a short waveform is passed through unchanged, and a long one is still downsampled.
- Screenshot coverage comes from the existing `@PreviewsDayNight` previews that already render an
  empty waveform, re-recorded in this branch.
- These do not prove the touch handling itself, which has no unit-test seam; that was reasoned from
  the zero-width range and is best confirmed by the video below.

## Screenshots / GIFs

|Before|After|
|-|-|
|||

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty (approved):
Thanks!

Issue #3493, opened 2024-09-19
media progress


johansmitsnl:
### Your use case

Currently on holiday seeing family members using Element and also WhatsApp and hearing the complaints about Element. Tested with the `0.6.2` release.

#### What would you like to do?

Very noticeable on slower connections is that you can't see where media download is.

#### How would you like to achieve it?

Add a download progress bar from 0 to 100 or circle that shows the real progress.

### Have you considered any alternatives?

No

### Additional context

Now with images you see a placeholder but you don't see the size or the progress of loading the media. They feel it is slow.

For videos you see a generic loading animation but it doesn't give you any idea how long you need to wait for the video to be loaded.

### Are you willing to provide a PR?

No

lub:
With disabled animations it's even worse - there is only a barely visible, tiny light grey bar at the top of the screen.

Also on phones with a camera notch part of the loading bar is even blocked by it, which makes it even less obvious imo.

PR #7366, opened 2026-08-09
Do not draw the media loading progress bar under the status bar

hayaksi1:
## Content

The media viewer is drawn edge to edge, and the linear progress indicator shown while a media item
downloads was the top-most child of a container that only applied navigation bar padding. It was laid
out at y=0, so it rendered behind the status bar and, on a device with a camera cutout, partly behind
the notch.

It now carries `WindowInsets.safeDrawing` horizontal and top padding, reusing the same idiom already
used by `MediaViewerBottomBar` in this file.

The top inset deliberately comes from `safeDrawing` rather than from the scaffold content padding:
the top app bar lives inside an `AnimatedVisibility`, so that padding collapses to zero whenever the
user taps to hide the overlay, which would put the bar back under the cutout.

## Motivation and context

Part of https://github.com/element-hq/element-x-android/issues/3493

The issue also asks for a real 0-100% download progress bar. That still needs download progress to be
exposed by the SDK, as the existing note in `rememberShowProgress` explains, so it is not addressed
here.

## Tests

- Open an image or video that is not yet cached and check the loading bar is drawn below the status
  bar, including on a device with a display cutout and in landscape.
- Tap the media to toggle the overlay and check the bar does not move.
- No unit test accompanies this. Paparazzi supplies zero window insets, so a snapshot assertion would
  compare identical output before and after and prove nothing.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerView.kt
+++ b/libraries/mediaviewer/impl/src/main/kotlin/io/element/android/libraries/mediaviewer/impl/viewer/MediaViewerView.kt
@@ -414,6 +414,7 @@ private fun MediaViewerPage(
             if (showProgress) {
                 LinearProgressIndicator(
                     modifier = Modifier
+                        .padding(WindowInsets.safeDrawing.only(WindowInsetsSides.Horizontal + WindowInsetsSides.Top).asPaddingValues())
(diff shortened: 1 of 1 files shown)


bmarty (approved):
Thanks

Issue #2925, opened 2024-05-28
Fix translation consistency - detected in French, but probably applicable to other languages.


bmarty:
<img width="300" alt="image" src="https://github.com/element-hq/element-x-android/assets/3940906/c90a02c6-bbb0-4df8-a321-26bbe3189009">

"Prévenez-moi dans ce salon pour Mettre en sourdine" is not consistent. We should add some context on Localazy for those 4 strings. Also maybe create a new key for "Mettre en sourdine", which is `common.mute` and is used elsewhere (or a new key for the 3 possible choices?). The value to replace "Mettre en sourdine" could be "Aucun message" or something better?

PR #7367, opened 2026-08-09
Use a dedicated string for the Mute room notification mode

hayaksi1:
## Content

The room notification settings screen offers three mutually exclusive modes. Two of them already had
screen-scoped string keys, but the third reused the shared `common_mute`, which also labels the Mute
action button on the room details screen. A translator could not fix one occurrence without breaking
the other, and in French the option reads as a verb phrase where the list needs a noun phrase.

This adds `screen_room_notification_settings_mode_mute` and uses it at both places on that screen: the
option list and the read-only default-setting row. The English value is unchanged, so nothing renders
differently and no screenshot needs re-recording.

The Mute action button on the room details screen keeps `common_mute`, which is the one context where
the generic verb is correct.

## Motivation and context

Part of https://github.com/element-hq/element-x-android/issues/2925

The issue also asks for translator context to be added on Localazy for the four strings of this screen,
and suggests the English wording of this option could be reconsidered now that it has its own key. Both
are core-team actions that cannot be done from a fork.

## Tests

- The English value is identical, so the existing screenshot tests for the room notification settings
  screen and the existing presenter tests all pass unchanged, which is what proves the rendering did not
  move.
- No test asserts on the old key for this screen.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/notificationsettings/RoomNotificationSettingsItem.kt
+++ b/features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/notificationsettings/RoomNotificationSettingsItem.kt
@@ -12,7 +12,6 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.res.stringResource
 import io.element.android.features.roomdetails.impl.R
 import io.element.android.libraries.matrix.api.room.RoomNotificationMode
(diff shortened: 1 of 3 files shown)


bmarty (approved):
Thanks. We will add the string to Localazy and add some context there separately.

Issue #1933, opened 2023-11-30
Handle call events in the timeline


bmarty:
Start a call in a room and the timeline will look like:
<img width="338" alt="image" src="https://github.com/vector-im/element-x-android/assets/3940906/232ddf2c-7caf-4fbe-b7ad-29f73f46596d">

We may want to just hide those Event, maybe in the context of https://github.com/vector-im/element-meta/issues/2236

bblacher:
This was fixed in https://github.com/element-hq/element-x-android/pull/2639, right?

I just want to add that the events for jitsi calls still appear in the timeline as io.element.video.member

toger5:
We want the time-line events to depend on the call member events and not just the ringing/notify events. Even though using the ring events is an acceptable stop gap.

PR #7368, opened 2026-08-09
Hide legacy Jitsi call membership events from the timeline

hayaksi1:
## Content

Timelines for rooms that have been used with the older Jitsi-based video rooms show one
`Custom event io.element.video.member` row per participant joining or leaving. That event type is the
predecessor of `m.call.member`, which is already excluded from the timeline; because the newer type is
unknown to the app, the formatter falls through to its custom-state branch and renders the raw developer
string.

This adds the legacy type to the timeline's excluded events so it is filtered out by the SDK-level
timeline filter, exactly as the modern one already is. The room list preview was already unaffected,
since that rendering mode returns nothing for custom state events.

## Motivation and context

Part of https://github.com/element-hq/element-x-android/issues/1933

The original report was addressed for `m.call.member`; this covers the remaining legacy type reported
in the follow-up comment. The broader discussion in that issue, about deriving the call tile from call
membership rather than notify events, is not addressed here.

## Tests

- In a room carrying these events, the timeline no longer shows `Custom event io.element.video.member`
  rows.
- No unit test accompanies this. The value crosses the SDK boundary through
  `stateEventTypeFromString`, which is not available in JVM unit tests — the existing
  `StateEventTypeTest` already documents that limitation for custom state event types, and no other
  entry in this list is covered either.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/appconfig/src/main/kotlin/io/element/android/appconfig/TimelineConfig.kt
+++ b/appconfig/src/main/kotlin/io/element/android/appconfig/TimelineConfig.kt
@@ -13,11 +13,14 @@ import io.element.android.libraries.matrix.api.room.StateEventType
 object TimelineConfig {
     const val MAX_READ_RECEIPT_TO_DISPLAY = 3
 
+    private const val JITSI_CALL_MEMBER_EVENT_TYPE = "io.element.video.member"
+
     /**
      * Event types that will be filtered out from the timeline (i.e. not displayed).
      */
     val excludedEvents = listOf(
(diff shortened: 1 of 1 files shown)


hayaksi1:
### Visual Changes
| View / State | Before | After |
| :--- | :--- | :--- |
| Light | 01-before-light | 02-after-light |
| Dark | 03-before-dark | 04-after-dark |

PR #7369, opened 2026-08-09
Hide space children you have left that can only be joined by invitation

### LINKED ISSUE #6130: Leaving an `invite` room of a space will render it as Joinable at first
### Steps to reproduce

1. Join an `invite` room of a space.
2. Go to the space and see the room listed there.
3. Leave the room.

#### What did you expect?

The room disappears from the list as it can no longer be joined without an invite.

#### What happened instead?

The room remains visible and gets a Join button. Once you go out of the space and come back, the room is gone.

iOS has the same behaviour.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

_No response_

### Homeserver

_No response_

### Will you send logs?

Yes

hayaksi1:
## Content

After leaving a room listed in a space, the room stayed in the space's child list and was rendered with
a Join button, even though it can only be entered with a new invitation. Leaving the space and coming
back made it disappear, because the server omits it from the refetched hierarchy.

The presenter now applies the same rule locally: a child the current user has left and whose join rule
is `invite` is dropped from the list straight away.

The filter is deliberately narrow. It keys on the membership being `LEFT` combined with an `invite`
join rule, not on the derived visibility — that would collapse `knock` and unknown join rules into
"private" and would wrongly remove the Join button from knockable rooms and from any child whose join
rule the SDK did not populate. Public, knock, restricted and unknown join rules keep their Join button,
and joined or invited rooms are unaffected.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/6130

## Tests

- `features/space/impl/src/test/.../root/SpacePresenterTest.kt` — left children with `invite`, `public`,
  `knock` and absent join rules are emitted, and only the `invite` one is filtered out; a second case
  pins that `invite` children which are joined, invited or of unknown membership are all kept.
- The existing presenter and view tests are unchanged, and the space view snapshots are driven by the
  state provider rather than the presenter, so none are affected.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/space/impl/src/main/kotlin/io/element/android/features/space/impl/root/SpacePresenter.kt
+++ b/features/space/impl/src/main/kotlin/io/element/android/features/space/impl/root/SpacePresenter.kt
@@ -35,6 +35,7 @@ import io.element.android.libraries.matrix.api.core.toRoomIdOrAlias
 import io.element.android.libraries.matrix.api.room.BaseRoom
 import io.element.android.libraries.matrix.api.room.CurrentUserMembership
 import io.element.android.libraries.matrix.api.room.join.JoinRoom
+import io.element.android.libraries.matrix.api.room.join.JoinRule
 import io.element.android.libraries.matrix.api.room.powerlevels.permissionsAsState
 import io.element.android.libraries.matrix.api.spaces.SpaceRoom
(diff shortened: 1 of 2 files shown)


bmarty (approved):
Thanks!

Issue #6253, opened 2026-02-25
All notifications repeat / duplicate periodically until message is read.


Zach-Pouska:
### Steps to reproduce

Default installation of element X on android 15 (oneplus's oxygenos) from google play, notifications using firebase by default.

#### What did you expect?
One notification for one message.

#### What happened instead?
Notifications periodically duplicate. Not checking messages for a while results in 3 or more copies.

For one message of "test", 

<img width="178" height="48" alt="Image" src="https://github.com/user-attachments/assets/d497c47e-36c4-48a9-9a22-76696ed00b5b" />

 the following notifications are seen 
<img width="1080" height="362" alt="Image" src="https://github.com/user-attachments/assets/5464714e-5a22-4e1e-acc9-fffb4d421cbe" />

### Your phone model

OnePlus 10T

### Operating system version

Android 15

### Application version and app store

Element X 26.02.0 (202602000) from google play store

### Homeserver

matrix.org

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

jmartinesp:
Thanks for the logs. We've seen similar issues before with UnifiedPush, but not with Firebase.

Sadly, the logs don't contain enough info (I can only see logs for 'push received', but they don't mention any identifiers that could help us track the events). 

Could you go to developer settings (they can be enabled by tapping 7 times in the app version in the settings screen), enable tracing logs for the 'notification client', restart the app and once you reproduce it a couple of times send them again? With those it should be easier to understand what's happening.

Also, changing the general log level to `debug` would also add more useful logs.

Zach-Pouska:
Should have just uploaded more logs!

jmartinesp:
Ok, we have some progress, I see several `Queueing notifications` log messages, which only happen when the WorkManager worker for fetching notifications runs:

```
2026-02-25T22:26:22.081697Z DEBUG elementx: Queueing notifications: [NotificationEventRequest(sessionId=@sessionid, roomId=!room_id, eventId=$event_2, providerInfo=Firebase), NotificationEventRequest(sessionId=@sessionid, roomId=!room_id, eventId=$event_1, providerInfo=Firebase), NotificationEventRequest(sessionId=@sessionid, roomId=!room_id, eventId=$event_3, providerInfo=Firebase)] |
[...]

PR #7370, opened 2026-08-09
Stop repeating messages that are already shown in a room notification

hayaksi1:
## Content

When a room notification is rebuilt, the notification that is already on screen is used as the
accumulator: its messaging style is extracted and the incoming events are appended to it. Every event
was appended unconditionally, so an event that was re-delivered or re-resolved while its notification
was still showing was added a second time, and the same message appeared twice or more in the shade.
The notification also re-alerted each time.

Events whose id is already present in the existing messaging style are now skipped, and the
notification is posted with `onlyAlertOnce` when the incoming batch contained nothing new.

Two paths are deliberately exempt. Smart reply failures re-emit the same synthesised event id on
purpose in order to append the "message not sent" line, so they are still appended. Filtering happens
per event rather than per message, so an image and its caption — two messages for a single event — are
kept or skipped together.

This addresses the duplicated lines and the repeated alerts. It cannot help once the notification has
been dismissed, since there is then no existing notification to compare against, so it is a fix for the
visible duplication rather than for whatever causes the event to be delivered again.

## Motivation and context

Part of https://github.com/element-hq/element-x-android/issues/6253

## Tests

- `libraries/push/impl/src/test/.../factories/DefaultNotificationCreatorTest.kt` — an event that is
  already displayed is not added again and the rebuild no longer alerts; a genuinely new event is still
  added, in order; a smart reply failure for an already displayed event is still appended; and an image
  with a caption is not duplicated.
- These do not cover the re-delivery itself, which happens before this code runs.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
@@ -196,11 +196,12 @@ class DefaultNotificationCreator(
             isThread = threadId != null,
             roomIsGroup = !roomInfo.isDm,
         )
-        messagingStyle.addMessagesFromEvents(events, imageLoader)
+        val newEvents = messagingStyle.filterOutAlreadyDisplayedEvents(events)
+        messagingStyle.addMessagesFromEvents(newEvents, imageLoader)
         return builder
             .setCategory(category)
             .setNumber(events.size)
-            .setOnlyAlertOnce(roomInfo.isUpdated)
+            .setOnlyAlertOnce(roomInfo.isUpdated || newEvents.isEmpty())
             .setWhen(lastMessageTimestamp)
             // MESSAGING_STYLE sets title and content for API 16 and above devices.
             .setStyle(messagingStyle)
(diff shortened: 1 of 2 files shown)


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
```diff
@@ -416,6 +417,15 @@ class DefaultNotificationCreator(
             .build()
     }
 
+    private fun MessagingStyle.filterOutAlreadyDisplayedEvents(
+        events: List<NotifiableMessageEvent>,
+    ): List<NotifiableMessageEvent> {
+        val displayedEventIds = messages.mapNotNullTo(mutableSetOf()) { it.extras.getString(MESSAGE_EVENT_ID) }
+        return events.filter { event ->
+            event.isSmartReplyError() || displayedEventIds.add(event.eventId.value)
```
bmarty: It will be a bit clearer like this:

```suggestion
            event.isSmartReplyError() || event.eventId.value !in displayedEventIds)
```

bmarty (changes requested):
Thanks, only one remark about code clarity

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/factories/NotificationCreator.kt
bmarty: I wonder if the previous code was not better, since it also filter duplicates in `events`... So maybe revert? Also the code does not compile now.

bmarty (changes requested):

hayaksi1:
Done!

bmarty (approved):

PR #7371, opened 2026-08-09
Show the alias or the direct message user's Matrix ID in the room picker

### LINKED ISSUE #5428: Make recipients in the "Send to" viewmore destinguishable

Safely select a recipient in the "Send to" view

 I would like to share text or a image with someone using Element X.

Being able to distinguish multiple entries with a similar display name e.g. using the Matrix ID

1. Simply show the matrix id of every entry (easy fix)
2. Show the Matrix ID only if entries are similar
3. Allow viewing the chat by clicking an entry (i.e. selection only works through clicking the radio button icon)
4. Works, because you can only select one recipient anyways: the overall UX could be reworked, instead of selecting someone and then clicking send:
    1. remove the radio buttons 
    2. Instead of selecting a recipient, clicking on a recipient opens the chat of that entry in draft mode: the text is in the text box, ready to send; likewise images, you still could give the image a description and which isn't possible with the current UX; after entering a description, allow the user to see the chat history before requiring to send
    3. Allow the user to press "back" to abort and select a different entry

### Additional context

_No response_

No

hayaksi1:
## Content

The forward and share room picker showed only a display name, so several direct messages, or rooms
sharing a name, were indistinguishable and it was easy to send to the wrong one.

Each row now carries a secondary line: the room's canonical alias when it has one, which is what was
already shown, and otherwise the other person's Matrix ID when the room is a direct message. Rooms that
are neither aliased nor a direct message keep showing the name alone.

`heroes` is populated for any room whose name is derived from its members, not only for direct
messages, so the single hero is not on its own proof of a peer: a two person room that is not a direct
message would have shown a member's Matrix ID. `SelectRoomInfo` therefore carries `isDm`, taken from
`RoomInfo`, and the hero is read only when it is set. Group rooms with no alias remain
indistinguishable, since they have no public identifier to show and the internal room id would be
noise.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/5428

## Tests

- `libraries/roomselect/impl/src/test/.../RoomSelectViewTest.kt` — an aliased room renders its alias,
  a direct message without an alias renders the other user's Matrix ID, which the previous alias-only
  branch could not do, and a room that has a single hero but is not a direct message renders no Matrix
  ID.
- The preview state provider gained a direct message entry so the new branch is covered by the
  screenshot tests as well.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/roomselect/impl/src/main/kotlin/io/element/android/libraries/roomselect/impl/RoomSelectView.kt
+++ b/libraries/roomselect/impl/src/main/kotlin/io/element/android/libraries/roomselect/impl/RoomSelectView.kt
@@ -260,7 +260,8 @@ private fun RoomSummaryView(
                 maxLines = 1,
                 overflow = TextOverflow.Ellipsis
             )
-            val subtitle = roomInfo.canonicalAlias?.value ?: roomInfo.heroes.singleOrNull()?.userId?.value
+            val otherUserId = roomInfo.heroes.singleOrNull()?.userId?.takeIf { roomInfo.isDm }
+            val subtitle = roomInfo.canonicalAlias?.value ?: otherUserId?.value
             if (subtitle != null) {
                 Text(
                     text = subtitle,
--- a/libraries/roomselect/impl/src/main/kotlin/io/element/android/libraries/roomselect/impl/RoomSelectView.kt
+++ b/libraries/roomselect/impl/src/main/kotlin/io/element/android/libraries/roomselect/impl/RoomSelectView.kt
@@ -260,10 +260,10 @@ private fun RoomSummaryView(
                 maxLines = 1,
                 overflow = TextOverflow.Ellipsis
             )
-            // Alias
-            roomInfo.canonicalAlias?.let { alias ->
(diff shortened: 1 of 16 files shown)


libraries/roomselect/impl/src/main/kotlin/io/element/android/libraries/roomselect/impl/RoomSelectView.kt
```diff
@@ -260,10 +260,10 @@ private fun RoomSummaryView(
                 maxLines = 1,
                 overflow = TextOverflow.Ellipsis
             )
-            // Alias
-            roomInfo.canonicalAlias?.let { alias ->
+            val subtitle = roomInfo.canonicalAlias?.value ?: roomInfo.heroes.singleOrNull()?.userId?.value
```
bmarty: `heroes` can be present for room which are not DM, so maybe this is not OK to render a heroe in this case.

bmarty (commented):
Thanks, one remark.

bmarty (approved):
Thanks!

PR #7372, opened 2026-08-09
Copy a blocked user's Matrix ID by long pressing them

### LINKED ISSUE #4893: Be able to copy mxid from blocked user list

I have locally blocked a user as emergency measure due to invite spam. Now, I would like to add the MXID to my policy server and modbots. The problem is, how to get at the user id? There is literally no way to copy it into the clipboard.

I don't like Child Porn and Cat Torture

Either short-click on a user in that list (currently the only possible action is 'unblock'), which could lead to another screen with possible actions, similar to the user detail page. Or Long-Press a user which could open a context menu. Currently, this does nothing.

_No response_

### Additional context

_No response_

Yes

hayaksi1:
## Content

Long-pressing an entry in Settings, Blocked users now copies that user's Matrix ID to the clipboard and
confirms with a snackbar. A short tap still opens the unblock confirmation, so existing behaviour is
unchanged.

The gesture reuses the copy-to-clipboard idiom already used on the user profile screen, and the row
declares a long-click label so the action is announced by a screen reader. No new string was added.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/4893

The Matrix ID is displayed on this screen but there was no way to get it out of the app, which someone
who has blocked a user for abuse needs in order to report them onwards.

## Tests

- `features/preferences/impl/src/test/.../blockedusers/BlockedUsersPresenterTest.kt` — the copy event
  puts the Matrix ID on the clipboard and posts the "Copied to clipboard" snackbar message.
- `features/preferences/impl/src/test/.../blockedusers/BlockedUserViewTest.kt` — a long press on a row
  emits the copy event, which the previous plain clickable row could not do.
- These do not cover the clipboard implementation itself, which is faked in tests.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/blockedusers/BlockedUsersEvents.kt
+++ b/features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/blockedusers/BlockedUsersEvents.kt
@@ -12,6 +12,7 @@ import io.element.android.libraries.matrix.api.core.UserId
 
 sealed interface BlockedUsersEvents {
     data class Unblock(val userId: UserId) : BlockedUsersEvents
+    data class CopyToClipboard(val userId: UserId) : BlockedUsersEvents
     data object ConfirmUnblock : BlockedUsersEvents
     data object Cancel : BlockedUsersEvents
 }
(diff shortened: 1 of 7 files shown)


bmarty (approved):
Thanks!

PR #7373, opened 2026-08-09
Expose global account data get and set on MatrixClient

### LINKED ISSUE #5794: Missing public API to set or access AccountData (setAccountData not exposed)
The Rust SDK FFI includes a set_account_data function, but this functionality is not exposed in the Android/Kotlin layer of Element-X.
There is currently no public API to set or read AccountData from the Android SDK.
This prevents any client-side data synchronization that relies on AccountData.
Please expose the corresponding APIs in the Android layer.

hayaksi1:
## Content

`MatrixClient` had no way to read or write global account data, although the underlying client supports
it. This adds `getAccountData(eventType)` and `setAccountData(eventType, content)` to the interface and
implements them in `RustMatrixClient`, following the same `withContext(sessionDispatcher)` /
`runCatchingExceptions` idiom as the neighbouring methods.

The content is passed through as the raw JSON string exactly as the SDK returns and accepts it, with no
mapping to a Kotlin type: event types are open ended, so a typed wrapper belongs to whichever caller
needs one. `getAccountData` returns null when the event is not set.

Scope is limited to global account data. The room-level variant and the change-notification APIs the SDK
also exposes are not included; they can follow when a feature needs them.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/5794

## Tests

- `libraries/matrix/impl/src/test/.../RustMatrixClientTest.kt` — the value returned by the client is
  passed through unchanged and the underlying call is made exactly once with the requested event type;
  setting forwards both the event type and the raw content exactly once.
- The fakes in `libraries/matrix/test` and the impl test fixtures gained the two methods so existing
  consumers still compile.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty (approved):
Thanks!

PR #7374, opened 2026-08-09
Fix the feature module template writing the impl build script into the api module

### LINKED ISSUE #4666: Gradle Build has Circular Dependency error when creating new feature
"Template Module Feature Build Gradle API.kts" not being used so gradle.build for Impl calls itself instead.
generate_templates.sh broken? But it hasn't changed since June 2024.
Bug with "Generate Module from Template" plugin?
None of the files in /templates/ have changed in months, but it is very broken.

I tried to follow the directions at: https://github.com/element-hq/element-x-android/blob/develop/docs/_developer_onboarding.md#template-and-naming

Steps I took:
1. right click on features directory in Project (Android) view > Create Module From Template
1a. fill fields with myfeature MyFeature and click ok a few times
2. gradle sysnc
2a. confirm features/myfeature/impl/kotlin/io.element.android.features.myfeature.impl visible in Project View and has some files in it
3. right click on io.element.android.features.myfeature.impl *1 > new *2 > Template Presentation Classes
3*1. I'm not sure if this is what the directions say to do. They are a little unclear. I tried a few other things that seemed very wrong and this put the expected files in the expected place.
3*2. This is an additional menu not in the directions. Directions should be updated.
3a. File name: MyFeature *3
3*3. This is NOT what the instructions say to do. The instructions say to prefix the base class name with the package name, but that makes no sense, and looked very wrong when I tried to do it (see MySecondFeature in branch linked below) and also produced circular dependency error
4. gradle sync
5. attempt to see preview of MyFeatureView, it wants me to build and refresh
6. Build Output:
`
Circular dependency between the following tasks:
:features:myfeature:api:bundleLibCompileToJarDebug
+--- :features:myfeature:api:compileDebugJavaWithJavac
|    +--- :features:myfeature:api:bundleLibCompileToJarDebug (*)
|    +--- :features:myfeature:api:compileDebugKotlin
|    |    +--- :features:myfeature:api:bundleLibCompileToJarDebug (*)
|    |    \--- :features:myfeature:api:kspDebugKotlin
|    |         \--- :features:myfeature:api:bundleLibCompileToJarDebug (*)
|    \--- :features:myfeature:api:kspDebugKotlin (*)
+--- :features:myfeature:api:compileDebugKotlin (*)
\--- :features:myfeature:api:kspDebugKotlin (*)
`
7. confirm preview for ConfirmAccountProviderView works
8. right click /features/login/.../screens/confirmaccountprovider > new > Template Presentation Classes > file name: "TestPresenter"
9. Build & Refresh on /confirmaccountprovider/TestPresenterView preview pane
10. Preview shows "TestPresenter feature view" in day and night

Resulting in this state: https://github.com/lex-neufeld/element-lex-android/tree/refactor-AccountProvider-MyFeature

The differences between /features/login/api/build.gradle.kts and /myfeature/api/build.gradle.kts are numerous. Most suspicious is that the namespace for myfeature/api ends in .impl instead of .api Indeed, it's not following the template in /element-x-android/tools/templates/files/fileTemplates/Template Module Feature Build Gradle API.kts at all

I think the generated template is broken. Or maybe [Generate Module from Template](https://plugins.jetbrains.com/plugin/13586-generate-module-from-template) is not working as expected.

Please fix templates and update _developer_onboarding.md.

hayaksi1:
## Content

Generating a new feature with the Generate Module from Template plugin produced a module that could not
build: `features/<name>/api/build.gradle.kts` was written from the impl template, giving the api module
a dependency on itself and an impl namespace.

`FeatureModule.json` declared two placeholders, `BUILD_GRADLE_API` and `BUILD_GRADLE_IMPL`, both
defaulting to the literal `build.gradle.kts`, and used them as the keys of the `fileTemplates` map. The
plugin resolves those keys to filenames before matching, so both collapsed to the same key and the
second entry silently overwrote the first.

The same matching code caused a second, quieter breakage: it falls back to a suffix match on the file
path, so `DefaultXEntryPoint` matched the key `XEntryPoint` and the impl module's entry point was
generated from the api template, with an api package.

Both files are now keyed by their directory, which the plugin supports for exactly this case, and the
two placeholders are removed. That also brings the generation dialog back to the two fields the
documentation describes. The file templates themselves were correct and are unchanged.

The onboarding document gains a note that an already imported template has to be removed before
re-importing, because the plugin keeps its own copy and will not pick up changes to the JSON.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/4666

## Tests

- Import the template, create a module from it, and check that the api build script uses the library
  plugin and an api namespace, that the impl entry point declares an impl package, and that a Gradle
  sync succeeds.
- No automated test is possible: the mapping is interpreted by the IDE plugin, not by the build.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty (approved):
Should be fine, thanks

PR #7376, opened 2026-08-09
Show a Message not found placeholder when a replied to message cannot be loaded

### LINKED ISSUE #6960: Reactions to messages a user does not have access to are handled correctly on Android
### Steps to reproduce

steps:
1. Have a room that contains messages and an account that does not yet have access to this room
2. Invite your second account to this room (make sure he won't have access to the past messages)
3. Accept the invitation on the second account
4. Have the first account respond to a message which the second account has no access to.
5. Look at this response on the second account

#### What did you expect?
Message: Unable to load event that was replied to, it either does not exist or you do not have permission to view it.
and then this message (like in web) or like in iOS where the text is replaces by 2 rectangles.

#### What happened instead?
\\\> The replied to message contains an error message

\\\> Clicking the (error) replied to message causes an unclear error popup.

### Your phone model

Google pixel 8

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

Yes

Yes

hayaksi1:
## Content

When the message a reply points at cannot be fetched, the reply preview rendered the raw error string
coming from the SDK, in critical red styling. Tapping it emitted a focus request that could never
succeed, so the user got a "Message not found" dialog with no way forward.

The preview now renders the localised "Message not found" string in the same neutral italic style
already used for redacted and undecryptable replies, and it is no longer clickable, so the dead-end
dialog can no longer be reached.

No new string was added: this reuses `error_message_not_found`, which is the wording of the dialog this
tap used to raise. A more specific "you do not have access" wording was rejected because this state is
produced for any failure to load the replied-to event, not only for a permission problem.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/6960

## Tests

- `features/messages/impl/src/test/.../timeline/TimelineViewTest.kt` — a reply to an event that could
  not be loaded shows the localised placeholder instead of the SDK string and emits nothing when
  clicked; a reply to a loaded event still emits the focus event, which pins that the click handler was
  not removed too broadly.
- The composer's reply preview uses the same composable and inherits the change.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty (approved):
Thanks!

I have asked for a design confirmation here: https://www.figma.com/design/G1xy0HDZKJf5TCRFmKb5d5/Compound-Android-Components?node-id=2019-5548#1877971766 CC @americanrefugee

bmarty:
I will update the UI separately

PR #7377, opened 2026-08-09
Do not render an empty timeline row for state events without text

### LINKED ISSUE #3575: Redacted join events display as an empty box in the timeline
### Steps to reproduce

1. Join a user to a room
2. Redact that user's join event
3. Observe that EX renders an empty "room change" line (e.g. https://matrix.to/#/!aCvOWKNDeXXpPgquKq:matrix.org/$cZVjfEH3VJM0AWToXwNWQwdFl4Oiwo27wXsNCVxb_Zg?via=matrix.org&via=tchncs.de&via=envs.net)

Alternatively wait for a spammer to join the EXA room and the Administrator account to ban them and redact their join event, like the linked message above.

#### What did you expect?
To read either `$user joined the room` or `$user joined the room and was removed` or something like that.
Element Web simply renders `$user joined the room`

#### What happened instead?
No text is rendered

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

Yes

No

hayaksi1:
## Content

A membership event that has already been redacted by the time it reaches the client arrives with no
membership change, and the event formatter correctly produces no text for it. The timeline still laid
out a state event row around that empty text, which appeared as an empty box, and the event was still
counted by the collapsed "N room changes" header.

State events with a blank body are now dropped before the items are grouped, so no row is laid out and
the collapsed count matches what is shown when the group is expanded. The same situation is reachable
for a third party invite with no display name, and for a profile change that changes neither the
display name nor the avatar; one check covers all three.

Rendering the event the way Element Web does is not possible here: the SDK does not compute the
membership change for an already redacted event and does not backfill the previous content, so
suppressing the row is the only behaviour available to the client today.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/3575

## Tests

- `features/messages/impl/src/test/.../timeline/factories/TimelineItemsFactoryTest.kt` — an
  unformattable state event produces no timeline item, and it is not counted in a group of state
  events, which is what keeps the collapsed header honest.
- The shared test formatter fixture now returns a non-empty string, matching every real formatter,
  which returns either nothing or a non-empty string but never the empty string.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


jmartinesp:
I think this seems reasonable. @mxandreas what do you think?

PR #7378, opened 2026-08-09
Keep manually revealed media visible when re-entering a room

### LINKED ISSUE #6853: Images manually made visibile forget they were visible when re-entering the room
### Steps to reproduce

1. Set the "Show media in timeline" setting to "in private rooms"
2. Go to a public chatroom with media (e.g. #off-topic:gnome.org)
3. Find an image and click Show
4. Go back to the room list
5. Go back to the room and find the image again and it'll be hidden

This would be less of an issue if the "show media in timeline" option had a per-room override in addition to a global setting. I've realized that I'm only in rooms that ban media, or heavily moderate them, so I've just switched to showing them all by default to workaround this.

#### What did you expect?
The image to stay visible after I clicked Show once

#### What happened instead?
It was re-hidden after leaving the room.

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

Yes

Yes

hayaksi1:
## Content

With media previews hidden, tapping "Show" on an individual image revealed it only until the room was
closed. The set of revealed event ids lived on `TimelineProtectionPresenter`, which is scoped to the
room graph, and that graph is rebuilt every time the room is opened — so on re-entry the set was empty
and every image was hidden again.

The set moves into a new session-scoped `TimelineProtectionStore`, so a reveal now survives leaving and
re-entering the room.

The store is deliberately in-memory rather than persisted. Revealing media is a privacy control and
there is no "hide again" action today, so keeping the reveals for the session leaves restarting the app
as a cheap way to undo one. Session scope also means the set is dropped on logout and never leaks
between accounts.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/6853

## Tests

- `features/messages/impl/src/test/.../timeline/protection/TimelineProtectionPresenterTest.kt` — a
  second presenter built over the same store already reports the previously revealed event on its first
  emission, which is the re-entering-the-room case. The existing cases are unchanged and now also pin
  that two presenters with separate stores stay isolated.
- These do not cover the DI wiring itself, which follows the existing session-scoped store idiom.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/protection/TimelineProtectionStore.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/protection/TimelineProtectionStore.kt
@@ -7,25 +7,26 @@
 
 package io.element.android.features.messages.impl.timeline.protection
 
-import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.mutableStateSetOf
+import androidx.compose.runtime.snapshots.SnapshotStateSet
 import dev.zacsweers.metro.ContributesBinding
 import dev.zacsweers.metro.SingleIn
 import io.element.android.libraries.di.SessionScope
 import io.element.android.libraries.matrix.api.core.EventId
 
 interface TimelineProtectionStore {
-    val allowedEventIds: Set<EventId>
+    val allowedEventIds: SnapshotStateSet<EventId>
     fun allowEvent(eventId: EventId)
 }
 
 @ContributesBinding(SessionScope::class)
 @SingleIn(SessionScope::class)
 class DefaultTimelineProtectionStore : TimelineProtectionStore {
-    private val allowedEvents = mutableStateOf(emptySet<EventId>())
+    private val allowedEvents = mutableStateSetOf<EventId>()
 
-    override val allowedEventIds: Set<EventId> get() = allowedEvents.value
+    override val allowedEventIds: SnapshotStateSet<EventId> = allowedEvents
 
     override fun allowEvent(eventId: EventId) {
-        allowedEvents.value = allowedEvents.value + eventId
+        allowedEvents += eventId
     }
 }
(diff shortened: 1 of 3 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/protection/TimelineProtectionStore.kt
```diff
@@ -0,0 +1,31 @@
...
+
+interface TimelineProtectionStore {
+    val allowedEventIds: Set<EventId>
+    fun allowEvent(eventId: EventId)
+}
+
+@ContributesBinding(SessionScope::class)
+@SingleIn(SessionScope::class)
+class DefaultTimelineProtectionStore : TimelineProtectionStore {
+    private val allowedEvents = mutableStateSetOf<EventId>()
```
bmarty: Can be a simple `mutableSetOf` instead of `mutableStateSetOf`.

bmarty (changes requested):
One remark, else this is fine!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/protection/TimelineProtectionStore.kt
hayaksi1: Done

bmarty: Changing 

```kotlin
private val allowedEvents = mutableStateSetOf<EventId>()
```

to

```kotlin
private val allowedEvents = mutableStateOf(emptySet<EventId>())
```

is not what I have asked. I want:

```kotlin
private val allowedEvents = mutableSetOf<EventId>()
```

bmarty (changes requested):

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/protection/TimelineProtectionStore.kt
bmarty: OK, sorry I was wrong, please revert to the first version of the code

bmarty: To avoid confusion, please change the type of `allowedEventIds` to `SnapshotStateSet<EventId>`.

bmarty (approved):

PR #7379, opened 2026-08-09
Do not display pinned events that the client cannot render

### LINKED ISSUE #7314: handle invalid pinned events gracefully
### Steps to reproduce

1. Set invalid data as one of the entries in the pinned events. E.g. by pinning a local echo using e-web 1.11
2. Also pin some valid events
3. Look at room in EXA

#### What did you expect?
Can see any valid pinned events but invalid ones are hidden.

https://spec.matrix.org/v1.19/client-server-api/#mroompinned_events specifically says
> Clients are responsible for determining if a particular event in the pinned list is displayable, and have the option to not display it if it cannot be pinned in the client.

#### What happened instead?
The whole event is displayed as "unsupported"

### Your phone model

_No response_

### Operating system version

_No response_

schildi next nightly based on exa 26.08.0

### Homeserver

synapse

### Will you send logs?

No

No

hayaksi1:
## Content

Pinned events whose content the client cannot render — unknown event types, and message-like or state
events that failed to parse — are no longer shown. They are dropped from both the pinned messages
banner and the pinned messages list, along with any day divider left without an event to introduce.
When no pinned event can be displayed, the list shows its empty state.

Events that can still be rendered are unaffected, including undecryptable events and redacted ones,
which carry information the user needs.

Such an event can still be unpinned from the room timeline, where it remains visible.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/7314

The pinned events timeline is built with an exclusion filter rather than an allowlist, so an event of a
type the app has no renderer for reaches both surfaces: the banner falls through to the formatter's
catch-all and reads "Unsupported event", and the list renders an unknown content tile. The
specification makes this the client's call — clients are responsible for deciding whether a pinned
event is displayable and may choose not to display it.

## Tests

- `features/messages/impl/src/test/.../pinned/banner/PinnedMessagesBannerItemFactoryTest.kt` — an item
  is produced for a message and not for any of the three unrenderable contents, nor for a virtual item.
- `features/messages/impl/src/test/.../pinned/list/PinnedMessagesListPresenterTest.kt` — an
  unrenderable pinned event and its orphaned day divider are both dropped and the count reflects only
  what is shown; a timeline whose only pinned event is unrenderable yields the empty state rather than
  a filled one containing just a day divider.
- `features/messages/impl/src/test/.../pinned/banner/PinnedMessagesBannerPresenterTest.kt` — the banner
  count and page follow the same rule.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/pinned/DisplayablePinnedEvents.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/pinned/DisplayablePinnedEvents.kt
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2026 Element Creations Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.features.messages.impl.pinned
+
+import io.element.android.libraries.matrix.api.timeline.MatrixTimelineItem
+import io.element.android.libraries.matrix.api.timeline.item.event.EventContent
+import io.element.android.libraries.matrix.api.timeline.item.event.FailedToParseMessageLikeContent
+import io.element.android.libraries.matrix.api.timeline.item.event.FailedToParseStateContent
+import io.element.android.libraries.matrix.api.timeline.item.event.UnknownContent
+import io.element.android.libraries.matrix.api.timeline.item.virtual.VirtualTimelineItem
+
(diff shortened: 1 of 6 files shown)


bmarty (approved):
LGTM, thanks!

ganfra:
> LGTM, thanks!

Should we not handle this at the sdk level instead?

bmarty:
> > LGTM, thanks!
> 
> Should we not handle this at the sdk level instead?

I'd say it's up to the app to decide what to do with such events. We could have a SDK configuration though.

PR #7380, opened 2026-08-09
Avoid the settings layout jumping when the developer options entry appears

### LINKED ISSUE #6703: Layout jumping while enabling developer options
### Steps to reproduce

1. Open Element X
2. Click on profile picture in the upper left corner
3. Scroll down to version
4. Click the version section at the button a few times (maybe a few too much)
5. Developer options get shown in list
6. Due to still clicking (see 5.) the developer options open and some settings may change

#### What did you expect?
The list gets scrolled to the bottom after the developer options appeared in the list, so the user still clicks on the version button, which does not have effect.

#### What happened instead?
Layout shifts because the developer options appears in list.

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

No

No

hayaksi1:
## Content

Unlocking developer settings takes seven taps on the version string at the bottom of the settings
screen. The entry was inserted into the scrollable column just above the version, so on the seventh tap
everything below the insertion point moved down by one row — including the version text under the
user's finger — and the freshly inserted "Developer options" row took its place. An eighth tap
therefore opened developer settings by accident.

The row is now always composed, and is invisible and inert until developer settings are unlocked, so
its height is reserved from the start. The version string does not move, the entry simply appears in
the gap that was already on screen, and any extra tap still lands on the version string.

While hidden the row is drawn at zero alpha, has its semantics cleared and hidden from accessibility
services, and has no click handler, so it is not announced by a screen reader and is not focusable.
This follows the placeholder idiom already used by the typing notification view.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/6703

Reserving the space is what the issue thread asked for, rather than moving the entry below the version
string: when the user is scrolled to the bottom, a row placed below the version renders past the
viewport edge, so seven taps would appear to do nothing.

## Tests

- `features/preferences/impl/src/test/.../root/PreferencesRootViewTest.kt` — the version string's
  position in the root is unchanged when the entry becomes visible, and the entry does exist afterwards
  so the assertion cannot pass vacuously.
- The existing case asserting the entry is not shown when developer settings are locked now also guards
  the accessibility behaviour, since it fails if the semantics clearing is dropped.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty (commented):
This is not OK to have the extra vertical space always displayed above the version on this important screen.

hayaksi1:
> This is not OK to have the extra vertical space always displayed above the version on this important screen.

So what should be approach to fix?

bmarty:
I have closed the issue, please see https://github.com/element-hq/element-x-android/issues/6703#issuecomment-5263629897

PR #7381, opened 2026-08-09
Show the account being verified on the session verification screen

### LINKED ISSUE #2690: Element X asks for my recovery key but does not give me any information what account it tries to recover/verify
I want to use Element X it is already logged in but I dont know with which account.
I havent used it for a while and just started it again.

_No response_

### Additional context

The app does not even give me an option to log out. The only thing i can do is ask another client or enter a recovery key but i dont know for which client.

Yes

hayaksi1:
## Content

The self-verification screen shown to an existing unverified session displayed only a lock icon and two
generic sentences, so someone with more than one account had no way to tell which account the app was
asking a recovery key for.

The screen now shows that account under the existing title and subtitle: avatar, display name and
Matrix ID. When the account has no display name, the Matrix ID is shown on its own.

The account is read from the client's user profile flow, so it stays correct if the profile is
refreshed while the screen is open. No new string is needed. The shared `BigIcon` component is
deliberately left alone: swapping the lock icon for the avatar would change a Compound component used
across the app and its own snapshot suite, so the account is added alongside the icon instead.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/2690

## Tests

- `features/ftue/impl/src/test/.../ChooseSessionVerificationModePresenterTest.kt` — the state carries
  the current user, and a later profile update is reflected in the next emission, which proves the flow
  is collected rather than read once.
- `features/ftue/impl/src/test/.../ChooseSessionVerificationModeViewTest.kt` — the display name and
  Matrix ID are both rendered, and an account with no display name still shows its Matrix ID.
- The recovery key entry screen itself is unchanged and still shows no account context.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty:
Issue has been closed, see https://github.com/element-hq/element-x-android/issues/2690#issuecomment-5251636543

PR #7382, opened 2026-08-09
Expand the emoji picker to full height when the keyboard is shown

### LINKED ISSUE #7270: Emoji picker is hidden by keyboard
### Steps to reproduce

1. Long press on a message.
2. Tap the emoji picker button.
3. Tap the "Search emojis" text box.
4. Search for an emoji.

#### What did you expect?

I expected to be able to see the emojis I’m searching for.

#### What happened instead?

All emojis are hidden beneath the keyboard, which is layered above the emoji picker.

### Your phone model

Fairphone 5

### Operating system version

Android 16

Element X 26.07.1 installed via Obtainium

### Homeserver

matrix.org

### Will you send logs?

Yes

No

hayaksi1:
## Content

The emoji picker bottom sheet now expands to full height while the soft keyboard is visible, so search
results are laid out above the keyboard instead of behind it.

The picker content is rendered with `fillMaxSize`, so the sheet measures itself to the full window
height. That makes the partially expanded anchor available, and the modal sheet auto-shows at that
anchor — half way down the screen. The sheet does reserve the IME inset through its content window
insets, but it applies that padding at its own bottom edge, which is off-screen while the sheet rests
half way down. The band that is both on screen and above the keyboard is then about a tenth of the
screen: enough for the search field and little else.

Expanding the sheet while the keyboard is up puts its bottom edge back at the screen edge, so the inset
it already reserves lands exactly on top of the keyboard.

`Modifier.imePadding()` on the sheet content is not an option here: the sheet applies the insets with
`windowInsetsPadding`, which consumes them, so a nested `imePadding()` reads zero.

Expanding the sheet also brought a blank strip between the results and the keyboard into view.
`EmojiPickerView` pads its search bar with `Modifier.padding(bottom = 10.dp)` to separate the collapsed
field from the category tabs underneath it, and that padding stayed applied once the search bar
expanded to fill the picker, so the picker stopped 10dp short of its own bottom edge. Until now that
strip sat below the visible part of a half-height sheet; against the keyboard it reads as a gap. The
padding is now applied only while the search bar is collapsed, which is the only state that has
anything below it.

This covers the reaction emoji picker, which is the one reported. The emoji picker on the user status
screen resolves its sheet height the same way and has the same problem; it is left for a separate
change because the sheet there is configured differently.

## Motivation and context

Closes https://github.com/element-hq/element-x-android/issues/7270

## Tests

- Long press a message, open the emoji picker and tap the search field: the sheet expands and the
  matching emoji are visible and tappable above the keyboard.
- Open the picker without searching: it still opens at its usual height.
- `features/messages/impl/src/test/.../customreaction/ExpandSheetWhenImeIsVisibleTest.kt` — dispatching
  IME insets to the composition expands the sheet, and leaves it at its resting height while the
  keyboard is hidden.
- `libraries/emoji/impl/src/test/.../picker/EmojiPickerViewTest.kt` — with the search active, the
  results grid reaches the bottom of the picker. It fails against the previous code, where the grid
  stopped 10dp short.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/emoji/impl/src/main/kotlin/io/element/android/libraries/emoji/impl/picker/EmojiPickerView.kt
+++ b/libraries/emoji/impl/src/main/kotlin/io/element/android/libraries/emoji/impl/picker/EmojiPickerView.kt
@@ -62,7 +62,7 @@ internal fun EmojiPickerView(
 
     Column(modifier) {
         SearchBar(
-            modifier = Modifier.padding(bottom = 10.dp),
+            modifier = Modifier.padding(bottom = if (state.isSearchActive) 0.dp else 10.dp),
             queryState = state.searchQuery,
             resultState = state.searchResults,
             active = state.isSearchActive,
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/customreaction/CustomReactionBottomSheet.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/customreaction/CustomReactionBottomSheet.kt
@@ -13,6 +13,7 @@ import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.isImeVisible
 import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.SheetState
 import androidx.compose.material3.SheetValue
 import androidx.compose.material3.rememberBottomSheetState
 import androidx.compose.runtime.Composable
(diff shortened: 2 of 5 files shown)


hayaksi1:
Done

jmartinesp:
Thanks. I think the way we'd have fixed this would be to use `modifier.imePadding()` to the `ModalBottomSheet` inside `CustomReactionBottomSheet`, since that makes the first few emojis visible, but expanding the bottomsheet when the keyboard is visible as done here makes more sense UX-wise 👍 .

The only problem I see seems to be this blank 'bar' here:

image


I guess some other insets might be conflicting there? Can I ask you to check if it's possible to remove it?

libraries/emoji/impl/src/main/kotlin/io/element/android/libraries/emoji/impl/picker/EmojiPickerView.kt
```diff
@@ -62,7 +62,7 @@ internal fun EmojiPickerView(
 
     Column(modifier) {
         SearchBar(
-            modifier = Modifier.padding(bottom = 10.dp),
+            modifier = Modifier.padding(bottom = if (state.isSearchActive) 0.dp else 10.dp),
```
jmartinesp, line 65: Ah, that was a tricky one. Good catch.

jmartinesp (approved):
Approving this, assuming the CI passes. Thanks again!

Issue #7321, opened 2026-08-05
[Task] Simplify the change/selection of account provider


mxandreas:
Please see and discuss the details in the meta issue.

bmarty:
Done in #7394

PR #7394, opened 2026-08-10
Simplify choosing the account provider during sign in, with autocomplete of previously used ones

jennaharris7:
## Content

Simplifies choosing the account provider during manual sign in and account creation

- Consolidates the account-provider confirmation, change and search screens into a single editable entry screen.
- Removes the Android-only homeserver resolver / search list (iOS parity).
- Adds inline autocomplete (grey ghost-text) of previously-used account providers, stored locally per device and never synced.
- Defaults the field to `matrix.org` on first launch, and to the last successfully-used provider afterwards.

## Motivation and context

Implements element-hq/element-meta#3283 (task element-hq/element-x-android#7321)

## Screenshots / GIFs

[Screen_recording_20260810_102939.webm](https://github.com/user-attachments/assets/12680d32-184a-4c9d-a197-4396d0c6e2a8)

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 16

## Checklist

    - [ ] No.


libraries/ui-strings/src/main/res/values/temporary.xml
```diff
@@ -10,4 +10,10 @@
     <string name="a11y_state_collapsed">"Collapsed"</string>
     <string name="a11y_active">Active</string>
     <string name="a11y_inactive">Inactive</string>
+    <string name="screen_select_server_title_login">"Enter your account provider"</string>
+    <string name="screen_select_server_title_register">"Choose an account provider"</string>
+    <string name="screen_select_server_textfield_header">"Account provider"</string>
+    <string name="screen_select_server_textfield_placeholder">"example.com"</string>
+    <string name="screen_select_server_textfield_footer_login">"If your account is provided by your company or a community, then change the account provider (example: companyname.com)."</string>
+    <string name="screen_select_server_textfield_footer_register">"Create an account on matrix.org for free. If you want to create an account with a different provider, then change it accordingly (example: communityname.org)."</string>
```
bmarty: The strings have been added to Localazy (ex: `screen.change_server.textfield_footer.register`)

bmarty (changes requested):
Thanks!

I spotted a bug:
- on the select server screen, when `https://matrix.org` is pre-filled, edit it:
- remove the `g`
- add the `g` back
- Press continue.

An error occurs:

Image

- Close the error dialog and press "Continue again: it works.

Also:
- I will sync the Localazy strings today, so that the strings for this feature will be available.
- `https://` could be added automatically. Now entering `matrix.org` does not work. We have an extension `ensureProtocol` which could be used for that.
- `matrix.org` could be in the autocompletion store by defaut, even when no successful login have been done on it.

libraries/ui-strings/src/main/res/values/temporary.xml
jennaharris7: Bug is fixed, ensureProtocol and matrix.org have been added, and the strings have been updated 👍

bmarty:
Thanks for the update! The feature is working pretty well.

I have done more tests and here are my remarks:

1. (major) when entering `element.io` for the homeserver, and login is successful, `https://element.io` is stored. But then when entering `ele` I have no completion to `element.io`. If I enter `htt`, I have completion to `https://element.io`. Entering `https://` should really be optional for the user, completion should work if they enter `ele`. Also see the acceptance criteria in https://github.com/element-hq/element-meta/issues/3283, `https://` is never mentioned. And first criteria is "`matrix.org` is shown by default", but actually `https://matrix.org` is shown by default.
2. (major) For enterprise build, if `EnterpriseService.defaultHomeserverList()` (which has been renamed) returns a list of servers plus `*` (for instance `listOf("element.io", "matrix.org", "*")`), this screen was displayed:

image

Now the user has no possibility to see the list of servers, and they are not used for the completion neither.

3. (minor) when entering `htt` and completion is displayed, if I press the "Continue" button, I think that it will be better if the completion is applied, i.e. the TextField does not render the completion, but the complete server. In this case, when pressing back when the MAS page is displayed, the full homeserver url is rendered.
4. (minor / not sure) when a completion is displayed, pressing "Enter" on the keyboard should maybe first apply the suggestion. When there is no suggestion, pressing "Enter" can keep the same current effect, i.e. submitting the form. To summarize 3. and 4., when a completion is displayed: when user presses "Continue", the suggestion is applied and the server is contacted (no need to press "Continue" twice), but "Enter" key on the keyboard has to be pressed twice for the same effect.
5. Potential bug, but can it really happen? if `https://a.b.c` is stored, it will not be possible for the user to login to `https://a.b`, because the suggestion is always applied and there is no way to reset the store.
6. The screenshots have to be recorded, but I guess you wait for the PR to be ready to merge to do this final step.

bmarty (approved):
Thanks for the update! I have tested the application and I believe that the behavior is correct.

I will trigger the screenshot recording.

bmarty (approved):

Issue #7393, opened 2026-08-10
Can't report a problem from within the apl


falkorichter:
### Steps to reproduce

1. Where are you starting? - wanted to start the app - it mentioned that it had crashed previously and I agreed to send logs.
3. What do you click? - send logs

<img width="1280" height="2856" alt="Image" src="https://github.com/user-attachments/assets/90997e15-6274-4111-ace8-0b8acb198929" />

4. I tried multiple times it always fails.

Here are the log files

[logcat.log](https://github.com/user-attachments/files/30905449/logcat.log)

[logs.2026-08-10-17.log](https://github.com/user-attachments/files/30905453/logs.2026-08-10-17.log)

#### What did you expect?
The report to be sent 
#### What happened instead?
Error pop up

### Your phone model

Pixel 10 pro

### Operating system version

Android 17

### Application version and app store

26.08.0

### Homeserver

Matrix.oeg

### Will you send logs?

Yes

### Will you submit a Pull Request?

No

PR #7395, opened 2026-08-10
Fix bug reports failing to send after the app has crashed

hayaksi1:
## Content

Sending a bug report fails with the server error `Could not save report` when the app has previously
crashed and "Send log files" is enabled. The full crash call stack is appended to the report
description, and the bug report server copies that description verbatim into the GitHub issue it
opens. GitHub refuses an issue body over 65536 characters, so the server's save step fails and the
app surfaces the error it returns.

The description is now truncated before being sent. The user's own text and the GitHub issue link
come first, so only the tail of the crash stack is dropped, and a marker is appended in its place.
Nothing is lost: the uncaught exception handler already writes the full stack through Timber, so it
is present in the attached log files.

The stored crash data is only cleared after a report has been sent successfully, which is why a
report failing this way keeps failing on every retry until the cache is cleared.

## Motivation and context

Relates to https://github.com/element-hq/element-x-android/issues/7393, and to the earlier report of
the same error in https://github.com/element-hq/element-x-android/issues/5739, where the cause was
not identified.

The existing limits bound the uploaded logs — `MAX_LOG_CONTENT_SIZE` and `MAX_LOG_LINES_SIZE`, added
in #6003 and #6075 — and the server rejects those with a different error. Nothing bounded the
description itself.

## Tests

- `DefaultBugReporterTest` — a description over the limit is truncated, carries the marker, and
  stays below GitHub's issue body limit.
- `DefaultBugReporterTest` — a description within the limit is sent unchanged, with the crash call
  stack intact.
- Both were confirmed to fail with the change reverted.

These cover the payload the app builds; they do not exercise the real bug report server, so
end-to-end submission is not proven by them.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/appconfig/src/main/kotlin/io/element/android/appconfig/RageshakeConfig.kt
+++ b/appconfig/src/main/kotlin/io/element/android/appconfig/RageshakeConfig.kt
@@ -35,4 +35,11 @@ object RageshakeConfig {
      * The maximum number of log lines a rageshake can contain.
      */
     const val MAX_LOG_LINES_SIZE = 1_000_000
+
+    /**
+     * The maximum size of the description. The bug report server copies it verbatim into a GitHub issue body,
+     * which GitHub rejects above 65536 characters. The remaining budget is left to the other reported fields,
+     * mainly the list of the omitted log file names.
+     */
+    const val MAX_DESCRIPTION_SIZE = 40 * 1024
 }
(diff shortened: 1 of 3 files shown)


jmartinesp (approved):

jmartinesp:
Makes sense, thanks again for the fixes. Could you fix the lint issue before merging this?

hayaksi1:
@jmartinesp Done

Issue #5203, opened 2025-08-21
Videos uploaded in the app are significantly degraded in quality compared to native uploads


Valenthyne:
### Steps to reproduce

1. Where are you starting? What can you see?

I begin in a chat in which I would like to upload a video file, specifically one to which I have already uploaded the video.

2. What do you click?

First, I download the video that was uploaded by tapping the video, tapping the "Information" icon in the top right, and tapping "Save".

Then, I tap the "+" to go to upload a video file by selecting "Photo & Video Library" and then the video in question. It is a .webm file of approximately 4.2 MB in size.

3. More steps…

I submit the video by clicking the green "Send" button. I go to check the video and notice that it is not only in much inferior video quality, but the size has also been crunched by around a third.

I also try uploading the video with "Optimize media quality" enabled, and the result is the same, leading me to believe the toggle is perhaps broken and optimization is forced regardless.

#### What did you expect?

I expected the video quality to be the same as it was when uploaded through the native Element desktop application (Linux) which appears to be the same as the file's original quality.

#### What happened instead?

The video quality is considerably degraded despite the fact that "Optimize media quality" is NOT enabled under "Advanced settings".

It would appear that optimization is taking place regardless of the option selected in the app. There doesn't seem to be any good reason why the app would force optimization despite native applications uploading them as normal.

I do see that the original Element app also forces compression (though in this case it creates a file three times the size, defeating the purpose).

If this forced optimization is intended behavior, I suppose this is less of a bug report and more of a suggestion to make optimization a choice instead. Further clarification may be necessary on the toggle as well.

### Operating system version

Android 16

### Application version and app store

Element X v2025.08.3 (GitHub Release APK)

### Will you send logs?

No

### Are you willing to provide a PR?

No

jmartinesp:
> The video quality is considerably degraded despite the fact that "Optimize media quality" is NOT enabled under "Advanced settings".

I believe this was fixed in https://github.com/element-hq/element-x-android/pull/5183 but hasn't made it to a release yet. Sorry for the uncaught regression!

> If this forced optimization is intended behavior, I suppose this is less of a bug report and more of a suggestion to make optimization a choice instead. Further clarification may be necessary on the toggle as well.

It's the intended behaviour for compatibility - I won't enter into details much here since I already did in other issues, but it's done for compatibility between platforms -, although we want to change how it works to add several presets the user can choose.

Valenthyne:
[...]

PR #7400, opened 2026-08-11
Update DefaultMediaOptimizationSelectorPresenter.kt

escix:
By Claude.

### Problem

When compressing a video before upload, the size the user is shown in the media-quality picker (e.g. "≈4 MB" for the Low preset) can be wildly optimistic compared to the actual uploaded file size (e.g. 19 MB, from a 74 MB source). This makes the size estimate misleading and can cause uploads to exceed a homeserver's upload limit even when the picker said they'd fit.

This is also visible as general "video quality/size is unpredictable after compression" reports, e.g. [element-hq/element-x-android#5203](https://github.com/element-hq/element-x-android/issues/5203) and [#4191](https://github.com/element-hq/element-x-android/issues/4191).

### Root cause

Two independent issues compound to make the estimate diverge from reality:

1. **`VideoCompressor` requests VBR, which many hardware encoders don't tightly bound.**
   In `VideoCompressor.kt`, the `DefaultEncoderFactory` is configured with `BITRATE_MODE_VBR`:

   ```kotlin
   VideoEncoderSettings.Builder()
       // Use VBR which is generally better for quality and compatibility, although slightly worse for file size
       .setBitrateMode(MediaCodecInfo.EncoderCapabilities.BITRATE_MODE_VBR)
       .setBitrate(newBitrate)
       .build()
   ```

   Under VBR, `setBitrate()` is treated by many device encoders as a rough average/target rather than a ceiling — for busier footage the encoder can spend significantly more bits than requested to preserve quality. This is a known characteristic of `androidx.media3` Transformer specifically (see [androidx/media#1707](https://github.com/androidx/media/issues/1707), where a user configured a lower bitrate/resolution/framerate and still got a *higher* output bitrate than the input). This is the dominant cause of the multi-times-larger-than-estimated output.

2. **The upload size estimate only accounts for the video bitrate, never the audio track.**
   In `DefaultMediaOptimizationSelectorPresenter.kt`, the per-preset size estimate is:

   ```kotlin
   val bitRateAsBytes = preset.compressorHelper().calculateOptimalBitrate(videoDimensions, 30) / 8f
   val durationInSeconds = duration.inWholeSeconds.toFloat()
   val calculatedSize = (bitRateAsBytes * durationInSeconds * 1.1f).roundToLong() // Adding 10% overhead for safety
   ```

   But the compressed output always includes an AAC audio track (`.setAudioMimeType(MimeTypes.AUDIO_AAC)` in `VideoCompressor.kt`), which this formula never accounts for. For longer videos this alone can be a multi-MB gap.

3. **Silent fallback handling made this hard to diagnose.**
   `Transformer.Listener.onFallbackApplied` was a no-op (`= Unit`). If a device's encoder can't honor the requested bitrate/mode and Media3 substitutes different settings, the app had no visibility into it at all.

### Changes

**`libraries/mediaupload/impl/.../VideoCompressor.kt`**
- Switch the video encoder from `BITRATE_MODE_VBR` to `BITRATE_MODE_CBR` so the computed bitrate is enforced as a target rather than treated as a loose average. This is the primary fix for output size unpredictability.
- Log `onFallbackApplied` (previously a no-op) so we can see when a device's encoder can't honor our requested transformation and substitutes something else.
- Log key `ExportResult` fields (`durationMs`, `fileSizeBytes`, `videoEncoderName`, `averageAudioBitrate`) in `onCompleted`, to give us real data for tuning the audio-bitrate assumption below across real devices, since Media3's `DefaultEncoderFactory` doesn't expose a public way to pin the audio bitrate up front.

**`features/messages/impl/.../DefaultMediaOptimizationSelectorPresenter.kt`**
- Add an `ASSUMED_AUDIO_BITRATE` constant (128 kbps, a reasonable default for AAC) and fold it into the per-preset size estimate, so the audio track is no longer silently excluded.

### Testing

- [x] Verified the patch applies cleanly against a fresh checkout of `26.08.0` (`git apply --check`).
- [ ] Manual test: compress the same video at Low/Standard/High before and after this change, on at least one device, and compare estimated vs. actual output size.
- [ ] Manual test: force an encoder fallback (e.g. request CBR on a device/codec combo that doesn't support it) and confirm `Timber.w` fires with the fallback details.
- [ ] Existing unit tests (`VideoCompressorConfigFactoryTest`, `DefaultMediaOptimizationSelectorPresenterTest`) should continue to pass unmodified — neither test hardcodes an expected `sizeInBytes`/bitrate value, only resize/no-resize behavior, which is unaffected by this change.

### Known limitations / follow-ups

- `ASSUMED_AUDIO_BITRATE` (128 kbps) is a reasonable default for Media3's AAC encoder, not a value read back from the actual encoder — `DefaultEncoderFactory`'s public API in media3 1.10.1 doesn't expose a way to pin the audio bitrate the way `VideoEncoderSettings` does for video. The new `onCompleted` logging of `averageAudioBitrate` is meant to let us confirm/tune this constant against real device data; if it's consistently off, a follow-up PR should adjust it (or explore whether a newer media3 version exposes audio encoder settings).
- Even with `BITRATE_MODE_CBR`, not every device/encoder combination is guaranteed to honor it — the new fallback logging will surface cases where a device silently reverts to a different mode, which we should monitor after this ships.
- This does not address the separate, larger design question raised in #5203 — that there is currently no way to fully skip video re-encoding, unlike images (`compressImages` bool). That would be a bigger, separate change and is out of scope here.

### Related issues

- Closes/relates to #5203 (videos degraded and forced-compressed regardless of settings)
- Relates to #4191 (compressed video ending up larger than original for some sources)

### DIFF
--- a/libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/VideoCompressor.kt
+++ b/libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/VideoCompressor.kt
@@ -1,5 +1,215 @@
 /*
  * Copyright (c) 2025 Element Creations Ltd.
+ * Copyright 2025 New Vector Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.features.messages.impl.attachments.video
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.produceState
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import dev.zacsweers.metro.Assisted
+import dev.zacsweers.metro.AssistedFactory
+import dev.zacsweers.metro.AssistedInject
+import dev.zacsweers.metro.ContributesBinding
+import io.element.android.libraries.architecture.AsyncData
+import io.element.android.libraries.core.mimetype.MimeTypes.isMimeTypeVideo
+import io.element.android.libraries.di.SessionScope
+import io.element.android.libraries.featureflag.api.FeatureFlagService
+import io.element.android.libraries.featureflag.api.FeatureFlags
+import io.element.android.libraries.mediaupload.api.MaxUploadSizeProvider
+import io.element.android.libraries.mediaupload.api.MediaOptimizationConfigProvider
+import io.element.android.libraries.mediaupload.api.compressorHelper
+import io.element.android.libraries.mediaviewer.api.local.LocalMedia
+import io.element.android.libraries.preferences.api.store.VideoCompressionPreset
+import kotlinx.collections.immutable.ImmutableList
+import kotlinx.collections.immutable.toImmutableList
+import timber.log.Timber
+import kotlin.math.roundToLong
+
+@AssistedInject
+class DefaultMediaOptimizationSelectorPresenter(
+    @Assisted private val index: Int,
+    @Assisted private val localMedia: LocalMedia,
+    @Assisted private val sendAsFile: Boolean,
+    private val maxUploadSizeProvider: MaxUploadSizeProvider,
+    private val featureFlagService: FeatureFlagService,
+    private val mediaOptimizationConfigProvider: MediaOptimizationConfigProvider,
+    private val videoCompressionPresetSelector: VideoCompressionPresetSelector,
+    mediaExtractorFactory: VideoMetadataExtractor.Factory,
+) : MediaOptimizationSelectorPresenter {
+    companion object {
+        /**
+         * Approximate bitrate (in bits per second) used by the AAC audio encoder for compressed videos.
+         * The video pre-processor doesn't override the audio encoder settings, so this should track
+         * whatever the platform default AAC bitrate ends up being. Used to avoid under-estimating the
+         * final file size, since the video bitrate alone doesn't account for the audio track.
+         */
+        private const val ASSUMED_AUDIO_BITRATE = 128_000
+    }
+
+    @ContributesBinding(SessionScope::class)
+    @AssistedFactory
+    interface Factory : MediaOptimizationSelectorPresenter.Factory {
+        override fun create(
+            index: Int,
+            localMedia: LocalMedia,
+            sendAsFile: Boolean,
+        ): DefaultMediaOptimizationSelectorPresenter
+    }
+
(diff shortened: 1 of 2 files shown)


libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/VideoCompressor.kt
```diff
@@ -1,4 +1,214 @@
...
+import io.element.android.libraries.mediaupload.api.compressorHelper
+import io.element.android.libraries.mediaviewer.api.local.LocalMedia
+import io.element.android.libraries.preferences.api.store.VideoCompressionPreset
+import kotlinx.collections.immutable.ImmutableList
+import kotlinx.collections.immutable.toImmutableList
+import timber.log.Timber
+import kotlin.math.roundToLong
+
+@AssistedInject
+class DefaultMediaOptimizationSelectorPresenter(
```
bmarty: Claude: you're confused: please update the file `DefaultMediaOptimizationSelectorPresenter.kt` instead

bmarty (changes requested):

libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/VideoCompressor.kt
escix: @bmarty 
Sorry for the error on my part.
Rectified

libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/VideoCompressor.kt
```diff
@@ -102,8 +102,10 @@ class VideoCompressor(
         val encoderFactory = DefaultEncoderFactory.Builder(context)
             .setRequestedVideoEncoderSettings(
                 VideoEncoderSettings.Builder()
-                    // Use VBR which is generally better for quality and compatibility, although slightly worse for file size
-                    .setBitrateMode(MediaCodecInfo.EncoderCapabilities.BITRATE_MODE_VBR)
+                    // Use CBR so the requested bitrate is treated as a hard target rather than an average.
+                    // VBR lets many hardware encoders spend significantly more bits than requested on complex
+                    // footage, which made the resulting file size unpredictable (and much larger than estimated).
+                    .setBitrateMode(MediaCodecInfo.EncoderCapabilities.BITRATE_MODE_CBR)
```
jmartinesp, line 108: I was looking at this PR and I think the rest of it is probably fine, but this is problematic: I remember in the past I tried using CBR (constant bitrate mode) but it seems like some devices couldn't handle it and [media transcoding completely failed in those](https://github.com/element-hq/element-x-android/issues/4857). VBR on the other hand seemed to be compatible with every device.

PR #7404, opened 2026-08-11
Add hooks for X.509 verification of user identities

richvdh:
Add an interface `io.element.android.libraries.matrix.api.x509.X509Provider`, which can be implmenented to perform automatic verification of users using X.509 certificates, based on the functionality added to matrix-rust-sdk in https://github.com/matrix-org/matrix-rust-sdk/pull/6727.

The default implementation does nothing, but the interface will be implemented in Enterprise builds.

## Checklist

    - [X] No.

### DIFF
--- a/app/src/main/kotlin/io/element/android/x/MainActivity.kt
+++ b/app/src/main/kotlin/io/element/android/x/MainActivity.kt
@@ -61,6 +61,9 @@ class MainActivity : NodeActivity() {
         setContent {
             MainContent(appBindings)
         }
+
+        val activity = this
+        lifecycleScope.launch { appBindings.x509Provider().onAppStartup(activity) }
     }
 
     @Composable
--- a/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/x509/X509Provider.kt
+++ b/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/x509/X509Provider.kt
@@ -10,14 +10,20 @@ package io.element.android.libraries.matrix.api.x509
 import android.app.Activity
 
 /**
- * Interface for things which can provide a [RawX509Signer] and/or a [RawX509Verifier]. Called during client creation.
+ * Interface for things which can provide a [RawX509Signer] and/or a [RawX509Verifier].
  */
 interface X509Provider {
-    /** Hook which is called during app startup. Allows the implementation to do any necessary initialization. */
-    suspend fun onAppStartup(parentActivity: Activity)
+    /**
+     * Hook which is called during app startup. Allows the implementation to do any necessary initialization.
+     *
+     * Note that the hook is run in the background: it will not block startup of the app.
+     *
+     * @param activity The activity which is active while the app starts (i.e. the `MainActivity`).
+     * */
+    suspend fun onAppStartup(activity: Activity)
 
     /**
-     * Provide a [RawX509Signer].
+     * Provide a [RawX509Signer]. Called during client creation.
      *
      * If a non-null result is returned, the implementation will be called to add a signature to any newly-created
      * digital identity (i.e., master cross-signing key).
@@ -25,7 +31,7 @@ interface X509Provider {
     suspend fun getRawX509Signer(): RawX509Signer?
 
     /**
-     * Provide a [RawX509Verifier].
+     * Provide a [RawX509Verifier]. Called during client creation.
      *
      * If a non-null result is returned, the implementation will be called to verify X.509 signatures on users'
      * digital identities (i.e. master cross-signing keys).
--- a/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/x509/X509Provider.kt
+++ b/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/x509/X509Provider.kt
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2026 Element Creations Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.matrix.api.x509
+
+import android.app.Activity
+
+/**
+ * Interface for things which can provide a [RawX509Signer] and/or a [RawX509Verifier]. Called during client creation.
+ */
+interface X509Provider {
+    /** Hook which is called during app startup. Allows the implementation to do any necessary initialization. */
+    suspend fun onAppStartup(parentActivity: Activity)
+
+    /**
+     * Provide a [RawX509Signer].
+     *
+     * If a non-null result is returned, the implementation will be called to add a signature to any newly-created
+     * digital identity (i.e., master cross-signing key).
+     */
+    suspend fun getRawX509Signer(): RawX509Signer?
+
+    /**
+     * Provide a [RawX509Verifier].
+     *
+     * If a non-null result is returned, the implementation will be called to verify X.509 signatures on users'
+     * digital identities (i.e. master cross-signing keys).
+     */
+    suspend fun getRawX509Verifier(): RawX509Verifier?
+}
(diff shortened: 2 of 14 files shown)


richvdh:
I'm not sure what Sonarqube is complaining about. It seems to be about existing code unrelated to this PR.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/x509/X509Provider.kt
```diff
@@ -0,0 +1,34 @@
...
+package io.element.android.libraries.matrix.api.x509
+
+import android.app.Activity
+
+/**
+ * Interface for things which can provide a [RawX509Signer] and/or a [RawX509Verifier]. Called during client creation.
+ */
+interface X509Provider {
+    /** Hook which is called during app startup. Allows the implementation to do any necessary initialization. */
+    suspend fun onAppStartup(parentActivity: Activity)
```
bmarty: Neat: `parentActivity` can be confusing (an Activity A can start activity B and from B POV, A is the parent Activity) , naming the parameter `activity` is a bit better.


```suggestion
    suspend fun onAppStartup(activity: Activity)
```

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/x509/X509Provider.kt
```diff
@@ -0,0 +1,34 @@
...
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.libraries.matrix.api.x509
+
+import android.app.Activity
+
+/**
+ * Interface for things which can provide a [RawX509Signer] and/or a [RawX509Verifier]. Called during client creation.
```
bmarty: Maybe move `Called during client creation.` to the doc of `getRawX509Signer` and/or `getRawX509Verifier`, since it's not applicable to `onAppStartup`

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/x509/DefaultX509Provider.kt
```diff
@@ -0,0 +1,35 @@
...
+import dev.zacsweers.metro.ContributesBinding
+import dev.zacsweers.metro.SingleIn
+import io.element.android.libraries.matrix.api.x509.RawX509Signer
+import io.element.android.libraries.matrix.api.x509.RawX509Verifier
+import io.element.android.libraries.matrix.api.x509.X509Provider
+
+/**
+ * Default implementation of [X509Provider], which does nothing.
+ */
+@SingleIn(AppScope::class)
```
bmarty: No need for the annotation `@SingleIn`. It is OK to create a new object `DefaultX509Provider` when necessary.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -236,7 +240,18 @@ class RustMatrixClientFactory(
...
+
+        val rawX509signer = x509Provider.getRawX509Signer()
+        if (rawX509signer != null) {
+            builder = builder.withRawX509Signer(RawX509SignerWrapper(rawX509signer))
+        }
+
+        val rawX509verifier = x509Provider.getRawX509Verifier()
+        if (rawX509verifier != null) {
+            builder = builder.withRawX509Verifier(RawX509VerifierWrapper(rawX509verifier))
+        }
```
bmarty, line 252: We could have a hook in EnterpriseService to tweak the client builder. It would allow to move `RawX509SignerWrapper` and `RawX509VerifierWrapper` into the private module.

app/src/main/kotlin/io/element/android/x/MainActivity.kt
```diff
@@ -61,6 +61,9 @@ class MainActivity : NodeActivity() {
         setContent {
             MainContent(appBindings)
         }
+
+        val activity = this
+        lifecycleScope.launch { appBindings.x509Provider().onAppStartup(activity) }
```
bmarty, line 66: The dialog that can be displayed by `onAppStartup` will not block the `MainContent` to run the UI, i.e. a Matrix client can be setup and started when the dialog is still displayed. Is it OK? Else let me know if you need help to ensure that the app does not go further until the dialog is not closed.

bmarty (changes requested):
Thanks, nice work! A few remarks, please let me know if something is not clear.

app/src/main/kotlin/io/element/android/x/MainActivity.kt
richvdh: > The dialog that can be displayed by `onAppStartup` will not block the `MainContent` to run the UI, i.e. a Matrix client can be setup and started when the dialog is still displayed. Is it OK? Else let me know if you need help to ensure that the app does not go further until the dialog is not closed.

Yeah, it's fine for now. It's certainly not elegant, and we may want to improve this later, so thank you for the offer.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/x509/X509Provider.kt
richvdh: done, with a bit of extra documentation.

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/x509/X509Provider.kt
richvdh: done

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/x509/DefaultX509Provider.kt
richvdh: done

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
richvdh: I'm not really sure how this would work, without adding a dependency on the rust-sdk types to `io.element.android.features.enterprise.api`. Adding an abstraction layer doesn't help.

The job of `RawX509SignerWrapper` is to map from our abstracted `RawX509Signer` into the interface required by the rust-sdk, `org.matrix.rustcomponents.sdk.RawX509Signer`.  If that mapping takes place in the private module, then we need to send a `org.matrix.rustcomponents.sdk.RawX509Signer` over the enterprise api - which I don't think we can do.

bmarty: I have shared a patch in our Matrix room so that it should unblock this - please let me know if it's not the case

libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/x509/X509Provider.kt
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package io.element.android.libraries.matrix.api.x509
+
+import android.app.Activity
+
+/**
+ * Interface for things which can provide a [RawX509Signer] and/or a [RawX509Verifier].
+ */
+interface X509Provider {
```
bmarty, line 15: `getRawX509Signer()` and `getRawX509Verifier()` should be moved to the private modules and `X509Provider` should be rename to `AppStartupHook` or something like that.

bmarty (changes requested):
(submitting a review)
Please see my comment at https://github.com/element-hq/element-x-android/pull/7404/changes#r3785176174.

At the end, this PR should not have any occurrences of "X509", everything should be moved to the private modules :)

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
richvdh: I think I'm having the same problem as Jorge in https://github.com/element-hq/element-x-android/pull/7495#discussion_r3831353034, which is that, for this to work, we need for the enterprise code to have access to the SDK classes. I'll wait for that PR to land and then take another look.

richvdh:
I'll close this; we're going to replace it with more implementation in the enterprise module and less in the foss module.

PR #7407, opened 2026-08-12
build-rust-sdk: print the command we run

richvdh:
It's helpful to know what exactly is being run.

Also some tweaks to reduce the amount of directory changing that needs to happen.


jmartinesp (approved):
Thanks.

PR #7410, opened 2026-08-12
Feature: global search

jmartinesp:
## Content

First implementation of the global search feature for rooms and messages.

Note: the design is not final, this is just an initial approximation.

## Screenshots / GIFs

In the PR.

## Tests

Enable the search feature in developer options and open the search screen from the home screen.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] Yes (tests). In this case, please request a review by Copilot.
    - [ ] No.

### DIFF
--- a/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/roomlist/RoomListPresenter.kt
+++ b/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/roomlist/RoomListPresenter.kt
@@ -30,7 +30,7 @@ import io.element.android.features.announcement.api.AnnouncementService
 import io.element.android.features.home.impl.datasource.RoomListDataSource
 import io.element.android.features.home.impl.filters.RoomListFiltersState
 import io.element.android.features.home.impl.filters.into
-import io.element.android.features.home.impl.search.GlobalSearchPresenter
+import io.element.android.features.home.impl.search.GlobalSearchState
 import io.element.android.features.home.impl.search.RoomListSearchEvent
 import io.element.android.features.home.impl.search.RoomListSearchState
 import io.element.android.features.home.impl.spacefilters.SpaceFiltersState
@@ -86,7 +86,7 @@ class RoomListPresenter(
     private val announcementService: AnnouncementService,
     private val coldStartWatcher: AnalyticsColdStartWatcher,
     private val spaceFiltersPresenter: Presenter<SpaceFiltersState>,
-    private val globalSearchPresenter: GlobalSearchPresenter,
+    private val globalSearchPresenter: Presenter<GlobalSearchState>,
     private val featureFlagService: FeatureFlagService,
 ) : Presenter<RoomListState> {
     private val encryptionService = client.encryptionService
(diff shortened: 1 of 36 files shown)


features/home/impl/src/main/kotlin/io/element/android/features/home/impl/roomlist/RoomListPresenter.kt
```diff
@@ -85,10 +86,15 @@ class RoomListPresenter(
     private val announcementService: AnnouncementService,
     private val coldStartWatcher: AnalyticsColdStartWatcher,
     private val spaceFiltersPresenter: Presenter<SpaceFiltersState>,
+    private val globalSearchPresenterFactory: GlobalSearchPresenter.Factory,
```
bmarty: I do not think that `GlobalSearchPresenter` needs a factory.
We could have this code instead:


```suggestion
    private val globalSearchPresenter: Presenter<GlobalSearchState>,
```

tests/uitests/src/test/snapshots/images/features.home.impl.search_GlobalSearchView_Day_3_en.png
bmarty, line 1: This problem already exists on `RoomListSearchView` previews, but it is annoying - but not critical - that the results do not contain any room with the word "Query". IRL this should never happen, and I'd like to have preview states that would actually exist at runtime

tests/uitests/src/test/snapshots/images/features.home.impl_HomeView_Day_12_en.png
bmarty, line 1: `state.globalSearchState.isEnabled` should be `false` in the state provider, so that this preview is not updated.

bmarty (commented):
Nice work, thanks!

One small issue I have spotted:
- open the global search screen
- enter some text
- the rooms are shown in the result
- switch to "Messages"
- the search is not triggered. It should be triggered with the entered terms.

When switching back to "Chats", the search is triggered (with the eventually edited query) and the new room list result is shown. 

Another thing, the segmented button state is kept between 2 search session, I wonder if it should always reset to "Chats". A question for product I guess.

Last point, I was not able to get result when searching for message, even after sending new message in e2e room.

jmartinesp:
About:

> Last point, I was not able to get result when searching for message, even after sending new message in e2e room.

Can you try again with the latest changes? It's odd, I thought I was having the same issue but it turns out I wasn't searching for already indexed values. Also, take into account this search uses full words, so if you search for "hel" it won't find a "helllo" word.

bmarty:
Thanks for the update.

- Still not able to get result of messages. Even by searching full word and including word that are sent to clear room.
- Also I notice that there is no "empty result" view for the "Chats"

jmartinesp:
> Still not able to get result of messages. Even by searching full word and including word that are sent to clear room.

To be honest, I have no idea about what could be happening. Do you see any logs related to search?

bmarty:
Clearing the cache has fixed the message search behavior.

Thanks for the latest fixes, it looks good now. We are missing a state for message search when query is empty, but this can be handled separately:

image

bmarty (approved):

ShadowRZ:
👀 This does not appears to be merged despite auto merge was enabled, likely due to ElementBot not having an exception on CLA so required CLA check is still pending

PR #7426, opened 2026-08-14
Add KDoc to all the interfaces of the api modules

bmarty:
## What

Adds KDoc to every interface in every `api` module, and to each of their functions and properties.

Coverage is complete for the agreed scope: **222 interfaces and 820 members**, across the 60 `api` modules that contain at least one in-scope interface.

| | interfaces | members |
| --- | --- | --- |
| `libraries/matrix/api` | 69 | 398 |
| the other 59 modules | 153 | 422 |
| **total** | **222** | **820** |

## Scope

In scope: non-`sealed` `interface` and `fun interface` declarations under `src/main/kotlin` of an `api` module, plus every abstract `fun`, `val` and `var` they declare (including nested interfaces such as DI factories and node proxies).

Deliberately excluded:

- **`sealed interface` ADTs** (114 of them, e.g. `JoinRule`, `RoomMembersState`, `BackupUploadState`) — union types whose subtypes carry the meaning.
- **Navigation plumbing** (~40) — `*EntryPoint`, nested `Callback : Plugin`, and `Presenter` implementors, where the docs would be near-identical boilerplate.

22 `api` modules therefore contain nothing in scope and are untouched.

## Approach

Each doc was written **after reading the implementation** — `Default*` in the sibling `impl` module, or `Rust*` in `libraries/matrix/impl` — so that the docs record behaviour a caller cannot infer from the signature. Some examples:

- `MatrixClient.getRoom` / `getJoinedRoom` build a **new closeable instance on every call**, owned by the caller.
- `createRoom` / `joinRoom` / `knockRoom` wait for the membership to arrive through sync and return a **`null` `RoomInfo` on timeout rather than a failure**.
- Collecting `Timeline.membershipChangeEventReceived` or `onSyncedEventReceived` is **what keeps the timeline subscribed** to the SDK.
- `Timeline.paginate` fails with `CannotPaginate` when the matching status forbids it; each status's initial `hasMoreToLoad` depends on the timeline mode.
- `EncryptionService`'s backup and recovery flows report a waiting-for-sync value until the session syncs; `isLastDevice` and `hasDevicesToVerifyAgainst` are **polled every 5s**, not observed.
- `WellknownRetriever` returns a stale entry immediately and refreshes in the background; an enterprise build with a hardcoded config short-circuits entirely.
- `SecretKeyRepository.getOrCreateKey`'s `requiresUserAuthentication` **only applies at creation** — an existing key keeps the protection it was created with.
- `AppNavigationStateService`'s `owner` parameter: the state is a stack, and a leave call whose owner does not match the current level is **silently ignored**.
- Preference getters state their **actual default values**, read from the impls.

Where an implementation is too intricate to compress into three sentences, the doc states the contract and the caller-visible edge cases and leaves the mechanism to the code.

## Documentation style

- 1–3 sentences per interface and per member (often 1).
- `@param` for every parameter, one sentence each; `@return` only where the return is not already obvious.
- Interface-level docs carry contracts shared by all members (e.g. "both methods are idempotent and never throw: errors are wrapped in the returned `Result`") instead of repeating them per member.
- Accurate existing docs were left untouched — a fair number of members already had good docs and are not in this diff.

## Inaccurate docs corrected

1. **`BaseRoom.getUpdatedIsEncrypted`** was documented as returning the room's directory visibility. It returns the encryption state.
2. **`NotificationSettingsService.notificationSettingsChangeFlow`** was documented as a state flow of room notification settings. It is a `SharedFlow<Unit>` that only signals that the push rules changed.
3. **`MatrixClient.resolveRoomAlias`** had a malformed KDoc block and a typo.

## One non-comment change

The whole diff is comments-only **except a single line**: fix (2) above removed the only reference to `RoomNotificationSettingsState` in that file, leaving an unused import that ktlint would reject, so the import is removed too. It is called out in that commit's message.

## Reviewing

Commits are one per module (`libraries/matrix/api` is split by subpackage, since a single commit would have been ~1000 additions; the single-interface modules are grouped so as not to produce a dozen near-empty commits). Each commit message names the behaviours taken from the implementation.

⚠️ **This exceeds the 500-addition guideline by a lot (~2750).** It is opened as a draft for that reason — the per-module commit boundaries should make splitting it into reviewable PRs mechanical.

Verification used while writing: a per-module scan for undocumented interfaces/members (now reporting 0/0), a check for imports whose only remaining use was a KDoc reference (which is what caught the case above), and a 160-column check.

### DIFF
--- a/features/announcement/api/src/main/kotlin/io/element/android/features/announcement/api/AnnouncementService.kt
+++ b/features/announcement/api/src/main/kotlin/io/element/android/features/announcement/api/AnnouncementService.kt
@@ -12,15 +12,34 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import kotlinx.coroutines.flow.Flow
 
+/**
+ * Queues the full-screen announcements shown to the user once, such as the notice about a new feature.
+ *
+ * A dismissed announcement is remembered so it is not shown again.
+ */
 interface AnnouncementService {
+    /**
+     * Queues an announcement, unless the user has already dismissed it.
+     *
+     * @param announcement the announcement to show.
+     */
     suspend fun showAnnouncement(announcement: Announcement)
 
+    /**
+     * Records that the user dismissed an announcement, so it is not queued again.
+     *
+     * @param announcement the announcement that was dismissed.
+     */
     suspend fun onAnnouncementDismissed(announcement: Announcement)
 
+    /** The announcements still waiting to be shown, in the order they should appear. */
     fun announcementsToShowFlow(): Flow<List<Announcement>>
 
     /**
      * Use this composable to render the announcement UI in Fullscreen.
+     * Draws nothing while there is no announcement to show.
+     *
+     * @param modifier layout modifier for the container.
      */
     @Composable
     fun Render(
--- a/features/cachecleaner/api/src/main/kotlin/io/element/android/features/cachecleaner/api/CacheCleaner.kt
+++ b/features/cachecleaner/api/src/main/kotlin/io/element/android/features/cachecleaner/api/CacheCleaner.kt
@@ -8,6 +8,9 @@
 
 package io.element.android.features.cachecleaner.api
 
+/**
+ * Removes the temporary files holding decrypted content, so plaintext media does not linger on disk.
+ */
 interface CacheCleaner {
     /**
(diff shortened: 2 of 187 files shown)


jmartinesp (approved):
Image

Jokes aside, this seems ok in general. I'm a bit afraid we'll run into the 'what's worse than no docs? wrong docs' situation, but the docs I've checked here and there seem accurate.

PR #7433, opened 2026-08-15
Fix outgoing session verification stuck with incomplete recovery setup

hardcore-sushi:
## Content

- In the outgoing verification state machine, when the `recoveryStateStateFlow` times out waiting for the `ENABLED` state, continue to the `Completed` state instead of aborting the transition.
- Stop waiting as soon as the recovery state is `ENABLED`, `DISABLED` or `INCOMPLETE` rather than only when it becomes `ENABLED`, so sessions for which recovery is not enabled do not wait 10 seconds for no reason.

## Motivation and context

On continuwuity-based home servers, I noticed that Element X got stuck indefinitely on the emoji verification screen, while the other user could see that the verification had completed successfully. This PR fixes this issue.

Server-side issues:
https://github.com/matrix-construct/tuwunel/issues/292
https://forgejo.ellis.link/continuwuation/continuwuity/issues/1476

## Screenshots / GIFs

Can provide if required.

## Tests

Investigation, fix and units tests are AI-generated. However, I manually reproduced the bug on a tuwunel server between one Element-Web user and one Element X  user on a fresh build without the fix (FOSS variant). All 3 unit tests were failing. With the fix, they pass, and the verification flow properly completes. Manually tested with and without a recovery key added to the Element X user account.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): LineageOS 23.2

## Checklist

    - [ ] No.

### DIFF
--- a/features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/outgoing/OutgoingVerificationStateMachine.kt
+++ b/features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/outgoing/OutgoingVerificationStateMachine.kt
@@ -15,7 +15,6 @@ import com.freeletics.flowredux.dsl.FlowReduxStateMachine
 import io.element.android.features.verifysession.impl.util.andLogStateChange
 import io.element.android.features.verifysession.impl.util.logReceivedEvents
 import io.element.android.libraries.core.bool.orFalse
-import io.element.android.libraries.core.data.tryOrNull
 import io.element.android.libraries.matrix.api.encryption.EncryptionService
 import io.element.android.libraries.matrix.api.encryption.RecoveryState
 import io.element.android.libraries.matrix.api.verification.SessionVerificationData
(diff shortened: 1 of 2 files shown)


features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/outgoing/OutgoingVerificationStateMachine.kt
```diff
@@ -95,10 +95,13 @@ class OutgoingVerificationStateMachine(
...
                     val hasBackup = encryptionService.doesBackupExistOnServer().getOrNull().orFalse()
                     if (hasBackup) {
-                        tryOrNull {
-                            encryptionService.recoveryStateStateFlow.filter { it == RecoveryState.ENABLED }
+                        try {
+                            encryptionService.recoveryStateStateFlow
+                                .filter { it == RecoveryState.ENABLED || it == RecoveryState.DISABLED || it == RecoveryState.INCOMPLETE }
                                 .timeout(10.seconds)
                                 .first()
+                        } catch (_: TimeoutCancellationException) {
```
jmartinesp, line 103: Good catch, we didn't realize `tryOrNull` would also treat the `TimeoutCancellationException` as a `CancellationException` and cancel the flow.

jmartinesp (approved):
Thanks.

PR #7436, opened 2026-08-16
Do not open the room again when the app is resumed from the recents list

### LINKED ISSUE #7205: Under certain conditions, the screen (activity) the user is on will be forced to jump.
### Steps to reproduce

1. Make sure the app can receive notifications;
2. Ask the other party to send you at least 1 message;
3. Wait for the notification to appear. Once it pops up, click it to make the app automatically enter the chat;
4. Keep pressing the back key (the navigation key provided by the system instead of the one in the upper left corner of the screen) until you return to the desktop;
5. Press the recent task key to enter the app again.

#### What did you expect?
After re-entering the app via step 5, the user should be left with the room list.

#### What happened instead?
After re-entering the app through step 5, the user is forced to the room they entered previously.

### Your phone model

OnePlus MT2110

### Operating system version

Oxygen OS 15 (Android 15)

Element X Android 26.7.0

### Homeserver

Synapse 1.156.0

### Will you send logs?

No

No

hayaksi1:
## Content

`MainActivity` is a `singleTask` activity that declares the `elementx://open` deep link, so when a
notification tap creates the task, that deep link becomes the task's base intent. Bringing the app
back from the recents list replays that base intent with `FLAG_ACTIVITY_LAUNCHED_FROM_HISTORY`,
`IntentResolver` turns it into a navigation again, and the user is pushed back into the room.

Ignore any intent the system replayed from the history, next to the launcher intent that is already
ignored. `RootFlowNode.handleIntent` already treats a `null` resolution as a no-op, so the app
simply stays where Appyx restored it.

## Motivation and context

Fixes #7205.

## Tests

- `./gradlew :appnav:testDebugUnitTest`
- New `IntentResolverTest.resolve intent replayed from the history should return null` builds an
  `ACTION_VIEW` intent with the history flag and a deeplink parser that would otherwise resolve a
  room. It fails without the change.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/intent/IntentResolver.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/intent/IntentResolver.kt
@@ -76,6 +76,7 @@ class IntentResolver(
 }
 
 private fun Intent.canBeIgnored(): Boolean {
+    if (flags and Intent.FLAG_ACTIVITY_LAUNCHED_FROM_HISTORY != 0) return true
     return action == Intent.ACTION_MAIN &&
         categories?.contains(Intent.CATEGORY_LAUNCHER) == true
 }
--- a/appnav/src/test/kotlin/io/element/android/appnav/intent/IntentResolverTest.kt
+++ b/appnav/src/test/kotlin/io/element/android/appnav/intent/IntentResolverTest.kt
@@ -46,6 +46,24 @@ class IntentResolverTest : RobolectricTest() {
     }
 
     @Test
+    fun `resolve intent replayed from the history should return null`() {
+        val sut = createIntentResolver(
+            deeplinkParserResult = DeeplinkData.Room(
+                sessionId = A_SESSION_ID,
+                roomId = A_ROOM_ID,
+                threadId = null,
+                eventId = null,
+            )
+        )
+        val intent = Intent(RuntimeEnvironment.getApplication(), Activity::class.java).apply {
+            action = Intent.ACTION_VIEW
+            addFlags(Intent.FLAG_ACTIVITY_LAUNCHED_FROM_HISTORY)
+        }
+        val result = sut.resolve(intent)
+        assertThat(result).isNull()
+    }
+
+    @Test
     fun `test resolve navigation intent root`() {
         val sut = createIntentResolver(
             deeplinkParserResult = DeeplinkData.Root(A_SESSION_ID)


jmartinesp:
I can't really reproduce the issue, can you provide some steps?

I tried:

1. Simulate the app being killed by the OS.
2. Receiving a notification an opening it, which starts the app:
```
elementx: [MainActivity] onCreate, with savedInstanceState: true, intent action: android.intent.action.MAIN, intent data: null
elementx: [MainActivity] onNewIntent, action: android.intent.action.VIEW, data: elementx://open/...
```
3. Then killing the app in several ways: by going back until it's closed, by manually force closing from settings, by simulating the OS killing it again...
4. Opened the app from recents, I just got the plain empty `onCreate` intent for all of them:
```
elementx: [MainActivity] onCreate, with savedInstanceState: true, intent action: android.intent.action.MAIN, intent data: null
```

Maybe this only happens on some Android versions, or I need to enable/disable something?

hayaksi1:
The difference is whether the deep link is the intent that *creates* the task's root activity. If the activity is still alive when the notification arrives, the link goes to `onNewIntent` and the task keeps its launcher base intent — which is exactly what your log shows (`onCreate` with `MAIN` + `savedInstanceState: true`, then `onNewIntent` with the link). Nothing is left for recents to replay in that case.

Reproduced on an API 36 emulator, using the deep link the notification uses:

```
adb shell am force-stop io.element.android.x        # no live activity in the task
adb shell am start -a android.intent.action.VIEW -d "elementx://open/<sessionId>/<roomId>"
adb shell input keyevent KEYCODE_HOME
adb shell am kill io.element.android.x              # force-stop would clear the task, this keeps it
# then open it from recents
```

After step 2 the task's base intent is the link, not the launcher intent:

```
baseIntent=Intent { act=android.intent.action.VIEW dat=elementx://open/... flg=0x10100000 cmp=.../io.element.android.x.MainActivity }
```

and opening it from recents starts it again with the link still attached:

```
ActivityTaskManager: START u0 {act=android.intent.action.VIEW dat=elementx://open/... flg=0x10100000 ...} with LAUNCH_SINGLE_TASK
```

`0x10100000` is `FLAG_ACTIVITY_NEW_TASK | FLAG_ACTIVITY_LAUNCHED_FROM_HISTORY`, so `canBeIgnored()` sees the room deep link again on every resume from recents. `MainActivity` being `singleTask` and owning both the LAUNCHER and the `elementx://open` filter is what lets whichever intent starts it become the task's base intent.

The reporter reaches the same state by pressing back out of the app and letting the process go, rather than by force-stopping it.

jmartinesp:
I can't reproduce it in a real device with Android 15, I got the same results as in my previous tests. I'll check later with the emulator in API 36, maybe something changed there.

jmartinesp (approved):
> I can't reproduce it in a real device with Android 15, I got the same results as in my previous tests. I'll check later with the emulator in API 36, maybe something changed there.

I've been able to reproduce this just once and I couldn't really test the fix, but it seems sensible.

PR #7437, opened 2026-08-16
Show an error when the room to leave cannot be found

### LINKED ISSUE #4594: I can't get out of a room that has been deleted
### Steps to reproduce

There seems to be a problem with the molly room (signal fork). Messages don't load and some errors say that this room doesn't exist anymore. So when I'm trying to remove it, it fails. It shouldn't fail. Right? 

### Your phone model

Galaxy s24+

### Operating system version

Android 15

25.04.1

### Homeserver

Matrix.org

### Will you send logs?

Yes

Yes

hayaksi1:
## Content

`LeaveRoomPresenter.showLeaveRoomAlert` ran its whole body inside
`client.getRoom(roomId)?.use { }`. When `getRoom` returns `null`, the body was skipped without
writing any state: `leaveAction` stayed `Uninitialized`, so `AsyncActionView` rendered nothing.
Long pressing a room and choosing **Leave room** appeared to do nothing at all, with no
confirmation dialog and no error.

The missing room is now handled explicitly and surfaced as a failure, which `AsyncActionView`
already renders with the existing generic error dialog, so no new string is needed. `leaveRoom`
used a `!!` for the same lookup and reported the miss as a `NullPointerException`; it now throws
the same named exception.

## Motivation and context

Fixes #4594 in part: the user gets an error instead of a dead menu entry. Actually leaving a room
the client cannot resolve is not possible from here, since leaving goes through `Room.leave()` and
`MatrixClient` has no room-id-only leave.

## Tests

- `./gradlew :features:leaveroom:impl:testDebugUnitTest`
- Two new tests in `LeaveBaseRoomPresenterTest` cover the confirmation and the direct leave path
  with a client that does not know the room. The confirmation one fails without the change.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

### DIFF
--- a/features/leaveroom/impl/src/main/kotlin/io/element/android/features/leaveroom/impl/LeaveRoomPresenter.kt
+++ b/features/leaveroom/impl/src/main/kotlin/io/element/android/features/leaveroom/impl/LeaveRoomPresenter.kt
@@ -60,7 +60,13 @@ class LeaveRoomPresenter(
         roomId: RoomId,
         leaveAction: MutableState<AsyncAction<Unit>>,
     ) = launch(dispatchers.io) {
-        client.getRoom(roomId)?.use { room ->
+        val room = client.getRoom(roomId)
+        if (room == null) {
+            Timber.e("Unable to find room $roomId to leave it")
+            leaveAction.value = AsyncAction.Failure(RoomNotFoundException(roomId))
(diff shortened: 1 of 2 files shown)


bmarty (approved):
LGTM, thanks!

Issue #4238, opened 2025-02-06
Count of people besides `People` room settings entry is incorrect after joining room


dkasak:
### Steps to reproduce

Immediately upon joining a room, the count of people besides People in room settings can be wrong. Actually entering the People menu shows the correct member list, but the count outside is wrong until the app is restarted.

![Image](https://github.com/user-attachments/assets/c2feca48-6699-4234-ad9a-7755d0eb59c1)

In the above situation there were actually 7 people in the room. Unfortunately I didn't reproduce this myself and the affected person cannot rageshake so there will be no logs.

#### What did you expect?

The correct count to be displayed.

#### What happened instead?

The incorrect count was displayed.

### Will you send logs?

No

### Are you willing to provide a PR?

No

frebib:
I wonder if this is related to/caused by the same thing as #4236

frebib:
Maybe tangentially, I do think it would be prudent for the app to render some kind of indication that the user count in the room is unknown (spinner, or `?`, or something) until the join is complete and the full state of the room is resolved on the homeserver

jmartinesp:
> Maybe tangentially, I do think it would be prudent for the app to render some kind of indication that the user count in the room is unknown (spinner, or `?`, or something) until the join is complete and the full state of the room is resolved on the homeserver

I don't think so, the value we use there comes from a completely different source, AFAICT. And if I'm not mistaken, this value is only updated [here](https://github.com/matrix-org/matrix-rust-sdk/blob/3428494468dc0b34397d74c976e05fcb5ede9284/crates/matrix-sdk-base/src/sliding_sync/mod.rs#L852) when we receive the room from sliding sync, then it's serialized and deserialized when needed.

PR #7438, opened 2026-08-16
Refresh the room members when a mention is started

### LINKED ISSUE #4236: Mention list doesn't update when a user updates a room
To reproduce:
1. Open a room
2. Have a user join whilst the room is open
3. Try to @ the user to welcome them
4. Notice they're not in the mention list
5. Optionally: close and reopen the room and notice they're present

It would be good to reload the mention list in an open timeline view when the room membership changes

hayaksi1:
## Content

The mention suggestions in the composer read `room.membersStateFlow`, which is a cache filled only
by an explicit `updateMembers()` call — the contract on `BaseRoom` says so. Nothing calls it while
a room is open, so a member who joins after the room was opened never shows up in the suggestion
list until the room is left and entered again.

`updateMembers()` is now called when a mention search starts. The trigger is
`distinctUntilChanged` on "the current suggestion is a mention", so it fires once per mention
rather than on every keystroke, and `RoomMemberListFetcher.fetchRoomMembers` already no-ops while a
fetch is in flight.

## Motivation and context

Fixes #4236.

The joined member count from `roomInfoFlow` was deliberately not used as the trigger: it comes from
sliding sync and is not reliably refreshed, which is the point jmartinesp makes on the related
#4238, so keying off it would miss exactly the joins this issue is about.

## Tests

- `./gradlew :features:messages:impl:testDebugUnitTest --tests "*MessageComposerPresenterTest*"`
- New `present - mention suggestions refresh the room members` asserts `updateMembers()` is called
  once when an `@` mention starts. It fails without the change. The three existing mention
  suggestion tests are unchanged apart from being given an `updateMembers` result on their fake.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
@@ -86,6 +86,7 @@ import io.element.android.libraries.textcomposer.model.MarkdownTextEditorState
 import io.element.android.libraries.textcomposer.model.Message
 import io.element.android.libraries.textcomposer.model.MessageComposerMode
 import io.element.android.libraries.textcomposer.model.Suggestion
+import io.element.android.libraries.textcomposer.model.SuggestionType
 import io.element.android.libraries.textcomposer.model.TextEditorState
 import io.element.android.libraries.textcomposer.model.rememberMarkdownTextEditorState
 import io.element.android.services.analytics.api.AnalyticsService
@@ -103,7 +104,9 @@ import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.debounce
+import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.filter
+import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.merge
 import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.launch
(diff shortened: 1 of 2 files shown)


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
```diff
@@ -459,6 +462,15 @@ class MessageComposerPresenter(
 
             val mentionTriggerFlow = merge(mentionStartTrigger, mentionCompletionTrigger)
 
+            // Refresh the room members, which are only loaded on demand, when a mention starts
+            launch {
+                suggestionSearchTrigger
+                    .map { it?.type == SuggestionType.Mention }
+                    .distinctUntilChanged()
+                    .filter { it }
```
jmartinesp: Nit, this could probably be:

```suggestion
                    .filter { it?.type == SuggestionType.Mention }
                    .distinctUntilChanged()
```

jmartinesp (approved):
Thanks, makes sense. I think we try to avoid calling `/members` too much because in large rooms it can be expensive, but here it makes sense to avoid out of date suggestions.

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/MessageComposerPresenter.kt
hayaksi1: Done!

Kept `distinctUntilChangedBy { it?.type }` before the filter rather than `distinctUntilChanged()` after it: the trigger carries the text typed after the `@`, so filtering first calls `/members` again on every keystroke instead of once per mention. The test now types after the `@` to hold that.

PR #7439, opened 2026-08-16
Exclude carriage returns from the login and password fields

hayaksi1:
## Content

The `sanitize()` helper applied to the login and password fields, and to the account deactivation
password field, promises in its own KDoc to remove "any new line characters" but only removes LF. A
value pasted from a source with CRLF line endings keeps its CR, which is invisible in the field and
is sent to the homeserver as part of the credential, so the login is rejected while typing the same
characters by hand works.

Both copies of the helper now drop every line separator. Whitespace that can legitimately be part
of a password, such as spaces and tabs, is deliberately left untouched.

## Motivation and context

Related to #5153. This closes a real CRLF paste hole, but the reporter's follow-up is not
conclusive that a carriage return was their case, so it may not be the full story of that issue.

The LF half of this helper was added in 732ec3f263 for #2263; CR was simply missed at the time.

## Tests

- `./gradlew :features:login:impl:testDebugUnitTest`
- `./gradlew :features:deactivation:impl:testDebugUnitTest`
- Two new tests in `LoginPasswordViewTest`, next to the existing LF ones, paste `a\r\nb` into the
  login and the password fields and assert the event carries `ab`. Both fail without the change.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

### DIFF
--- a/features/deactivation/impl/src/main/kotlin/io/element/android/features/logout/impl/AccountDeactivationView.kt
+++ b/features/deactivation/impl/src/main/kotlin/io/element/android/features/logout/impl/AccountDeactivationView.kt
@@ -299,10 +299,10 @@ private fun Content(
 }
 
 /**
- * Ensure that the string does not contain any new line characters, which can happen when pasting values.
+ * Ensure that the string does not contain any line separator, which can happen when pasting values.
  */
 private fun String.sanitize(): String {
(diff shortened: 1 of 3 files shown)


bmarty (approved):
OK, thanks.

Issue #7066, opened 2026-06-19
Declining Invites


TemoCrevlis:
### Your use case

If you open an invite and decline it, it should take you back to the Room list.
Currently you have to press back yourself after the invite is declined.

#### What would you like to do?
1. Let declining an invite in full-screen navigate you back to the Room list.

2. "Decline and block" should remove all other invites from the now blocked user.

### Are you willing to provide a PR?

No

PR #7440, opened 2026-08-17
Close the invite screen once the invite has been declined

hayaksi1:
## Content

Declining an invite from the full screen invite left the user on that same screen, so the only way back to the room list was to press back afterwards.

Both ways of declining are affected. The plain decline is handled by `AcceptDeclineInviteView`, which `JoinRoomFlowNode` renders with an empty `onDeclineInviteSuccess`. The decline and block screen is pushed onto the join room flow's own backstack and calls `navigateUp` on success, which pops back into the join room screen rather than out of it.

The flow node now navigates up on a successful decline, exactly as it already does after forgetting a room, and `DeclineInviteAndBlockEntryPoint` gets a callback so its screen can report a successful decline separately from a back press.

Part 2 of the issue, removing the other invites of a user that has just been blocked, is not covered here: it depends on what the homeserver does with the ignore list rather than on this navigation.

## Motivation and context

Part of #7066.

## Tests

- `features/invite/impl/.../declineandblock/DeclineAndBlockViewTest.kt`: a successful decline invokes the new callback and not the back callback.
- `features/invite/impl/.../declineandblock/DefaultDeclineAndBlockEntryPointTest.kt`: the callback is passed to the created node as a plugin.
- The wiring in `JoinRoomFlowNode` itself is not covered: the callbacks are passed from the node's `View`, which has no test harness in this module.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/HomeFlowNode.kt
+++ b/features/home/impl/src/main/kotlin/io/element/android/features/home/impl/HomeFlowNode.kt
@@ -269,6 +269,11 @@ class HomeFlowNode(
                     parentNode = this,
                     buildContext = buildContext,
                     inviteData = navTarget.inviteData,
+                    callback = object : DeclineInviteAndBlockEntryPoint.Callback {
+                        override fun onDeclineSuccess() {
+                            backstack.pop()
+                        }
+                    },
                 )
             }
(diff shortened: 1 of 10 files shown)


jmartinesp (approved):
Thanks, works as intended.

Issue #5978, opened 2026-01-07
VC 202512000 fails to replace an `"msgtype": "m.image"`'s `body:`'s by-default-`text/plain` `format:` with `/markdown`, upon retroactive caption addition; and removes it upon retroactive caption modification.


RokeJulianLockhart:
#### How To Reproduce It

1.	~~~sh
	#!/system/bin/sh
	monkey -p io.element.android.x 1
	~~~

1.	Enter a chat.

-	As I was directed to report, at [`element-web/issues/11083#issuecomment-3719762859`](https://github.com/element-hq/element-web/issues/11083#issuecomment-3719762859):

	1.	Upload an image, but enter no caption.

	1.	Modify it, to add a caption. Include CommonMark hyperlink syntax.

-	As I discovered, in [`!rLtgkIdZhxPpreEWYI:mozilla.org/$jWRGHhCdZFz3GBpcfdL2OnlWwo_ZD2XLTxiB0EE4pKE`](https://matrix.to/#/!rLtgkIdZhxPpreEWYI:mozilla.org/$jWRGHhCdZFz3GBpcfdL2OnlWwo_ZD2XLTxiB0EE4pKE?via=mozilla.org&via=matrix.org&via=matrix.vulpin.com), after filing this issue:

	1.	Upload an image, but enter a caption that utilises CommonMark `code` syntax.

	1.	Modify the content of that caption.

	I have yet to submit a screencast that demonstrates reproduction of this.

#### The Outcome

-	#### What I Expected To Occur

	Retroactively adding a caption that contains to-be-rendered CommonMark (and/or HTML5) markup should cause that markup to be rendered.

-	#### What Occurred

	It solely gains `"format": "org.matrix.custom.html"` if I upload the image *with* a CommonMark caption:

	https://github.com/user-attachments/assets/bd778a3e-b6c8-404f-9c21-5ba296d53c04

	<p></p><details open><p></p><blockquote>

	~~~JSON
	{
	  "content": {
	    "body": "Another question: I want to test [`id=2004557`](https:\/\/bugzilla.mozilla.org\/show_bug.cgi?id=2004557#c0) (and, _if_ reproducible, [`id=2001852`](https:\/\/bugzilla.mozilla.org\/show_bug.cgi?id=2001852#c0)) in MozRegression. However, I've no known good build.",
	    "filename": "1000048141.png",
	    "format": "org.matrix.custom.html",
[...]

PR #7441, opened 2026-08-17
Convert markdown to HTML when a caption is edited

hayaksi1:
## Content

Adding or editing a caption on an already sent media message from the plain composer sent the caption as text only, so markdown in it was never rendered. Sending the very same caption with the media does render it.

The plain composer has no HTML to pass, so `formattedCaption` is null, and `RustTimeline.editCaption` built `EditedContent.MediaCaption` by hand with a null formatted body. The send path does not do that: `UploadParameters` is turned into a caption by the SDK, which derives the formatted body from the markdown when there is none.

`editCaption` now builds the edit with the SDK's `createCaptionEdit`, which runs the caption through the same conversion. Removing a caption still passes a null caption and is unaffected.

## Motivation and context

Part of #5978.

## Tests

- Send an image with no caption, add a caption such as `some *emphasis*` from the plain composer, and check that the emphasis is rendered, in this client and in one that displays `formatted_body`.
- Edit that caption and check that the formatted body is updated rather than dropped.
- No unit test: the conversion happens behind the FFI boundary, the same reason `sendMessage` and `editMessage`, which use the SDK's markdown constructors through `MessageEventContent`, have none either.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty (approved):
Thanks!

Issue #6429, opened 2026-03-21
fix-pg-map-id.py is not needed anymore


linsui:
The dex format is changed and the regex in https://github.com/element-hq/element-x-android/blob/908ad8a102cb9b689aa9514871db95c10a76f8a6/tools/release/fix-pg-map-id.py#L25 doesn't match anymore. Since the build is still reproducible, this is not needed anymore and can be removed. Thanks!

jmartinesp:
@bmarty does this mean we can remove the whole `inplace-fix.py` script too? I see it mentions other commands, but none of them are included in the repo AFAICT.

bmarty:
Yes, I'd remove both files.

linsui:
Thanks!

PR #7442, opened 2026-08-17
Remove the F-Droid dex patching step from the release script

hayaksi1:
## Content

`fix-pg-map-id.py` rewrites the `pg_map_id` of a dex file so that the F-Droid build stays reproducible. The dex format changed and the regex it matches on no longer matches anything, so the step does nothing, and the build is reproducible without it. `inplace-fix.py` was only vendored to drive that one command; none of its other commands are used in this repository.

Both scripts are removed, along with the four calls in `release.sh` that were the only caller. The release script now goes straight from unzipping the unsigned F-Droid artifact to signing it.

One side effect worth stating: `--page-size 16` made `inplace-fix.py` repack each APK with `zipalign -P 16` after the patch. AGP aligns the release APKs at packaging time and `apksigner` is already invoked with `--alignment-preserved true`, so the signed output stays 16 KB aligned without the repack. Worth confirming on the next release build.

## Motivation and context

Part of #6429, which also has the maintainer's answer that both files can go.

## Tests

- `bash -n tools/release/release.sh` parses.
- `grep` over the repository finds no remaining reference to either script outside `CHANGES.md` history.
- There is no automated coverage: `tools/release` is not on any Gradle source set and is not run by any workflow, so this needs a real release run to confirm end to end.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty (approved):
It should be fine, right @licaon-kter ?

licaon-kter:
Ping us when we can test a release :)

Issue #4519, opened 2025-04-02
Verification request still shows 'client does not support in-chat key verification'


escix:
### Steps to reproduce

Trying to trigger a verification process from element-web to element-x

#### What did you expect?

The verification process to succeed without any issues

#### What happened instead?
The verification process goes through but a message is triggered saying the in-chat key verification is not supported by this client.

Also on a second note, the verification process was taking about 15 seconds to display the emojis, it did display the emoji when I clicked on the desktop session, so I am unsure if the desktop element triggered the element-x to display the emojis or it was a pure coincidence. The desktop element showed the verification popup but I did not do anything on that.

### Your phone model

Galaxy S23 Ultra

### Application version and app store

25.03.4

### Homeserver

1.127.1

### Will you send logs?

No

### Are you willing to provide a PR?

No

d-navikas:
Had this issue when trigger a verification process between two element-x and from desktop to x. Strange to see it's still alive and marked as minor.

domestomas:
I have the same isssue between two instances of element x android latest version, can not verify user without using element web; how is this minor?

PR #7443, opened 2026-08-17
Do not render in-chat key verification requests in the timeline

hayaksi1:
## Content

A `m.key.verification.request` sent from a client that supports in-chat verification carries a plain text body aimed at clients that do not, along the lines of "… but your client does not support in-chat key verification". Element X has no msgtype of its own for that event, so it falls through to `OtherMessageType` and the timeline prints that body, telling the user their client cannot do something they may have just done through the session verification flow.

The event is now dropped while the timeline items are built, so it is not rendered at all, which is what Element X iOS does for the same msgtype.

The room list preview and the notification for such an event are left as they are; they are separate formatters and were not part of what was reported.

## Motivation and context

Part of #4519.

## Tests

- `features/messages/impl/.../timeline/factories/TimelineItemsFactoryTest.kt`: a timeline holding a key verification request and a normal message renders only the normal message.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty (approved):
OK, thanks

Issue #5966, opened 2026-01-01
Unsupported <details> HTML element. Missing line break after <summary> element


pkorba:
### Steps to reproduce

Element X Android is missing support for \<details\> and \<summary\> HTML tags.

1. Send a message with HTML \<details\> element
```
{  
  "content": {
    "body": "",
    "format": "org.matrix.custom.html",
    "formatted_body": "<details><summary>Summary example</summary>Text inside details</details>",
    "msgtype": "m.notice"
  },
  "origin_server_ts": 1767282410169,
  "sender": "@[redacted]",
  "type": "m.room.message",
  "unsigned": {
    "membership": "join"
  },
  "event_id": "$[redacted]",
  "room_id": "![redacted]"
}
```

#### What did you expect?
Behavior like in Element Web - full support for both elements:
- automatically inserting line break between the content of \<summary\> tag and the text of \<details\> tag
- \<details\> tag in closed state by default, switches to open state when tapped
- visual triangle indicating whether the element is in open/closed state
<img width="249" height="110" alt="Image" src="https://github.com/user-attachments/assets/f93d9416-d967-4ef9-9b40-363cf5cbcc1a" />

#### What happened instead?
Missing support for both tags. No line break between the content of \<summary\> element and the text of \<details\> element.
<img width="902" height="326" alt="Image" src="https://github.com/user-attachments/assets/bf78a325-7fe1-45ab-a370-6e7b60104db0" />

### Your phone model

Samsung Galaxy S24 FE

### Operating system version

Android 16

### Application version and app store

Element X Android 25.12.0

### Homeserver

matrix.org

### Will you send logs?

No

### Are you willing to provide a PR?

No

PR #7444, opened 2026-08-17
Show the summary of a details element on its own line

hayaksi1:
## Content

The HTML safelist used to render formatted messages does not contain `summary`, so jsoup dropped the tag and kept its text. The summary of a `details` element and the text inside it were then run together on a single line, which is how the reported message rendered.

Before cleaning, a `summary` is now rewritten as a paragraph, the closest tag the safelist supports, so the two parts are separated the way they are on Element Web. Messages that contain no `summary` take the same path as before.

This is the line break half of the issue only. Rendering `details` as a collapsed section with a disclosure triangle needs real support for the element in the renderer and is not attempted here.

## Motivation and context

Part of #5966.

## Tests

- `libraries/matrixui/.../messages/ToHtmlDocumentTest.kt`: a `details` element ends up with its summary in a paragraph of its own and the body text after it, and a document with no summary keeps its paragraphs untouched.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/ToHtmlDocument.kt
+++ b/libraries/matrixui/src/main/kotlin/io/element/android/libraries/matrix/ui/messages/ToHtmlDocument.kt
@@ -67,8 +67,19 @@ private fun fixMentions(
 private object CustomHtmlToDomParser {
     fun document(html: String): Document {
         val outputSettings = OutputSettings().prettyPrint(false).indentAmount(0)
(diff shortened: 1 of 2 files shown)


jmartinesp:
Could this be instead part of the [rich text editor library](https://github.com/element-hq/matrix-rich-text-editor/) for now, just like https://github.com/element-hq/matrix-rich-text-editor/pull/146? The HTML parser is there and it's better to just handle it properly than adding workarounds in the app.

hayaksi1:
@jmartinesp Done.

Issue #2617, opened 2024-03-28
Fails to show user ID with slash


shadow-identity:
### Steps to reproduce

1. Open a [message](https://matrix.to/#/!ppWIVhhqgxBReHcvfZ:intelfx.name/$shCfE2Qp_sk8Ylw06TWQQ5-P9_XBRRdWuhE9x80qcYI) containing user mention. User ID should have slash "/" in it, for example `@s/ash:matrix.org`. 
2. Tap on user ID

#### What did you expect?
To see the user info, like with any other user.

#### What happened instead?
It opens `https://matrix.to/#/@it/sme:matrix.org`. Matrix.to fails to show the user (incorrectly reporting "wrong URL"). So user can not see any user info.

Please note that the slash in the user id `@it/sme:matrix.org` is not escaped.
But even escaped slash URL matrix.to returns the error.

### Operating system version

Android 14

### Application version and app store

Element X 0.4.6

### Will you send logs?

No

### Are you willing to provide a PR?

No

shadow-identity:
BTW same happens at least in Element Web and Element for macOS

bmarty:
Thanks for reporting.

`/` is accepted in the Matrix userId localpart grammar: https://spec.matrix.org/latest/appendices/#user-identifiers, so this will have to be fixed at some point.

bmarty:
For the issue with matrix.to not working properly, you may want to report it in this project https://github.com/matrix-org/matrix.to. Thanks!

shadow-identity:
Thanks, I found that they already have it: https://github.com/matrix-org/matrix.to/issues/288

PR #7445, opened 2026-08-17
Open the user profile of a user id that contains a slash

hayaksi1:
## Content

A slash is valid in the local part of a user id, but in a permalink it also separates the identifier from an event id. `parseMatrixEntityFrom` cannot tell the two apart, so a mention of a user such as `@it/sme:matrix.org` failed to parse and the link was treated as an ordinary URL: tapping it opened matrix.to in a browser instead of the user profile.

Only a user id can start with an `@`, so the identifier is now recovered before the URI is handed to the SDK. The recovery is applied only when the identifier really contains a slash and is a valid user id, so every other permalink still takes exactly the same path as before.

## Motivation and context

Part of #2617.

## Tests

- `libraries/matrix/impl/.../permalink/DefaultPermalinkParserTest.kt`: a matrix.to link, a matrix.to link with the slash escaped, and an Element Web link all resolve to a user link, while a URL that is not a permalink stays a fallback link.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/permalink/DefaultPermalinkParser.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/permalink/DefaultPermalinkParser.kt
@@ -8,11 +8,13 @@
 
 package io.element.android.libraries.matrix.impl.permalink
 
+import android.net.Uri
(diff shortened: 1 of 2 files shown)


bmarty (approved):
OK, thanks!

Issue #6518, opened 2026-04-02
Red dot shown for encryption


floviolleau:
### Steps to reproduce

1. Go in settings of the Android Element X app

#### What did you expect?
All encryption steps are configured on all devices and all devices are verified

#### What happened instead?
I still see a red dot next to encryption

![Image](https://github.com/user-attachments/assets/c55ea1ac-dc7e-4f81-9fe9-58f4b9f73ae4)

### Your phone model

Samsung S24+

### Operating system version

Android 16.0

### Application version and app store

Element X version 26.0.3

### Homeserver

Synapse 1.150.0

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

bmarty:
Hello @floviolleau , this indicator can be shown if your key backup state is unknown, or if your recovery key state is unknown or incomplete. What do you see when you navigate to the "Encryption" screen?

floviolleau:
Hello @bmarty,

Thanks for caring this issue.

Here is the capture:

![Screenshot_20260402_121105_Element X.jpg](https://github.com/user-attachments/assets/424965d7-c35c-48bb-83a4-707392063e75)

floviolleau:
Maybe linked with https://github.com/element-hq/element-web/issues/33031

PR #7447, opened 2026-08-17
Hide the encryption badge when key storage is set up on the server

hayaksi1:
## Content

The badge next to Encryption in Settings was shown for every unknown backup state. A session whose key backup exists on the server but has not been reported as enabled by the SDK therefore kept a red dot, while the Encryption screen itself showed key storage as on and offered nothing to do. That is the mismatch reported in the issue.

The indicator now resolves an unknown backup state the way `SecureBackupRootState` already does, by asking whether a backup exists on the server, and only badges the entry when there is none. The recovery half of the indicator is unchanged, and the room list top bar indicator follows because it reuses this one.

## Motivation and context

Part of #6518.

## Tests

- `libraries/indicator/impl/.../DefaultIndicatorServiceTest.kt`: an unknown backup state with a backup on the server no longer shows the indicator, and an unknown backup state without one still does.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/indicator/impl/src/main/kotlin/io/element/android/libraries/indicator/impl/DefaultIndicatorService.kt
+++ b/libraries/indicator/impl/src/main/kotlin/io/element/android/libraries/indicator/impl/DefaultIndicatorService.kt
@@ -9,11 +9,14 @@
 package io.element.android.libraries.indicator.impl
 
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.State
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
 import dev.zacsweers.metro.ContributesBinding
 import io.element.android.libraries.di.SessionScope
 import io.element.android.libraries.indicator.api.IndicatorService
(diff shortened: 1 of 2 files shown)


bmarty:
We need a product decision for this change. CC @mxandreas

mxandreas:
> The indicator now resolves an unknown backup state the way `SecureBackupRootState` already does, by asking whether a backup exists on the server, and only badges the entry when there is none. The recovery half of the indicator is unchanged, and the room list top bar indicator follows because it reuses this one.

Reading the description - this seems to be more like technical/implementation discrepancies where different sources of information are used on different screens for the same thing. I think someone from the Crypto team needs to confirm if this is sensible. cc: @andybalaam

andybalaam:
> > The indicator now resolves an unknown backup state the way `SecureBackupRootState` already does, by asking whether a backup exists on the server, and only badges the entry when there is none. The recovery half of the indicator is unchanged, and the room list top bar indicator follows because it reuses this one.
> 
> Reading the description - this seems to be more like technical/implementation discrepancies where different sources of information are used on different screens for the same thing. I think someone from the Crypto team needs to confirm if this is sensible. cc: @andybalaam

Without reading the code, I absolutely agree that the indicator next to the "Encryption" option in settings should be consistent with the indicator that is shown once you open the Encryption screen in settings.

@bmarty please can you articulate the decision you want from product? I can tell you that we want to be consistent. Are you asking which state we should use to decide whether an indicator is visible?

bmarty:
Thanks for the feedback, and confirming that consistency is what we want. I am not sure why we had some inconsistency there...

bmarty (approved):

Issue #4702, opened 2025-05-09
No notifications and Error: HTTP 507 on Test Push loop back


Loppansson:
### Steps to reproduce

1. I log into the app 🙂
2. I get a message from my friend, but I do not get a notification 
3. I run the "trouble shoot notifications"

#### What did you expect?

I expected there to not be a problem.

#### What happened instead?

I get "Error: HTTP 507" on Test Push loop back, and I do not know what that means 🥲

![Image](https://github.com/user-attachments/assets/3ef6e128-e296-4798-9459-723411dddd1b)

### Your phone model

FP5

### Operating system version

13

### Application version and app store

Element X version 25.04.3

### Homeserver

matrix.org

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

B08Z:
![Screenshot_20250509_195318_Element X.jpg](https://github.com/user-attachments/assets/5b5e7e12-3115-4660-9529-bdc54c1c1ec9)

I too am struggling with notifications. 

I get a different error and the first time I ran the trouble shouting I got a notification but with the final http error. 


Run it again and I get a visual motivation sometimes.

jmartinesp:
> I get "Error: HTTP 507" on Test Push loop back, and I do not know what that means 🥲

This HTTP error means 'insufficient storage': https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/507, although it seems like the NTFY gateway implementation uses it to warn you that you've been rate-limited: https://github.com/binwiederhier/ntfy/blob/630f2957deb670dcacfe0a338091d7561f176b9c/docs/config.md?plain=1#L1098

> I get a different error and the first time I ran the trouble shouting I got a notification but with the final http error.

This may be related, but it's a different error: the [429 HTTP status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/429) is for 'too many requests', so you're being rate-limited, either you personally or the Element X client. The solution should be the same, use a different gateway, as for the previous issue.

Loppansson:
Thank you for the clarification! 🙂

How do I go about using another gateway?

B08Z:
Same question, I'm using Firebase as I thought this would be more reliable. But could switch to self hosted nfty.

I can't see how element x can be rate limited on firebase and if it's my client do you know what the limit is. I am only using to combine what's app and signal and I'm not exactly a social butterfly with hundreds of messages a day :)

Sent with Spark

[...]

PR #7448, opened 2026-08-17
Explain a rate limiting push gateway in the notification troubleshooter

hayaksi1:
## Content

The push loop back test printed whatever HTTP status the gateway answered with, so a user throttled by the public ntfy.sh gateway was told "Error: HTTP 507". That reads as a fault on the server rather than as a quota the user can wait out or move away from, and 507 is documented by ntfy as its rate limiting answer.

The two statuses that mean the notification was throttled, 429 and ntfy's 507, are now mapped at the gateway boundary onto a typed `PushGatewayFailure.RateLimited`, and the troubleshooter gets a branch for it with a message that says what to do. Every other status keeps the existing behaviour.

This covers the confusing error half of the issue. Not getting notifications at all, when the gateway is throttling the device, remains a gateway side matter.

## Motivation and context

Part of #4702.

## Tests

- `libraries/push/impl/.../pushgateway/DefaultPushGatewayNotifyRequestTest.kt`: 429 and 507 raise the new failure, and another status is passed through untouched.
- `libraries/push/impl/.../troubleshoot/PushLoopbackTestTest.kt`: the new failure ends the test in a failure state carrying the new message.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


jmartinesp (approved):
Thanks, the error is way easier to understand now.

I've added the missing string to localazy along with another missing one that went unnoticed, so this should be ready to merge.

Issue #4112, opened 2025-01-05
Unable to share screenshot from Google Pay to a room (unsupported media type image/*)


rajil:
### Steps to reproduce

1. Make a payment in google pay and share the transaction using 'Share screenshot' button in google pay to Element X
2. Select the appropriate room and click on Send

#### What did you expect?
Screenshot will get shared
#### What happened instead?
Got an error 'Your message was not bridged: failed to convert message: unsupported media type image/* in image message'

I tried again with old regular Element and it successfully shared the screenshot

### Your phone model

Samsung S23 Ultra

### Operating system version

Android 14

### Application version and app store

Element X Version 0.7.5

### Will you send logs?

No

### Are you willing to provide a PR?

No

rajil:
This seems to have resolved in version 0.7.6

rajil:
Tried this again with 0.7.6 today and received error,

`Your message was not bridged: unsupported media type (image/* in m.image)`

PR #7449, opened 2026-08-17
Send a concrete mime type for media shared with a wildcard one

hayaksi1:
## Content

Some apps share an image with the mime type `image/*` rather than a concrete one. That value was carried straight through the upload and into the event, so the message reached the room with a media type that is a filter rather than a media type, and bridges rejected it as "unsupported media type image/*".

A wildcard subtype is now resolved to a default for its top level type before the upload is processed. The mapping is the one the download path already applies so the SDK can pick a sensible extension, and it has been moved to `MimeTypes` so both sides use the same code rather than two copies.

Values that already carry a subtype are untouched, so only shares that were previously unusable change.

## Motivation and context

Part of #4112.

## Tests

- `libraries/core/.../mimetype/MimeTypesTest.kt`: concrete types are kept, `image/*`, `video/*` and `audio/*` get a concrete default, and anything else falls back to `application/octet-stream`.
- `libraries/mediaupload/impl/.../AndroidMediaPreProcessorTest.kt`: an image processed with `image/*` is uploaded as `image/jpeg`.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


libraries/core/src/main/kotlin/io/element/android/libraries/core/mimetype/MimeTypes.kt
```diff
@@ -63,4 +63,17 @@ object MimeTypes {
         val subType = components.last()
         return subType.isNotBlank() && subType != "*"
     }
+
+    /**
+     * A mime type with a wildcard subtype is a filter, not a media type, so replace it with a
+     * concrete default for its top level type.
+     */
+    fun String?.withDefaultSubtype(): String = when {
```
bmarty: The name `withDefaultSubtype` may confuse developers because if the mimetype receiver has a sub type, it will not be replaced by the default one. Maybe rename to `ensureDefaultSubtype`?

bmarty (approved):
Thanks!

libraries/core/src/main/kotlin/io/element/android/libraries/core/mimetype/MimeTypes.kt
hayaksi1: Done

bmarty:
Thanks for the update. Can you fix the conflict please?

hayaksi1:
@bmarty Done.

Issue #7059, opened 2026-06-19
Relax image compression appropriately.


YamatoRyou:
#### What would you like to do?
As the title suggests.

#### Why would you like to do it?
Whether taking photos within the app or selecting photos from the album, with "Media Upload Optimization" enabled, the app excessively compresses photos (resulting in a resolution of only 768x1024, and a file size typically not exceeding 100 KB). The iOS version, under the same conditions, can achieve 1536x2048 resolution, with a file size of several hundred KB.

#### How would you like to achieve it?
n/a

### Have you considered any alternatives?

n/a

### Additional context

n/a

### Are you willing to provide a PR?

No

PR #7451, opened 2026-08-17
Relax image compression so photos and screenshots keep more detail

hayaksi1:
## Content

`AndroidMediaPreProcessor` downscales images against a 640px reference. The KDoc next to it says the intent is to send images of up to 1280x1280, but because `inSampleSize` is a power of two the reference behaves as a half-size target: a 12000x3916 photo comes out as 3006x979, and a 1818x1178 screenshot as 909x589. Text in screenshots becomes unreadable and photo detail is lost.

Raise the reference to 1280 and correct the KDoc to describe what the value actually bounds. Since `inSampleSize` only doubles, every image moves exactly one step up rather than to a fixed size, and images already below the threshold are unaffected.

The trade-off is memory during upload: a very large source now decodes at `inSampleSize=4` instead of `8`, so the transient bitmap is roughly four times larger while it is being rotated and compressed. That is the same order of magnitude the in-app camera path already handles.

## Motivation and context

Part of #7059.

## Tests

`libraries/mediaupload/impl/.../AndroidMediaPreProcessorTest.kt`: the existing JPEG expectations move from 3006x979 to 6012x1958, which is the observable effect of the change. Run with `./gradlew :libraries:mediaupload:impl:testDebugUnitTest`.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/AndroidMediaPreProcessor.kt
+++ b/libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/AndroidMediaPreProcessor.kt
@@ -62,10 +62,10 @@ class AndroidMediaPreProcessor(
         /**
          * Used for calculating `inSampleSize` for bitmaps.
          *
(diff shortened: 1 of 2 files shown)


bmarty (approved):
LGTM, thanks!

Issue #3964, opened 2024-11-28
Alpha background is black when using os sharing vs in app sharing


toger5:
### Steps to reproduce

1. Where are you starting? What can you see?
2. What do you click?
3. More steps…

### Will you send logs?

Yes

### Are you willing to provide a PR?

Yes

PR #7452, opened 2026-08-17
Keep transparency when an image is shared into the app with an imprecise mime type

hayaksi1:
## Content

The upload pre-processor picks the bitmap encoder from the mime type it is handed: PNG stays PNG, everything else is re-encoded as JPEG. When an image arrives from another app the declared type is frequently broader than the image really is — a share intent that only says the top-level type, or a file whose declared type does not match its bytes. A transparent PNG that arrives that way is re-encoded as JPEG and its alpha channel is flattened to black, which is what the reporter sees when sharing from outside the app but not from inside it.

Resolve the concrete format from the image header before deciding anything, and use it both for the "can this be compressed at all" check (which is what keeps GIF and WebP untouched) and for the encoder choice. A declared PNG or JPEG is trusted as-is, so the common path does not read the header at all.

This deliberately does not widen the PNG branch of `mimeTypeToCompressFormat`: mapping every image to PNG would re-encode ordinary photos as PNG and inflate uploads.

## Motivation and context

Part of #3964.

## Tests

`libraries/mediaupload/impl/.../AndroidMediaPreProcessorTest.kt`: a PNG asset processed with a wildcard image mime type now reports `image/png`, where before it was re-encoded as JPEG. Run with `./gradlew :libraries:mediaupload:impl:testDebugUnitTest`.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/AndroidMediaPreProcessor.kt
+++ b/libraries/mediaupload/impl/src/main/kotlin/io/element/android/libraries/mediaupload/impl/AndroidMediaPreProcessor.kt
@@ -85,8 +85,9 @@ class AndroidMediaPreProcessor(
             val result = when {
                 mimeType == MimeTypes.Svg -> processSvgImage(uri, mimeType)
                 mimeType.isMimeTypeImage() -> {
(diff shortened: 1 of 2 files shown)


jmartinesp (approved):
Thanks!

hayaksi1:
@jmartinesp Done.

Issue #4639, opened 2025-04-25
Issue when switching to System theme


bmarty:
### Steps to reproduce

1. Set the application to dark theme
2. Element X UI is dark
3. Set the system to light theme.
4. Element X UI is still dark (this is OK)
5. Set the application to "System"

#### What did you expect?

Element X should follow the system and render the UI in light.

#### What happened instead?

Element X is still rendered in dark.

Note that switching twice the system theme fixes the issue.

### Your phone model

S24

### Operating system version

Android 15

### Application version and app store

25.04.3

### Homeserver

NA

### Will you send logs?

No

### Are you willing to provide a PR?

No

bmarty:
It seems that the method `isSystemInDarkTheme()` from `androidx.compose.foundation` is sometimes liing...

PR #7453, opened 2026-08-17
Apply the system theme when switching back to it from a forced light or dark theme

hayaksi1:
## Content

`Theme.System` is resolved with `isSystemInDarkTheme()`, which reads `LocalConfiguration` — the activity configuration. When the user picks Light or Dark, `ElementThemeApp` calls `AppCompatDelegate.setDefaultNightMode`, and AppCompat rewrites exactly that configuration to the forced night mode. Selecting System afterwards sets `MODE_NIGHT_FOLLOW_SYSTEM`, but the activity configuration still reports the previously forced value, so the app keeps showing the old theme.

Resolve the system night mode from the application configuration instead. AppCompat leaves it untouched, and it is the same configuration AppCompat itself reads to expand `MODE_NIGHT_FOLLOW_SYSTEM`. The activity configuration is still read, as the recomposition key, so a system-wide theme change still repaints.

## Motivation and context

Part of #4639.

## Tests

`libraries/compound/.../ThemeTest.kt`: the existing System cases now drive the application configuration, and a new case pins the fix — with the application configuration in dark and the activity configuration forced to light, `Theme.System.isDark()` returns true, and the mirror case returns false. Run with `./gradlew :libraries:compound:testDebugUnitTest --tests "*ThemeTest*"`.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/compound/src/main/kotlin/io/element/android/compound/theme/Theme.kt
+++ b/libraries/compound/src/main/kotlin/io/element/android/compound/theme/Theme.kt
@@ -8,8 +8,11 @@
 
 package io.element.android.compound.theme
 
-import androidx.compose.foundation.isSystemInDarkTheme
+import android.content.res.Configuration
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.remember
+import androidx.compose.ui.platform.LocalConfiguration
+import androidx.compose.ui.platform.LocalContext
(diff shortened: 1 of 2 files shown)


bmarty (approved):
LGTM, thanks for the fix!

Issue #6276, opened 2026-03-03
Voice message getting silent in android auto


SvenXP:
### Steps to reproduce

1. Pixel 8 pro connected to car via android auto. Playing music with synfonium. 
2. Pressing play on voice message. Message starts, music fades out. After a few seconds message gets silent. But still "running"
3. When starting music, also voice message can be heard again. Music fades again and message gets silent after another few seconds.

#### What did you expect?
Music fades, message is played and I can hear it. 

#### What happened instead?
Music fades, message is played and I can hear it only a few seconds.

### Your phone model

Pixel 8 pro

### Operating system version

Android 16

### Application version and app store

26.03.0

### Homeserver

Synapse non public home server

### Will you send logs?

No

### Are you willing to provide a PR?

No

jmartinesp:
Maybe related to https://github.com/element-hq/element-x-android/pull/6194 ?

PR #7454, opened 2026-08-17
Keep playing a voice message when another app plays a short sound

hayaksi1:
## Content

Two things about voice message audio focus, one per commit.

**A short sound from another app stops the message.** A notification, a navigation prompt or
anything else that takes focus for a moment sends `AUDIOFOCUS_LOSS_TRANSIENT` or
`AUDIOFOCUS_LOSS_TRANSIENT_CAN_DUCK`, and `DefaultMediaPlayer` pauses on it. The message stops part
way through and has to be found and restarted. Recording already ignores those two, because the
capture path is not affected by focus; voice message playback now does the same. A permanent
`AUDIOFOCUS_LOSS`, a real phone call, still pauses, and the media viewer is untouched. The requester
test moved into a named function rather than growing the inline condition.

**The declared usage does not match what is rendered.** `DefaultAudioFocus` groups voice message
playback with Element Call and voice message recording, so it asks for focus with
`USAGE_VOICE_COMMUNICATION` (and `STREAM_VOICE_CALL` on the pre-O path), while the player that
renders the audio uses the default media attributes. Routing decisions are made from the declared
usage, so a head unit that sees a voice-communication request switches to its call context and
routes the media output away while the player keeps running. That matches the two details in the
report that a pause does not: the message goes silent *while still running*, and it becomes audible
again when the music app resumes. Playback now declares `USAGE_MEDIA` / `STREAM_MUSIC`, matching
what it renders and what the media viewer already declares. Recording keeps
`USAGE_VOICE_COMMUNICATION`, which drives the capture path and the exclusive transient request next
to it. Element Call is unchanged.

The ducking policy is left alone, so `willPausedWhenDucked()` still returns `true` for voice
messages. Happy to split the second commit out if the two are better reviewed apart.

## Motivation and context

Part of #6276.

## Tests

`libraries/audio/impl/.../DefaultAudioFocusTest.kt`, run with
`./gradlew :libraries:audio:impl:testDebugUnitTest`:

- A transient focus loss, of either kind, does not pause a voice message; a permanent one does.
- A transient focus loss still does not stop a recording, and still pauses the media viewer.
- Each requester declares the usage it should: media for voice message playback and the media
  viewer, voice communication for recording and Element Call.

Reverting only the `VoiceMessage` line of `pausesOnTransientFocusLoss` fails the transient case and
nothing else. This required adding the common test dependencies to the module, which had no test
source set.

The routing behaviour itself needs a car head unit and could not be reproduced here; that half is
argued from the declared-versus-rendered mismatch in the code and from the symptoms in the report.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


jmartinesp:
Sorry, but AFAICT this doesn't fix the issue the user was reporting: my theory is a notification sound paused their voice message, or maybe some other short-lived sound that requested focus. 

The right fix I think is probably to treat `AudioFocusRequester.VoiceMessage` as we do with  `AudioFocusRequester.RecordVoiceMessage`  and ignore `onFocusLost()` for it when we get `AudioManager.AUDIOFOCUS_LOSS_TRANSIENT` or  `AudioManager.AUDIOFOCUS_LOSS_TRANSIENT_CAN_DUCK` so the playing isn't paused.

hayaksi1:
@jmartinesp Done.

jmartinesp (approved):

Issue #4582, opened 2025-04-11
Autoscroll the screen to the TextField with focus


bmarty:
It can happen that on some devices the keyboard hides the TextField which has the focus. In this case the user does not see what they are typing. This has been observed on the password field of the login screen for instance.

We should revisit all the places in the code where this can occurs, and ensure that the TextField is visible when it gets the focus.

PR #7455, opened 2026-08-17
Keep the focused login field visible above the keyboard

hayaksi1:
## Content

On the login and account provider screens the software keyboard can cover the field being typed into. Both screens put their fields in a scrollable container, but nothing scrolls the content back once the keyboard reduces the viewport, so on shorter screens the focused field ends up behind the keyboard. The maintainer named the login password field specifically on the issue.

Apply the existing `Modifier.bringIntoViewOnImeVisible()` to those fields. It is already used by the recovery key screen for the same purpose: it waits for the IME to actually be visible before requesting `bringIntoView`, so the scroll happens against the reduced viewport rather than the full one.

Scope is deliberately limited to the login flow fields the issue is about, rather than a blanket change to every text field in the app.

## Motivation and context

Part of #4582.

## Tests

No new automated test. The behaviour depends on the IME insets actually changing and on the field being taller than the remaining viewport, neither of which the module's Robolectric view tests can produce; `ExpandSheetWhenImeIsVisibleTest` is the closest existing harness and covers the modifier's sibling rather than layout outcomes. The existing `:features:login:impl` tests pass unchanged.

Manual check: focus the password field on a short screen and confirm it scrolls above the keyboard.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty (approved):
Thanks!

Issue #3756, opened 2024-10-30
Emptiing the composer when editing the message allow to record a voice message


bmarty:
### Steps to reproduce

1. Send a message like "Test"
2. Edit the message

<img width="390" alt="image" src="https://github.com/user-attachments/assets/f44820a0-fe93-4356-bd6f-5aab541041da">

3. Empty the composer
4. Observe that it's now possible to record a voice message

<img width="390" alt="image" src="https://github.com/user-attachments/assets/3747ebf7-f683-4e5c-b721-979e6698c0a6">

#### What did you expect?

When editing a message it should not be possible to record a voice message

#### What happened instead?

The microphone is displayed, so I can record a voice message

### Your phone model

Any

### Operating system version

Any

### Application version and app store

0.7.2 and previous

### Homeserver

NA

### Will you send logs?

No

### Are you willing to provide a PR?

No

PR #7456, opened 2026-08-17
Keep the send button when the composer is emptied while editing a message

hayaksi1:
## Content

`rememberEndButtonParams` decides what the composer's end button does. It checks whether there is sendable text *before* it checks whether the user is editing a message, so clearing the text of a message being edited falls into the empty-composer branch and the send button is replaced by the voice recorder button. Starting a recording from there abandons the edit, and there is no way back to confirming it.

Check the editing branch first and let it render a disabled send button when there is nothing to send. This is exactly what `rememberEndButtonParamsFormatting` already does for the formatting composer — the two functions simply disagreed on branch order, and this makes the plain composer follow the one that is already right.

Note for the screenshot bot: the edit and caption composer previews include an empty-text state, so their goldens change — the empty editing row now shows a disabled send button instead of the microphone. I have not re-recorded them, because Paparazzi renders focused text fields differently on my machine and recording here would bake in unrelated diffs.

## Motivation and context

Part of #3756.

## Tests

New `libraries/textcomposer/impl/.../TextComposerEndButtonTest.kt`:

- editing with text shows the send edited message button;
- emptying the composer while editing keeps it, and does not offer to record a voice message;
- the same for an empty caption edit;
- an empty composer outside of edition still offers to record a voice message, which is the behaviour this must not regress.

Run with `./gradlew :libraries:textcomposer:impl:testDebugUnitTest`.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


jmartinesp (approved):
Thanks!

jmartinesp:
Ah, we need to record new screenshot firsts. Could you do that before merging, please?

hayaksi1:
@jmartinesp Done.

Issue #1270, opened 2023-09-11
Add a timeout to crypto verification screens


michaelkaye:
### Steps to reproduce

Install on a test server. Attempt to verify two devices (element web + element x android)

Sit waiting on a spinner for 5+min

#### What did you expect?
"looks like it didn't work; do you want to retry?"

#### What happened instead?
Spun for over 5min until i got bored -

### Your phone model

Pixel XL

### Application version and app store

Built from develop

### Homeserver

internal latest synapse

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

bmarty:
Done in #7457

PR #7457, opened 2026-08-17
Stop waiting on a verification request the other session never accepts

hayaksi1:
## Content

The outgoing verification flow has no client-side timeout. Once a request is sent the screen sits on "Waiting to accept request" indefinitely: if the other session never answers, or is not running at all, nothing in the app ever tells the user the request can no longer succeed, and the only way out is to press back.

Bound that wait to ten minutes, the lifetime the Matrix specification gives a `m.key.verification.request`. Past that point the other session can no longer accept it, so waiting longer cannot succeed. On expiry the request is cancelled and the machine moves to its existing canceled state, which already offers to start again.

The duration is anchored to the specification rather than chosen for UX reasons: cancelling earlier would abort verifications that would still have completed once the user picked up their other device.

## Motivation and context

Part of #1270.

## Tests

`features/verifysession/impl/.../OutgoingVerificationPresenterTest.kt`: a new case asserts that a request nobody accepts stays in the waiting state at nine minutes, then cancels the verification and moves to the canceled state once the timeout passes.

The shared helper that drives a verification to the verifying state now advances time by a bounded amount instead of `advanceUntilIdle()`, which would otherwise skip straight past the new timeout.

Run with `./gradlew :features:verifysession:impl:testDebugUnitTest`.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty (approved):
Tested OK, thanks!

bmarty:
It seems that there is an unused import:

/home/runner/work/element-x-android/element-x-android/features/verifysession/impl/src/test/kotlin/io/element/android/features/verifysession/impl/outgoing/OutgoingVerificationPresenterTest.kt:30:1: The import 'kotlinx.coroutines.test.TestScope' is unused. [UnusedImports]

hayaksi1:
@bmarty Done — restored the file-level `@file:Suppress("UnusedImports")` this branch had dropped. The import is genuinely needed: `TestScope` only appears in the `context(testScope: TestScope)` parameter of `requestVerificationAndAwaitVerifyingState`, which the ktlint rule does not look at, so the suppression was there for exactly this reason and removing it was an accident on my side.

`:features:verifysession:impl:ktlintCheck`, `detekt` and `testDebugUnitTest` are green locally.

features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/outgoing/OutgoingVerificationStateMachine.kt
```diff
@@ -22,10 +22,12 @@ import io.element.android.libraries.matrix.api.verification.SessionVerificationS
 import io.element.android.libraries.matrix.api.verification.VerificationRequest
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.TimeoutCancellationException
 import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.timeout
+import kotlin.time.Duration.Companion.minutes
```
bmarty, line 30: ktlint is now complaining that imports are not ordered correctly. Can you fix please?

bmarty (changes requested):

features/verifysession/impl/src/main/kotlin/io/element/android/features/verifysession/impl/outgoing/OutgoingVerificationStateMachine.kt
hayaksi1: Done!

features/verifysession/impl/src/test/kotlin/io/element/android/features/verifysession/impl/outgoing/OutgoingVerificationPresenterTest.kt
```diff
@@ -141,6 +143,26 @@ class OutgoingVerificationPresenterTest {
...
+    @Test
+    fun `present - A verification request the other session never accepts is canceled`() = runTest {
+        val cancelVerificationRecorder = lambdaRecorder<Unit> {}
+        val service = unverifiedSessionService(
+            requestDeviceVerificationLambda = { },
+            cancelVerificationLambda = cancelVerificationRecorder,
+        )
+        val presenter = createOutgoingVerificationPresenter(service)
+        presenter.test {
+            awaitItem().eventSink(OutgoingVerificationViewEvents.RequestVerification)
```
bmarty:
```suggestion
            awaitItem().eventSink(OutgoingVerificationViewEvent.RequestVerification)
```

bmarty (approved):
Fixing issue after merging develop

PR #7461, opened 2026-08-17
Sync strings - New language supported: Arabic

bmarty:
https://github.com/element-hq/element-x-android/actions/runs/31981105744/job/95247915868 failed as expected because of string duplication.


jmartinesp:
There is another duplicate: `notification_knock_request_body`.

jmartinesp (approved):
Thanks!

Issue #4460, opened 2025-03-23
Click on notification always jump to end


xundeenergie:
### Steps to reproduce

1. I receive a notification via ntfy
2. I click on this notification

#### What did you expect?
The timeline jumps to the notified message

#### What happened instead?
It always jumps to the end in a room.
Sometimes i think i can see, that it first jumps to the notified message and immediately it jumps to the end.

### Your phone model

Pixel7

### Operating system version

GrapheneOS

### Application version and app store

Element X all installed versions and Schildichat Next

### Will you send logs?

No

### Are you willing to provide a PR?

No

PR #7463, opened 2026-08-17
Keep a focused timeline focused until the event has been rendered

hayaksi1:
## Content

Opening a room from a notification asks the timeline to focus the notified event. When that event is outside the loaded live window, `TimelineController` creates a detached timeline focused on it, and a detached timeline always carries a forward loading indicator as its last chronological item. The items list is reversed for display, so that indicator is the bottom-most row and is composed on first layout, and merely composing it emits a forward pagination request nobody asked for. Because a notified event sits close to the live end, that single page reaches the end, and the controller responds by closing the detached timeline. The live window replaces the item list while it is still anchored at the bottom, so the user lands on the newest message and the focus request is discarded.

Automatic forward pagination is now deferred while a focus request is still pending. Once the focused event has been rendered the viewport has moved off the indicator row, so only a real scroll back to the live end switches to the live timeline, which is what the controller's existing behaviour is for.

Backward pagination and the explicit jump-to-live action are untouched, and `TimelineController` itself is unchanged, so the intended detached-to-live switch still happens when the user actually scrolls to the end. The loading indicator's timestamp is re-stamped on every timeline emission, so the deferred request is not lost — it fires again on the next emission if the row is still on screen.

## Motivation and context

Part of #4460.

## Tests

`features/messages/impl/.../timeline/TimelinePresenterTest.kt`: a forward pagination while the focused event is not rendered yet is ignored and the detached timeline stays open; the same pagination is performed once `OnFocusEventRender` has arrived, and the timeline then switches back to live; a backward pagination is still performed while the focus request is pending.

`TimelineControllerTest` is unchanged and still green, which is what proves the intended switch to live was not broken.

These are unit tests over the presenter, so they pin the pagination gate rather than the on-device scroll position.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelinePresenter.kt
@@ -166,9 +166,14 @@ class TimelinePresenter(
         fun handleEvent(event: TimelineEvent) {
             when (event) {
                 is TimelineEvent.LoadMore -> {
-                    if (event.direction == Timeline.PaginationDirection.FORWARDS && timelineMode is Timeline.Mode.Thread) {
-                        // Do not paginate forwards in thread mode, as it's not supported
-                        return
+                    if (event.direction == Timeline.PaginationDirection.FORWARDS) {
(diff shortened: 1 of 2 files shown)


bmarty (approved):
Thanks!

Issue #4095, opened 2024-12-25
Don't show unknown/custom state events in rooms by default


tulir:
#### What would you like to do?
Stop users from seeing confusing "custom events"

#### Why would you like to do it?
Many users don't know how Matrix works, and some users have strong adverse reactions to anything unexpected in their chat view

#### How would you like to achieve it?
Don't show any unknown event types in the room timeline, except if the user opts into viewing hidden events via a developer option (#3760)

### Have you considered any alternatives?

Users could be educated to understand Matrix better and not get scared about unknown things, but that is likely more difficult than hiding unknown events

### Are you willing to provide a PR?

No

PR #6862, opened 2026-05-25
Do not show "Custom event" and "Made no changes" events if "View source" isn't enabled

bxdxnn:
## Motivation and context

Fixes #5209

1. These events are purely technical
2. "User made no changes" literally doesn't mean anything to the user

## Tests

- Test if you see these events with "View source" enabled
- Test if you don't see these events with "View source" disabled

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

- This PR was made with the help of AI:
    - [x] Yes. In this case, please request a review by Copilot.
    - [ ] No.
- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly defines what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #7464, opened 2026-08-17
Do not render unknown state events in the timeline

hayaksi1:
## Content

An unknown or custom state event was rendered in the timeline as the untranslated developer string `Custom event <type>`, while the room list already filtered the same event out. Any room whose members use a custom state event therefore shows rows of internal event types to every user, in English, whatever their locale.

The formatter now declines custom state events in both rendering modes, and the timeline drops events the formatter cannot render instead of coercing the missing text to an empty string. Both halves are needed: on its own, returning null from the formatter turns the wrong string into an empty tappable pill, because the timeline had no notion of "this event has nothing to display".

The filter is applied to the raw items before the item factory runs, rather than by returning null later. That matters for two reasons: the hidden event never becomes a neighbour when bubble grouping is computed, so the messages either side of it are not split into separate blocks for no visible reason; and a day divider whose only remaining event was hidden is dropped with it. This mirrors the existing `keepDisplayablePinnedEvents` precedent.

The SDK-side timeline filter cannot express this. It excludes an enumerated list of state event types, and the set of custom types is unbounded, so client-side filtering is the only option.

The issue also asks for a developer option to view hidden events (#3760). No such option exists in the codebase yet, so this ships the plain hide; the opt-in would be purely additive later.

## Motivation and context

Part of #4095.

## Tests

`features/messages/impl/.../timeline/factories/DisplayableTimelineEventsTest.kt`: a custom state event is dropped, a known state event and a message are kept, a day divider is dropped when its only event was hidden and kept when the day still has an event or a read marker.

`features/messages/impl/.../timeline/factories/TimelineItemsFactoryTest.kt`: a custom state event is not emitted and does not split the group of the messages around it. This fails on develop with `None, Middle, None` instead of `First, Last`, so it pins both the filtering and the fact that it happens before the group position is computed.

`libraries/eventformatter/impl/.../StateContentFormatterTest.kt`: a custom state event is not rendered in either rendering mode.

Run with `./gradlew :features:messages:impl:testDebugUnitTest :libraries:eventformatter:impl:testDebugUnitTest`.

Read receipts and permalinks that land on a hidden event are not covered: as with the state types the SDK already excludes, they have nowhere to resolve to.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


frebib:
Can I ask that instead of unconditionally removing these from the timeline, there's still an option to show them? Maybe a toggle in developer options?

jmartinesp:
I think we need some product decision for this. @mxandreas this is waiting for when you come back.

jmartinesp:
That said, this seems something that should use `RustTimelineEventFilterFactory` to filter out state events based on some `FilterTimelineEventCondition`. While there's not any for all custom/unknown event types, probably adding that condition in the SDK wouldn't be that hard.

mxandreas:
> The issue also asks for a developer option to view hidden events (#3760). No such option exists in the codebase yet, so this ships the plain hide; the opt-in would be purely additive later.

Happy to help but not sure if I fully understand the core problem correctly. So, there can always be a case where the client just does not know an event type (perhaps it is a new event type and the client is old or it is some kind of custom event as mentioned) - correct?

I think hiding such an event isn't good because that can make "messages" disappear and create confusion. But I also agree that a hardcoded English message is also not great. Thus, ideally:

1) The message is translatable and something a regular user can read, e.g. "[Sender] sent a message that can't be displayed".
2) There's an option to hide/ignore such events - in case for some reason a user is starting to receive them a lot. I'd say in most cases it should be rare or?

mxandreas:
Also, I assume this applies only to non-state events and that in case the event has the fallback `body`, we display that instead?

bxdxnn:
I think this duplicates #6862

hayaksi1:
@mxandreas To clarify the scope: this only affects custom **state** events (`OtherState.Custom`). A state event has no fallback `body`, so there is nothing to display instead — that is why it hides rather than substitutes. If you prefer a translatable "[Sender] sent a message that can't be displayed" plus an option to hide, I will rework it that way, just say which one you want.

@jmartinesp Agreed that is the better layer. `StateEventType.Custom(type)` already covers a named type, but there is no condition for the unbounded "all unknown types" set: `FilterTimelineEventCondition` in matrix-rust-sdk currently has only `EventType`, `MembershipChange` and `ProfileChange`. If that condition lands in the SDK, `DisplayableTimelineEvents.kt` here should be dropped in favour of it.

@frebib A developer option can be added on top of this without changing the filtering.

On #6862: the overlap is real for the formatter half. One thing that PR needs either way — `TimelineItemContentStateFactory` does `TimelineItemStateEventContent(text.orEmpty().toString())`, so a `null` from the formatter renders a blank tappable row rather than removing it. That is why this PR filters the timeline list as well, which also takes the day divider that is left with no events.

mxandreas:
> To clarify the scope: this only affects custom **state** events (`OtherState.Custom`). A state event has no fallback `body`, so there is nothing to display instead — that is why it hides rather than substitutes. If you prefer a translatable "[Sender] sent a message that can't be displayed" plus an option to hide, I will rework it that way, just say which one you want.

If this is about custom **state** events, then I think we should hide them.

jmartinesp:
I've added https://github.com/matrix-org/matrix-rust-sdk/pull/6969 that should cover this and simplify the logic.

jmartinesp:
With the new SDK bindings, you can change the current code to just adding a new timeline filter condition that excludes `AnyCustomStateEvent`. Let me know if you want to handle it in this PR or if we can close it and create a separate PR ourselves.

Issue #5224, opened 2025-08-26
Use text entered before selecting an image to send as image caption


noantiq:
#### What would you like to do?

I would like to use my text that I entered in the composer as an image caption.

#### Why would you like to do it?

Convenience. Because I forget to select the image before writing a caption or because I'm writing a message and then deciding to send an supporting image.

#### How would you like to achieve it?

- If I entered text that I did not send as a message yet (the composer is not empty) and then select an image, I want the entered text from before to be used as the image caption - meaning that the text is copied and used to pre-fill the image caption text input.
  - If I then send the image with that caption, it should be removed from the message composer where I first added it
  - If I cancel sending the image, eg. by using the back button, the text should not be lost, but instead it should still be in the message composer where I first entered it
    - If I only entered text for the image caption after selecting an image, that should not be copied to the message composer

### Additional context

I want to have an familiar interaction flow - other apps, like Signal and Whatsapp also behave as described.

### Are you willing to provide a PR?

No

beedell-roke:
This is all the more important, considering [`issues/5978`](https://github.com/element-hq/element-x-android/issues/5978#issue-3789561836).

PR #7465, opened 2026-08-17
Use the text already in the composer as the attachment caption

hayaksi1:
## Content

Typing a message and then picking an image opened the attachment preview with an empty caption editor, so the text had to be retyped, and the original was still sitting in the composer after the image was sent.

The composer's text now travels forwards to the preview as the initial caption, carried on the existing parcelable navigation target. A single room-scoped bit travels back to say that a handed-over caption was actually sent, keyed by `Timeline.Mode`, and the composer clears itself and deletes its saved draft on its next composition.

Clearing on send rather than on hand-over is deliberate. Cancelling the preview reports nothing, so the typed text is still there, restored by the composer's own saveable state and by the draft that Appyx's pause of the off-screen composer persists. If the back-channel bit is ever lost, for example to process death while the preview is open, the user keeps a duplicate copy of their text, which is a better failure than losing it.

Two cases are deliberately left alone. A pending edit hands over nothing and is not cleared, so the text being edited still edits the original on the next send. And `MessageComposerEvent.SendUri`, the share-into-app and keyboard-image path, bypasses the preview entirely and is unchanged.

Keying the hand-over on the timeline mode is what keeps a live composer and a thread composer from clearing each other; they share one room-scoped composer context, so a plain flag would not have been safe.

## Motivation and context

Part of #5224.

## Tests

`features/messages/impl/.../attachments/AttachmentsPreviewPresenterTest.kt`: a handed-over caption pre-fills the caption editor and no caption leaves it empty; sending reports the caption as sent, sending without one reports nothing, and cancelling reports nothing; sending in a thread reports the caption for that thread only.

`features/messages/impl/.../messagecomposer/MessageComposerPresenterTest.kt`: picking an image hands the composer text over, hands over nothing when the composer is empty, and hands over nothing while editing; the composer is cleared and its draft deleted once a handed-over caption has been sent; a caption sent in a thread does not clear the live composer.

Rich-text formatting is flattened to markdown by the hand-over, which is what a caption can carry today; converting a caption to HTML is separate scope.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesFlowNode.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesFlowNode.kt
@@ -169,7 +169,12 @@ class MessagesFlowNode(
         ) : NavTarget
 
         @Parcelize
-        data class AttachmentPreview(val timelineMode: Timeline.Mode, val attachments: ImmutableList<Attachment>, val inReplyToEventId: EventId?) : NavTarget
+        data class AttachmentPreview(
+            val timelineMode: Timeline.Mode,
+            val attachments: ImmutableList<Attachment>,
+            val inReplyToEventId: EventId?,
+            val caption: String?,
+        ) : NavTarget
 
         @Parcelize
         data class LocationViewer(val mode: ShowLocationMode) : NavTarget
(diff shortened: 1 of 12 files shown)


bmarty (commented):
LGTM, but I prefer to request a product approval for this change. CC @mxandreas

mxandreas:
> LGTM, but I prefer to request a product approval for this change. CC @mxandreas

This is fine with me on the condition that we also do it on iOS now.

Issue #4770, opened 2025-05-23
Improve VoiceMessageComposerPresenterTest


bmarty:
For instance:
- `RECORDING_STATE` should be splitted into several intermediate states, because `FakeVoiceRecorder` emits all the states at the same time in the method `startRecord()`. A delay should be added [here](https://github.com/element-hq/element-x-android/blob/develop/libraries/voicerecorder/test/src/main/kotlin/io/element/android/libraries/voicerecorder/test/FakeVoiceRecorder.kt#L50).
- `FakeVoiceRecorder` could take some lambda in the constructor parameter to follow our current pattern. It coulde replace the method `FakeVoiceRecorder.assertCalls` and `FakeVoiceRecorder.givenThrowsSecurityException`
- Doing the change above makes some tests to fails, in particular `error("Previous recording was not cleared")` is invoked. This should be fixed, maybe `FakeVoiceRecorder` is a bit too clever?

PR #7466, opened 2026-08-17
Model the fake voice recorder on a cancellable recording job

hayaksi1:
## Content

`FakeVoiceRecorder` pushed a whole recording synchronously from a single suspending call, which made it both weaker and stricter than `DefaultVoiceRecorder`. It now models a cancellable recording job, the way the real recorder does.

Every level emission carried the same elapsed time and the emissions were coalesced by Molecule, so no intermediate recording state was observable. `present - recording state - number of levels is limited` had to compensate with `skipItems(numberOfLevels / 2 - 1)`, a magic number that only worked because of the two-to-one coalescing ratio. Each level now advances the test time source and suspends for its share of the recording duration, so the states arrive one at a time, `present - recording state` asserts both of them, and that magic number is gone.

Call verification moved to the constructor lambdas used elsewhere in the repo, replacing the hand-rolled `startedCount`/`stoppedCount`/`deletedCount` and `assertCalls`. The counters discarded the argument, so nothing pinned that cancelling a recording stops it with `cancelled = true`; `present - abort recording` now asserts exactly that. `givenThrowsSecurityException` is replaced by a throwing `startRecordResult`, and Truth is no longer a dependency of the fixture module.

Finally, a second `startRecord()` no longer raises `error("Previous recording was not cleared")`, where `DefaultVoiceRecorder` only warns and returns, and stopping or deleting now terminates an in-flight emission loop the way cancelling the real recording job does — so a stale `Recording` state can no longer arrive after `Finished`, which is what made adding a delay to the fake break tests before.

No production code changes.

## Motivation and context

Part of #4770.

## Tests

Test-only change. `DefaultVoiceMessageComposerPresenterTest` is 28 tests green with the reworked fake.

Behavioural gains, rather than a rename: `present - recording state` now asserts the intermediate `Recording(500ms, [0.1f])` as well as the final state; `present - recording state - number of levels is limited` no longer depends on the coalescing ratio; `present - abort recording` asserts `stopRecord(cancelled = true)`, which the old counters could not see.

Checked that the new `lambdaError()` defaults do not fire through `FakeDefaultVoiceMessageComposerPresenterFactory` by running `MessagesPresenterTest` unchanged, and that `:libraries:voicerecorder:test` and `:features:messages:test` still compile.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl/voicemessages/composer/DefaultVoiceMessageComposerPresenterTest.kt
+++ b/features/messages/impl/src/test/kotlin/io/element/android/features/messages/impl/voicemessages/composer/DefaultVoiceMessageComposerPresenterTest.kt
@@ -47,6 +47,7 @@ import io.element.android.libraries.voicerecorder.api.VoiceRecorder
 import io.element.android.libraries.voicerecorder.test.FakeVoiceRecorder
 import io.element.android.services.analytics.test.FakeAnalyticsService
 import io.element.android.tests.testutils.WarmUpRule
+import io.element.android.tests.testutils.consumeItemsUntilTimeout
 import io.element.android.tests.testutils.lambda.any
 import io.element.android.tests.testutils.lambda.lambdaRecorder
 import io.element.android.tests.testutils.lambda.value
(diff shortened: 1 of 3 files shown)


jmartinesp (approved):
Looks OK if the CI passes.

Issue #3790, opened 2024-11-02
Leaving a room sometimes doesn't remove it from the room list 


frebib:
This has happened several times to me in the past few months:

Steps to reproduce:
1. Leave a room
2. Observe that instead of being dumped back in the room list, you're still in the room timeline even though it's clear that you left:![Screenshot_20241102-111128.png](https://github.com/user-attachments/assets/f050cf60-7f1e-4836-b8ff-63b3d45b92a4)
3. Leaving the room again makes it disappear from the room list

stevecs:
Just noticed this now on element version 1.12.10.   Left a room and it still shows up on the room list.   When you select it it shows that you are 'previewing' and asks if you want to join but there is no means to remove the room from the list.

PR #7467, opened 2026-08-17
Dismiss the room screen whichever leave signal arrives last

hayaksi1:
## Content

Leaving a room from the room screen produces two independent signals: sync reports the membership as `LEFT`, and `RoomMembershipObserver` reports the local leave. `RoomFlowNode` decided what to do by sampling the second one at the moment the first arrived, peeking at a shared flow's replay cache. When sync won that race the peek came back empty, so a room the user had just left looked like a room left elsewhere, and the join room screen was shown instead of returning to the room list. Nothing re-evaluated the decision when the local signal did arrive, so the user was left on a screen for a room they had already left.

The two signals are now combined rather than one being sampled, so whichever arrives second re-evaluates the decision and the dismissal is not lost. Waiting for the local signal is not an option: it never arrives for a leave from another device, which is why commit e52f2e2274 replaced a suspending read with the peek in the first place. Combining keeps that fix — the latch simply stays false for a remote leave — while making the outcome independent of arrival order.

The decision itself moved into its own function, so both orderings can be tested without standing up the navigation graph. The node keeps only the navigation.

This addresses the navigation half of the issue. The other half in the title, a left room lingering in the room list, is room-list filtering in the SDK and is not touched here.

## Motivation and context

Part of #3790.

## Tests

`appnav/src/test/kotlin/io/element/android/appnav/room/RoomMembershipNavigationTest.kt`: a joined room enters the room and an invited room shows the join room screen; leaving from this device dismisses the flow in both arrival orders; leaving from another device shows the join room screen; a local leave for a different room is ignored.

The room-info-first ordering is the regression guard — it fails with the previous replay-cache peek and passes with the combined flow.

Run with `./gradlew :appnav:testDebugUnitTest`.

These are flow-level tests, so they pin the decision rather than the resulting Appyx transition.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/room/RoomMembershipNavigation.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/room/RoomMembershipNavigation.kt
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2026 Element Creations Ltd.
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+ * Please see LICENSE files in the repository root for full details.
+ */
+
+package io.element.android.appnav.room
+
+import io.element.android.libraries.core.coroutine.withPreviousValue
+import io.element.android.libraries.matrix.api.core.RoomId
+import io.element.android.libraries.matrix.api.room.CurrentUserMembership
+import io.element.android.libraries.matrix.api.room.RoomInfo
+import io.element.android.libraries.matrix.api.room.RoomMembershipObserver
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.filter
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.stateIn
+import java.util.Optional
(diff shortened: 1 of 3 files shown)


bmarty (commented):
LGTM, but maybe @ganfra can have a second look.

appnav/src/main/kotlin/io/element/android/appnav/room/RoomMembershipNavigation.kt
```diff
@@ -0,0 +1,64 @@
...
+internal fun roomMembershipNavigation(
+    roomId: RoomId,
+    roomInfoFlow: Flow<Optional<RoomInfo>>,
+    membershipUpdates: Flow<RoomMembershipObserver.RoomMembershipUpdate>,
+    scope: CoroutineScope,
+): Flow<RoomMembershipNavigation> {
+    val leftFromCurrentDeviceFlow = membershipUpdates
+        .filter { it.roomId == roomId && !it.isUserInRoom }
+        .map { true }
+        .stateIn(scope, started = SharingStarted.Eagerly, initialValue = false)
```
ganfra, line 48: Maybe change to 
```
val leftFromCurrentDeviceFlow = membershipUpdates
      .filter { it.roomId == roomId }
      .map { !it.isUserInRoom }
      .stateIn(scope, started = SharingStarted.Eagerly, initialValue = false)
```

ganfra (commented):
One remark otherwise LGTM

Issue #4657, opened 2025-04-29
Classes that extend PreviewParameterProvider<T> confusingly suffixed with just "Provider"


lex-neufeld:
Proposal: suffix them with PreviewParam instead. 
"Preview" alone is not a good candidate as is already used as a suffix for functions as part of a well established pattern. 
"Params" alone is not a good candidate as is already used as a data class and is maybe a bit too general. 
"PreviewParams" as a string is only present in one .java file for the Camera so little chance for confusion there.
"PreviewParam" with no -s doesn't appear to be used anywhere.

Why: make dev onboarding less painful, speed up debugging

Status Quo:
If I understand correctly, Previews get their "mock" data from PreviewParameterProviders which are configured in a separate file/class per View and called from that View. At first glance, the naming conventions is <PackageName>View / <PackageName>Provider. These classes for creating "mock" data for Previews are not production code. However, Matrix and Element-X both contain numerous features that are called "Provider" as well as the concept being a strong pattern in production code all over the project. Sometimes a "Provider" suffix means "not production code, just preview stuff" and sometimes it is production code. So you have to double think when searching for usages. Additionally, we get brain cancer inducing names when a provider feature also has a View and thus has a Provider for the Preview for the View of the Provider. 
Examples: 
AccountProviderProvider is just the preview param for the View for AccountProvider
changeaccountprovider is a production feature. It has a View, 
but there is no ChangeAccountProviderProvider, but rather
ChangeAccountProviderStateProvider for the preview param for ChangeAccountProviderView
not to be confused with: changeAccountProviderNodeProvider (which is part of a factory)

Could instead be:
AccountProviderPreivewParam for the View for AccountProvider
changeaccountprovider has has a ChangeAccountProviderView
for which there is, naturally, a ChangeAccountProviderPreviewParam. I don't need to parse what type each View operates on while scanning lists. 
Then PreviewParam classes are  immediately understandable and can be safely ignored by new and old devs alike unless they are actually working on previews.

[...]

PR #7468, opened 2026-08-17
Rename PreviewParameterProvider subclasses to the PreviewParam suffix

hayaksi1:
## Content

The `Provider` suffix on `PreviewParameterProvider` subclasses reads as a dependency provider rather than as preview data, and it collides with the real providers in the codebase — `StringProvider`, `PickerProvider`, and the login feature's own `AccountProvider` domain type — so the name does not say which of the two a class is. This renames them all to a `PreviewParam` suffix, as agreed on the issue.

194 classes move, along with the 185 files named after them, the konsist rule that enforces the convention, the two presentation-class file templates plus their settings, and the onboarding documentation.

Rebased on develop by merge. Upstream deleted the account provider search and change screens, so the two preview classes this branch had renamed there go with them, and the two preview classes develop added in the meantime — `GlobalSearchStateProvider` and `SendFailureDialogStateProvider` — are renamed here too, which keeps the totals above unchanged.

It cannot be split. `Konsist.scopeFromProduction()` scans the whole project, so flipping the assertion before every class has moved leaves the rule red, and a rule accepting either suffix during a transition would be the kind of compatibility layer this change exists to avoid.

Two judgement calls inside the rule are worth calling out. The `name.endsWith("PreviewProvider").not()` clause is dropped, because its only job was to reject a redundant `Preview` in front of `Provider` and the new suffix already contains `Preview`; no class currently ends with `PreviewProvider`, so this changes nothing today. And `EditableAvatarViewUriProvider` is renamed in the exclusion list rather than removed, even though no such class exists in this repository — it presumably comes from the modules overlaid for the enterprise build, so any `PreviewParameterProvider` still named `*Provider` over there will need the same rename before the flipped rule passes for that variant.

Nothing else keyed off the old suffix. No snapshot filename contains a provider class name, so there is no Paparazzi churn, and the Kover variants match on `*Presenter`, `*State` and `*ViewKt`, so the coverage bounds do not move.

## Motivation and context

Part of #4657.

## Tests

`tests/konsist/.../KonsistClassNameTest.kt` is the test for this convention: the rule now requires the `PreviewParam` suffix, and its "more than 100 classes found" guard still holds. Run with `./gradlew :tests:konsist:testDebugUnitTest --rerun` — konsist scans sources at execution time, so a cached run proves nothing.

`./gradlew :tests:uitests:compileDebugUnitTestKotlin` compiles every module the rename touches (188 of them) and is green, and `./gradlew ktlintFormat` handled the import re-ordering that the renames forced.

`git status --porcelain -- tests/uitests/src/test/snapshots` is empty, which is the check that no snapshot changed.

No new test: a naming convention has nothing behavioural to assert beyond the konsist rule.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


jmartinesp:
@bmarty what do you think? I'm in favor of this change, since `StateProvider` seems like something we'd use internally in the app, and not something intended for the UI previews.

bmarty:
@copilot resolve the merge conflicts in this pull request

bmarty:
> @bmarty what do you think? I'm in favor of this change, since `StateProvider` seems like something we'd use internally in the app, and not something intended for the UI previews.

Yes, we [agreed](https://github.com/element-hq/element-x-android/issues/4657#issuecomment-2886949321) to do this change.

hayaksi1:
@bmarty Done.

bmarty (approved):
Thanks

PR #7469, opened 2026-08-17
Use the right base URL for the ESS config/well known endpoints

jmartinesp:
## Content

- Rename `UrlContentFetcher` to `ClientUrlContentFetcher` and add a couple `server` and `homeserverUrl` properties to it, so anything using the fetcher can use any of those as base URLs. This is needed for a Pro feature.
- When we try to load a remote config from an endpoint and we don't find a valid value, store it as `NotFound(needsRefresh = false)` so we know we've already checked and there was nothing there. This should still trigger a request to fetch the value, but on background, not blocking. This was causing the session restoration to be slower than needed, since `ContentScannerUrlProvider` was trying to fetch the remote config.

FOSS part of https://github.com/element-hq/element-android-enterprise/pull/45

## Motivation and context

- Fix an issue where the incorrect ESS config URL was being created.
- Fix the delay when restoring a client.

## Tests

With Element Pro, check the right ESS config endpoint and alternatively the right well-known endpoints are checked.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.

### DIFF
--- a/features/enterprise/api/src/main/kotlin/io/element/android/features/enterprise/api/EnterpriseService.kt
+++ b/features/enterprise/api/src/main/kotlin/io/element/android/features/enterprise/api/EnterpriseService.kt
@@ -10,7 +10,7 @@ package io.element.android.features.enterprise.api
 
 import androidx.compose.ui.graphics.Color
 import io.element.android.compound.colors.SemanticColorsLightDark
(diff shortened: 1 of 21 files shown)


bmarty (approved):
LGTM, thanks!

Issue #7275, opened 2026-07-25
Clearing cache from developer options is too destructive


bxdxnn:
### Steps to reproduce

1. Clear cache from developer options

#### What did you expect?

Not much

#### What happened instead?
"Add account" screen until I've restarted the app
Rooms are missing and existing rooms show members name instead of room names

### Will you send logs?

Yes

### Will you submit a Pull Request?

No

bxdxnn:
Everything syncs again when I clear cache one more time from developer options.

Destinyg133:
> Everything syncs again when I clear cache one more time from developer options.

Im having same problem, i have to do it twice because first time doesnt work it just deletes everything and its stuck on empty EX

jmartinesp:
I *believe* #7472 should have fixed this. Please reopen if the issue still exists.

PR #7472, opened 2026-08-18
Don't restart the sync service if the client is shutting down

jmartinesp:
## Content

Add `val MatrixClient.isShuttingDown: Boolean` field to make the `SynchOrchestrator` ignore any changes to start/stop the sync.

## Motivation and context

The current behaviour was causing issues in the room list when clearing the cache, randomly:

1. The client would ask the SDK to clear the caches.
2. The SDK would stop the sync service, set the `pos` value of sliding sync to 0 so the HS returns all the required state the next time the client is used, then clear the caches.
3. However, between the sync service being stopped with the `pos=0` change and the caches being cleared, there was enough time for the `SyncOrchestrator` to decide the sync service was stopped for no good reason since there is network connectivity and the app is on foreground, so it restarted the sync, with `pos=0` (requesting all the state events) and performed a request immediately.
4. The new `pos` value was persisted, and the next time the new client did an 'initial sync' it had a `pos != 0`, which the HS uses to avoid returning some required state it already fetched.
5. The room list now lacks state events for lots of rooms and room names, avatars, members, etc. are now missing.

*Should* fix https://github.com/element-hq/customer-success/issues/820, as this was most likely caused by the user either clearing the cache manually or enabling/disabling the threads feature in labs.

For users affected by this, clearing the cache in a version with this patch should fix it.

## Tests

Clear the cache or toggle the threads feature as many times as you want. The initial sync should always fetch the room list items in the same order and contain the full info.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
@@ -101,7 +101,7 @@ class SyncOrchestrator(
         val isAppActiveFlow = combine(isAppActiveFlows) { actives -> actives.any { it } }
         combine(
             // small debounce to avoid spamming startSync when the state is changing quickly in case of error.
-            matrixClient.syncService.syncState.debounce(100.milliseconds),
+            syncService.syncState.debounce(100.milliseconds),
             networkMonitor.connectivity,
             isAppActiveFlow,
         ) { syncState, networkState, isAppActive ->
--- a/appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
@@ -16,7 +16,7 @@ import io.element.android.features.networkmonitor.api.NetworkMonitor
 import io.element.android.features.networkmonitor.api.NetworkStatus
 import io.element.android.libraries.core.coroutine.CoroutineDispatchers
 import io.element.android.libraries.core.coroutine.childScope
-import io.element.android.libraries.matrix.api.sync.SyncService
+import io.element.android.libraries.matrix.api.MatrixClient
 import io.element.android.libraries.matrix.api.sync.SyncState
 import io.element.android.services.analytics.api.AnalyticsService
 import io.element.android.services.analytics.api.recordTransaction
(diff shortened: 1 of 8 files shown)


appnav/src/main/kotlin/io/element/android/appnav/di/SyncOrchestrator.kt
```diff
@@ -94,15 +96,20 @@ class SyncOrchestrator(
...
             appForegroundStateService.isSyncingNotificationEvent,
             appForegroundStateService.hasRingingCall,
-            appForegroundStateService.isSharingLiveLocation
+            appForegroundStateService.isSharingLiveLocation,
         )
         val isAppActiveFlow = combine(isAppActiveFlows) { actives -> actives.any { it } }
         combine(
             // small debounce to avoid spamming startSync when the state is changing quickly in case of error.
-            syncService.syncState.debounce(100.milliseconds),
+            matrixClient.syncService.syncState.debounce(100.milliseconds),
```
bmarty: Could use `syncService` directly, no?


```suggestion
            syncService.syncState.debounce(100.milliseconds),
```

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomInfoMapper.kt
```diff
@@ -37,6 +38,9 @@ import org.matrix.rustcomponents.sdk.RoomPowerLevels as RustRoomPowerLevels
 
 class RoomInfoMapper {
     fun map(rustRoomInfo: RustRoomInfo): RoomInfo = rustRoomInfo.let {
+        if (it.id == "!Kk3Pee-WrV8kK3vFLJnyYj3xVT4Kvy3bMG3u7TLZxnA") {
+            Timber.d("Found !Kk3Pee-WrV8kK3vFLJnyYj3xVT4Kvy3bMG3u7TLZxnA")
+        }
```
bmarty: Was this added for test, or do you mean pushing this to nightly/prod?

bmarty (approved):
Thanks. I guess this could also fix #7275 ?

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomInfoMapper.kt
jmartinesp: Oops. I forgot that was there. Thanks.

jmartinesp: https://github.com/element-hq/element-x-android/pull/7472/commits/35ec93a431c7266e3fb9757bc97655a0f0d1af65

jmartinesp:
> Thanks. I guess this could also fix #7275 ?

I hope so, but I'm not sure, to be honest.

PR #7494, opened 2026-08-21
Move strings to the login module

bmarty:
Just a change in the Localazy config.json, so that the string of the new server selection screen are routed to the correct module.
Then import that Localazy strings and their already existing translation.


jmartinesp (approved):
Thanks!

PR #7495, opened 2026-08-21
Add `ClientEnterpriseHook` and `ClientBuilderEnterpriseHook`

jmartinesp:
## Content

- Use the new hooks to not expose the enterprise internal when setting up custom configs for both `ClientBuilder` and `Client`.
- Remove `ContentScannerUrlProvider` since it's no longer needed.
- Simplify the DI for several enterprise components, by using `RemoteEnterpriseConfigProvider.Origin` to provide the needed dependencies, depending on the logic we want to use to fetch the remote config data.
- Add a dummy `MatrixClientBuilder` interface to be able to expose the SDK's `ClientBuilder` outside of the `:libraries:matrix:impl` module.

## Motivation and context

FOSS part of https://github.com/element-hq/element-android-enterprise/pull/46.

## Tests

Check the ESS config is still fetched or used from the cache, as before.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.


libraries/matrix/impl/build.gradle.kts
```diff
@@ -21,12 +21,12 @@ android {
...
 dependencies {
-    releaseImplementation(libs.matrix.sdk)
+    releaseApi(libs.matrix.sdk)
     if (file("${rootDir.path}/libraries/rustsdk/matrix-rust-sdk.aar").exists()) {
         println("\nNote: Using local binary of the Rust SDK.\n")
-        debugImplementation(projects.libraries.rustsdk)
+        debugApi(projects.libraries.rustsdk)
     } else {
-        debugImplementation(libs.matrix.sdk)
+        debugApi(libs.matrix.sdk)
```
bmarty: Not sure we want this change. It will expose the API of the FFI layer. Is it mandatory? Maybe another solution would be that private modules add a ref to the rust library when necessary.

bmarty (approved):
One remark about dependencies, else, nice to see less pro code there!

libraries/matrix/impl/build.gradle.kts
jmartinesp: It only does so for modules that implement `:libraries:matrix:impl`, not all of them AFAICT. I don't see another way to both expose the SDK classes in the enterprise modules so we can modify them there and be able to use a custom build like we do in `:libraries:matrix:impl`... unless we duplicate the same logic of checking if the library AAR is present?

bmarty: Yes, duplicating the logic is better I think. I'll do it

PR #7500, opened 2026-08-21
Let users enter their full Matrix user ID on the account provider screen

jennaharris7:
## Content

Allow entering a full Matrix user ID (e.g. `@alice:example.org`) in the account provider field on the sign-in screen. When a user ID is entered, we sign in to its homeserver (the domain part, preserving any port) and pass the ID to the authentication server as a login hint (`mxid:@alice:example.org`) so it can pre-fill the user's identity. A plain account provider (e.g. `matrix.org`) continues to work exactly as before.

## Motivation and context

Re-adds the "type your full Matrix ID" capability from #7375 (which was reverted), in a privacy-preserving form. Only the **homeserver** is ever persisted to the account-provider history — never the user ID — so no identifying username is written to app-global preferences before authentication.

Notes / known limitations:
- The hint pre-fills the identity on the authentication (MAS) page but does not override an already-active MAS session — if a different user is already signed in there, the server still shows that session.

## Tests

New presenter tests in `ConfirmAccountProviderPresenterTest`:
- a full Matrix ID resolves to its homeserver and passes the `mxid:` login hint,
- a Matrix ID with a port keeps the port in the homeserver URL,
- a plain account provider passes no login hint.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 16

## Checklist

    - [ ] No.

### DIFF
--- a/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/confirmaccountprovider/ConfirmAccountProviderPresenter.kt
+++ b/features/login/impl/src/main/kotlin/io/element/android/features/login/impl/screens/confirmaccountprovider/ConfirmAccountProviderPresenter.kt
@@ -30,6 +30,7 @@ import io.element.android.features.login.impl.login.LoginModeEvent
 import io.element.android.features.login.impl.login.LoginModeState
 import io.element.android.libraries.architecture.Presenter
 import io.element.android.libraries.core.uri.ensureProtocol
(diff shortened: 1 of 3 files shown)


bmarty:
Thanks, will review. FTR the MAS limitation about ignoring hint when another account is already logged in is tracked by https://github.com/element-hq/matrix-authentication-service/issues/5525.

bmarty (approved):
LGTM, thanks!

PR #7501, opened 2026-08-21
Update Enterprise Copyright in idea settings, so that new files get an expected header.

bmarty:
What the title says.


jmartinesp (approved):
Thanks!

PR #7510, opened 2026-08-22
Fix crash at startup after upgrading from an older version (run database migrations on upgrade)

### LINKED ISSUE #7509: Crash at startup after upgrading from an older version: SQLDelight migrations never run (`SqlCipherDriverFactory.onUpgrade` does not call `super`)
### Steps to reproduce

1. Install an older build whose session database predates migrations 9–11 (e.g. the 25.05.4 release APK).
2. Launch it once (the session database is created at schema level 9). Signing in is not required.
3. Install a current build over it (reproduced with builds of v26.08.2 and of today's `develop` — `libraries/encrypted-db` is unmodified in our fork).
4. Launch.

### Outcome

**What did you expect?**
The session database is migrated (migrations 9, 10, 11) and the app starts.

**What happened instead?**
Crash before the login screen, 100 % reproducible:
```
android.database.sqlite.SQLiteException: no such column: SessionData.position (code 1): , while compiling: SELECT SessionData.userId, ... SessionData.position, SessionData.lastUsageIndex, ... FROM SessionData ORDER BY lastUsageIndex DESC
    at net.zetetic.database.sqlcipher.SQLiteConnection.nativePrepareStatement(Native Method)
    ...
    at app.cash.sqldelight.driver.android.AndroidSqliteDriver.executeQuery
    at io.element.android.features.migration.impl.migrations.AppMigration02.migrate
    at io.element.android.features.migration.impl.MigrationPresenter$present$1$1.invokeSuspend
```

**Cause.** `libraries/encrypted-db/src/main/kotlin/io/element/encrypteddb/SqlCipherDriverFactory.kt` overrides `AndroidSqliteDriver.Callback.onUpgrade` to invoke the re-key callback (since #6774) but never calls `super.onUpgrade(db, oldVersion, newVersion)`, which is what runs `schema.migrate()`. So no `.sqm` migration is ever executed on upgrade, while SQLite's `user_version` is still bumped afterwards — the database then *claims* to be current with the old schema, so a later fixed build does not retry the migration either. Users updating monthly rarely hit this (they had migrations 9/10 applied by earlier callback implementations), but anyone skipping a few releases does, and every future migration will crash upgrading users.

**Fix.** Call `super.onUpgrade(db, oldVersion, newVersion)` before the re-key callback. Verified on device: 25.05.4 → patched build starts and migrates correctly. PR follows.

### Your phone model

Pixel 8 Pro

### Operating system version

Android 17

Fork build of v26.08.2 (jeheja/feral-android, `libraries/encrypted-db` unmodified); also present in develop @ 477aa36640

### Homeserver

feralisme.fr (Synapse)

### Will you send logs?

No (stack trace above)

Yes

jeheja:
## Content

`SqlCipherDriverFactory`'s `AndroidSqliteDriver.Callback.onUpgrade` override now calls `super.onUpgrade(db, oldVersion, newVersion)` before the re-key callback, so the SQLDelight `.sqm` migrations actually run on upgrade (the base implementation is what executes `schema.migrate()`).

## Motivation and context

Fixes #7509: upgrading from a build whose session database predates migrations 9–11 (e.g. 25.05.4) crashed at startup with `SQLiteException: no such column: SessionData.position` (in `AppMigration02`), because no migration was executed while `user_version` was still bumped. One-line fix; no user-facing change other than the crash disappearing.

Note for maintainers: databases already opened by an affected build have `user_version` stamped at the current schema with the old columns, so this fix alone does not repair them; a self-heal that inspects `PRAGMA table_info(SessionData)` and applies the missing migrations is what we ship in our fork — happy to contribute it as a follow-up if wanted.

## Screenshots / GIFs

No UI change.

## Tests

- Installed the official 25.05.4 release APK, launched it once (creates the v9 session database).
- Installed a release build with this patch over it: app starts, migrations applied, relaunch healthy.
- Same steps without the patch: crash 100 % (stack trace in #7509).
- `./gradlew :libraries:session-storage:impl:verifySqlDelightMigration` passes.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 17 (Pixel 8 Pro)

## Checklist

    - [ ] No.
- [ ] Changes have been tested on an Android device or Android emulator with API 24 (tested on API 37)
- [ ] UI change has been tested on both light and dark themes (no UI change)
- [ ] Accessibility has been taken into account (no UI change)
- [ ] Pull request includes screenshots or videos if containing UI changes (none)

Suggested label: `PR-Bugfix` (I cannot add labels on this repository).

### DIFF
--- a/libraries/encrypted-db/src/main/kotlin/io/element/encrypteddb/SqlCipherDriverFactory.kt
+++ b/libraries/encrypted-db/src/main/kotlin/io/element/encrypteddb/SqlCipherDriverFactory.kt
@@ -55,6 +55,9 @@ class SqlCipherDriverFactory(
                 schema
             ) {
             override fun onUpgrade(db: SupportSQLiteDatabase, oldVersion: Int, newVersion: Int) {
+                // The base implementation runs the SQLDelight migrations (schema.migrate); without it
+                // the database keeps its old schema after an upgrade and the next query crashes.
+                super.onUpgrade(db, oldVersion, newVersion)
                 onUpgradeCallback?.invoke(db, oldVersion, newVersion)
             }
         })


bmarty (changes requested):
Thanks. I checked the fix by adding a new migration to one of our database. Without the fix, the application crashes, and with the fix the migration is performed and no crash.

Can you sign the CLA please? Without that we cannot merge the PR.

bmarty:
Closed in favor of #7570.

PR #7511, opened 2026-08-23
Index older room history for message search, with a Start indexing control in developer settings

hayaksi1:
## Content

The local search index is attached at `ClientBuilder` time and is written only as a side effect of
events passing through the SDK's event cache, so it holds exactly what arrived after it was created.
Nothing in the app ever asks the SDK for anything older on the index's behalf. A user who enables
message search on an existing account therefore searches an index that starts at the moment they
enabled it, gets nothing back for every message they actually remember, and reasonably concludes
that search does not work.

This adds a bounded background sweep that back-paginates rooms so their history reaches the index,
and a developer-settings control to run that sweep on demand and watch it. The sweep plans a queue
from the room-list summaries (most recently active first, low-priority rooms last, capped at 200
rooms), walks each room from its live end, and persists its position in a durable cursor so process
death costs at most one room of progress. Every execution is bounded in pages and rooms rather than
events — 200 pages per room, 300 per execution, a five-minute deadline, 45 seconds per room —
because `paginate()` reports only whether it reached the start of a room. In the background it runs
under `NetworkType.UNMETERED` with battery-not-low and storage-not-low, deliberately not expedited,
and is cancelled by tag on logout.

Those constraints are right for a download nobody asked for, but they also mean there is frequently
no moment on a real device where all three hold at once, and nothing reports whether the sweep is
waiting, executing, or long since finished. Developer options therefore gains a "Message search
index" section, rendered only while the flag is on: Start indexing when idle, Resume indexing naming
the room the previous sweep stopped at, a progress bar with Cancel while running, and a summary of
rooms swept and pages fetched when the queue drains. A user-initiated run keeps the same unique work
name as the background sweep, so the one-sweep-per-session invariant and the single cursor still
hold; it differs only in what an explicit tap justifies — `ExistingWorkPolicy.REPLACE`,
`NetworkType.CONNECTED`, linear backoff at 30 seconds, and promotion to a `dataSync` foreground
service with a progress notification. The larger 1000-page/30-minute budget applies only when that
promotion succeeds; an unpromoted run keeps the conservative budget and retries.

A finished sweep is deliberately never reported as "your history is searchable". A room whose
history is already in the local event-cache store re-hydrates from disk on pagination and the SDK
drops those updates before the indexer sees them, while `paginate()` returns the same success as a
network fetch that indexed everything. `SearchBackfillStatus` has no `Complete` member and
`RoomSweepOutcome.REACHED_START` means "there was nothing left to ask for" — the UI says swept and
fetched, never findable.

This adds `android.permission.FOREGROUND_SERVICE_DATA_SYNC` to the merged manifest, which is
app-wide and appears on store listings. A service type can only be set by merging onto the
`SystemForegroundService` androidx.work already declares, so it cannot be scoped to this worker
alone. That is a permission-surface change for a capability no shipping user can reach, and it is
the part of this most deserving of an explicit decision. With `feature.message_search` off nothing
is scheduled, no pagination is issued, the section is not rendered, and every other WorkManager
request type keeps its existing tag, constraints and cancellation behaviour.

## Motivation and context

In-room search has been asked for repeatedly, and the reports are specific: search is present in
unencrypted rooms and absent in encrypted ones. This is one layer underneath that surface — it makes
older messages reachable by the index at all, and makes the indexing observable while the feature is
being built — so it closes nothing on its own. The sweep is expected to be superseded by
matrix-org/matrix-rust-sdk#6805, and the developer-settings section is meant to be deleted along
with it rather than to become the product's indexing trigger.

Part of element-hq/element-x-android#3709
Relates to element-hq/element-x-android#6557
Relates to element-hq/element-x-android#4149
Relates to element-hq/element-x-android#7072
Relates to element-hq/element-meta#3252

## Screenshots / GIFs

|State|Light|Dark|
|-|-|-|
|Flag on, index not attached yet|||
|Idle|||
|Waiting for a connection|||
|Running, still planning the queue|||
|Running, room 3 of 9|||
|Paused mid-queue|||
|Finished|||

## Tests

- `SearchBackfillRunnerTest` — 15 cases over the loop shape: pagination stops at the start of a room
  and is never attempted when it cannot paginate; the per-room page cap, per-room time limit,
  execution page budget and execution deadline each stop the sweep and are recorded; a repeatedly
  failing room is abandoned and the sweep continues; an unjoined room is skipped; the room is always
  released; a stored cursor resumes without revisiting; a drained cursor starts a new generation.
- `SearchBackfillPlannerTest` — 8 cases, mostly negative: ordering by recency, low-priority rooms to
  the tail, exclusion of spaces, tombstoned rooms, unjoined rooms and rooms with no latest event.
- `SearchBackfillRequestBuilderTest` — 4 Robolectric cases over both request shapes: unique per
  session, tagged for logout cancellation, constrained to free network; and the user-initiated shape
  keeping the unique name while switching to REPLACE with both tags and relaxed constraints.
- `DefaultMessageSearchIndexerTest` — 5 Robolectric cases, including the one the second tag exists
  for: the background sweep's permanently-enqueued shared tag must not read as "waiting".
- `DefaultWorkManagerSchedulerTest` — every request type is cancelled when a session is deleted, so
  a future type that leaks background work after logout fails here.
- `DeveloperSettingsPresenterTest` — 5 cases over the status derivation and the start/cancel events.

What these do not prove: nothing exercises the chain from `paginate()` through the event cache into
tantivy — there is no app-level observation point and no index API over FFI — and none provokes a
real foreground-service promotion, which the platform decides at runtime. Both were checked on
device: a closed room produced 108 index writes carrying send dates up to three weeks old where none
arrived before, and a fresh login swept 9 of 9 rooms over 112 pages and made messages from 23 days
before the session existed searchable.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Android 16 (API 36) emulator (x86_64)

## Checklist

    - [ ] No.


jmartinesp:
Thanks for the contribution, it's quite impressive, but there is already some ongoing work on the SDK to do all the heavy lifting of indexing rooms automatically there. Once that's merged we'll have it either running automatically while the app is open or running manually using background tasks with `WorkManager`, and that part of the code could be interesting for us to keep.

Issue #2056, opened 2023-12-19
Latest release shows dev options independently of the "developer mode" toggle


bnjbvr:
### Steps to reproduce

Download latest release of ElementX Android on google play.

### Outcome

Developer options are always shown, although I haven't toggled on the "developer mode" option in the advanced settings. In fact, it's always shown, whether the toggle is on or off.

### Will you send logs?

No

### Are you willing to provide a PR?

No

bmarty:
Are you sure the version you're using is coming from the PlayStore?

`Developer options` are shown on `Debug` and `Nightly` builds, not `Release` builds as per [this code](https://github.com/element-hq/element-x-android/blob/main/features/preferences/impl/src/main/kotlin/io/element/android/features/preferences/impl/root/PreferencesRootPresenter.kt#L95), there is not link with `Advanced settings` / `Developer mode` (which has been renamed to `View source`) switch.

I have checked with the version from the PlayStore and `Developer options` is not shown.

bnjbvr:
Oops you're right, my bad it's a nightly, and quite an old one.

PR #7513, opened 2026-08-24
Keep the Developer options entry in Settings until it is switched off

hayaksi1:
## Content

Unlocking the developer options with seven taps on the version number only flipped an in-memory flag on `ShowDeveloperSettingsProvider`. That provider is injected per presenter, so a new one was built every time the settings screen was entered: the entry disappeared again as soon as the user navigated away, and always after an app restart.

The unlock is now the `showDeveloperSettings` preference in `AppPreferencesStore`, defaulting to `true` off release builds and `false` on them, so it survives navigation and process death. Writes go through the app coroutine scope so leaving the screen cannot cancel them.

The Developer options screen opens with an On/Off switch, the same shape as the platform's own Developer options screen. While it is on the entry stays in Settings; turning it off removes the entry, and the row then says how to bring it back.

`MultipleTapToUnlock` deliberately stays unlocked once it has fired and `MultipleTapToUnlockTest` asserts that, which was invisible while nothing could re-lock the entry. The version string's click handler is re-enabled the moment developer options go hidden, so a spent counter let a single tap undo the switch. The provider now takes a fresh counter after each successful unlock.

The unrelated `developerMode` preference behind Advanced settings, "View source", is untouched: it still gates event source, debug info and `/crash`, and it still has no bearing on this entry.

## Motivation and context

Seven taps on the version number is the advertised way to reach developer options, but the result did not stick, so the sequence had to be repeated on every visit. Relates to https://github.com/element-hq/element-x-android/issues/2056, where the same entry was reported as behaving independently of the Advanced settings toggle.

## Screenshots / GIFs

|  | On | Off |
|-|-|-|
| Light |  |  |
| Dark |  |  |

## Tests

- `DefaultAppPreferencesStoreTest` — the preference defaults to `true` off release builds, `false` on release builds, and persists an update.
- `ShowDeveloperSettingsProviderTest` — an unlock, and a switch-off, are both still in effect for a provider constructed afterwards; six taps do not unlock; a single tap after a switch-off does not bring the entry back. Each assertion is made against a provider whose build type gives the opposite initial value, so a pass proves the stored preference won rather than the default.
- `PreferencesRootPresenterTest` — the entry is hidden while the preference is off, and appears after seven taps on the version.
- `DeveloperSettingsPresenterTest` — switching off writes the preference.
- `DeveloperSettingsViewTest` — the switch emits `SetShowDeveloperSettings`.

These do not prove the value survives a real process restart; that needs a device.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty (approved):
Why not, thanks!

bmarty:
@copilot resolve the merge conflicts in this pull request

hayaksi1:
@bmarty Done.

PR #7519, opened 2026-08-24
Remove the reference to the decommissioned OIDC playground

sandhose:
## Content

Drop reference to https://github.com/element-hq/oidc-playground in `docs/oauth.md`.

## Motivation and context

The OIDC playground is being decommissioned, so that link no longer points at anything usable. Arguably that documentation could be reworked quite a bit, but at this point I'm mostly looking into scraping any references to the OIDC playground, and redirect people to just use matrix.org for example.

See https://github.com/element-hq/oidc-playground/pull/6

Reason for Claude co-authoring the commit is because I've asked it to sweep for references and it prepared a bunch of patches for me, including this one.


bmarty (approved):
Thanks!

Issue #1874, opened 2023-11-23
Certain types of images does not render on chat room


olmari:
### Steps to reproduce

1. Receive an "telegram custom emoji" in an room where telegram bridge and thus telegram users exist.
2. Observe blank message in EXA, whereas old EA and Element-desktop shows something.

#### What did you expect?
As this "telegram custom emoji" is after all an html image tag in matrix-perspective, I expect to see the image, or at least the "alternative" normal emoji the hinting also provides with (unformatted) body and img alt tag.

#### What happened instead?
Seeing blank message.

Exploration from data, for the message screenshots there are 2 messages, firs ont is reply to an image in an room, the very next one is separate "normal" message through bridge.

First an screenshot from Element Desktop, somewhat the "this is how it should look":
![1](https://github.com/vector-im/element-x-android/assets/5422161/0174fd9d-0f86-4411-8f82-99a949c598a8)

Then screenshot from EXA:
![2](https://github.com/vector-im/element-x-android/assets/5422161/4515e40b-6ea8-4be7-b4da-95a53c91c8fa)

Also as an mention, EXAs orooms list that shows latest message preview which was the then latest "telegram custom emoji", this shows the "alternate" picture / emoji message also provides:
![3](https://github.com/vector-im/element-x-android/assets/5422161/b28808ca-4637-4d08-b297-3e1b5efe6566)

Bonus mention, no screenshot:
EA shows the "alt" emojis in the room, which is at least something, but generally the actual provided images should be shown in EXA too like they do in Element Desktop.

Both events with some private data redacted:

```json
{
  "content": {
    "body": "> <geezetatee> an image\n\n😜",
    "external_url": "hps//t.me/c/<number>",
    "fi.mau.telegram.source": {
      "chat_id": <number>,
      "id": <number>,
      "peer_type": "channel",
      "space": <number>
    },
    "format": "org.matrix.custom.html",
[...]

PR #7523, opened 2026-08-25
Fall back to the message body when the formatted body has nothing to render

hayaksi1:
## Content

`FormattedBody.toHtmlDocument()` sanitises the HTML with a jsoup `Safelist` that does not include
`img`, so a formatted body whose only content is an `` — a custom emoji, a bot posting an
image inline — cleans down to an empty document. The document is still non-null, so every consumer
that already has an `?: body` fallback never takes it: the timeline renders an empty bubble, the
room list preview and the notification show an empty string, and `plainText` gives TalkBack nothing
to read.

`toHtmlDocument()` now returns `null` when nothing renderable survives the sanitisation, which is
what the callers already expect from a body they cannot parse. The timeline then falls back to the
plain text body, which for these messages is the `alt` text the sender provided.

Rendering `` inline is a separate, larger feature and is not attempted here.

## Motivation and context

Part of #1874.

## Tests

`libraries/matrixui/.../ToHtmlDocumentTest.kt`: a body containing only an `` produces `null`;
a body where text survives next to the `` still produces a document.

`libraries/matrixui/.../ToPlainTextTest.kt`: `TextMessageType.toPlainText()` returns the body for an
image-only formatted body, which is the room list preview and notification path.

`features/messages/impl/.../TimelineItemContentMessageFactoryTest.kt`: the timeline item for such a
message carries the body as its formatted body and a null html document.

Run with `./gradlew :libraries:matrixui:testDebugUnitTest :features:messages:impl:testDebugUnitTest`.

These are unit tests over the parser and the item factory, so they do not prove how the sanitised
HTML looks once the rich text editor renders it on screen.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty (approved):
Make sense, thanks.

Issue #262, opened 2023-03-27
Recomposition & State issues - nothing gets `remember { ... }`'ed


theIDinside:
### Steps to reproduce

Preface: I don't know much Android-development at all and I have minimal Kotlin experience.

# Irrelevant apropo
I ran into this issue, because I'm starting to implement a rudimentary syntax highlighting engine, similar to the one we can see on the normal web app.

The syntax highlighting stuff is trivial (and I do think the android app should mimick the browser/desktop version), whether one should handroll per language, or do some regex stuff - either way, reducing 3rd party dependencies is always a good thing. Always. 

The thing is though, re-running syntax highlighting on a message should only be done once. So how would I go about doing this? Using `remember { ... }` right? Which brings me to:

### Outcome

# Actual issue
Every time recomposition happens, the `remember { }` stuff never actually remembers anything. This got me thinking; I have very little experience in the JVM/Android platform, so I might not be doing it right, at all.

But then I set some breakpoints, and it seems, this functionality is not working as expected _anywhere_.

For instance, in the [`TimelineItemTextView`](https://github.com/vector-im/element-x-android/blob/develop/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/components/event/TimelineItemTextView.kt#L63) composable, there's a `remember(content.body)` - and it gets called _every time_  recomposition happens, making it completely unnecessary to have it in a remember block, no?

In other words, the data never changes, yet all these styles need to be applied every time, doing costly operations, over, and over, and over again.

Now, I'm not well-versed in Compose (or Kotlin for that matter) but this seems to be a pretty huge design flaw issue? The client has the data, yet it has to go through all the hoops of parsing it into a document etc applying styles, etc.

The culprit seems to be [`TimelineView`](https://github.com/vector-im/element-x-android/blob/develop/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineView.kt#L105-L110) which refires every recomposition and thus forces _everything_ below it to update as well.

Now, it makes sense that recomposition should happen, when scrolling, but every subsequent child in that "`Timeline` " has their operations applied fully, not just their positions or layout in the UI, but as previously show, the `linkify` stuff as well (and in my case - the reparsing of the syntax + applying colors on "code blocks"), the traversing of already parsed html, etc. The data of all the subcomponents should only be calculated once, it seems to me, only the UI should care.

Is this the intended design? What can be done about it? If anyone has ideas on how to solve it?

[...]

PR #7531, opened 2026-08-25
Stop rescanning message bodies for emoji only on every recomposition

hayaksi1:
## Content

Every text message in the timeline decides whether it is emoji-only by comparing its formatted body
to its plain body and then running a grapheme scan over the whole message. None of that is
remembered, so it runs again on every recomposition of every visible message, allocating two strings
and building a `GraphemeMatcher` each time — for a screen full of long messages that is the most
expensive thing in the row that does not need to happen twice.

The decision is now remembered against the body and the formatted body, so it is computed once per
message and recomputed only when the message itself changes. The preview branch stays where it was,
inside the same computation, so previews still render the emoji-only style. `containsOnlyEmojis` had
a single caller and existed only to read `LocalInspectionMode`, so its composable wrapper is gone and
the plain function keeps the name.

## Motivation and context

Part of #262, which asks for the costly per-message work in this file to stop being redone on every
recomposition. The rest of that issue is stale: the timeline list has been keyed and content-typed
since it was filed.

## Tests

`features/messages/impl/.../utils/EmojiTest.kt` covers the detection itself and is unchanged apart
from the name.

The `TimelineItemTextView` screenshots, which include the emoji-only state, still match byte for
byte, which is what proves the behaviour did not move.

Run with `./gradlew :features:messages:impl:testDebugUnitTest` and
`./gradlew :tests:uitests:verifyPaparazziDebug --tests '*TimelineItemText*'`.

No test counts recompositions, so the saving itself is argued from the code, not measured.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


jmartinesp (approved):
Thanks for the optimization.

Issue #3434, opened 2024-09-09
Better sending of mentions support


erikjohnston:
There are two annoyances when trying to mention someone:

1. Clicking their name on the timeline takes you to their member info, rather than inserting a mention like in the legacy apps (and the legacy apps have a "mention" menu item in the member info)
2. When you type `@...` it displays in alphabetical order(?), which can display really old unused accounts before recently active accounts (e.g `@matthew:ara.....` vs `@matthew:matrix....`)

Aquathing:
> Clicking their name on the timeline takes you to their member info, rather than inserting a mention like in the legacy apps (and the legacy apps have a "mention" menu item in the member info)

I would keep this behaviour, it's more natural. 
What about implement something like telegram's long press, where if u long press on somebody's profile picture it'll show u a list of options? 
![image](https://github.com/user-attachments/assets/7d5df1a6-c7df-43d5-b10c-c3300aae1dd1)

PR #7536, opened 2026-08-25
Rank the mention suggestions by how well they match what was typed

hayaksi1:
## Content

The member suggestions offered after `@` were emitted in whatever order the room member list
happened to be in, filtered but never ranked, so a member who matches only in the middle of their
name can be offered above one whose name starts with what was typed.

Matches are now ordered by how well they match the query: display name prefix first, then user ID
localpart prefix, then any other match. The sort is stable, so members of equal rank keep the order
the member list gave them, and the batch cap of 100 members is applied before the sort as it was
before, so nothing new runs over the whole member list.

The `@room` suggestion was compared case sensitively, so typing `@Ro` or `@ROOM` dropped it while
`@ro` kept it. It now matches the query the same way the members do.

## Motivation and context

Part of #3434. This covers the second half of that issue, the ordering of the suggestions. The
first half, inserting a mention from the member info screen, is a separate change and one commenter
on the issue disputes the behaviour it asks for, so it is left alone here.

Ordering by relevance is not ordering by recent activity: a `RoomMember` carries no activity
information, so the older of two equally matching accounts can still come first.

## Tests

`features/messages/impl/.../suggestions/SuggestionsProcessorTest.kt` gains two tests: one that the
three kinds of match come back in rank order, one that an uppercase query still offers `@room`.

Both fail on develop — the ordering one returns the members in input order, the `@room` one returns
an empty list.

Run with `./gradlew :features:messages:impl:testDebugUnitTest --tests '*SuggestionsProcessorTest*'`.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/suggestions/SuggestionsProcessor.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/messagecomposer/suggestions/SuggestionsProcessor.kt
@@ -115,13 +115,26 @@ class SuggestionsProcessor(
                 .filterUpTo(MAX_BATCH_ITEMS) { member ->
                     isJoinedMemberAndNotSelf(member) && memberMatchesQuery(member, query)
                 }
+                .sortedBy { member -> matchRank(member, query) }
                 .map(ResolvedSuggestion::Member)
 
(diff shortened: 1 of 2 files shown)


bmarty (approved):
LGTM, thanks!

PR #7537, opened 2026-08-25
Search the rooms of a space

### LINKED ISSUE #6566: Search rooms inside a space ("explore rooms")
### Your use case

Note: Im not a programmer, just someone who would like a feature, please

Element classic had the functionality to search a room inside a space called "explore rooms".

We are using matrix as university internal communication. As there are many lectures, we have 400+ rooms inside a space which is impossible to go through by hand.

This is the large feature we need, to be able tu use element X on a large scale. So it is somethin we desperately need!

By implementing it again :)

Maybe automatic alphabetic ordering of rooms when viewing the space-overview?

### Additional context

Dont know what a PR is, so I clicked "no"

No

hayaksi1:
## Content

A space with hundreds of children could only be explored by scrolling: the space screen had no way
to look for a room by name. This adds a search action to the space top bar which swaps the screen
into a search mode, built from the `SearchBar` the "add rooms to a space" screen already uses.

Children are matched on their name and on their canonical alias, case insensitively. The space room
list is drained in the background by `loadAllIncrementally`, so the search covers the whole
hierarchy rather than only the page currently on screen. The rows in the results are the same
`SpaceRoomItemView` rows as the list itself, so joining and answering an invite work from a search
result too.

## Motivation and context

Closes #6566. The reporter has 400+ rooms in a university space and cannot find one by hand; the
only workaround so far is the room directory, which does not know about the space.

The `SpaceRoomList` the SDK exposes has no server side search — `paginate`, `reset` and `rooms` are
all it offers — so the matching is client side, over the children loaded so far.

## Tests

`features/space/impl/.../root/SpaceSearchPresenterTest.kt` covers the three states: a query matching
on a name and on an alias, a query matching nothing, and leaving the search clearing the query.

All three fail on develop with the matching neutralised, which is what shows they guard the
filtering rather than the plumbing.

Run with `./gradlew :features:space:impl:testDebugUnitTest`.

The screenshot test of the search mode comes from the two new `SpaceStatePreviewParam` states.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bxdxnn:
See https://github.com/element-hq/element-x-android/pull/6567#issuecomment-4808917298

Issue #1612, opened 2023-10-20
No reason can be given when deleting the message


SandResidence:
#### What would you like to do?
The protocol supports the reason for deleting the message. The application does not ask for a reason, it just deletes the message.

#### Why would you like to do it?
Because the specification allows it.

#### How would you like to achieve it?
Idk

### Have you considered any alternatives?

No

### Additional context

The reason for deletion is also not visible in the message bubble, but available in JSON as `redacted_because`.

### Are you willing to provide a PR?

No

PR #7538, opened 2026-08-25
Let the user give a reason when removing a message

hayaksi1:
## Content

A redaction carries a reason in the protocol and other Matrix clients let the sender give one, but
here removing a message sent the redaction straight away, with no reason and no confirmation.

Removing a message now opens a dialog with an optional reason, and the redaction is only sent once
it is confirmed. A blank reason is sent as none, so confirming without typing anything behaves
exactly as it did before. A message which was never sent has nobody to give a reason to, so the
Remove button of the send failure dialog still redacts it immediately.

Showing the reason of a redacted message in the timeline, the other half of the issue, is not part
of this change.

## Motivation and context

Part of #1612, which is labelled `help wanted`. The dialog reuses the existing `TextFieldDialog` and
the existing `common_reason` and `action_remove_message` strings, so there is nothing new to
translate.

## Tests

`features/messages/impl/.../MessagesPresenterTest.kt` covers the four paths: confirming with a
reason sends it, a blank reason is sent as `null`, cancelling sends no redaction at all, and a
message which was never sent is redacted without a dialog.

The first three fail on develop, where the action redacts immediately.

Run with `./gradlew :features:messages:impl:testDebugUnitTest --tests '*MessagesPresenterTest*'`.

The dialog itself is covered by a new `MessagesStatePreviewParam` state.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesState.kt
```diff
@@ -61,6 +62,8 @@ data class MessagesState(
     val successorRoom: SuccessorRoom?,
     val threads: Threads,
     val showLiveLocationShareBanner: Boolean,
+    /** The event the user asked to remove, waiting for them to confirm and optionally give a reason. */
+    val eventToRedact: TimelineItem.Event?,
```
bmarty: Could be `redactEventAction: AsyncAction<Unit>` instead for code consistency, with a new state

```
data class ConfirmingRedaction(
    val eventId: EventId
) : AsyncAction.Confirming
``` 

even if in this particular case all the states of the `AsyncAction` will not be rendered.

bmarty (changes requested):
I've discussed with the team and here is what we want:

Image

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/MessagesState.kt
hayaksi1: Done!

hayaksi1:
@bmarty Done.

bmarty (approved):
Thanks for the update!

bmarty (approved):
Thanks, merging in my branch so I can update the wording

Issue #5356, opened 2025-09-15
Entire username shown for every user even if no username is duplicated


reazndev:
### Steps to reproduce

1. Where are you starting? What can you see?

Regular room with multiple people. 

No one has the same display name.

#### What did you expect?

For each user only the display name should be shown.

#### What happened instead?

The entire handle (username:matrix.org) is shown for every single user even though there are no duplicate display names.

### Your phone model

Google Pixel 9a

### Operating system version

Android 16

### Application version and app store

Element X version: 25.09.0

### Homeserver

hilfikers.com

### Will you send logs?

No

### Are you willing to provide a PR?

No

PR #7539, opened 2026-08-25
Name a sender the timeline has no profile for from the room member list

hayaksi1:
## Content

A timeline event carries whatever profile the SDK resolved for its sender, and when that is
`Unavailable`, `Pending` or an error the row falls back to the raw Matrix ID — for the name, the
avatar letter and the reply preview alike. The room member list the same screen already holds
usually knows that user, so the ID is shown even though the display name is available.

The timeline item factory now falls back to the matching room member when the sender profile is not
resolved. It does so both when an item is first built and in the update path that runs when the
member list arrives afterwards, so an event rendered before the members loaded settles on the
display name rather than staying on the ID. A sender the member list does not know keeps the
profile the timeline gave, so nothing changes for genuinely unknown users.

## Motivation and context

Part of #5356. What this does not change is the disambiguation itself: a profile the SDK reports as
`Ready` with `displayNameAmbiguous = true` still renders the ID next to the name, since that flag is
a passthrough of the SDK value.

## Tests

`features/messages/impl/.../factories/TimelineItemsFactoryTest.kt` gains two tests: a sender with an
unavailable profile is named from the member list, and a sender the member list does not know keeps
the unresolved profile.

The first fails on develop, where the profile stays `Unavailable` and the avatar keeps the user ID.

Run with `./gradlew :features:messages:impl:testDebugUnitTest --tests '*TimelineItemsFactoryTest*'`.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


bmarty (approved):
LGTM, thanks!

bmarty:
@hayaksi1 conflicts need to be fixed on the test.

hayaksi1:
@bmarty Done.

PR #7541, opened 2026-08-25
Add a setting to leave the rooms of a Space out of the room list

### LINKED ISSUE #6857: Option to hide Space sub-groups on the home screen

Add an option to hide sub-groups/rooms inside Spaces from the default home screen view, so only top-level Spaces are shown until a Space is opened.

Currently, all groups/rooms inside Spaces are visible directly on the home screen, which can become cluttered when using many Spaces.

This would provide:

* A cleaner and less overwhelming UI
* Easier navigation in large setups
* Better focus on the currently used Space
* A more compact and organized home screen

Add a setting or toggle that collapses Space contents by default and only shows the rooms/sub-groups after opening the Space.

_No response_

### Additional context

_No response_

No

hayaksi1:
## Content

Every room a Space contains is listed on the home screen next to the rooms which belong to no Space,
so joining a large Space floods the room list.

Advanced settings gains a switch, off by default, which leaves those rooms out of the room list.
They stay reachable from the Spaces tab, and picking a Space filter still shows them, since choosing
a Space is an explicit request to see its rooms.

The room list filters the SDK offers have no negation and nothing for "not in a Space", so the
exclusion is applied to the emitted summaries, using the `descendants` the space filters already
carry. That means it only covers the Spaces the user has joined: a room in a Space they are not in
stays listed.

## Motivation and context

Closes #6857. The setting and its placement were agreed on the issue: @mxandreas — "I think it is
probably safe to try out this setting" — and @americanrefugee — "Yes, I agree to adding a setting.
My recommendation is to add the option to Advanced Settings between the first block of options and
Moderation and Safety." That is where the switch is.

The two new strings are in `temporary.xml` per AGENTS.md.

## Tests

`features/home/impl/.../roomlist/RoomListPresenterTest.kt` gains a pair of tests: a room which is a
descendant of a Space is hidden with the setting on and listed with it off.
`features/preferences/impl/.../advanced/AdvancedSettingsPresenterTest.kt` covers the switch writing
the preference.

Both fail on develop with the filtering neutralised and with the event ignored, respectively.

Run with `./gradlew :features:home:impl:testDebugUnitTest :features:preferences:impl:testDebugUnitTest`.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


mxandreas:
I think this is alright if we can get the same on iOS. I'd like the apps not to deviate more than they currently do. cc @americanrefugee

bxdxnn:
Duplicates #6965

PR #7545, opened 2026-08-25
Let the user turn off Markdown formatting when sending a message

### LINKED ISSUE #5022: Allow.disable Markdown Formatting
Add a button like Element Client.

Image

The Markdown format causes the message text to be scrambled.

_No response_

### Additional context

_No response_

No

hayaksi1:
## Content

A message typed in the plain composer is sent with no HTML body, so the SDK turns whatever looks
like Markdown in it into formatting. Text which merely contains asterisks or underscores comes out
scrambled and there is no way to stop it.

Advanced settings gains a switch, on by default, which sends the message as plain text instead. The
rich text composer is unaffected: it builds the HTML itself rather than relying on the Markdown
being interpreted, so an explicitly formatted message still arrives formatted.

The switch covers sending a new message. Replies, edits and captions still go through the SDK's
Markdown handling, because `Timeline.replyMessage` and `Timeline.editMessage` have no plain text
flag to pass — `sendMessage` already did. Extending those is a separate change to the matrix layer.

## Motivation and context

Closes #5022, which asks for the control Element Web has in its composer. It is a setting here
rather than a composer button, matching where the other sending preferences live.

The new strings are in `temporary.xml` per AGENTS.md.

## Tests

`features/messages/impl/.../MessageComposerPresenterTest.kt` gains a pair of tests: a plain message
is sent with `asPlainText = true` when the setting is off, and with `false` when it is on.
`AdvancedSettingsPresenterTest` covers the switch itself.

The first fails on develop, which always sends `asPlainText = false`.

Run with `./gradlew :features:messages:impl:testDebugUnitTest :features:preferences:impl:testDebugUnitTest`.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


frebib:
> and there is no way to stop it

Yes there is- enable the Text Formatting (rich text) mode

Issue #4475, opened 2025-03-26
Remove placeholder from the Element Call custom URL form.


bmarty:
The URL path isn't constrained. So, I would recommend not having the placeholder value at all. This is consistent with the [iOS implementation](https://github.com/element-hq/element-x-ios/pull/3939).

_Originally posted by @hughns in https://github.com/element-hq/element-x-android/pull/4470#discussion_r2013785875_

jmartinesp:
I'm not sure this is true.

Without the `/room` path using just `https://call.element.dev`:

<details>
<summary>Screenshot</summary>
<img width="442" alt="Image" src="https://github.com/user-attachments/assets/1f053308-7890-4055-952f-e99bc9e84cc2" />
</details>

With it:

<details>
<summary>Screenshot</summary>
<img width="442" alt="Image" src="https://github.com/user-attachments/assets/6d26f9d0-433a-4ea8-a916-0a61d50c5596" />
</details>

Without the `/room` suffix we don't get the user to actually join the room call. And as far as I know this is the default behaviour on EC now, the only reason why we're not directly appending the `/room` suffix ourselves is because someone could have a subdomain in a reverse proxy like `call.my.tld` pointing to the actual `https://1.2.3.4/room` URL.

jmartinesp:
Maybe I should ping @hughns so he can confirm: should we somehow make sure the URL contains the `/room` suffix so joining a call takes you to the room instead of the lobby? And if so, should we use something different from the placeholder, or use a different value in the placeholder?

hughns:
Whilst it is true that if you want to use `call.element.dev` then you need to use `https://call.element.dev/room` as the URL. But, if you unpack the embedded tarball and host it somewhere then the URL might be `https://call.foo.tld/some/path/somewhere`.

i.e. in future it will really depend on how the server has been deployed.

Rather than attempting to make this developer option fully self explanatory inline within the app (which might take a lot of text), what I would suggest is adding something to the docs in the EX repositories that explains how this developer option can be used. e.g. somewhere in https://github.com/element-hq/element-x-android/tree/develop/docs and https://github.com/element-hq/element-x-ios/tree/develop/docs

PR #7546, opened 2026-08-25
Document how the Element Call base URL developer option is used

hayaksi1:
## Content

Adds `docs/element_call.md`. It covers where the Element Call base URL developer option lives, that
the value is used verbatim as the widget base URL and therefore has to be the URL of the page that
joins a room (`https://call.element.io/room`, not `https://call.element.io`), why the application
cannot append the room path itself, and where the value is read in the code.

The placeholder itself is unchanged: as explained in the issue thread, the path is a property of the
deployment, so there is no suffix the application could safely add.

## Motivation and context

Relates to #4475. The issue asks to remove the placeholder from the field, but the thread settled on
a different resolution — hughns: *"Rather than attempting to make this developer option fully self
explanatory inline within the app (which might take a lot of text), what I would suggest is adding
something to the docs in the EX repositories that explains how this developer option can be used."*
This is the Android half of that. element-x-ios still needs its own page.

## Tests

- `python3 tools/docs/generate_toc.py --verify ./*.md docs/*.md` (the `checkDocs` gate) reports `OK`
  for the new file with its generated table of contents.
- Documentation only — nothing here is covered by a unit test.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s): n/a — documentation only

## Checklist

    - [ ] No.

### DIFF
--- a/docs/element_call.md
+++ b/docs/element_call.md
@@ -1,0 +1,53 @@
+# Element Call
+
+<!--- TOC -->
+
+* [Overview](#overview)
+* [Setting a custom base URL](#setting-a-custom-base-url)
+* [Why the application does not add the room path itself](#why-the-application-does-not-add-the-room-path-itself)
+* [Where the value is read](#where-the-value-is-read)
+
+<!--- END -->
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


bmarty (approved):
Thanks

Issue #7310, opened 2026-08-01
Please restore the ability to see membership changes in public rooms


frebib:
Ideally make it configurable because many probably don't want to see it. Some do (hi 👋🏼) in some cases and hiding it entirely removes the ability to see some information about what is going on in the room.

networkException:
Having invites hidden regardless of the join rules makes the client unusable for moderating a room

PR #7025, opened 2026-06-15
Refactor: replace client-side event filtering for public rooms with SDK

jmartinesp:
## Content

Replace `FilterPublicEmptyDayPostProcessor` and `FilterPublicMembershipChangesPostProcessor` with some simple SDK timeline filters.

## Motivation and context

Replace and simplify https://github.com/element-hq/element-x-android/pull/6360

## Tests

The app should behave just as it did: join and leave events and display name / avatar url changes in publicly accessible, non-encrypted rooms should be hidden. Since the filter happens in the SDK, day dividers for empty days should also be removed automatically.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

- [x] I am aware of the [etiquette](https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#etiquette).
- This PR was made with the help of AI:
    - [ ] Yes. In this case, please request a review by Copilot.
    - [x] No.
- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly defines what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #7547, opened 2026-08-25
Only hide membership changes in rooms that are actually publicly joinable

hayaksi1:
## Content

`RustTimelineEventFilterFactory` hides membership changes and profile changes from the timeline, and
the comment above the guard says it does so when "the room is publicly joinable and not encrypted".
The guard itself was `joinRule !is JoinRule.Invite`, which is much broader — it also matched knock
rooms, restricted rooms, `KnockRestricted`, custom join rules, and rooms whose join rule the app had
not learnt yet (`joinRule == null`). None of those are publicly joinable, so joins, leaves and
display-name changes were being hidden in rooms where they carry real information.

This narrows the guard to `JoinRule.Public`, which is what the comment already describes, and moves
the decision into `hidesMembershipAndProfileChanges` so it can be unit tested — `TimelineEventFilter`
itself is a uniffi object and needs the native library, whereas the guard is plain Kotlin.

Genuinely public rooms are unchanged: they still hide the events, deliberately, as #7025 intended.

## Motivation and context

Relates to #7310. The issue asks for membership changes to come back in public rooms and ideally to
be configurable; this only recovers the rooms the filter was never meant to cover, so the public-room
half of the request stays open — that one is a product decision rather than a defect.

## Tests

- `libraries/matrix/impl/src/test/kotlin/.../room/TimelineEventFilterFactoryTest.kt` — four cases:
  public + unencrypted hides the events; invite / knock / restricted / knock-restricted / custom show
  them; encrypted or unknown-encryption public rooms show them; unknown join rule shows them.
- Negative control: with the old `joinRule !is JoinRule.Invite` guard restored, two of the four fail
  (`...are shown in rooms that are not publicly joinable` and `...when the join rule is unknown`),
  so the tests are guarding the change rather than restating it.
- `./gradlew :libraries:matrix:impl:testDebugUnitTest :libraries:matrix:impl:ktlintCheck :libraries:matrix:impl:detekt` — green.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): unit tests only, no UI change

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/TimelineEventFilterFactory.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/TimelineEventFilterFactory.kt
@@ -37,7 +37,7 @@ class RustTimelineEventFilterFactory : TimelineEventFilterFactory {
         }
         // If the room is publicly joinable and not encrypted, we also want to exclude membership changes and profile changes,
         // as they will pollute the timelines since they're quite common and not add much value.
-        val excludedMembershipChanges = if (joinRule !is JoinRule.Invite && isEncrypted == false) {
+        val excludedMembershipChanges = if (hidesMembershipAndProfileChanges(joinRule, isEncrypted)) {
             listOf(
                 FilterTimelineEventCondition.MembershipChange(MembershipChangeFilter.JOIN),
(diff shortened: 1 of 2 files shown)


bxdxnn:
The logic is "don't hide for invite-only" because for other types, it's not much important to know that someone has a different avatar/name or that some joined for security reasons, see https://github.com/element-hq/element-x-android/pull/6360#issuecomment-4507961974

Issue #4556, opened 2025-04-08
"we couldn't reach this homeserver" is incredibly vague and hard to debug


anarcat:
### Steps to reproduce

1. i'm trying to login to matrix.debian.social with element x
2. i follow the login process and enter the homeserver URL and fail

#### What did you expect?

I expected to be able to login to the homeserver. It's running synapse 1.127.1 so it should have sliding sync (right?).

#### What happened instead?

Instead, it now fails on the MAS stuff, as it gets a 404 on `/_matrix/client/unstable/org.matrix.msc2965/auth_metadata`.

It seems to me this could be relayed back to the user. There are a couple of similar bug reports in this issue tracker, yet they all vary subtly:

- #1097 is about missing sliding sync support (same error message!)
- #3447 seems to be about proxy issues
- #954 is similarly unclear but seemed to be about SSO

In generaly, percolate back to the user the actual error, including the HTTP status code and the URL it tried to reach. Telling users to "contact their homeserver admin" is not really useful, especially when the user *is* the homeserver admin. I only found out what the issue was by connecting to a separate IP (so i could find the traffic in the log) and grepping `homeserver.log`, which is not really intuitive.

I am also, fundamentally, puzzled that Element X already *requires* MAS given that this was deployed only *last week* on Matrix.org. Maybe I'm missing something?

### Your phone model

irrelevant

### Operating system version

Android 15 / CalyxOS

### Application version and app store

element x from f-droid

### Homeserver

1.127.1

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

alexander-potemkin:
Same on iOS - spend a few days, have no glue how to troubleshoot it at all

alexander-potemkin:
As it's turned out, it could also mean that the server accept TLS 1.3, which is not supported: https://github.com/element-hq/element-x-ios/issues/786

ShadowRZ:
@anarcat

> I am also, fundamentally, puzzled that Element X already _requires_ MAS given that this was deployed only _last week_ on Matrix.org. Maybe I'm missing something?

Yes you missed something.

`https://matrix.debian.social/_matrix/client/v3/login` Looks like this:

```
{
  "flows": [
    {
      "type": "m.login.sso",
      "identity_providers": [
        {
          "id": "oidc",
          "name": "Salsa",
          "brand": "salsa"
        }
      ]
    },
    {
      "type": "m.login.token"
    },
    {
      "type": "m.login.application_service"
    }
[...]

PR #7553, opened 2026-08-25
Name the homeserver's administrator when it cannot be reached

hayaksi1:
## Content

"We couldn't reach this homeserver" tells the user to contact an administrator without saying who
that is. The Matrix spec has a place for that answer:
[`GET /.well-known/matrix/support`](https://spec.matrix.org/v1.16/client-server-api/#getwell-knownmatrixsupport),
served by the delegating domain rather than the homeserver API, so it usually answers even when the
API does not — which is exactly the case in the report.

On the failure path only, the login flow now reads that file and adds the administrator's email
address, Matrix ID or support page to the dialog. An admin contact is preferred over the other roles,
an email address over a Matrix ID, and the support page is the last resort. A homeserver that
advertises nothing, or that cannot be reached at all, keeps today's text — the lookup never fails the
login, and it never runs on the success path.

Deliberately out of scope: raising the log level during authentication and offering the logs from the
dialog, the other half of manuroe's comment.

## Motivation and context

Part of #4556. manuroe, reopening it as an enhancement: *"And to really finish the work, we need to
add the admin details in this error popup as adverstised by `GET /.well-known/matrix/support`."*

## Tests

- `HomeserverSupportContactProviderTest` — role preference, email over Matrix ID, support page as the
  last resort, and nothing advertised.
- `OnBoardingPresenterTest` — an unreachable homeserver produces
  `ChangeServerError.InvalidServer(supportContact = …)` when one is advertised and
  `ChangeServerError.InvalidServer()` when not.
- Negative control: dropping the enrichment makes `present - an unreachable homeserver reports the
  administrator it advertises` fail.
- `./gradlew :features:login:impl:testDebugUnitTest :features:login:impl:ktlintCheck :features:login:impl:detekt` — green.
- The dialog variant is covered by a new `ChangeServerErrorPreviewParam` value, appended last so the
  existing goldens keep their indices; the two new ones were recorded by CI.

## Tested devices

- [ ] Physical
- [x] Emulator

## Checklist

    - [ ] No.


bmarty (changes requested):
Thanks for the enhancement.

Since the Matrix specs is defining where the administrator can be reached, the change has to be located in the SDK first. Probably `ClientBuildException.ServerUnreachable` should carry the admin email if available.

hayaksi1:
Moved the lookup into the SDK as suggested: matrix-org/matrix-rust-sdk#6924. I'll rework this PR on top of it once that lands.

Issue #5727, opened 2025-11-13
[Task] Change role from member profile screen


manuroe:
Please see and discuss the details in the meta issue.

bmarty:
Closing as we [decided](https://github.com/element-hq/element-meta/issues/3028#issuecomment-3822893349) to not implement this, and bot are picking the task.

PR #7556, opened 2026-08-25
Change a member's role from their moderation actions

hayaksi1:
## Content

Promoting or demoting somebody meant leaving the member you were looking at, going to Roles and
permissions, and editing a list. The actions sheet for a member now offers the roles they do not
already hold — Make admin, Make moderator, Make member — and applies the choice with
`updateUsersRoles` after a confirmation.

It reuses the sheet the moderation actions already live in rather than adding a screen. The offer is
gated on the current user being able to send `m.room.power_levels`, enabled only for a target they
already outrank (the same rule kick and ban use), and never shown for an owner, whose role this
screen cannot safely change.

## Motivation and context

Part of #5727, which points at a meta issue I cannot read, so the placement and wording here are my
own — happy to move them if the design says otherwise. The strings are in `temporary.xml` for the
Localazy import.

## Tests

- `RoomMemberModerationPresenterTest` — the roles offered exclude the one the member already holds;
  nothing is offered without the power levels permission; nothing is offered for an owner; confirming
  calls `updateUsersRoles` exactly once with the expected `UserRoleChange`, and not before.
- Negative control: removing the permission gate and the `updateUsersRoles` call fails two of the four.
- `./gradlew :features:roommembermoderation:impl:testDebugUnitTest :features:roomdetails:impl:testDebugUnitTest`
  plus ktlint and detekt — green. `RoomMemberModerationPermissions` gains a field, so
  `RoomMemberListStatePreviewParam` in `features/roomdetails` is updated to match.

## Tested devices

- [ ] Physical
- [x] Emulator

## Checklist

    - [ ] No.

### DIFF
--- a/features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListStatePreviewParam.kt
+++ b/features/roomdetails/impl/src/main/kotlin/io/element/android/features/roomdetails/impl/members/RoomMemberListStatePreviewParam.kt
@@ -109,11 +109,13 @@ internal fun aRoomMemberListState(
 fun aRoomMemberModerationState(
     canBan: Boolean = false,
     canKick: Boolean = false,
+    canChangeRoles: Boolean = false,
 ): RoomMemberModerationState {
     return object : RoomMemberModerationState {
         override val permissions: RoomMemberModerationPermissions = RoomMemberModerationPermissions(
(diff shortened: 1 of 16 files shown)


bmarty:
Meta issue is public and we decided to not implement this, see https://github.com/element-hq/element-meta/issues/3028#issuecomment-3822893349

Issue #2789, opened 2024-05-02
Feat: Jump back to message after jumping to reply


cyrneko:
#### What would you like to do?

When jumping to a reply, often I will wanna go back to the message that I was previously at, where I got the reply from.

#### Why would you like to do it?

Usually I check a reply to get context, and as such doing this serves the purpose of getting context without having to scroll back to a specific conversation manually. Often, the message with the reply is not from an ongoing conversation

#### How would you like to achieve it?

After hitting a reply, I would like to either have a separate button, or have the "jump to bottom" button act as a button to jump to the message I got the reply from. 

### Have you considered any alternatives?

Scrolling is tideous, especially when it is a long chain of replies.

### Additional context

Preferably, this would stack so that even when you hit multiple replies, you could backtrack until eventually you came back to the first reply you clicked.

### Are you willing to provide a PR?

No

PR #7558, opened 2026-08-25
Offer a way back after jumping to a replied-to message

hayaksi1:
## Content

Tapping a reply scrolls the timeline to the quoted message. Getting back to what you were reading
means scrolling by hand, or jumping to the bottom and losing the place entirely.

The timeline now remembers the first visible event each time a reply is followed, and shows a third
scroll button that returns to it. The stack pops one entry per tap, so a chain of replies unwinds in
the order it was followed. Only reply taps are remembered — a jump from a notification or a permalink
is not — and jumping to live clears the stack.

It rides the machinery that already exists: `TimelineEvent.FocusOnEvent` gains a `fromReply` flag,
the return jump re-dispatches the same focus request, and the button is another
`JumpToPositionButton` in the column that already holds the other two.

## Motivation and context

Part of #2789. The affordance itself has not been through design — the issue has no comments — so a
third button is my reading of "jump back", and I am happy to move it or fold it into an existing
control instead.

## Screenshots / GIFs

Covered by a new `@PreviewsDayNight` preview, so the button appears in the file diff with the other
two above it.

## Tests

- `TimelinePresenterTest` — following a reply makes `canJumpBack` true and `JumpBack` clears it; a
  focus that did not come from a reply does not; jumping to live forgets the stack.
- Negative control: not pushing the anchor fails two of the three.
- `./gradlew :features:messages:impl:testDebugUnitTest :features:messages:impl:ktlintCheck :features:messages:impl:detekt` — green.
- One existing assertion changes: `TimelineViewTest`'s `clicking on a reply to a loaded event emits
  the expected Event` now expects `FocusOnEvent(AN_EVENT_ID, fromReply = true)`. That assertion is the
  behaviour this PR changes, and the stricter form is the point rather than collateral.

## Tested devices

- [ ] Physical
- [x] Emulator

## Checklist

    - [ ] No.


bmarty (commented):
Thanks.

I agree that it can be painful to navigate to a replied event and lose the current read position.

Regarding the UI, I wonder if we really need a new button. Maybe using the system Back button or the back gesture is enough to navigate back to history.

Anyway I would need a confirmation of product before going further, and ideally this will have to be implemented in other Element clients.

Issue #6959, opened 2026-06-04
Pinning a message another user does not have access to is handled incorrectly


PietG1:
### Steps to reproduce

steps:

1\. Have a room that contains messages and an account that does not yet have access to this room

2\. Invite your second account to this room (make sure he won't have access to the past messages)

3\. Accept the invitation on the second account

4\. Have the first account pin a message which the second account has no access to.

5\. Look at the chat on the second account

#### What did you expect?
1. The are no pinned messages or it is not shown
2. In the settings of the screen you see 0 and there are non shown if you click on it

#### What happened instead?
1. The pinned message keeps on loading
2. In the settings of the screen you see 1 at Pinned messages  and if you click on it it stays loading

### Your phone model

google pixel

### Will you send logs?

Yes

### Are you willing to provide a PR?

Yes

PR #7559, opened 2026-08-25
Count only the pinned messages the reader can actually see

hayaksi1:
## Content

The pinned messages row in room details counted the event ids in `m.room.pinned_events`. A room can
pin an event a given member has no access to — one sent before they joined, in a room that does not
share its history — and such an event never reaches the pinned timeline. So the row promised a
message that the pinned list could not show, which is exactly what the reporter saw: room details
saying one pinned message, the list showing none.

The count now comes from the pinned timeline, the same source the pinned banner and the pinned list
already use, so the three agree by construction. Until that timeline has loaded, the row falls back
to the ids in the room state, so it still shows a number straight away rather than a spinner.

## Motivation and context

Part of #6959.

Deliberately not fixed here, and worth saying plainly: the other half of that report — the pinned
list sitting on a spinner instead of saying it has nothing to show — is not app-side. The list leaves
its loading state as soon as the pinned timeline emits anything at all, including an empty list; an
event that cannot be fetched means the SDK's pinned timeline never emits, and nothing in this
repository can turn that into an empty state.

## Tests

- `RoomDetailsPresenterTest` — the count is the number the pinned timeline can show, and a room whose
  only pinned event the user cannot see reports none.
- The two existing assertions on `pinnedMessagesCount` are unchanged and still pass, because the
  fallback keeps today's behaviour whenever the pinned timeline has not loaded.
- Negative control: restoring `roomInfo.pinnedEventIds.size` fails both new tests.
- `./gradlew :features:roomdetails:impl:testDebugUnitTest :features:messages:impl:testDebugUnitTest`
  plus ktlint and detekt on the four modules — green. `:tests:konsist:testDebugUnitTest --rerun-tasks`
  — green.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): unit tests only, no UI change

## Checklist

    - [ ] No.

### DIFF
--- a/features/messages/api/src/main/kotlin/io/element/android/features/messages/api/pinned/PinnedEventsTimelineProvider.kt
+++ b/features/messages/api/src/main/kotlin/io/element/android/features/messages/api/pinned/PinnedEventsTimelineProvider.kt
@@ -9,10 +9,20 @@
 package io.element.android.features.messages.api.pinned
 
 import io.element.android.libraries.matrix.api.timeline.TimelineProvider
+import kotlinx.coroutines.flow.Flow
 
 /**
  * A [TimelineProvider] whose active timeline holds the pinned events of the room, rather than its live events.
  *
  * It exists as its own type so that it can be injected where only the pinned timeline is wanted.
  */
-interface PinnedEventsTimelineProvider : TimelineProvider
+interface PinnedEventsTimelineProvider : TimelineProvider {
+    /**
+     * How many of the room's pinned events the user can actually see, or `null` while that is not known yet.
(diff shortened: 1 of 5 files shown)


bmarty (approved):
Thanks!

bmarty:
@mxandreas do you agree with this change? All the users may not see the same number of pinned events, depending if they have actually access to it. It can be confusing, because if no pinned events are seen, the pin message banner will not be displayed. Waiting for your decision before merging the PR.

bmarty (commented):
Waiting for product approval.

Issue #5209, opened 2025-08-23
Hide "User made no changes" timeline events


noantiq:
#### What would you like to do?

Hide timeline events that say that a user made no changes.

#### Why would you like to do it?

[As reported for Element Android Classic](https://github.com/element-hq/element-android/issues/1430) the timeline events that say that a user made no changes do not contain any useful information for other users and therefore are very annoying. Especially in combination with https://github.com/element-hq/element-x-android/issues/2623

#### How would you like to achieve it?

Do not show them in the timeline or room list per default, only show them when activating a corresponding developer option.

### Are you willing to provide a PR?

No

Lenni-builder:
This still seems to exist and can be really annoying. Shouldn't it be a relatively simple fix?

PR #7560, opened 2026-08-25
Stop showing membership events that changed nothing in the timeline

hayaksi1:
## Content

A `m.room.member` event whose only effect is to re-send the same membership renders in the timeline
as "Alice made no changes". It carries nothing for anyone else in the room, and clients that rewrite
their own membership leave runs of these rows between real messages.

`TimelineItemsFactory` now drops the event, next to the in-chat key verification request it already
drops. An earlier revision of this PR returned `null` from `RoomMembershipContentFormatter` instead;
that does not work. `TimelineItemContentRoomMembershipFactory` builds
`TimelineItemRoomMembershipContent(text.orEmpty().toString())` and `TimelineItemStateView` renders
the body in a centred `Text` with no emptiness guard, so a `null` from the formatter leaves a blank
row that still takes part in bubble grouping and still keeps its day separator alive. The formatter
is left untouched.

The room list preview is unchanged. It shows whatever the latest event was, so hiding the text there
would leave a stale preview rather than no preview.

## Motivation and context

Relates to #5209.

## Tests

- `TimelineItemsFactoryTest.a membership event that changed nothing is not rendered` — a `NONE`
  membership change, a `JOINED` one and a message go in; the `NONE` one is not emitted and the other
  two are.
- Run against the unfixed code, where it fails.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): API 36

## Checklist

    - [ ] No.


bxdxnn:
Duplicates #6862

hayaksi1:
It seems so, so I have updated this one to fix the part #6862 does not cover.

The formatter change is gone: `RoomMembershipContentFormatter` and both its callers are back to their
`develop` state, and the event is now dropped in `TimelineItemsFactory` next to the in-chat key
verification request it already drops.

The reason is that returning `null` from the formatter does not remove the row.
`TimelineItemContentRoomMembershipFactory` builds `TimelineItemRoomMembershipContent(text.orEmpty().toString())`
and `TimelineItemStateView` renders that body in a centred `Text` with no emptiness guard, so the row
is still there, blank, still grouped and still keeping its day separator alive.
`TimelineItemContentStateFactory` does the same thing, so both of the `null` returns in #6862 land on
it too.

Issue #5813, opened 2025-11-25
Do not show read marker for the person who wrote the message


srgcdev:
### Steps to reproduce

1. Receive a message from someone
2. Inmediately I receive a marker indicating that person has read the message he actually wrote

#### What did you expect?
I expect no read marker appear for the person who wrote the message because is redundant.

### Will you send logs?

No

### Are you willing to provide a PR?

No

bmarty:
The marker can't stay on any previous message, can it?

srgcdev:
That's it. When a user write a message implies that he/she has already read all previous messages.

jmartinesp:
I guess we *could* hide the read marker client-side if we consider it redundant, but this needs a product decision.

bmarty:
> That's it. When a user write a message implies that he/she has already read all previous messages.

But if other messages are sent by other people, the read receipt of the first user (`someone`) will be missing (assuming they have not read the new messages)? It will have to be displayed under the message that this user has sent, so I do not think that this is a good idea.

srgcdev:
When other people send new message, if no read marker appears it's asssumed that the last message he/she read was the message he/she wrote.

Other clients like fluffychat do that.

bmarty:
If a Bob is looking for Alice Read Receipt, Bob will look at the Read Receipt, i.e. at the right of the screen. Bob can miss Alice's message , because Alice's avatar is displayed on the left of the screen. I do not think this is a good idea, closing the issue.

PR #7561, opened 2026-08-25
Stop showing a sender's read receipt on their own message

hayaksi1:
## Content

A message shows a read receipt avatar for the person who wrote it. A user's receipt marks the last event they have read, so it lands on their own message the moment they send one, and the row then reports that the author has read what they just wrote.

The receipt is dropped where the receipts are computed, so the avatar row, the grouped-event row and the read receipt bottom sheet all agree. Nothing is lost: a receipt sitting on its own author's message is the only place that author's receipt can be, so hiding it hides no position the reader could otherwise have seen.

`TimelinePresenterTest`'s avatar propagation case used the same user as both the sender and the only receipt holder, so with this change it had nothing left to assert. Its receipt now comes from another member; both of its assertions are unchanged.

## Motivation and context

Relates to #5813.

## Tests

- `TimelineItemsFactoryTest` — an event with a receipt from its own sender and one from another member keeps only the other member's.
- The rest of `:features:messages:impl` still passes, including `TimelinePresenterTest`.

These do not cover the case where the SDK stops reporting a sender's own receipt at all; the filter is then simply a no-op.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): API 36

## Checklist

    - [ ] No.

### DIFF
--- a/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemEventFactory.kt
+++ b/features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/factories/event/TimelineItemEventFactory.kt
@@ -191,6 +191,7 @@ class TimelineItemEventFactory(
         }
         return TimelineItemReadReceipts(
             receipts = event.receipts
+                .filterNot { it.userId == event.sender }
                 .map { receipt ->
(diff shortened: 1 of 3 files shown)


bmarty:
I have closed the underlying issue.

Issue #1548, opened 2023-10-02
Standardise how a room converted to a DM shows events in room list preview


kittykat:
### Steps to reproduce

1. Create a room
2. Convert to DM
3. Send location or poll

#### What did you expect?
Both platforms show the same preview for the event

#### What happened instead?
EX Android shows: "username: event" in the room list preview
EX iOS shows: "event" in the room list preview

### Application version

Nightly

### Homeserver

matrix.org

manuroe:
@amshakal we need to double the production decision. Android and iOS are doing the opposite of what the description is saying.

amshakal:
We need to not show username in the preview when there are only 2 people there. It gets tricky when bots are there in a DM.

PR #7562, opened 2026-08-25
Drop the sender's name from the room list preview of two-person rooms

hayaksi1:
## Content

The room list preview prefixes the sender's name unless the room is a direct message. A room created as a group and later converted to a direct message is not marked as one by the server, so its preview kept naming the only other person in it on every message, while iOS showed no name at all.

The prefix now depends on how many people are actually in the room rather than on the direct message flag: no name while only two people can be talking, the name back as soon as a third joins. That is the behaviour asked for in the issue, and it also keeps the name for a two-person room that has a bot in it, since the bot counts.

## Motivation and context

Relates to #1548.

## Tests

- `RoomListRoomSummaryFactoryTest` — a two-member room formats its preview without the sender, a three-member room formats it with the sender.

These do not cover a room whose member count has not synced yet; the preview then follows whatever count is known, and corrects itself when the count arrives.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): API 36

## Checklist

    - [ ] No.


bmarty (approved):
OK for me, thanks!

Issue #5797, opened 2025-11-24
following room upgrades loses notification seeing


HarHarLinks:
### Steps to reproduce

1. Join a room such as !vlnjqGLpLJlFmBSkfQ:matrix.org
2. Customize notifications top this room, e.g. to pings only
3. Wait for the room to get upgraded
4. Click jump to new room !KDiNABdMiGixPIk0YCQ7syudVPr8hqpxpJl_qEUlxWQ
5. Join it
6. Check room notification settings

#### What did you expect?
The "same" room will have the same notification settings

#### What happened instead?
Default notifications on the new room

### Application version and app store

25.11.3

### Will you send logs?

No

### Are you willing to provide a PR?

No

bmarty:
Thanks for reporting. I have also noticed that the banned users are not banned again in the new room. I guess this is how room upgrade works...
@mxandreas do you think it worth spending time on this issue? I do not think so personnaly even if I agree that upgrading a room should have the less impact as possible on the user experience / user custom configuration.

HarHarLinks:
The same applies to Element Web btw.

mxandreas:
> do you think it worth spending time on this issue? I do not think so personnaly even if I agree that upgrading a room should have the less impact as possible on the user experience / user custom configuration

Yes, in an ideal world, various room "settings" or "configurations" would not be lost during the upgrade. However, this is not only an EX issue (AFAIK), as well as implementation-wise it would something that probably the server would be better to handle, in particular as no one will ever trigger upgrade from a mobile client.

The topic would require a more comprehensive approach (e.g. what data needs to be copied over), it is not something we can fix on the go on mobile side only.

HarHarLinks:
Note this issue occurred where the room was not upgraded using the upgrade API and also I'm on the same homeserver as the new room owner.

> Thanks for reporting. I have also noticed that the banned users are not banned again in the new room.

I think this is related but a different problem and actually considered best practice in T&S circles.

sandhose:
This is actually a server problem when the upgrade is done 'manually', which is what the Trust and Safety team is currently doing, read: https://github.com/element-hq/synapse/issues/19199

PR #7563, opened 2026-08-25
Keep a room's notification setting when it is upgraded

hayaksi1:
## Content

A room that has been upgraded is a different room to the server, so the notification rule a user set on the old one does not apply to the new one. Following the upgrade banner into the replacement room therefore silently returned that conversation to the default, and a room someone had deliberately set to mentions only started notifying for everything again.

The setting is now carried over in `JoinRoom`, which every join already goes through, so it applies however the replacement room is reached rather than only from the banner.

Only a rule the user set themselves is copied. A room that was following the default keeps following it, rather than being pinned to whatever the default happened to be at the moment of the upgrade.

## Motivation and context

Relates to #5797.

## Tests

- `DefaultJoinRoomTest` — joining the successor of a room with its own notification mode sets that mode on the new room; joining a room with no predecessor, or whose predecessor was on the default, changes nothing.

These do not cover the case where the predecessor room is no longer in the client's store, in which case nothing is copied and the new room keeps the default.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): API 36

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/join/DefaultJoinRoom.kt
+++ b/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/join/DefaultJoinRoom.kt
@@ -10,9 +10,11 @@ package io.element.android.libraries.matrix.impl.room.join
 
 import dev.zacsweers.metro.ContributesBinding
 import im.vector.app.features.analytics.plan.JoinedRoom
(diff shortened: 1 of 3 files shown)


bmarty (commented):
Make sense, but the fix could maybe be located in the SDK?

hayaksi1:
Moved the fix into the SDK as suggested: matrix-org/matrix-rust-sdk#6923. I'll rework this PR on top of it once that lands.

Issue #6031, opened 2026-01-17
[Feature Request] PIN Pad Randomiser


saradasama:
### Your use case

Please consider adding the possibility (in the form of a toggle option) to randomise the numbers (in their original positions) of the _PIN pad_. This modern and neat solution can improve the privacy aspect of unlocking the application every time it is needed.

### Are you willing to provide a PR?

No

PR #7564, opened 2026-08-25
Offer to shuffle the digits of the PIN keypad

hayaksi1:
## Content

The PIN keypad always draws 1 to 0 in the same nine positions, so anyone standing close enough to watch the movement of a finger can read the PIN back without seeing the screen at all. A new switch under Screen lock draws the digits in a random order instead, keeping the same grid so the pad still looks and behaves the way it did.

The order is drawn once per unlock screen rather than after every key, so a PIN can still be entered without hunting for each digit twice. The setting is off by default and lives next to the biometric unlock switch, in the same store.

## Motivation and context

Relates to #6031.

## Tests

- `PinKeypadTest` — a shuffled keypad still offers every digit exactly once, and clicking a digit still emits that digit rather than the one that used to sit in that position.
- The existing keypad and unlock tests are unchanged and still pass with shuffling off.

These do not assert that a shuffled pad differs from an unshuffled one, which is not deterministic; they assert that shuffling never loses, duplicates or mislabels a key.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): API 36

## Checklist

    - [ ] No.


bmarty (commented):
Nice. We will need a product validation first.

Also, the current layout with the `0` at the bottom is quite a standard digit-keypad, but with a random number at this position, this could look weird...

Issue #4099, opened 2024-12-28
Warn when the server is unavailable rather than just saying "synchronizing"


ara4n:
### Steps to reproduce

1. Block your server somehow (network failure; server offline)
2. Fg the app
3. Watch a neverending "Synchronizing" chip and timeline spinner
4. Wonder why your app won't sync

#### What did you expect?

"Offline" or "Server unavailable" or "Can't connect" or similar.

#### What happened instead?

Neverending "Synchronizing" spinner, implying that the server is accessible but is taking a long time to sync.

We should be able to distinguish the two (e.g. Threema distinguishes red & orange connectivity state) to inform the user as to what's going on.

### Will you send logs?

No

### Are you willing to provide a PR?

No

PR #7567, opened 2026-08-25
Say when the server cannot be reached instead of showing nothing

hayaksi1:
## Content

While the homeserver is unreachable but the device still has a working network, the app showed no indication at all: the syncing chip hides as soon as the SDK reports the session offline, and the offline banner only appears when the device itself has no connectivity. A blocked or down server was therefore indistinguishable from a quiet, healthy one.

The chip now has three states rather than a boolean, and the third says "Server unreachable" for exactly that case. When the device has no network the chip stays hidden, because the offline banner already says so and two overlapping messages would be worse than one.

## Motivation and context

Relates to #4099.

## Tests

- `LoggedInPresenterTest` — an offline sync with the device connected reports the server as unreachable; an offline sync with the device disconnected reports nothing.
- The existing sync spinner case is unchanged apart from now naming the state it expects.

These do not distinguish a server that is down from one that is merely slow past the SDK's own threshold; the app shows whatever the SDK's sync state says.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): API 36

## Checklist

    - [ ] No.


bmarty (approved):
Thanks!

Issue #7062, opened 2026-06-19
Notification Setting: Invitations


TemoCrevlis:
### Your use case

Every group includes some form/hint of the word 'notification' (sound, notify etc) except in 'Additional settings' which only has a toggle next to the 'Invitations' label.
It's possible a user might think they're turning off invitations entirely and not just notifications for it.

<img width="1600" height="900" alt="Image" src="https://github.com/user-attachments/assets/e6ca4962-ac05-4fd7-adcf-523e32b72403" />

#### What would you like to do?
1. Put "Invitations" under "Notify me for".
2. Remove the toggle when/if Invite notification filters are added.
#### Why would you like to do it?
Placing it under "Notify me for" and following the selection pattern of the other options would improve clarity and consistency with the rest of the page.

### Are you willing to provide a PR?

No

PR #7569, opened 2026-08-25
Move the invitations and @room mention toggles in with the other notification settings

hayaksi1:
## Content

The switch for invitations sat on its own under "Additional settings", labelled only "Invitations". Nothing around it said the setting was about notifications, so turning it off reads as though it stops invitations arriving rather than stopping them notifying.

It now sits under "Notify me for", next to the group chat and direct chat entries, which is what the rest of that section is for. "Additional settings" held nothing else — the only other entry there is the call toggle, which is commented out until call notifications are supported — so the category is gone.

The "Mentions" category held only the @room switch, which answers the same question as the three entries above it, so it has moved under "Notify me for" too and that category is gone as well.

This is the first half of the issue. The second half, removing the toggle once invitation notification filters exist, is not addressed here.

## Motivation and context

Relates to #7062.

## Tests

- `NotificationSettingsViewTest` is unchanged and still finds and toggles both the invitations and the @room mention switches, so the controls are intact in their new place.

These do not assert which section the switches are in; the Paparazzi snapshots of `NotificationSettingsView` show that.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): API 36

## Checklist

    - [ ] No.


bmarty (commented):
That change makes sense to me, @americanrefugee @mxandreas do you agree?

americanrefugee:
@bmarty I agree with the changes, but we should also do the same for iOS then. Here's the issue I created to track that: https://github.com/element-hq/element-x-ios/issues/6073

americanrefugee (approved):
LGTM

americanrefugee:
@hayaksi1 We should also move `Notify me on @room` to the same list of options under `Notify me for`, below `Invitations`.

americanrefugee (changes requested):
Please move `screen_notification_settings_room_mention_label` to be included in the list for "Notify me for" as well, instead of on its own. 

Meanwhile, `screen_notification_settings_room_mention_label` has already been updated with the new string "**Mentions of @room**"

Here's a screenshot for reference:

Image

hayaksi1:
@americanrefugee Done.

americanrefugee:
@bmarty Should I wait for the change on Localazy to be merged before I approve this issue, or is it okay to approve now?

(For context, the setting is now in the correct place but the label is still wrong.)

bmarty:
The new label will be synced next week, let's wait

PR #7570, opened 2026-08-25
Run the database migrations again when a database is upgraded

hayaksi1:
## Content

Upgrading over an older install crashed the app before the login screen with
`SQLiteException: no such column: SessionData.position`. The session, cache and push databases were
opened at their new schema version but their tables were never migrated, so the first query against
a column added by migrations 9–11 failed.

`SqlCipherDriverFactory` overrides `AndroidSqliteDriver.Callback.onUpgrade` so that the database key
can be replaced, but it never called through to the base implementation — and that base
implementation is the only thing that runs `schema.migrate(...)`. Every SQLDelight migration in the
app was therefore skipped. The override now calls `super.onUpgrade` after the re-key, and it moved
out of an anonymous object into a named `MigrationCallback` so the behaviour can be tested.

The re-key still runs first, so a database that crosses the re-key version is re-encrypted before
its tables are migrated, exactly as before.

## Motivation and context

Part of #7509

## Tests

- `libraries/encrypted-db/src/test/kotlin/io/element/encrypteddb/MigrationCallbackTest.kt` — asserts
  the schema migrations run on upgrade, and that the key replacement happens before them.
- Both tests were run against the unfixed code and both fail there.
- They do not prove that a real 25.05.4 SQLCipher database upgrades end to end; that still needs a
  device.

## Tested devices

- [ ] Physical
- [x] Emulator

## Checklist

### DIFF
--- a/libraries/encrypted-db/build.gradle.kts
+++ b/libraries/encrypted-db/build.gradle.kts
@@ -5,6 +5,8 @@
  * SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
  * Please see LICENSE files in the repository root for full details.
  */
+import extension.testCommonDependencies
+
 plugins {
     id("io.element.android-library")
 }
@@ -27,4 +29,6 @@ dependencies {
     implementation(libs.google.tink)
 
     implementation(projects.libraries.androidutils)
+
+    testCommonDependencies(libs)
 }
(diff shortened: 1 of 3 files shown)


bmarty (approved):
Thanks. Was waiting for https://github.com/element-hq/element-x-android/pull/7510 author to sign the CLA.

bmarty (approved):

Issue #6730, opened 2026-05-05
Editing a caption saves it as plain text


bxdxnn:
### Steps to reproduce

1. Send a media with a formatted caption
2. Edit the caption after sending

#### What did you expect?

The caption is saved formatted

#### What happened instead?

All of the caption's text becomes unformatted.

### Will you send logs?

No

### Are you willing to provide a PR?

No

nonepuc:
nice idea
And I'd also like to add that it should be possible to send multiple photos at once

PR #7572, opened 2026-08-25
Keep the formatting of a caption that is being edited

hayaksi1:
## Content

Editing the caption of a media message dropped all of its formatting. Bold, links and the rest came
back as plain text in the composer, and saving then wrote that plain text over the formatted caption,
so the formatting was lost for everyone in the room.

The composer was being opened on the plain caption. Editing a message body already starts from its
HTML when rich text editing is on; the caption path was the only one that did not, and it had no HTML
to start from — the timeline content kept only the already-parsed `formattedCaption`, which is a
rendered `CharSequence`, not the caption's source.

Media contents now carry the caption's HTML alongside the parsed one, the same way text contents
already carry `htmlBody`, and editing a caption starts from it when rich text editing is on. The save
path already sent the composer's HTML, so it is unchanged.

## Motivation and context

Part of #6730

## Tests

- `MessagesPresenterTest.present - handle action edit caption starts from the formatted caption` —
  editing a caption that has HTML opens the composer on that HTML.
- The test was run against the unfixed code and fails there.
- Two existing `TimelineItemContentMessageFactoryTest` expectations gained the new field; they compare
  whole content objects, so any added property has to be spelled out in them.
- Nothing here covers the plain-text composer, which keeps reading the plain caption as before.

## Tested devices

- [ ] Physical
- [x] Emulator

## Checklist


bmarty (approved):
Thanks!

Issue #6970, opened 2026-06-05
Room List: Tap Room avatar to open larger view


TemoCrevlis:
#### What would you like to do?
View a Room avatar in a larger format directly from the Room list.
#### Why would you like to do it?
The current flow requires three taps to view a Room avatar in full size, with two of those interactions located at or near the top of the screen. This increases interaction cost and makes the process less convenient, particularly on larger devices. Providing quicker access to the avatar would improve discoverability and reduce unnecessary navigation.
#### How would you like to achieve it?
Allow users to tap a Room avatar directly from the Room list to open the avatar in a full-screen image viewer/modal.

### Have you considered any alternatives?

Allow users to tap a Room avatar directly from the Room list to open a bottom sheet containing a larger preview of the Room avatar.

### Image Showing 2 Modals & 1 Bottom Sheet With Larger Image Preview

<img width="1182" height="822" alt="Image" src="https://github.com/user-attachments/assets/e4554748-12fa-4138-98e7-6a55f442a3a5" />

### Additional context
The Room name sits on top the image to prevent any obstruction.
- For that reason a stronger blur or shadow effect may be used around the name to ease readability.

PR #7573, opened 2026-08-26
Let the room list avatar be tapped to see it larger

hayaksi1:
## Content

Tapping a room's avatar opens it full screen everywhere else in the app — in the timeline top bar, in
the room details, and on a user's profile. In the room list it had no meaning of its own: the tap fell
through to the row and opened the room.

The room list avatar now opens the same media viewer, through a new `AvatarPreview` target on the home
flow, reusing `MediaViewerEntryPoint.createParamsForAvatar` exactly as the other three screens do.

It is only clickable when there is an avatar to show. A room with a generated letter avatar, and an
invitation whose avatar is being hidden, still open the room on a tap as before, so the row keeps its
whole tap target in the cases where a larger view would show nothing.

## Motivation and context

Part of #6970

## Tests

- `RoomListViewTest.clicking on the avatar of a room that has one invokes the avatar callback`
- `RoomListViewTest.the avatar of a room that has none is not clickable`
- Both were run against the unfixed code and both fail there.
- They cover the row's behaviour, not the media viewer itself, which is unchanged and already covered.

## Screenshots / GIFs

The only golden that moved is `features.home.impl_HomeViewA11y_en.png`: the accessibility overlay
draws the avatar's new click action. Every ordinary snapshot is byte-identical, since nothing about
the row's layout or colours changed.

## Tested devices

- [ ] Physical
- [x] Emulator

## Checklist


features/home/impl/src/main/kotlin/io/element/android/features/home/impl/components/RoomSummaryRow.kt
```diff
@@ -79,6 +82,7 @@ internal fun RoomSummaryRow(
     isInviteSeen: Boolean,
     onClick: (RoomListRoomSummary) -> Unit,
     modifier: Modifier = Modifier,
+    onAvatarClick: ((RoomListRoomSummary) -> Unit)? = null,
```
bmarty: As a general rule we do not like callback to be optional, because it can lead to bug, and it make the code a bit more convoluted  (`enabled = onAvatarClick != null`...). Maybe the file AGENTS.md could be updated to add this rule.

hayaksi1: Done.

frebib:
As a person that is predominantly left-handed: I absolutely detest this "feature" in WhatsApp. I often click the avatar to open the room because that's where my finger happens to land on the left-hand side of the screen. If anything I'd argue this is right-hand-preferential and as such am not a fan.

Issue #3835, opened 2024-11-07
Contacts not displayed in Direct Share menu


EmanuelLoos:
# Your use case

## What would you like to do?

I would like Matrix contacts to be shown in the Direct Share menu like they are when using the old Element app so that the new app does not lack functionality that the old app provides.

## Why would you like to do it?

Because it makes sharing files faster and simpler. It also boosts the odds that Element is used for instant messaging rather than other apps that have this functionality.

## How would you like to achieve it?

I only know that the old Element app does achieve it, not how. But maybe someone knows how the old app does it? This might be of help as well: [https://developer.android.com/training/sharing/direct-share-targets?hl=de](https://developer.android.com/training/sharing/direct-share-targets?hl=de)

## Are you willing to provide a PR?

No

kevincox:
A bit of additional context from Element classic. Bug: https://github.com/element-hq/element-android/issues/2029 Implementation: https://github.com/element-hq/element-android/pull/2244

PR #7574, opened 2026-08-26
Offer conversations in the system share sheet

hayaksi1:
## Content

Sharing something from another app listed Element X once, as a single entry, with no way to pick a
conversation. The Direct Share row of the system share sheet never showed any of them.

The app already publishes a long-lived shortcut per room, categorised as a conversation, which is most
of what the sharing shortcuts API asks for. What was missing is the declaration that ties those
shortcuts to the share intent: the launcher activity carried no `android.app.shortcuts` meta-data, so
there was no `<share-target>` for the system to match them against, and the shortcuts were only ever
used for the launcher long-press menu and for notification bubbles.

It now ships one, pointing at the activity that already receives `ACTION_SEND` and `ACTION_SEND_MULTIPLE`,
and the published shortcuts carry the category it names.

The system offers these from Android 10 onwards, where sharing shortcuts are handled natively. Covering
Android 7 to 9 as well would additionally need the `androidx.sharetarget` library; that is a dependency
this PR deliberately does not add.

## Motivation and context

Part of #3835

## Tests

- `DefaultNotificationConversationServiceTest.the shortcut is offered to the share sheet as a conversation`
  — the published shortcut carries both the conversation category and the share-target category.
- The test was run against the unfixed code and fails there.
- It cannot cover the manifest side: that a `<share-target>` matches is a decision the platform makes.
  Verified instead by building the APK and checking `res/xml/shortcuts.xml` is packaged and the
  `android.app.shortcuts` meta-data is present in the merged manifest.

## Tested devices

- [ ] Physical
- [x] Emulator

## Checklist

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -139,6 +139,12 @@
                 <data android:host="user" />
                 <data android:host="room" />
             </intent-filter>
+            <!--
+              Conversation shortcuts offered in the Direct Share row of the system share sheet.
+            -->
+            <meta-data
+                android:name="android.app.shortcuts"
+                android:resource="@xml/shortcuts" />
             <!-- Incoming share simple -->
             <intent-filter>
                 <action android:name="android.intent.action.SEND" />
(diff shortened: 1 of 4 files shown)


bmarty (approved):
Thanks!

PR #7576, opened 2026-08-26
Fix/animated avatar compression

escix:
By Claude.
Fixed, animated images getting converted if the compression is turned on.


bmarty:
Thanks. As per our [AI policy](https://github.com/element-hq/element-x-android/blob/main/CONTRIBUTING.md#etiquette), please open an issue first.

We can then discuss there if we allow animated content to be set for user/room avatar, check what the spec says, etc.

PR #7590, opened 2026-08-28
Call ClientBuilderEnterpriseHook whenever we create a client

richvdh:
Currently, the `ClientBuilderEnterpriseHook` is only called when we restore a previous session, rather than when performing a new login.

I'm putting this up for discussion rather than in the expectation it will land as-is. I don't really think it's the right solution, because it means that the session ID is not set for clients used for a login (and hence the functionality in the Enterprise hook which depends on that does not work correctly).

However, since we're trying to create the client before we know the user ID, it's hard to see what would be better.

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -120,6 +125,15 @@ class RustMatrixClientFactory(
...
     }
 
+    /**
+     * Given a Rust-side [Client], wrap it into an application-side [RustMatrixClient].
+     *
+     * This is a helper for the other `create` method, as well as being called directly from
+     * [io.element.android.libraries.matrix.impl.auth.RustMatrixAuthenticationService] after successful
+     * authentication.
+     *
+     * TODO: feels like this could do with a better name, since it doesn't actually create a client.
```
bmarty, line 135: The `create` methods do not create a `Client`, but they create a `RustMatrixClient`. The methods could return `MatrixClient` though.

bmarty (commented):
I think that the changes make sense. It's quite complex code, since we are using the same method from the SDK to either create a client to authenticate, or create a client to restore an existing session.

Maybe we could revisit the code and let the client used to authenticate be destroyed, and always create a fresh once the user has just logged in, as if a session has been restored.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
richvdh: It feels strange that we have two different methods called `create` with very different behaviours. I feel like it would be much clearer if this was called `wrapClient` or something of that nature.

richvdh:
@bmarty has opened #7591 to explore the ideas he suggests above. I'm pretty happy with whatever solution we end up with here.

I'll keep this PR open for now as it may be helpful for the crypto team while we are working on a feature that needs this hook to work.

richvdh:
#7591 works for us.

PR #7591, opened 2026-08-28
Iterate on client creation

bmarty:
## Content

Make sure that a RustMatrixClient is always created the same way, either when restoring a session or when signing in.

It ensure that the hook are correctly applied.

(See https://github.com/element-hq/element-x-android/pull/7590#pullrequestreview-5051828460).

## Motivation and context

Ensure that hook works the same way.

## Tests

- Login flows and session restoration should still work as expected.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/features/enterprise/api/src/main/kotlin/io/element/android/features/enterprise/api/ClientBuilderEnterpriseHook.kt
+++ b/features/enterprise/api/src/main/kotlin/io/element/android/features/enterprise/api/ClientBuilderEnterpriseHook.kt
@@ -13,6 +13,23 @@ import io.element.android.libraries.matrix.api.core.SessionId
 /**
  * A hook that can be used to customize the [MatrixClientBuilder] for enterprise features.
  */
-fun interface ClientBuilderEnterpriseHook {
-    suspend operator fun invoke(clientBuilder: MatrixClientBuilder, sessionId: SessionId): MatrixClientBuilder
+interface ClientBuilderEnterpriseHook {
+    /**
+     * Customize the [MatrixClientBuilder] for enterprise features.
+     * This method is invoked everytime a new [MatrixClientBuilder] is created.
+     *
+     * @param clientBuilder The [MatrixClientBuilder] to customize.
+     * @return The customized [MatrixClientBuilder].
+     */
+    suspend fun tweakClientBuilder(clientBuilder: MatrixClientBuilder): MatrixClientBuilder
+
+    /**
+     * Customize the [MatrixClientBuilder] for enterprise features.
+     * This method is invoked when a new [MatrixClientBuilder] is created to build a client for a specific session.
+     *
+     * @param clientBuilder The [MatrixClientBuilder] to customize.
+     * @param sessionId The [SessionId] for which the [MatrixClientBuilder] is being created.
+     * @return The customized [MatrixClientBuilder].
+     */
+    suspend fun tweakClientBuilder(clientBuilder: MatrixClientBuilder, sessionId: SessionId): MatrixClientBuilder
 }
--- a/features/enterprise/api/src/main/kotlin/io/element/android/features/enterprise/api/ClientBuilderEnterpriseHook.kt
+++ b/features/enterprise/api/src/main/kotlin/io/element/android/features/enterprise/api/ClientBuilderEnterpriseHook.kt
@@ -21,15 +21,15 @@ interface ClientBuilderEnterpriseHook {
      * @param clientBuilder The [MatrixClientBuilder] to customize.
      * @return The customized [MatrixClientBuilder].
      */
-    suspend fun tweakClientBuilder(clientBuilder: MatrixClientBuilder): MatrixClientBuilder
+    suspend fun beforeClientCreation(clientBuilder: MatrixClientBuilder): MatrixClientBuilder
 
     /**
      * Customize the [MatrixClientBuilder] for enterprise features.
-     * This method is invoked when a new [MatrixClientBuilder] is created to build a client for a specific session.
+     * This method is invoked as well as the other method when a new [MatrixClientBuilder] is created to build a client for a specific session.
      *
      * @param clientBuilder The [MatrixClientBuilder] to customize.
      * @param sessionId The [SessionId] for which the [MatrixClientBuilder] is being created.
      * @return The customized [MatrixClientBuilder].
      */
-    suspend fun tweakClientBuilder(clientBuilder: MatrixClientBuilder, sessionId: SessionId): MatrixClientBuilder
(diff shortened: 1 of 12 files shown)


libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/auth/RustMatrixAuthenticationService.kt
```diff
@@ -166,18 +169,7 @@ class RustMatrixAuthenticationService(
                         passphrase = pendingKey.formattedAsString(),
                         sessionPaths = currentSessionPaths,
                     )
-                val matrixClient = rustMatrixClientFactory.create(client, sessionData, isMessageSearchAvailable())
-
-                // Apply enterprise hooks to the newly created client as soon as possible
-                clientEnterpriseHook(matrixClient)
-
-                newMatrixClientObservers.forEach { it.invoke(matrixClient) }
-                sessionStore.addSession(sessionData)
```
bmarty, line 174: Worth noting that `matrixClient.waitForKnownVerificationState()` was not called in this case.

libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/RustMatrixClientFactory.kt
```diff
@@ -117,14 +116,6 @@ class RustMatrixClientFactory(
...
         client.restoreSession(sessionData.toSession())
 
-        create(client, sessionData, isMessageSearchAvailable)
-    }
-
-    suspend fun create(
-        client: Client,
-        sessionData: SessionData,
-        isMessageSearchAvailable: Boolean,
-    ): RustMatrixClient {
```
jmartinesp, line 128: Note by doing this we're not reusing an already valid client and are forcing the creation of a different client, which can mean ~1s for opening the SDK DBs again I think. It's not that important since it happens on the login flow, but that's why we had 2 methods and tried to avoid destroying the current client in the log in flows.

bmarty: I see. So maybe we could have a second hook to update the client once it's logged in? Not sure if this will work for @richvdh .

To let this PR focused on this change, I'll extract the first commit to its own PR.

bmarty: I have measured 580ms spent on the new method `finalizeClientCreation`, but only 80 ms for the method `restoreSession` in it, so the impact is actually very small (?)

richvdh: In theory we could work around it, but for my usecase it would be much better to run the hook before the client is created. I don't need to know the userid though, so provided the hook actually gets called whenever a client is created, that would be fine.

bmarty: I do not understand, how the hook using `MatrixClient` can be called before `MatrixClient` is created? Are you talking about `ClientEnterpriseHook` or something else?

richvdh: I'm talking about `ClientBuilderEnterpriseHook`. I need a way to call methods on the underlying rust-side `ClientBuilder` each time we build a Client.

I could *probably* work around it by adding methods to the rust-side `MatrixClient` and calling them after the client is created, but I'd much prefer to avoid that.

features/enterprise/api/src/main/kotlin/io/element/android/features/enterprise/api/ClientBuilderEnterpriseHook.kt
```diff
@@ -13,6 +13,23 @@ import io.element.android.libraries.matrix.api.core.SessionId
...
+     * This method is invoked everytime a new [MatrixClientBuilder] is created.
+     *
+     * @param clientBuilder The [MatrixClientBuilder] to customize.
+     * @return The customized [MatrixClientBuilder].
+     */
+    suspend fun tweakClientBuilder(clientBuilder: MatrixClientBuilder): MatrixClientBuilder
+
+    /**
+     * Customize the [MatrixClientBuilder] for enterprise features.
+     * This method is invoked when a new [MatrixClientBuilder] is created to build a client for a specific session.
```
richvdh: would be good to be explicit if this is called **as well as** the other method, or **instead of** it.

(also imho the two implementations should have different names, since they serve somewhat different purposes, but ymmv)

bmarty: The documentation of the other method is explicit: "This method is invoked everytime a new [MatrixClientBuilder] is created.", but yes, I can be explicit here to.

Regarding the different name, I think the different signature is enough. Do you have a suggestion for a better name?

richvdh: personally I'd call them something like `beforeClientCreation` and `beforeClientCreationWithSession`. (h/t @andybalaam who helped me here)

(nb you don't actually *have* to "tweak" the client builder in the methods -- you could just use them as an informative hook and return the builder untweaked. Hence, it feels like `beforeX` is a better name.)

But I really don't mind, it's very much up to you!

bmarty: I have done the renaming in [a817b6f](https://github.com/element-hq/element-x-android/pull/7591/commits/a817b6f219a1b9a3b8acd20065a858e4616a7ed4)

ganfra (commented):
Ok for me like this. Have you tried using QRCode login? Is there some risks to lose data when closing the client quickly after login?

jmartinesp:
> Ok for me like this. Have you tried using QRCode login? Is there some risks to lose data when closing the client quickly after login?

AFAICT we haven't started any sync or encryption task at that point, so I *think* it should be ok.

ganfra:
> > Ok for me like this. Have you tried using QRCode login? Is there some risks to lose data when closing the client quickly after login?
> 
> AFAICT we haven't started any sync or encryption task at that point, so I _think_ it should be ok.

Is there not any verification thing done automatically when doing QR code login?

bmarty:
> Ok for me like this. Have you tried using QRCode login? Is there some risks to lose data when closing the client quickly after login?

I have checked again the QrCode login flow and it works fine, the session is created and verified.

ganfra (approved):

PR #7606, opened 2026-08-31
[release] Check that there is no unmerged PR with the label "Z-NextRelease", else exit

bmarty:
One more improvement on release script v2: Check that there is no unmerged PR with the label "Z-NextRelease", else exit


jmartinesp (approved):
Thanks!

PR #7412, opened 2026-08-12
Remove the legacy web-based account creation for matrix.org

jmartinesp:
## Content

Remove all code related to the workaround to register an account in matrix.org using the web form.

## Motivation and context

This was added as a temporary measure to allow matrix.org users to use Element X before it supported MAS, but it hasn't been needed for a very long time.

## Tests

There should be no changes in practice.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

- [x] I am aware of the [etiquette](https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#etiquette).
- This PR was made with the help of AI:
    - [ ] Yes. In this case, please request a review by Copilot.
    - [x] No.
- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly defines what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #7608, opened 2026-08-31
Remove ExternalSession, fallback login with Element Web is gone.

bmarty:
Finish cleanup started in https://github.com/element-hq/element-x-android/pull/7412

(extracted from #7591)


jmartinesp (approved):
Thanks! There are some lint issues to fix though.

PR #6477, opened 2026-03-26
Support multi-select polls

bxdxnn:
## Content

Support `max_selections` for polls

## Motivation and context

https://github.com/element-hq/element-meta/issues/2694
Fixes #6278

## Tests

- Test if you can create polls with `max_selections`
- Test if you can select only up to the count of `max_selections`

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

- This PR was made with the help of AI:
    - [x] Yes. In this case, please request a review by Copilot.
    - [ ] No.
- [x] Changes have been tested on an Android device or Android emulator with API 24
- [x] UI change has been tested on both light and dark themes
- [x] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly defines what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #7613, opened 2026-08-31
Support creating polls with multiple answers and send response with multiple answers

bmarty:
Complement for #6477 , see description there


jmartinesp (approved):
Thanks!

PR #7624, opened 2026-09-02
Improve detection of invalid placeholders in localazy strings

jmartinesp:
## Content

- Only detect `%` as an invalid placeholder if there are other placeholders present in the string.
- When there are no placeholders present and we find '%%', mention it as a possible mistake in a warning log.
- Only check for invalid placeholders once per string: previously this was incorrectly added inside a loop and ran several times.

## Motivation and context

When we use `Resources.getString(id)` without any params, this will internally just fetch the string and return it as is. In this case, having a `%` symbol is fine, and having `%%` would print both symbols when rendered.

However when `Resources.getString(id, arg1, arg2, ...)` is called, this will use `String.format(result, arg1, arg2, ...)` under the hood and a single `%` symbol will be detected as an invalid placeholder and crash in runtime. In this case we need to check if any valid placeholder exists in the string before checking whether `%` or `%%` is present.

## Tests

I just modified a few strings with several cases: no placeholder and a single %, no placeholder and several % (2-3), and the same with actual placeholders.

## Checklist

    - [x] No.

### DIFF
--- a/tools/localazy/checkForbiddenTerms.py
+++ b/tools/localazy/checkForbiddenTerms.py
@@ -59,11 +59,34 @@ _PLACEHOLDER_MATCHES = re.compile(r'%%|%[0-9$.A-Za-z]*')
 def find_invalid_placeholders(value):
     """Return the list of malformed placeholder tokens found in `value`."""
     invalid = []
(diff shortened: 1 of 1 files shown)


bmarty (approved):
Thanks for the iteration on this!

PR #7631, opened 2026-09-02
Make sure `enforce_element_pro` is taken into account in FOSS

jmartinesp:
## Content

Have `DefaultEnterpriseService` actually implement `isElementProEnforced`.

## Motivation and context

This was previously only implemented in the enterprise version of this interface, which missed the point: whether it's enabled or disabled there, any checks for this on Element Pro will pass since it *is* using Element Pro. The important target was FOSS, which I missed 🤦 .

## Tests

With Element X Android (FOSS), either point at some HS with the `enforce_element_pro` option in the Element well known file as true, or have a MITM proxy that overrides the response, and ensure the right error is displayed.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

    - [x] No.

### DIFF
--- a/features/enterprise/impl-foss/build.gradle.kts
+++ b/features/enterprise/impl-foss/build.gradle.kts
@@ -10,6 +10,7 @@ import extension.testCommonDependencies
  */
 plugins {
     id("io.element.android-compose-library")
+    alias(libs.plugins.kotlin.serialization)
 }
 
 android {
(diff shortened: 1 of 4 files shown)


bmarty (approved):
Thanks. Also `EnterpriseEnterpriseService.isElementProEnforced(...)` should always return `false`.

**EDIT** Actually I am not sure, since the call is also conditionned by `isEnterpriseBuild().not()`. Not sure what's best.

jmartinesp:
> Thanks. Also `EnterpriseEnterpriseService.isElementProEnforced(...)` should always return `false`.
> 
> **EDIT** Actually I am not sure, since the call is also conditionned by `isEnterpriseBuild().not()`. Not sure what's best.

Yeah, that's why I left it there. The behaviour works, and it might be handy to check that in some other case in the future. For now I think we can keep it.

PR #7639, opened 2026-09-03
Keep the push foreground service alive while notifications are fetched

Johennes:
## Content

`FetchPendingNotificationsWorker` currently stops the foreground service, and releases its wakelock, as soon as it starts, so resolving the pushed events (sliding sync, room key retrieval, decryption and, for ringing calls, waiting for the room state) runs under the WorkManager job budget only. This change keeps the service until the worker has finished, on every exit path, so the process stays in the foreground service state for the whole fetch.

Additionally, this change enables expedited work from Android 12 (API 31) instead of 13. `WorkManager` only falls back to a foreground service with a visible notification _below_ API 31, which is what the previous gate was meant to avoid but didn't achieve.

## Motivation and context

Once the worker starts, the only thing keeping the process alive is the `WorkManager` job. Expedited jobs run on a quota-based budget that Android sizes for work that finishes within a few minutes, and on API 31 and 32 the job is not expedited at all, so it can be deferred by Doze. In practice this has shown up as pushes that time out before the event is fetched, with the SDK tuned to very short timeouts to compensate. For ringing calls it means the call goes to a plain notification or nothing.

Keeping the foreground service running changes the process state for the whole fetch. While a foreground service is active the app is treated as foreground for scheduling purposes, so the expedited quota does not apply, Doze does not defer the work, and the wakelock keeps the CPU on while the sliding sync, key retrieval and decryption complete. The three-minute short-service limit still bounds it. Stopping the service in a `finally` also fixes two smaller issues: the invalid-input and network-retry paths previously left the service running until its own timeout, and a system cancellation of the worker skipped the stop altogether.

## Screenshots / GIFs

None.

## Tests

See unit tests.

## Tested devices

None.

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/workmanager/FetchPendingNotificationsWorker.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/workmanager/FetchPendingNotificationsWorker.kt
@@ -9,6 +9,7 @@
 package io.element.android.libraries.push.impl.workmanager
 
 import android.content.Context
+import android.os.Build
 import androidx.work.CoroutineWorker
 import androidx.work.WorkerParameters
 import dev.zacsweers.metro.AppScope
@@ -39,6 +40,7 @@ import io.element.android.services.analytics.api.AnalyticsService
 import io.element.android.services.analytics.api.finishLongRunningTransaction
 import io.element.android.services.analytics.api.recordTransaction
 import io.element.android.services.analyticsproviders.api.AnalyticsTransaction
+import io.element.android.services.toolbox.api.sdk.BuildVersionSdkIntProvider
 import io.element.android.services.toolbox.api.systemclock.SystemClock
 import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.flow.first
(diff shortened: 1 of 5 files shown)


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/workmanager/FetchPendingNotificationsWorker.kt
```diff
@@ -62,14 +64,26 @@ class FetchPendingNotificationsWorker(
...
+    }
+
+    private suspend fun doWorkInternal(): Result {
         // RunCatching for test in debug mode
         val sessionId = runCatchingExceptions {
             inputData.getString(SyncPendingNotificationsRequestBuilder.SESSION_ID)?.let(::SessionId)
         }.getOrNull() ?: return Result.failure()
 
-        // We can stop the foreground service and unlock the wakelock, since the work is now running and the device should be kept awake
-        fetchPushForegroundServiceManager.stop()
```
jmartinesp, line 71: I think we don't want this for *every* API, if the only affected ones are 31 and 32: the foreground service is there only to make sure  the just scheduled worker runs as fast as possible, and usually it doesn't even have enough time to display a 'fetching push notifications...' notification.

If we change this, every time you receive a push, this ongoing notification will be displayed for several seconds instead while the notifications are fetched, which seems very disruptive.

jmartinesp (changes requested):

libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/workmanager/FetchPendingNotificationsWorker.kt
Johennes: Good points. I have changed things around in 5a8c6eb7c66022c80890f9fb5156d18b103dc64f so that:

- below Android 12, the service is released as soon as the work runs, as before
- on Android 12+ the service is released right after the results have been handed over, before the opportunistic sync

As far as I understand, on 12+ the system defers a foreground service's notification by 10 seconds. So with the sync no longer inside the window the notification should only appear when resolving the events takes longer than that, which is the slow case where keeping the process in the foreground service state actually helps.

jmartinesp: > As far as I understand, on 12+ the system defers a foreground service's notification by 10 seconds. 

Oh, TIL. Then I agree this change could be useful: worst case scenario, it won't do anything, but it *should* improve reliability in case WorkManager decides the worker shouldn't be expedited (too many notifications?).

jmartinesp: Also, if it turns out this new behaviour is problematic we can always revert it before the next RC. We'll test this in the nightlies.

jmartinesp (approved):

PR #7640, opened 2026-09-03
Make NotificationResultProcessor.emit wait until the batch is processed

Johennes:
## Content

This change makes `NotificationResultProcessor` own its lifecycle instead of relying on the push handler having been constructed first. `emit` starts the collector if it is not running, and results go through an unlimited channel, which holds items until a receiver takes them, so nothing is lost while the collector is starting or between it being launched and actually subscribing. `emit` also suspends until its batch has been processed, so the worker only completes once the notification or ringing call has been set up, and the wakelock or foreground service the worker holds covers the whole path rather than ending while the batch is still queued. Finally, a batch that throws while being processed is logged instead of cancelling the collector, which previously stopped all further notifications for the life of the process.

## Motivation and context

`NotificationResultProcessor` currently hands resolved push events from the fetch worker to a collector running on the app scope. That collector is only started from `DefaultPushHandler`, which is only created when a push receiver runs in the current process, and the shared flow in between has no replay, so values emitted while nobody is subscribed are dropped. A `FetchPendingNotificationsWorker` that `WorkManager` runs in a fresh process, for example a retry after the process was killed or an expedited job that was deferred, therefore fetches and decrypts the events and then emits them into nothing. No notification is shown and nothing is logged.

## Screenshots / GIFs

None.

## Tests

See unit tests.

## Tested devices

None.

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationResultProcessor.kt
+++ b/libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/NotificationResultProcessor.kt
@@ -12,6 +12,7 @@ import dev.zacsweers.metro.ContributesBinding
 import dev.zacsweers.metro.SingleIn
 import io.element.android.features.call.api.CallData
 import io.element.android.features.call.api.ElementCallEntryPoint
+import io.element.android.libraries.core.extensions.runCatchingExceptions
 import io.element.android.libraries.di.annotations.AppCoroutineScope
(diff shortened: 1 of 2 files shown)


bmarty:
This makes sense, but I'd like @jmartinesp to have a second look please 🙏

jmartinesp (approved):
Yes, it makes sense, thanks. I think the problem here is when I wrote this I thought `extraBufferCapacity` was applied even when the flow was not subscribed to, nice job spotting the issue!

PR #7551, opened 2026-08-25
Say so when this session has no secrets to verify another one with

hayaksi1:
## Content

When an incoming verification request arrives, this session offers "Start verification" whatever
state its own cross-signing secrets are in. If the recovery state is `INCOMPLETE` the secrets are not
on this device, so the flow can only fail — and it fails at the very end, after the user has compared
the emoji.

The screen now shows the alert step from the design instead — "Verification with this device is
currently unavailable" / "Please use another device or recovery key." — with only the Ignore action.
This applies to the initial step only; a verification already in progress is untouched, as is the
request-acknowledgement the presenter already sends.

`RecoveryState.INCOMPLETE` is the same signal the room list already uses to raise the "confirm your
recovery key" banner, so no new notion of "missing secrets" is introduced.

## Motivation and context

Part of #5344, using the copy from the design attached to the issue.

## Tests

- `IncomingVerificationPresenterTest` — a session with `RecoveryState.INCOMPLETE` reaches
  `Step.Unavailable` and never calls `acceptVerificationRequest`; a session with `ENABLED` still
  reaches `Step.Initial`.
- Negative control: with the step mapping reverted, `present - a session missing its secrets cannot
  be used to verify and offers no verification action` fails, so the test guards the change.
- `./gradlew :features:verifysession:impl:testDebugUnitTest :features:verifysession:impl:ktlintCheck :features:verifysession:impl:detekt` — green.
- The new state is added to `IncomingVerificationStatePreviewParam`, so the existing
  `@PreviewsDayNight` previews cover it and the screenshot was recorded by CI.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): unit tests and Composable previews

## Checklist

- [x] I am aware of the [etiquette](https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#etiquette).
- This PR was made with the help of AI:
    - [x] Yes. In this case, please request a review by Copilot.
    - [ ] No.
- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [x] UI change has been tested on both light and dark themes
- [x] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly defines what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #7644, opened 2026-09-04
Revert "Say so when this session has no secrets to verify another one with"

bmarty:
Reverts element-hq/element-x-android#7551 as this specific case has to be handled by the SDK first.


ganfra (approved):

bmarty:
(merging ignoring unrelated flaky test)

PR #7649, opened 2026-09-04
Replace Danger with actions/github-script

bmarty:
Same idea as element-hq/element-x-ios#6080: reimplement the PR checks with `actions/github-script`, with zero external dependencies.

- **PR metadata checks** move from `tools/danger/dangerfile.js` to a new `PR Checks` workflow, running on `pull_request_target` with no checkout and no secret. All the existing checks are ported as-is, and a single comment is kept up to date (and deleted once the PR is clean).
- **Lint reporting** moves from `tools/danger/dangerfile-lint.js` to a step in the quality workflow, which parses the ktlint/detekt/lint XML reports and emits inline annotations instead of comments.
- `danger.yml`, both dangerfiles, the `Gemfile` and `docs/danger.md` are deleted.

Because neither part needs `DANGER_GITHUB_API_TOKEN` anymore, the checks now also run on pull requests from forks — `danger.yml` was skipped there entirely.

Three bugs fixed along the way:
- the png warning was dead code (a `filter` callback with no `return`), so it never fired;
- `upload_reports` had no `always()`, so it was *skipped* whenever ktlint/detekt/lint failed — i.e. exactly when there was something to report;
- Android lint's `lint-results-*.xml` is not Checkstyle, so `danger-plugin-lint-report` silently never parsed it. Both formats are handled now.

### Needed before/after merge

- Branch protection: the required check `Danger main check` no longer exists, it should be replaced by `PR Checks`. Note that `danger.yml` also ran on `merge_group`, which `pull_request_target` has no equivalent for (same trade-off as iOS).
- The `DANGER_GITHUB_API_TOKEN` secret must stay: it is still used by `recordScreenshots.yml`, `sync-localazy.yml` and `gradle-wrapper-update.yml`. Renaming it can be a follow-up.


jmartinesp (approved):
🤞

PR #7657, opened 2026-09-07
Sync Strings

ElementBot:


jmartinesp:
I just removed a problematic translation from localazy missing a placeholder and triggered a new sync.

jmartinesp (approved):
This looks better now. I haven't fully checked the large diff though.

PR #7661, opened 2026-09-07
Fix two crashes triggered by opening the Showkase browser

bmarty:
TL;DR; This PR fixes crashes when opening the Showkase browser.

Showkase screen now have a root menu:

image

The rest is AI generated v.

## Content

Two independent crashes, both surfaced by opening the Showkase browser from the developer options.

**1. `IllegalStateException: Scoped provider was invoked recursively` (`DefaultNotificationChannels`)**

`DefaultNotificationChannels` created its channels from `init`, and that work blocks with `runBlocking`. `runBlocking` steals pending work from the calling thread's coroutine event loop — deliberately, and [even when a dispatcher is passed](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/concurrent/src/Builders.concurrent.kt), since it picks up `ThreadLocalEventLoop.currentOrNull()`.

Appyx resolves nodes synchronously from inside an active event loop on the main thread (`EventLoop.processUnconfinedEvent` is in the stack trace). So when node creation reached this `@SingleIn` binding, the stolen continuation re-entered the DI graph and asked for `NotificationChannels` again while Metro's `DoubleCheck` was still uninitialised. That guard is a reentrant JVM monitor, so a second instance was built and Metro threw.

Fixed by moving the work to an idempotent `ensureChannelsCreated()` called from every public entry point, so the provider is published before anything can block. `getSilentChannelId()` is now overridden too — it used the interface default, so nothing guaranteed that channel existed.

**2. `IllegalArgumentException: Navigation destination … SHOWKASE_CATEGORIES cannot be found`**

Showkase 1.0.5 builds a reduced nav graph when only one of its three categories is populated, and that graph has no `SHOWKASE_CATEGORIES` destination. The groups screen navigates to it on back press regardless — its `onRootScreen` flag is computed before the `NavController` has a graph, so it is never `true`. 1.0.5 is the latest release, so there is no version to upgrade to.

Our root only had components because `:libraries:compound` declared a `@ShowkaseTypography` token set **and** a `@ShowkaseRoot` but never applied the Showkase processor, so none of it was ever generated — those annotations had no effect at all. This PR applies the processor there and drops the unused root, leaving `DesignSystemShowkaseRootModule` as the single root. It now aggregates the typography tokens, which populates a second category and keeps the categories screen in the graph.

Side effects worth reviewing: the browser gains a Typography category, `compound`'s own previews are now aggregated as components (166 → 338), and Showkase's categories screen lists an empty `Colors (0)` row since we have no `@ShowkaseColor` tokens. A unit test asserts two categories stay populated, since dropping back to one would silently bring the crash back.

## Motivation and context

Crash 1 is reported in element-hq/element-x-android-rageshakes#15042 (nightly `26.09.1`, sha `66c50285`, i.e. current `develop`).

Crash 2 is reproducible on `develop`: open the Showkase browser, then press back.

Neither is caused by the other; fixing only the navigation crash would leave the DI crash in place.

## Screenshots / GIFs

No UI change in the app itself. The only visual difference is inside the Showkase developer tool, which now opens on its categories screen (Components / Colors / Typography) instead of going straight to the component groups.

## Tests

Unit tests:

- `:libraries:push:impl:testDebugUnitTest` — 291 tests pass. Four existing tests asserted the channel work happened in `init`, so they were renamed `init - ` → `first use - ` and now trigger an accessor; one of them was silently vacuous without that. Added: construction creates no channel, and first use creates exactly 4 channels once.
- `:libraries:designsystem:testDebugUnitTest` — new `DesignSystemShowkaseRootModuleTest` asserts at least two Showkase categories are populated.
- `ktlintCheck` clean on `:libraries:push:impl`, `:libraries:compound`, `:libraries:designsystem`.

Manual, on the emulator, running the exact sequence that crashed before:

- Step 1: launch the app, open the developer options from the onboarding screen.
- Step 2: tap "Open Showkase browser" → the browser now opens on the categories screen.
- Step 3: tap "Components" → the component groups screen opens.
- Step 4: press back → returns to the categories screen (this is where it crashed before).
- Step 5: press back again → the browser closes and returns to the app.
- Step 6: checked `adb logcat -b crash` throughout — zero fatal exceptions.

Not verified on device: crash 1 depends on a DI/event-loop timing window that I could not reproduce locally, so that fix is argued from the stack trace and the coroutines/Metro sources rather than from a live repro.

## Tested devices

- [ ] Physical
- [x] Emulator

## Checklist

    - [ ] No.

Left unchecked on purpose: tested on API 36 rather than API 24; no UI change, so the light/dark, accessibility and screenshot items do not apply.


libraries/push/impl/src/main/kotlin/io/element/android/libraries/push/impl/notifications/channels/NotificationChannels.kt
```diff
@@ -105,8 +105,25 @@ class DefaultNotificationChannels(
...
+     * still publishing this `@SingleIn` provider lets that stolen work re-enter the graph and build
+     * a second instance, which Metro rejects with "Scoped provider was invoked recursively".
+     */
+    private fun ensureChannelsCreated() {
+        if (channelsCreated) return
+        synchronized(recreateLock) {
+            if (channelsCreated) return
+            // Published before creating, not after: the `runBlocking` below can still re-enter this
+            // method, and channel creation is idempotent, so a no-op is better than recursing.
+            channelsCreated = true
```
jmartinesp, line 124: Can it? The code just seems to check and maybe set some values in `AppPreferencesStore`.

jmartinesp:
2 things about these changes:

1. We're calling this in `runBlocking` because we need to perform IO operations and this code in synchronous. However, looking at the different callers of the methods in `NotificationChannels` I *think* all are either suspend functions or can be turned into one with not a lot of effort. Maybe it's better to make the initialization of these channels async with a proper `Mutex`?
2. I'm ok with fixing the Showkase error, but I'm pretty sure no one is using Showkase anymore in the project. AFAICT no designer uses it, we don't use it (at least I don't) and I don't think anyone else is interested. Maybe it's time to remove it? There are alternatives to generate something like the interactive MacOS app iOS has to showcase their components, but I believe they have some constraints that would make them difficult to use by us.

bmarty:
Happy to remove Showkase entirely. Will do in a separate PR.

PR #7672, opened 2026-09-09
Send to the conversation picked in the share sheet

Xant3s:
## Content

The Direct Share row of the system share sheet already offered conversations, but picking one did nothing with that choice: the app still showed the account picker (when there is more than one account), then the room picker. The system names the chosen shortcut in `Intent.EXTRA_SHORTCUT_ID`, which nothing read. The share now carries that id, and the session it belongs to resolves the room it was built for.

Follows #7574, which declared the `<share-target>` that makes the row appear.

## Motivation and context

Part of #3835

## Tests

Unit tests cover the id format (`ConversationShortcutIdTest`), the id reaching `ShareIntentData` for both plain text and uris (`DefaultShareIntentHandlerTest`), and the presenter sending without a selection — including the case where the shortcut belongs to a different account than the one handling the share (`SharePresenterTest`).

Manually:

- Send a message in a room, so its conversation shortcut is published.
- Share text from another app and pick that conversation in the Direct Share row.
  The message is sent to that room, with no account or room picker.
- Same with an image.
## Tested devices

- [ ] Physical
- [X] Emulator
- OS version(s): API 34 (Android 14)

## Checklist

    - [ ] No.
- [X] UI change has been tested on both light and dark themes
- [X] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [X] Pull request includes screenshots or videos if containing UI changes

### DIFF
--- a/appnav/src/main/kotlin/io/element/android/appnav/RootFlowNode.kt
+++ b/appnav/src/main/kotlin/io/element/android/appnav/RootFlowNode.kt
@@ -65,6 +65,7 @@ import io.element.android.libraries.matrix.api.core.toRoomIdOrAlias
 import io.element.android.libraries.matrix.api.permalink.PermalinkData
 import io.element.android.libraries.oauth.api.OAuthAction
 import io.element.android.libraries.oauth.api.OAuthActionFlow
+import io.element.android.libraries.push.api.notifications.conversations.conversationShortcutRoomId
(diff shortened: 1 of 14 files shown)


Susurrus:
I just opened #7704 yesterday having seen this issue so linking it here.

PR #7667, opened 2026-09-08
Try improving renovate issues with Maven Central rate limiting

jmartinesp:
## Content

- Try not to make too many concurrent requests to Maven Central: the logs display rate limit responses followed by OK responses, which could mean we're hammering the Maven Central repo with too many concurrent requests.
- Also, try not hitting Maven Central for `androidx.` libraries.

## Motivation and context

Fix Renovate being rate-limited, which makes it hard to update dependencies sometimes.

## Tests

I think the only way to test this is to merge it 🫤 .

## Checklist

- [x] I am aware of the [etiquette](https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#etiquette).
- This PR was made with the help of AI:
    - [ ] Yes. In this case, please request a review by Copilot.
    - [x] No.
- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [ ] Pull request title will be used in the release note, it clearly defines what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #7677, opened 2026-09-09
Improve Renovate configuration to fix issues with Maven Central

jmartinesp:
## Content

- Don't group `androidx` dependencies.
- Make sure `android.*`, `androidx.*` and `com.android.tools*` dependencies are only fetched from Google's Android repo. It should make both renovate and gradle sync faster.
- Tweak the concurrent requests options to higher values that still don't seem to trigger the Maven Central rate limit.

## Motivation and context

https://github.com/element-hq/element-x-android/pull/7667 didn't really work as expected.

## Tests

I tested this locally with `--dry-run` and it worked fine AFAICT.

## Checklist

    - [x] No.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -25,7 +25,19 @@ dependencyResolutionManagement {
                 includeModule("com.github.philburk", "jsyn")
             }
         }
-        google()
+        // Check for official Android-related packages only in Google's maven repo
+        exclusiveContent {
+            forRepository {
+                google()
+            }
+            filter {
+                includeGroupByRegex("android\\..*")
+                includeGroupByRegex("androidx\\..*")
+                includeGroupByRegex("com\\.android\\.tools.*")
+                includeGroupByRegex("com\\.google\\.firebase.*")
+                includeGroupByRegex("com\\.google\\.android.*")
+            }
+        }
         mavenCentral()
         maven {
             url = uri("https://repo1.maven.org/maven2/")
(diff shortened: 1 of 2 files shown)


jmartinesp:
Let's try setting it as a proper PR instead of a draft.

jmartinesp:
@bmarty seems fine I think?

---

image

All these were present before my changes AFAICT. We can fix them in a separate PR or as part of this one, your call.

bmarty:
> @bmarty seems fine I think?
> 
> All these were present before my changes AFAICT. We can fix them in a separate PR or as part of this one, your call.

I checked a few weeks ago, these are 2 GitHub actions, we can ignore the warning. We are already using the latest version (which are a bit old).

settings.gradle.kts
```diff
@@ -25,7 +25,19 @@ dependencyResolutionManagement {
...
+        exclusiveContent {
+            forRepository {
+                google()
+            }
+            filter {
+                includeGroupByRegex("android\\..*")
+                includeGroupByRegex("androidx\\..*")
+                includeGroupByRegex("com\\.android\\.tools.*")
+                includeGroupByRegex("com\\.google\\.firebase.*")
+                includeGroupByRegex("com\\.google\\.android.*")
```
bmarty, line 38: I wonder why there are 5 items in this list and only 3 in `"matchPackageNames"` in the `renovate.json5` config file.

bmarty (approved):
Thanks, just one point to clarify.

settings.gradle.kts
jmartinesp: Ah, good catch. I started with the top 3 at the beginning in both sides, then realised Gradle needed the other two, but I never added them to renovate. Will do it now!

PR #3125, opened 2024-07-02
Improve screenshot testing with ComposablePreviewScanner

jmartinesp:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [x] Other : Improve CI processes

## Content

- Replaces `Showkase` for screenshot tests with [ComposablePreviewScanner](https://github.com/sergio-sastre/ComposablePreviewScanner). Showkase is only used now for the browser in `:libraries:designsystem`.
- Use reflection to 'fix' the generated screenshot names with Paparazzi so we don't have any more `_null_1_2_` suffixes in the screenshot filenames.
- Add a very crude 'sharding' for recording and verifying screenshots, which allows for parallelisation and should improve CI and local test times.

## Motivation and context

Improve DevX. It should also fix the case where adding/removing a single preview in a module renamed every single other one.

## Screenshots / GIFs

There are a few in this PR 😇 .

## Tests

We should test this against the Element Gallery.

## Tested devices

-

## Checklist

- [ ] Changes have been tested on an Android device or Android emulator with API 23
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly define what will change for the user
- [x] Pull request includes screenshots or videos if containing UI changes
- [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
- [x] You've made a self review of your PR

PR #7678, opened 2026-09-09
Remove Showkase and its in-app component browser

bmarty:
Showkase stopped being used for screenshot testing in #3125, when preview enumeration moved to `ComposablePreviewScanner`. All that was left was the debug-only "Open Showkase browser" developer option, plus dead wiring: an inert `@ShowkaseRoot` in `:libraries:compound`, an unused dependency in `:tests:uitests`, a ProGuard keep file, and stale Kover excludes and docs.

Removing it drops the KSP processor from `:libraries:designsystem`, which was generating 327 files per variant.

Please see the [contributing etiquette](https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#etiquette).


jmartinesp (approved):
🔥

PR #7686, opened 2026-09-10
Add enter animation to new timeline events

jmartinesp:
## Content

Add the `WithEnterAnimation` wrapper to add an slide-up + fade in animation to timeline events with SYNC or LOCAL origin so new items either coming from sync or sent by the current users animate when added to the timeline. Other items will work the same.

## Motivation and context

Implement https://github.com/element-hq/element-x-android/issues/6966.

FWIW, I tried a more thorough implementation with `Modifier.animateItem()` but there were lots of weird animations, as we had before: 

- Grouped events overlapping with any events above them when being expanded.
- When you send several items in a row, the latest event with the 'sent' indicator will collapse with an animation, and the new item being added will be animating from what it looks like the middle of that event, which makes them overlap in an awkward way.
- Probably more I haven't noticed.

So I decided to drop that implementation and do something simpler.

## Screenshots / GIFs

https://github.com/user-attachments/assets/731a1d2e-0b03-4c82-87d4-bcc85140aaf7

## Tests

Test sending and receiving new events in the timeline, as well as redacting them or paginating.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 15

## Checklist

- [x] I am aware of the [etiquette](https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#etiquette).
- This PR was made with the help of AI:
    - [ ] Yes. In this case, please request a review by Copilot.
    - [x] No.
- [ ] Changes have been tested on an Android device or Android emulator with API 24
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-x-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request title will be used in the release note, it clearly defines what will change for the user
- [ ] Pull request includes screenshots or videos if containing UI changes
- [x] You've made a self review of your PR

PR #7688, opened 2026-09-10
Add smooth animation for new messages

### LINKED ISSUE #6966: [Task] Add a slide-up animation on new messages
See https://github.com/element-hq/element-meta/issues/3238

bxdxnn:
## Content

Refactor the timeline row logic a bit to allow a smooth timeline animation for new rows 
## Motivation and context

Fixes #6966

## Tests

- Write a message

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


jmartinesp:
This is an improvement from what we have, at the cost of making the timeline logic a bit more complex and presumably less performant - extracting read receipts when the timeline items change, I don't think it remembers anything to prevent new computations?

Also, it still has the issue about expanding a group of events overlapping with the items above it. It's not terrible, but it does look weird:

https://github.com/user-attachments/assets/d49e9b4c-313d-4a43-baad-39a3b6b55608

I imagine there can be other issues like this, where components suddenly changing sizes can cause overlaps, but I couldn't find any, so maybe it's the only case where this could happen.

So I'm a bit torn: it feels smoother than my implementation in https://github.com/element-hq/element-x-android/pull/7686, but it's also more prone to having animation issues.

bxdxnn:
@jmartinesp OK, I'll fix the issues you've mentioned. Are you generally OK with this PR though?

jmartinesp:
I'm not sure if splitting the timeline items and the read receipts is a good idea, but it seems like it's the only thing that could make inserting an event seamless: I tested some animations a bit more without splitting the events and read receipts, and while they looked better than before, there's a point where the previous event has lost its sent state and the new one receives it where the timeline item sizes change and there's more movement than I'd like.

Also, with the code in this PR there's a weird issue with read receipts that happens randomly: the animations for them are smooth at first, but then at some point they start being overlapped by the item above them and they kind of 'slide down' from them. I guess this too is caused by the reversed LazyList layout 🫤 :


https://github.com/user-attachments/assets/097646a6-ed51-4d85-bd37-6d3761e71780

bxdxnn:
@jmartinesp I've fixed the regressions you've mentioned, seems to be working so far. I'll regenerate screenshots to see what's up.

jmartinesp:
This looks really good, the `TimelinePlacementAnimationCoordinator` trick is quite clever. I tried achieving something similar without separating the read receipts, but I couldn't, there was always some 'spring' animation because of the read receipts disappearing resizing the actual row.

I'm still not a fan of separating the read receipts into their own items, and I'm a bit scared of the complexity of the mapper, but the result is **really** good. I'd be in favour of merging this, but I'd want to know what @bmarty thinks first.

jmartinesp:
We've discussed this with the team, and given we're about to start working on a multi-message selection feature that could conflict with this PR, we'll implement that first and then try to figure out if this solution still fits, since having the read receipts as a separate timeline item could be a problem in this case.

frebib:
I've been running this for a short while and I really love the new fluidity it brings to the app. It's certainly not perfect, but it's a huge improvement over the very clunky pop-in that the app has before this. I'd love to see this branch kept up-to-date

bxdxnn:
> It's certainly not perfect

Maybe you can point out (send a recording or something) what you didn't like so that I can fix it?

frebib:
I don't have the patch applied since my rebase today as there were too many conflicts and I couldn't be bothered to resolve them. From memory
- Adding a reaction causes the message to jump because the reaction isn't size-animated. This causes messages to briefly overlap
- The loading spinner at the top of the list when getting to the "top" of a room as it backpaginates slides across new messages as they appear. It'd probably be better if it just disappeared (?)
- In general things get kinda funky in rooms with multiple users. It's not _bad_ per-se, things just feel a bit off. I can't really quantify this one; it's more just a gut feeling. In DM rooms it works very well.

PR #7692, opened 2026-09-11
Fix blank screen at launch when the homeserver expires the sliding sync session

### LINKED ISSUE #7689: Blank screen at launch after the homeserver expires the sliding sync connection (400 M_UNKNOWN_POS)
### Steps to reproduce

1. Sign in to Element X Android and use it normally.
2. Leave the app unopened for more than 7 days. On several Android OEM builds (Xiaomi, realme, TECNO) aggressive battery management freezes the app, so it never wakes up in the background even with push configured.
3. Open the app.

### Outcome

**What did you expect?**

The app notices that the homeserver no longer knows its sliding sync position, restarts the sync from scratch and renders the room list.

**What happened instead?**

The app stays on a blank white screen indefinitely. The only workaround our users found is to clear app data / reinstall and sign in again.

That workaround is expensive: a fresh session restores the timeline from key backup, so every older message then carries the "The authenticity of this encrypted message can't be guaranteed on this device." shield.

**Server side**

Synapse deletes sliding sync connections that have not been used for `CONNECTION_EXPIRY = 7 days` (`synapse/storage/databases/main/sliding_sync.py`, cleanup runs hourly). After that the client's request

```
POST /_matrix/client/unstable/org.matrix.simplified_msc3575/sync?pos=<stale>&timeout=0
```

is answered with `400 M_UNKNOWN_POS` ("Unknown position"). The app does not appear to recover from that response.

**Scale on one self-hosted server** (Synapse 1.160.0, ~250 users, ~470 active sessions):

- 12 to 18 occurrences per day
- 11 distinct Android devices over 48 hours
- zero iOS devices, although iOS is roughly a third of the sessions here
- affected versions range from 26.05.1 to 26.09.1, so upgrading does not help

### Your phone model

Xiaomi (23028RNCAG, 2409BRN2CY, 23053RN02Y, 22120RN86G, M2006C3MNG, 24117RN76O), realme RMX3890, TECNO CL7, Samsung SM-A065F

### Operating system version

Android 10 through 16

Element X Android 26.05.1 to 26.09.1 (Google Play). Every observed version is affected, including the newest.

### Homeserver

Self-hosted Synapse 1.160.0 with Matrix Authentication Service

### Will you send logs?

No

No

bmarty:
## Content

Two commits:

1. `emptyNode()` gets a `delayBeforeShowingContent` parameter. When it is finite, a progress indicator and a "Loading…" label fade in after that delay (reusing the existing `DelayedVisibility` helper). It defaults to `Duration.INFINITE`, so the splash screen is unchanged. The two placeholders which are not guaranteed to be left — the FTUE one and the `LoggedInFlowNode` one — now use a 0.5 second delay.
2. `SyncOrchestrator` restarts the sync service when it is in the `Error` or `Terminated` state, with an exponential backoff from 1s to 30s.

## Motivation and context

Fixes #7689.

Synapse deletes sliding sync connections that have been unused for 7 days. When an OEM battery manager has kept the app frozen for longer than that, the next sliding sync request is answered `400 M_UNKNOWN_POS`.

The Rust SDK handles that: it resets `pos` and, in offline mode, restarts the sync. When that recovery gives up, the sync service ends up in `Error`/`Terminated` and stops for good — it is then up to the application to call `start()` again. Element X iOS does it ([`ClientProxy.swift`](https://github.com/element-hq/element-x-ios/blob/develop/ElementX/Sources/Services/Client/ClientProxy.swift): `if case .error = state { restartServices() }`), we did not: `SyncOrchestrator` only restarted the sync service when its state was `Idle`, and it is the only `startSync()` caller in the app. So the sync stayed dead for the rest of the process lifetime.

Nothing recovered from there, and that is what made the screen *blank* rather than stale: the sync never reaches `Running`, so `waitForE2eeInitializationTasks()` never returns, `sessionVerifiedStatus` stays `Unknown`, and the FTUE is stuck on `WaitingForInitialState`, which rendered an empty Node. Clearing the app data was the only workaround, which is expensive for users since it loses the local store.

The backoff is only reset once the sync service has been running for a while, because `startSync()` moves the state to `Running` right away — resetting on any `Running` value would defeat it.

## Screenshots / GIFs

Covered by the new `EmptyViewLoadingPreview` Composable Preview.

## Tests

- 6 new tests in `SyncOrchestratorTest` covering: restart on `Error`, <del>restart on `Terminated`</del>, growing delay when the sync keeps dying, delay reset after a steady recovery, and no restart while backgrounded or without network.
- Both were checked against negative controls: removing the `Error`/<del>`Terminated`</del> branch fails the positive tests, and flattening the backoff fails the "delay increases" test.
- Not reproduced end to end against a live homeserver yet.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.


appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
```diff
@@ -324,7 +325,9 @@ class LoggedInFlowNode(
 
     override fun resolve(navTarget: NavTarget, buildContext: BuildContext): Node {
         return when (navTarget) {
-            NavTarget.Placeholder -> emptyNode(buildContext)
+            // Rendered while the FtueState is Unknown. Render a loading state rather than nothing at all,
+            // since there is no guarantee that this state will be left.
+            NavTarget.Placeholder -> emptyNode(buildContext, delayBeforeShowingContent = 500.milliseconds)
```
ganfra: PR description says 1 seconds, here it's 500.milliseconds. Guess it's ok, just raising this.

appnav/src/main/kotlin/io/element/android/appnav/session/SyncOrchestrator.kt
```diff
@@ -143,8 +181,34 @@ class SyncOrchestrator(
...
+ */
+private fun SyncState.isDead() = this == SyncState.Error || this == SyncState.Terminated
+
+/**
+ * Exponential backoff, from [FIRST_RESTART_DELAY], doubling on every attempt, up to [MAX_RESTART_DELAY].
+ */
+private fun restartDelay(attempt: Int): Duration {
+    val exponent = attempt.coerceIn(0, 5)
+    return (FIRST_RESTART_DELAY * (1 shl exponent)).coerceAtMost(MAX_RESTART_DELAY)
+}
```
ganfra, line 208: This is hard to understand

appnav/src/main/kotlin/io/element/android/appnav/session/SyncOrchestrator.kt
```diff
@@ -143,8 +181,34 @@ class SyncOrchestrator(
...
+/**
+ * Duration during which the sync service has to stay running before the restart backoff is reset.
+ */
+@VisibleForTesting
+internal val RESTART_BACKOFF_RESET_DELAY = 30.seconds
+
+/**
+ * Return true if the sync service has stopped and will not restart by itself.
+ */
+private fun SyncState.isDead() = this == SyncState.Error || this == SyncState.Terminated
```
ganfra: Not sure we should restart automatically when Terminated? Also, SDK is already supposed to retry on his own, should we not try to fix this in the SDK if there is something fishy?

appnav/src/main/kotlin/io/element/android/appnav/LoggedInFlowNode.kt
bmarty: Description updated.

appnav/src/main/kotlin/io/element/android/appnav/session/SyncOrchestrator.kt
bmarty: Using the `pow` function is not better I think:
```kotlin
return (FIRST_RESTART_DELAY * (2.toDouble().pow(exponent)).toInt()).coerceAtMost(MAX_RESTART_DELAY)
```

We could also have a list of 6 hard-coded values if you prefer.

ganfra (approved):

appnav/src/main/kotlin/io/element/android/appnav/session/SyncOrchestrator.kt
ganfra: What about investigating in the sdk?

PR #7701, opened 2026-09-14
Reduce parallel JVMs spawned for tests

jmartinesp:
## Content

- For normal unit tests, reduce the number of parallel forks to 1 again.
- Split Roborazzi config and its step in the CI.

## Motivation and context

A large amount of our tests are Robolectric ones, which needs to be set up once per module and JVM, and takes 3-10s to do so, and it's slower the more JVMs are using the disk in parallel to do so.

When tested which was faster locally, if running all the tests using 1 or 4, 8 or the max available number of JVM forks, 1 was almost 2.5x faster than the slowest one.

## Checklist

    - [ ] No.


jmartinesp:
Note this would get even better with Robolectric `4.17.1` (not released yet), which contains a kind of shared cache between JVMs. Although that makes you think then spawning several JVMs would get faster (and it does), having a single JVM per module is still way faster.

ganfra (approved):

PR #7702, opened 2026-09-14
Fix flaky test in `TimelinePresenterTest`

jmartinesp:
## Content

Provide a coroutine scope and dispatcher values to `TimelineController` so we can explicitly control what runs when in unit tests. It's also more correct than the previous implementation, I think.

## Motivation and context

Fix a flaky test in `TimelineControllerPresenter`: `present - a forward pagination while the focused event is not rendered yet is ignored`.

This test usually worked when running in isolation, but when running the whole test suite, it sometimes failed.

## Checklist

    - [ ] No.


features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineController.kt
```diff
@@ -47,8 +49,10 @@ import java.util.Optional
 class TimelineController(
     private val room: JoinedRoom,
     @LiveTimeline private val liveTimeline: Timeline,
+    @RoomCoroutineScope private val roomCoroutineScope: CoroutineScope,
+    dispatchers: CoroutineDispatchers,
```
bmarty: Since this constructor is never injected, I believe that all the annotations could be just removed (`@SingleIn`, `@ContributesBinding`, `@RoomCoroutineScope`).

bmarty (approved):
Thanks!

features/messages/impl/src/main/kotlin/io/element/android/features/messages/impl/timeline/TimelineController.kt
jmartinesp, line 53: I believe it's injected, no? Either as `TimelineProvider` or `TimelineController`, like in `MessagesFlowNode `.

```
element-x-android/app/src/main/kotlin/io/element/android/x/di/AppGraph.kt:23:11 [Metro/MissingBinding] No binding found for TimelineController

  trace (in
  io.element.android.x.di.AppGraph.Impl.SessionGraphImpl.RoomGraphImpl.SpaceFlowGraphImpl):
      TimelineController is injected at
          io.element.android.features.messages.impl.MessagesFlowNodeAssistedFactory.create()
```

PR #7713, opened 2026-09-15
Do not report detekt LargeClass issues on test classes

bmarty:
detekt's default config excludes test source sets from `TooManyFunctions` and `StringLiteralDuplication`, but not from `LargeClass`. Add that exclusion and remove the 24 now-redundant `@Suppress("LargeClass")` from test files. The 2 suppressions in production code are kept.


jmartinesp (approved):
Makes sense, thanks!

PR #7714, opened 2026-09-15
feat(mediaviewer): restore floating minimized video player

### LINKED ISSUE #5421: Feature Request: System-wide floating video player (overlay service)

aliIsazadeh:
## Content

Restores the floating/minimized video player (YouTube-style PiP) for videos in the media viewer, allowing playback to continue in a draggable overlay while navigating the rest of the app.

## Motivation and context

Closes #5421.

This is a clean redo of #5423, which had to be closed after an accidental `develop` merge into the feature branch left the commit history unrecoverable. This PR reapplies the feature from scratch with the review feedback from that PR addressed:

- Replaced `VideoDataRepository` (a presenter-scoped singleton that leaked memory and used a non-thread-safe map) with direct `Uri` passing through the service `Intent`.
- Video is resolved from the downloaded `LocalMedia.uri` instead of falling back to `MediaSource.getUri()`, since Matrix `mxc://` URIs for encrypted media aren't natively playable.
- Added boundary clamping on drag gestures so the floating window can't be dragged off-screen, plus a safe aspect-ratio fallback to avoid division-by-zero/NaN dimensions.
- Player teardown now explicitly stops playback (`onRelease`), and playback errors dismiss the overlay instead of showing an unrecoverable system dialog.
- The expand/collapse icon now reflects the actual minimized state instead of always showing "expand".
- Added Compose previews and unit tests covering the floating window size/position math.

## Screenshots / GIFs

https://github.com/user-attachments/assets/cde0108d-39cb-47b7-9c3b-2752c87a48a9

## Tests

- Open a video in the media viewer.
- Tap the minimize action to float the player.
- Drag the floating player around the screen; verify it stays fully within screen bounds.
- Tap to toggle between minimized and fullscreen; verify the icon reflects the current state.
- Let playback finish, and trigger a playback error; verify the overlay dismisses cleanly in both cases.
- `./gradlew test` passes, including the new `ScreenSizeHelpersTest`.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

    - [ ] No.

### DIFF
--- a/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/system/SystemUtils.kt
+++ b/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/system/SystemUtils.kt
@@ -196,6 +196,24 @@ fun Context.openGooglePlay(
     }
 }
 
+/**
+ * Opens the system settings for the current app to allow drawing over other apps.
+ */
+fun Context.openSystemOverlaySettings() {
+    val intent = Intent(
+        Settings.ACTION_MANAGE_OVERLAY_PERMISSION,
+        "package:$packageName".toUri()
+    ).apply {
+        addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+    }
+    try {
+        startActivity(intent)
+    } catch (e: ActivityNotFoundException) {
+        toast(getString(R.string.error_no_compatible_app_found))
+    }
+}
+
+
 // Not in KTX anymore
 fun Context.toast(resId: Int) {
     Toast.makeText(this, resId, Toast.LENGTH_SHORT).show()
--- a/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/system/SystemUtils.kt
+++ b/libraries/androidutils/src/main/kotlin/io/element/android/libraries/androidutils/system/SystemUtils.kt
@@ -213,7 +213,6 @@ fun Context.openSystemOverlaySettings() {
     }
 }
 
-
 // Not in KTX anymore
 fun Context.toast(resId: Int) {
     Toast.makeText(this, resId, Toast.LENGTH_SHORT).show()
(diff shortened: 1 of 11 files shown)


aliIsazadeh:
Could a maintainer add the `PR-Feature` label so the changelog check passes? Thanks! 
@bmarty

PR #7727, opened 2026-09-17
Fix QR code login not checking the local network permission

### LINKED ISSUE #7694: QR sign-in silently fails on a LAN homeserver: ACCESS_LOCAL_NETWORK never requested in the QR flow (Android 16/17 Local Network Protection)
### Steps to reproduce

Fresh Element X (never granted the "local network devices" permission), on the sign-in screen. Homeserver is self-hosted (Synapse + Matrix Authentication Service, MSC3861/OIDC, msc4108_enabled) reachable only on a private/LAN address (e.g. https://matrix.example.lan → 192.168.x.x). On a second, already-signed-in device (Element Desktop) open Settings → Sessions → Link new device → Show QR code.
2. What do you click?
In Element X: Sign in → "Sign in with QR code" → scan the QR shown by Element Desktop.
3. More steps…
Element X opens the camera and decodes the QR (logcat shows libzxingcpp load + camera open/close, so the rendezvous URL is read fine), then shows "Establishing a secure connection" and spins forever.

#### What did you expect?
The QR flow should request the Android local-network permission (ACCESS_LOCAL_NETWORK) — as the manual "Enter your account provider" flow does — then connect to the rendezvous and finish signing in.

#### What happened instead?
The rendezvous URL is a private/LAN IP, and Element X hasn't been granted android.permission.ACCESS_LOCAL_NETWORK, so Android silently blocks the connection. Element X sends zero requests to the rendezvous (verified server-side: the QR-showing device polls and gets only 304s; the phone never contacts it). The QR flow never prompts for the permission, so nothing indicates why it fails.
Proof of cause: start a manual sign-in (Sign in → type the homeserver address) — that flow does show the "allow connections to devices on your local network" prompt. Grant it, back out, rescan the QR → connects instantly. Revoking the permission reproduces the hang (dumpsys package io.element.android.x → ACCESS_LOCAL_NETWORK: granted=false); granting it fixes it.

### Your phone model

Google Pixel 7a

### Operating system version

Android 17 (API 37)

Element X 26.09.1 - F-Droid (org.fdroid.fdroid)

### Homeserver

Self-hosted Synapse + Matrix Authentication Service (MSC3861 delegated auth, experimental_features.msc4108_enabled: true), on a LAN IP. Server-side MSC4108/rendezvous verified working, Element Desktop completes the flow against the same server.

### Will you send logs?

Yes,

```
### logcat during a QR scan while ACCESS_LOCAL_NETWORK = granted=false
### (device/network identifiers redacted)

08:00:53.988 I/InputDispatcher: ... io.element.android.x/io.element.android.x.MainActivity ...
08:00:57.595 D/CoreBackPreview: Window{io.element.android.x/io.element.android.x.MainActivity}: Setting back callback
08:00:59.618 I/CameraService: CameraService::connect call (PID <redacted> "io.element.android.x", camera ID 0) Camera API version 2
08:00:59.713 I/CameraService: Start camera streaming for io.element.android.x (UID 10274)
08:00:59.832 D/nativeloader: Load .../io.element.android.x-.../base.apk!/lib/arm64-v8a/libzxingcpp_android.so ... : ok
08:00:59.915 I/CameraService: Stop camera streaming for io.element.android.x (UID 10274)
08:00:59.950 W/DefaultDispatch: avc: denied { search } ... app=io.element.android.x   (SELinux cgroup noise, unrelated)
```

No

jmartinesp:
## Content

This fixes a bug in Android 17 where a homeserver in a local network was used with QR code login, and this flow didn't request the local network access permission.

Now, after we get a server name from the QR code, the local network access permission is checked and requested if missing. If the user actively denies it, we go back to the previous screen.

There is also a bugfix in `LocalNetworkPermissionGate` to not trigger this accidentally when the permission has been denied but we haven't requested a permission check yet.

## Motivation and context

Fixes https://github.com/element-hq/element-x-android/issues/7694.

## Tests

- On Android 17, on a fresh start, start a QR code login.
- Check the actual flow and it should work just asking the camera permission.
- Then modify `QrCodeScanPresenter` so the `urlOf` lambda returns something like `"http://localhost"` and try again. This time, the local network permission should be requested.
- If you deny the permission, the scan QR screen will be dismissed.

You can also check the flow in previous devices, but given it's reusing an existing component that's been tested before, it should just work™.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 17

## Checklist

    - [x] No.


features/login/impl/build.gradle.kts
```diff
@@ -72,6 +72,7 @@ dependencies {
     implementation(projects.libraries.preferences.api)
     implementation(projects.libraries.uiUtils)
     implementation(projects.libraries.wellknown.api)
+    implementation(projects.services.toolbox.api)
```
ganfra: Do we need this?

underclockeddev:
que rico, muchas gracias Jorge!

PR #7732, opened 2026-09-18
Stop rewriting screenshots that did not meaningfully change

bmarty:
## Content

Recording the screenshots for a change with no visual effect rewrote 72 PNGs, and that keeps happening: recent "Update screenshots" commits touch 72, 102, 72, 29, 63 and 121 files, always with as many insertions as deletions. The screenshots flap between two fixed renderings. Across the 72 files of the last such commit, 70 have reverted to a byte-identical earlier state at least once.

Half of them differ by about 0.001% of their pixels, a fraction of a pixel of coverage on the rounded corners of the message bubbles. That is 10 times below the `maxPercentDifference` that `verifyPaparazziDebug` accepts, so CI already considers the difference irrelevant and would never fail on it. But recording had no tolerance at all and rewrote every screenshot unconditionally, turning renderer noise into a Git LFS blob rewrite.

Paparazzi has `app.cash.paparazzi.overwriteOnMaxPercentDifference` for exactly this: when recording, keep the existing screenshot if it is within tolerance. It only works if the screenshots survive into the record run, so deleting them all beforehand has to go. Stale screenshots are now pruned afterwards instead, from the list of previews Paparazzi reports it has just rendered. That task refuses to run if it cannot read the reports or if it would delete an implausible number of files, so a format change breaks the build rather than wiping the screenshots.

This does not address the other half of the churn, where a preview with an animation or a bottom sheet records an entirely different image, sometimes a blank one. That is a LayoutLib bug, and it cannot be fixed here: the fix and the removal of an API Paparazzi still calls both landed in LayoutLib 16.2.3, so forcing that version makes every screenshot test fail. Documented in LayoutLibErrorFilterStatement so the next person does not retry it.

We will need https://github.com/cashapp/paparazzi/pull/2375 to fix the screenshot instability issue completely.

## Tests

- I'll add several times the record screenshot label

## Tested devices

NA

## Checklist

    - [ ] No.


bmarty:
Tests are OK, recording screenshot again to check stability

tests/uitests/src/test/kotlin/ui/LayoutLibErrorFilterStatement.kt
```diff
@@ -20,6 +20,17 @@ import org.junit.runners.model.Statement
...
+ * Note: forcing layoutlib >= 16.2.3 ourselves does not work, and this is not just a matter of
+ * declaring the version. The `setThreadNiceness` fix and the removal of `Bridge.prepareThread()`,
+ * which Paparazzi 2.0.0-alpha05 still calls, both landed in 16.2.3, so every screenshot test fails
+ * with `NoSuchMethodError: Bridge.prepareThread()`. 16.2.2 is the last compatible release and it does
+ * not have the fix. We have to wait for a Paparazzi release.
+ *
+ * Swallowing the error is not harmless: HandlerThread.run() publishes its Looper and then dies
+ * before Looper.loop(), on a background thread racing the render, so a screenshot can be recorded
+ * from a partially rendered frame. That is why previews with animations or a bottom sheet flip
+ * between two results (sometimes an entirely blank image) on every recording.
```
jmartinesp, line 33: Is this comment useful? I think it references stuff that's no longer here?

build.gradle.kts
```diff
@@ -215,19 +215,87 @@ tasks.register("generateDocsToc", Exec::class.java) {
...
+            check(expected.isNotEmpty()) {
+                "Could not read any snapshot name out of the Paparazzi run reports in $reportsDir. " +
+                    "The report format has probably changed. Refusing to delete anything."
+            }
+
+            val existing = imagesDir.listFiles { file -> file.extension == "png" }.orEmpty()
+            val obsolete = existing.filterNot { it.name in expected }
+            // Deleting a screenshot is a tracked file deletion, and a bad `expected` set would wipe thousands
+            // of them. Anything beyond a trickle means something is wrong rather than a few deleted previews.
+            check(obsolete.size <= existing.size / 10) {
```
jmartinesp, line 282: This might be a bit too low, but we can test it 👍 .

jmartinesp (approved):
Let's try this out!

tests/uitests/src/test/kotlin/ui/LayoutLibErrorFilterStatement.kt
bmarty: I think we should really wait for Paparazzi to be released with the fix in layoutlib 16.2.3 before going further on this. I'll revisit the PR when it will be the case.

PR #7735, opened 2026-09-18
Remove Roborazzi, use Paparazzi for Compound

jmartinesp:
## Content

The new screenshots will be recorded with the rest of the Paparazzi ones. Note the names changed, since Paparazzi enforces a `$packagename.$testname.$name.$extension` format

## Motivation and context

There is no need for 2 different screenshot test libraries in the project. At the time we used Roborazzi because it allowed larger screenshots, which Paparazzi failed to.

## Screenshots / GIFs

In the PR.

## Checklist

    - [ ] No.


jmartinesp:
Agh, I reduced the screen density and that broke the self-sizing. The latest changes *should* fix that.

libraries/compound/src/test/snapshots/images/io.element.android.compound.screenshot_MaterialColorSchemeTest_screenshots_material3_colors_-_dark.png
jmartinesp, line 1: We might want to give these a max width 😬 .

.github/workflows/tests.yml
```diff
@@ -79,7 +79,7 @@ jobs:
         run: ./gradlew :tests:uitests:verifyPaparazziDebug $CI_GRADLE_ARG_PROPERTIES
 
       - name: ⚙️ Run Compound screenshot tests
-        run: ./gradlew :libraries:compound:verifyRoborazziDebug $CI_GRADLE_ARG_PROPERTIES
+        run: ./gradlew :libraries:compound:verifyPaparazziDebug $CI_GRADLE_ARG_PROPERTIES
```
bmarty: Does it still need to be a separated task? It's not covered by the task `Run screenshot tests` above?

bmarty (commented):
One remark. Can we avoid having specific Paparazzi for the compound library as per any other module?

.github/workflows/tests.yml
jmartinesp: > One remark. Can we avoid having specific Paparazzi for the compound library as per any other module?

Sorry, I'm not sure what you mean: a specific test, Gradle task config or step in the CI?

libraries/compound/src/test/kotlin/io/element/android/compound/screenshot/CompoundIconTest.kt
```diff
@@ -51,7 +49,7 @@ class CompoundIconTest : RobolectricTest() {
                 )
             }
         }
-        captureRoboImage(file = screenshotFile("Compound Vector Icons - Dark.png")) {
+        paparazzi.snapshot(name = "Compound Vector Icons - Dark") {
```
bmarty, line 52: Do we still need to invoke `paparazzi.snapshot`? Using `@Preview` should be enough, now? Maybe due to the fact that this module was a separate project a long time ago? If yes, maybe keep this in case we roll back (for Element Call maybe?).

bmarty (approved):
Thanks for the update

.github/workflows/tests.yml
bmarty: Yes, what you did in https://github.com/element-hq/element-x-android/pull/7735/commits/861ca9d2efbd03c902bd0f704e6d5ef5cb0356f2

libraries/compound/src/test/kotlin/io/element/android/compound/screenshot/CompoundIconTest.kt
jmartinesp: > Do we still need to invoke paparazzi.snapshot? Using @Preview should be enough, now? 

We won't have to if we include `:libraries:compound` in the list of modules `:tests:uitests` includes, if I'm not mistaken. But if we do that we'll lose the titles, and some of the tests would have to be re-done, I believe, since their setup is a mix of the previews and a custom wrapper when using `snapshot`.

